Issue #22, opened 2019-08-18
Add GraphQL query for getting feeds


idoshamun:
Currently there are many endpoints for getting different feeds:
- Main feed which queries posts based on publications and tags and rank them by popularity (actually inside, it has two branches one for anonymous users and one for registered users)
- A single tag feed which obviously returns all the posts of a single tag ordered by time
- The same goes for a single publication feed

The current design is very limited for the applications as every change in one of the feeds must coordinated with the API. We would like to present a query which will provide the frontend a single yet flexible endpoint it need to create ad-hoc feeds and will lay the ground for a future search mechanism as well.

The query should support the following:
- Filters (everything is optional):
  - Before and after - used to filter posts based on the creation time
  - Exclude publications - posts which belong to these publications will be filtered out
  - Include publication - posts will belong only to these list of publications
  - Tags - posts must contain at least one tag in the list
- Ordering:
  - Popularity - rank posts by combining number of views and time since creation
  - Time - rank posts only by time

Query must also support pagination (probably can be achieved with before and after filters).

This issue is complex and can be split to smaller tasks as in the list below.
Would love to get the community thoughts on this one as well.

- [x] Write a GraphQL schema
- [x] Create a generic function to query the data with SQL
- [x] Refactor existing endpoints to use the mentioned above function
- [x] Choose a GraphQL server package (with Koa support)
- [x] Add a GraphQL query

PR #28, opened 2019-09-05
feat: adding GraphQL

Andrei0872:
This is related to #22.

This PR contains the GraphQL setup alongside with the implementation of the `post` entity.


src/graphql/helpers/post.js
```diff
@@ -0,0 +1,44 @@
...
+    fields: userId ? post.defaultUserFields : post.defaultAnonymousFields,
+    filters: Object.assign({}, { before: args.latest }, filters),
+    rankBy,
+    userId,
+    page: args.page,
+    pageSize: args.pageSize,
+  };
+};
+
+const splitArrayStr = str => (str ? str.split(',') : null);
```
idoshamun: Duplication of code from https://github.com/dailynowco/daily-api/blob/master/src/routes/posts.js#L13.
Maybe you can use the same function in both files

src/graphql/helpers/post.js
```diff
@@ -0,0 +1,44 @@
...
+
+const assignType = type => x => Object.assign({}, x, { type });
+
+const fetchToiletAd = async (ip, config) => {
```
idoshamun: Duplication of code from https://github.com/dailynowco/daily-api/blob/master/src/routes/posts.js#L97.
Maybe you can use the same function in both files

src/graphql/resolvers/post.js
```diff
@@ -0,0 +1,153 @@
...
+        feedParams.filters = Object.assign({}, feedParams.filters, {
+          publications: { include: helpers.splitArrayStr(args.pubs) },
+          tags: { include: helpers.splitArrayStr(args.tags) },
+        });
+      }
+
+      return await post.generateFeed(feedParams);
+    },
+
+    async post(parent, args, { models, user }) {
```
idoshamun: Please follow the same convention you used of object destructuring: `{ models: { post }, user }`

src/graphql/resolvers/post.js
```diff
@@ -0,0 +1,153 @@
...
+      throw new EntityNotFoundError('post', 'id', args.id);
+    },
+
+    async bookmarks(parent, { params: args }, { user, models: { post } }) {
+      if (!user) {
+        throw new ForbiddenError();
+      }
+
+
+      args.latest = args.latest ? new Date(args.latest) : null;
```
idoshamun: Too many blank lines before and after this line

src/graphql/resolvers/post.js
```diff
@@ -0,0 +1,153 @@
...
+      }
+
+      await post.hide(user.userId, model.id);
+
+      return id;
+    },
+  },
+
+  Post: {
+    created_at(parent) {
```
idoshamun: Please use camelCase convention rather then underscore case. I know database schema uses underscore case but this is just because it's a MySQL convention

src/graphql/resolvers/post.js
```diff
@@ -0,0 +1,153 @@
...
+      return parent.createdAt && parent.createdAt.toISOString();
+    },
+
+    read_time(parent) {
```
idoshamun: Please use camelCase convention rather then underscore case. I know database schema uses underscore case but this is just because it's a MySQL convention

src/graphql/resolvers/post.js
```diff
@@ -0,0 +1,153 @@
...
+      return parent.readTime;
+    },
+
+    published_at(parent) {
```
idoshamun: Please use camelCase convention rather then underscore case. I know database schema uses underscore case but this is just because it's a MySQL convention

src/graphql/schema/post.js
```diff
@@ -0,0 +1,82 @@
+const sharedPostInputFields = `
+  latest: String!
+  page: Int!
+  pageSize: Int!
+`;
+
+export default `
+  type Post {
```
idoshamun: Please use camelCase convention rather then underscore case. I know database schema uses underscore case but this is just because it's a MySQL convention

src/graphql/schema/post.js
```diff
@@ -0,0 +1,82 @@
...
+    id: ID!
+    title: String!
+    url: String!
+    image: String
+    ratio: Float
+    placeholder: String
+    publication_id: ID!
```
idoshamun: The post object which returned from `generateFeed` doesn't have a `publication_id` property although the schema does contain it. IMO we should remove this field.

src/graphql/schema/publication.js
```diff
@@ -0,0 +1,9 @@
+export default `
+  type Publication {
+    id: ID!
+    name: String!
+    image: String!
+    enabled: Boolean
+    twitter: String
```
idoshamun: This field is better be removed for now

src/graphql/schema/publication.js
```diff
@@ -0,0 +1,9 @@
+export default `
+  type Publication {
+    id: ID!
+    name: String!
+    image: String!
+    enabled: Boolean
```
idoshamun: This field is better be removed for now

src/graphql/helpers/post.js
Andrei0872: I could create an `utils/` directory where I'd put the common functions. Is it okay?

src/graphql/schema/publication.js
Andrei0872: If we remove the field above, we won't be able to perform queries like this one
```graphql
{
  post(id: 1) {
    title
    publication {
      name
    }
  }
}
```

src/graphql/schema/publication.js
Andrei0872: This is related to the comment above :D.

src/graphql/helpers/post.js
idoshamun: Yes, perfect

src/graphql/schema/publication.js
idoshamun: I mean specifically the `enabled` property not the publication field

src/graphql/schema/publication.js
idoshamun: I mean specifically the `twitter` field

Issue #46, opened 2020-03-04
Remove read articles from the main feed


idoshamun:
As stated in https://github.com/dailydotdev/daily/issues/142, we would like to filter out the read articles from the main feed of the extension.
The first step is to add a filter that will do exactly this in the `generateFeed` function (https://github.com/dailydotdev/daily-api/blob/master/src/models/post.js#L261). This function is the core that generates all the feeds we can browse. The good news is that a field which indicates whether the post has been read already exists and is called `read` (obviously :P), now we just need to filter by it.
In the second step, we have to add a parameter to the graphql query which states whether the user would like to filter out the read articles or not (https://github.com/dailydotdev/daily-api/blob/master/src/graphql/schema/post.js#L93).
Lastly, we need to add the user preference so we can sync it across the devices (https://github.com/dailydotdev/daily-api/blob/master/src/models/settings.js). Please make sure to create a knex migration file.
It is highly important to add tests to check everything, this is a core functionality of the API and we have to make sure it works and will keep working on every change.
I would love to provide more information if needed :)

prateek3255:
Okay, @idoshamun I'll go through the instructions mentioned above and will let you know if any help is needed.

idoshamun:
Thank you :)

prateek3255:
Hey, @idoshamun I have opened a draft PR #47 where I'll be pushing the changes, can you please review the changes so that I know I am moving in the right direction.

PR #47, opened 2020-03-04
feat(posts): Remove read articles from feed

prateek3255:
Fixes: #46

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -12,10 +12,4 @@ const add = (type, userId, postId, referer, agent, ip, timestamp = new Date()) =
     .then(() => obj);
 };
 
-const view = (userId, postId) => {
-  const obj = { type: 'view', userId, postId };
-  return db.insert(toSnakeCase(obj)).into(table)
-    .then(() => obj);
-};
-
-export default { getAll, add, view };
+export default { getAll, add };
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -12,4 +12,10 @@ const add = (type, userId, postId, referer, agent, ip, timestamp = new Date()) =
     .then(() => obj);
 };
 
-export default { getAll, add };
+const view = (userId, postId) => {
+  const obj = { type: 'view', userId, postId };
+  return db.insert(toSnakeCase(obj)).into(table)
+    .then(() => obj);
+};
+
+export default { getAll, add, view };
--- a/migrations/20200324164902_add_read_setting.js
+++ b/migrations/20200324164902_add_read_setting.js
@@ -1,0 +1,9 @@
+exports.up = knex =>
+  knex.schema.table('settings', (table) => {
+    table.boolean('show_only_not_read_posts').defaultTo(false);
+  });
+
+exports.down = knex =>
+  knex.schema.table('settings', (table) => {
+    table.dropColumn('show_only_not_read_posts');
+  });
--- a/src/graphql/helpers/post.js
+++ b/src/graphql/helpers/post.js
@@ -4,7 +4,7 @@ const getFeedParams =
 
     return {
       fields: userId ? [...post.defaultUserFields, 'read'] : post.defaultAnonymousFields,
-      filters: Object.assign({}, { before: args.latest }, filters),
+      filters: Object.assign({}, { before: args.latest }, filters, typeof args.read === 'boolean' && userId && { read: args.read }),
       rankBy,
       userId,
       ignoreUserFilters,
--- a/src/graphql/schema/post.js
+++ b/src/graphql/schema/post.js
@@ -63,6 +63,7 @@ export default `
     pubs: String
     tags: String
     sortBy: String
+    read: Boolean
   }
 
   input ToiletInput {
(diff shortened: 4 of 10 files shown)


prateek3255:
@idoshamun  I have completed the first part of the issue with test cases. Can you please check if its correct?

idoshamun:
@prateek3255 looks good, could you add or modify a test case that also requests the `read` field. I want to verify there is no conflict between the two.

prateek3255:
> @prateek3255 looks good, could you add or modify a test case that also requests the read field. I want to verify there is no conflict between the two.
Updated

prateek3255:
@idoshamun I have added the read filter to the graphql schema, can you please check if it's correct. Also, I am a bit confused regarding the test case, so it would be really nice if you could help me with it.

idoshamun:
@prateek3255 thank you very much! Regarding the test case what I had in my mind is a test case that requests the fields `id` and `read` for example and apply your brand new `read` filter as well. I'm having thoughts that the `read` field and `read` filter might create a conflict in the SQL. I want it to be validated with a test case. It's as easy as adding the `read` field to your new test cases (`should return only (not) read posts`).

prateek3255:
@idoshamun Sorry for the late reply got a little busy. I understood writing the test case part, one thing I am still not sure about, is how to make the fixture posts mark as read in the db for the test case.
I couldn't find any relevant function for it in the [models/post.js](https://github.com/dailydotdev/daily-api/blob/master/src/models/post.js) file.

idoshamun:
No rush, to mark a post as read you have to create a `view` event and to provide the relevant user and post id. A function to create an event is available here:
https://github.com/dailydotdev/daily-api/blob/master/src/models/event.js#L7

prateek3255:
@idoshamun I added the test cases as you said but for some reason, the not read test case is passing successfully, but the read one is failing (it's returning an empty array). Can you please check why this is happening?

idoshamun:
Have you tried to look at the generated SQL query by Knex? I think something with joining the events table is not right. Once you generate the query you can share it here if you can't find the issue.
Hopefully, we will manage to find the issue together :)

prateek3255:
It has something to do with the tags, because it's not returning the posts that don't have tags in them. I updated the fixtures such that the output only contains the posts with tags, then it works fine. Is this expected?

prateek3255:
> Have you tried to look at the generated SQL query by Knex? I think something with joining the events table is not right. Once you generate the query you can share it here if you can't find the issue.
Hopefully, we will manage to find the issue together :)

Just saw this right now, I'll check the generated SQL query.

idoshamun:
If you don't use the tags filter than it shouldn't mind the tags. It should return posts that have no tags as well. Let's look at the generated SQL and we'll understand better the situation :)

prateek3255:
Here's the generated query when the read filter is true @idoshamun 
```sql
select `posts`.`id`, `posts`.`title`, `posts`.`url`, `posts`.`published_at`, `posts`.`created_at`, `posts`.`image`, `posts`.`ratio`, `posts`.`placeholder`, `posts`.`views`, `posts`.`read_time`, `publications`.`id` as `pub_id`, `publications`.`image` as `pub_image`, `publications`.`name` as `pub_name`, (select GROUP_CONCAT(tags.tag ORDER BY tags_count.count DESC SEPARATOR ',') from `tags` inner join `tags_count` on `tags`.`tag` = `tags_count`.`tag` where `tags`.`post_id` = `posts`.`id` group by `tags`.`post_id`) as `tags`, bookmarks.post_id IS NOT NULL as bookmarked, (select count(*) from `events` where `events`.`type` = ? and `events`.`user_id` = ? and `events`.`post_id` = `posts`.`id`) as `read` from `posts` inner join `publications` on `posts`.`publication_id` = `publications`.`id` inner join `events` on `events`.`post_id` = `posts`.`id` and `events`.`user_id` = ? where `events`.`type` = ?
```
I don't understand it completely, can you check what's wrong here

idoshamun:
By looking at the SQL query, I couldn't find any issue. The edge-case of someone who doesn't have tags filter is extremely rare. I will try to dig deeper into this issue, in the meantime, you can proceed with the workaround you have found.
Thank you for contributing during this difficult time :heart:

prateek3255:
@idoshamun I have added the setting and updated the corresponding test cases as well.

prateek3255:
I think this finishes the API part, do let me know if there are still any changes required.

src/models/event.js
```diff
@@ -12,4 +12,10 @@ const add = (type, userId, postId, referer, agent, ip, timestamp = new Date()) =
     .then(() => obj);
 };
 
-export default { getAll, add };
+const view = (userId, postId) => {
```
idoshamun: One thing catches my attention here, this function is made especially for testing as it ignores the other parameters of an event. I'm afraid we can forget about it in the future and use it as is. What do you think?

prateek3255: In that case, I think we can just use the add function and remove the view function. I just noticed that I used the add function for marking posts as read for the in the `post` test. https://github.com/dailydotdev/daily-api/blob/de492bd5e2d9b7d66cd5b4a442891885632f5c8a/test/integration/models/post.js#L175
We can use it in the graphql tests as well, as that's the only place we are using the view function right now.

idoshamun: Ok so let's proceed with removing the function :)

prateek3255: Done!

prateek3255: Done @idoshamun

idoshamun:
Thank you for everything, in the coming days I'll create an issue for the extension and make sure to tag you. Meanwhile, we will reach out to you via Twitter to send a little gift :)

PR #733, opened 2021-09-21
feat: upvoted by queries for both posts and comments

sshanzel:
- [x] API for Post's Upvote(s)
- [x] API for Comment's Upvote(s)

DD-30


src/common/upvoteGenerator.ts
```diff
@@ -0,0 +1,29 @@
...
+import { getCursorFromAfter } from './pagination';
+import { base64 } from './base64';
+import { IBaseUpvote } from '../entity/BaseUpvote';
+
+export interface UpvotesPage extends Page {
+  timestamp?: Date;
+}
+
+export const upvotePageGenerator: PageGenerator<
+  IBaseUpvote,
```
idoshamun, line 12:
```suggestion
 { createdAt: Date },
```

idoshamun (changes requested):
I don't like the fact that we require entities to inherit from another class. It might mess with reflection things that TypeORM does. So I think we should revert back to a more explicit form of entities declare user and createdAt. But what we can do to achieve the same thing with the page generator is just to force the generic argument to have the `createdAt` field.
I added an inline suggestion, please make sure it works.
Other than that it looks fabulous!

**A friendly reminder to add tests**

sshanzel:
This should be now ready for review. I just revised the test cases I made to make the snapshot contain the some actual data to have better integrity. I am starting to love to write unit tests now 😆

idoshamun (changes requested):
I think it's better to add one more item in the tests so we can check that it returns a list ordered by `createdAt`. What do you think?

sshanzel:
True, we should do that, since we specifically want it be ordered in a descending manner. I'll write another one for this. Thanks for the thought!

idoshamun:
@sshanzel you can use the same test, I think

sshanzel:
I am starting to love writing unit-tests! Lol. I have now applied an item that will determine the order of the result.

idoshamun (changes requested):
Tests are now failing because you saved to the snapshot dates that are based on the current time. My recommendation is to use fixed dates. This will make sure the snapshot data is persistent and the tests will pass. After this, we can merge! 🚀

sshanzel:
Weird. Earlier I have tried a constant date but the dates kept on mismatching, though it was like a day difference on the post's test but not on the comment's test 🤪  Strange that it is working now. Lol. Pushing in a few seconds..

idoshamun (approved):
Woohoo! 🚀

sshanzel:
Oh yeah! 🎸🤘 Thank you so much! 🙏

PR #744, opened 2021-09-28
refactor: generic page generator for paginated queries

sshanzel:
Jira Ticket: [DD-213#done](https://dailydotdev.atlassian.net/browse/DD-213)

I have tried to abstract as much as possible and just left the `QueryBuilder` for consumers to focus only on the result of their actual query.


src/common/pageGenerator.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface PageGeneratorConfig {
+  maxSize?: number;
+}
+
+export interface PageGeneratorEntity {
+  createdAt: Date;
+}
+
+const DEFAULT_PAGE_SIZE = 300;
```
idoshamun, line 22:
```suggestion
const DEFAULT_PAGE_SIZE = 100;
```

src/common/pageGenerator.ts
```diff
@@ -0,0 +1,96 @@
...
+const DEFAULT_MAX_SIZE = 300;
```
idoshamun, line 23:
```suggestion
const DEFAULT_MAX_SIZE = 100;
```

src/common/pageGenerator.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export function createPageGenerator({
```
idoshamun, line 25: Not sure about the name because it's not an all-purpose page generator. It's a page generator for entities with `createdAt` property. What do you think?

src/schema/comments.ts
```diff
@@ -297,99 +296,59 @@ export const resolvers: IResolvers<any, Context> = {
...
-          if (page.timestamp) {
-            builder.queryBuilder = builder.queryBuilder.andWhere(
-              `${builder.alias}."createdAt" > :timestamp`,
-              { timestamp: page.timestamp },
-            );
-          }
-          return builder;
-        },
-      );
+      const pageGenerator = new GQLPageGenerator<GQLComment>();
```
idoshamun, line 304: We can create all these instances of the page generator outside the resolver. What do you think?

idoshamun (changes requested):
Please also make sure everything works locally otherwise the pagination might break in production

src/common/pageGenerator.ts
sshanzel: Ah, yes. I will think of a much more precise name for this. Thanks for the thought!

src/schema/comments.ts
sshanzel: I think that is a good idea, to have less memory footprint. Then we let's allow passing down of config on a much later stage of the generator's life.

sshanzel:
After checking the affected queries, locally, all seemed to be working as intended 💯

idoshamun (approved):
Woop woop!! 🚀

PR #745, opened 2021-09-29
fix: generic page generator order condition based on parameter

sshanzel:
DD-219#done

The Generic Page Generator is reliant on the order of the actual query. So the condition must come from the parameter that the consumer would send.

Since `orderBy` overrides the existing parameters, I used `addOrderBy` so consumers would still be able to apply their own set of rules for the order in the query builder.


idoshamun (approved):
Great addition!

PR #750, opened 2021-10-07
Feat: bookmark search

rebelchris:
This is the GraphQL change that goes with the apps folder changes for the same issue DD-78

In this version we introduced two new queries
- Search bookmarks (general search)
- Search suggestions for bookmarks (title search for the autocomplete)

Should be tested with the apps directory
DD-78 #done 

Note:
Advanced test needs to be made for the new features


rebelchris:
@idoshamun @sshanzel Added my test for the new queries, would love a review of that ❤️

PR #757, opened 2021-10-14
feat: alert filter status - notify topic subscribers on update

sshanzel:
DD-251 #done

- Created a pubsub topic `alerts-updated`
- `Alerts` entity with a property `filter`, default = `true`
- Include tracking of changes on debezium `application.properties`
- New case in CDC for `Alerts` table
- `publish` a `message` on alerts topic
- Create tests for the new CDC
- Create tests for `alerts` and ensure query and mutation are working.

### DIFF
--- a/src/schema/alerts.ts
+++ b/src/schema/alerts.ts
@@ -65,13 +65,13 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
   },
   Query: {
     userAlerts: async (_, __, ctx): Promise<GQLAlerts> => {
-     if (ctx.userId) {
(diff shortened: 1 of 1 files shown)


sshanzel:
I have made the necessary changes for using `filter` rather than `sidebar`. This should be ready 🎉

sshanzel:
Hi @idoshamun, I have now made the necessary changes based on the comments 🎉  I have one question though. I have used the ChangeObject generic type as to how it was used on sending a Post payload. Would that be sufficient or I will have to explicitly convert the payload into a JSON string then to an actual JSON Object?

PR #758, opened 2021-10-15
feat: advanced feed settings

sshanzel:
DD-248 #done

As part of the “Revamp Filtering Experience”, Advanced Settings was introduced. The main purpose is to tag a Source in which kind of “Editorial” they produce (e.g. Tech magazines, Release notes, etc.). This would allow us to have a more personalized feed based on what the user actually wants to see.

### DIFF
--- a/__tests__/__snapshots__/feeds.ts.snap
+++ b/__tests__/__snapshots__/feeds.ts.snap
@@ -352,9 +352,15 @@ Object {
   "advancedSettings": Array [
     Object {
       "defaultEnabledState": true,
-      "description": "Description for Tech magazines",
-      "id": 1,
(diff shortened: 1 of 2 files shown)


sshanzel:
By default, the `disabled` property should be null as the relevant values are non-existent. I think we can check the snapshots as it has some samples on how it would actually look like.

rebelchris:
@idoshamun From my perspective the global setup here looks valid.
But perhaps you have some more detailed issue you spot with your experience in the API development.

sshanzel:
Thank you for the awesome review @rebelchris! If that's the case, I'll be tagging @idoshamun to ask for review 🎉 🚀

PR #771, opened 2021-10-29
fix: feed filters query with 2 or more false advanced settings

sshanzel:
DD-286 #done

While working on the mutations of the UI, I bumped into a case where the user will have two or more false `enabled` values for the feed advanced settings.

- Included a commit where the test would fail if there are 2 or more false `enabled` values.
- Also included a commit where the tests would pass after the fix.


idoshamun (approved):
An important fix! Thanks

PR #779, opened 2021-11-03
fix: return not found on soft deleted post

sshanzel:


__tests__/posts.ts
```diff
@@ -499,8 +499,8 @@ describe('query post', () => {
     }
   }`;
 
-  it('should throw not found when cannot find post', () =>
-    testQueryErrorCode(client, { query: QUERY('notfound') }, 'NOT_FOUND'));
+  it('should throw not found when post was soft deleted', () =>
```
idoshamun, line 502: Please add it as another test, don't replace the existing test 🙏

__tests__/fixture/post.ts
```diff
@@ -4,6 +4,17 @@ import { Post, PostKeyword, PostTag } from '../../src/entity';
 const now = new Date();
 
 export const postsFixture: DeepPartial<Post>[] = [
+  {
```
idoshamun, line 7: Usually not a good idea to add a new fixture here.
This is used everywhere so it will cause other tests to fail I think.

idoshamun (changes requested):
Thanks for the super quick hot fix!!!
See my comments

__tests__/posts.ts
sshanzel: No way. I was so excited to writing test first doing "TDD" and I immediately got it wrong 🤣

__tests__/fixture/post.ts
sshanzel: Oh, right, I'll do this as a specific addition on that specific case only.

idoshamun (approved):
LGTM!

sshanzel:
Thank you for another lesson on when should we be adding items to our fixtures! 🚀

PR #787, opened 2021-11-08
fix: if feed is missing on feed advanced settings mutation

sshanzel:
DD-302 #done

For newly created users, when feed doesn't exist yet. The mutation might fail since the `feedId` is an FK at `FeedAdvancedSettings` entity going to `Feed` entity.


idoshamun (changes requested):
Please add a test to check this case

sshanzel:
Woop! Woop! The test turned green! 🚀

idoshamun (approved):
Yes!!!

PR #792, opened 2021-11-09
feat: comprehensive reading history

sshanzel:
DD-67 #done

I have created a new page generator based on **any** `Date` property. It is type-strict, so if you pass a property of the expected entity that was not a `Date` type, it won't pass tslint.

Reading history:
- query `readHistory` -  a paginated result of user's reading history
- mutation `hideReadHistory` - update the specified post with timestamp of the user's view history to be `hidden` = true


__tests__/users.ts
```diff
@@ -557,6 +564,79 @@ describe('query userReadHistory', () => {
...
+    }
+  }`;
+
+  it('should not authorize when not logged in', () =>
+    testQueryErrorCode(client, { query: QUERY }, 'UNAUTHENTICATED'));
+
+  it("should return user's reading history in descending order", async () => {
+    loggedUser = '1';
+    const createdAtOld = new Date('2020-09-22T07:15:51.247Z');
+    const createdAtNew = new Date('2021-09-22T07:15:51.247Z');
```
rebelchris, line 593: Maybe also test this one by passing them the other way around? So assigning the new one first then the old?

src/entity/View.ts
```diff
@@ -21,7 +21,10 @@ export class View {
   @Column({ type: 'text', nullable: true })
   referer?: string;
 
-  @PrimaryColumn({ default: () => 'now()' })
+  @PrimaryColumn({ default: () => 'now()', precision: 3 })
```
rebelchris, line 24: Why do we need this precision?

rebelchris (changes requested):
Couple of questions, but really nice work Lee!

src/entity/View.ts
sshanzel: Oh, sorry I will elaborate further the current description in the Docs Page.

Most languages follow the [ISO format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) in date which is 3 digits maximum for milliseconds, and when we store 6 (default) it won't match due to the missing additional 3 digits of milliseconds in the database.

__tests__/users.ts
sshanzel: That makes sense. I can probably just switch the first `view` item to have the new date and the second one to have the older date since I think if we create another test, it will serve the same purpose.

sshanzel: That makes sense. I can probably just switch the first `view` item to have the new date and the second one to have the older date since I think if we create another test, it will serve the same purpose. What do you think?

src/entity/View.ts
idoshamun: But do we actually need a ms level precision? I'm not sure

__tests__/users.ts
idoshamun: I'd like to see a test for a consequent page fetch please

idoshamun (changes requested):
Great work. Love the fact that you stuck to the docs. Made the outcome much more predictable :)

src/schema/users.ts
```diff
@@ -275,6 +349,21 @@ export const resolvers: IResolvers<any, Context> = {
...
+      { postId, timestamp }: { postId?: string; timestamp: Date },
+      ctx: Context,
+    ): Promise<void> => {
+      const args = { postId, timestamp, userId: ctx.userId };
+      const repo = ctx.getRepository(View);
+      const history = await repo.findOne(args);
+
+      if (!history) {
+        throw new NotFoundError('Read history not found');
+      }
```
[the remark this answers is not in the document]
sshanzel: That makes sense. One thing though, it is possible for the user to view the same `postId` with a different timestamp, and if we won't include it in the condition, we will be updating two values which should not be expected.

src/entity/View.ts
sshanzel: Good point. We can then set this to `precision: 0` since if we compare a date vs a timestamp with `000` milliseconds (appended by most formatter as shown below) it would still provide the expected result.

Screen Shot 2021-11-11 at 7 15 25 AM

src/schema/users.ts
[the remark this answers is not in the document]
idoshamun: At this point, it shouldn't happen

__tests__/users.ts
```diff
@@ -587,7 +587,7 @@ describe('query readHistory', () => {
   it('should not authorize when not logged in', () =>
     testQueryErrorCode(client, { query: QUERY }, 'UNAUTHENTICATED'));
 
-  it("should return user's reading history in descending order", async () => {
+  it("should return user's reading history", async () => {
```
idoshamun, line 590: Make sure to run lint fix, you use double quotes instead of single

src/entity/View.ts
```diff
@@ -21,7 +21,7 @@ export class View {
   @Column({ type: 'text', nullable: true })
   referer?: string;
 
-  @PrimaryColumn({ default: () => 'now()', precision: 3 })
+  @PrimaryColumn({ default: () => 'now()', precision: 0 })
```
idoshamun, line 24: I prefer to not introduce any change to existing columns in this table as it contains a lot of data.
Unless it's mandatory

idoshamun (changes requested):
Please add a test case for pagination in read history as well

__tests__/users.ts
sshanzel: I think the reason this passed is due to the apostrophe being used in the actual text. Let me rephrase it to avoid the usage of double-quotes.

src/entity/View.ts
sshanzel: Understood, with what, whenever we query data for comparison of timestamp, we will have to strongly type the actual query every time even for basic usage so we can parse it by removing the milliseconds and having it matched. I will update the current mutation.

I just felt the process in this PR can benefit future developments as we won't have to bother to do the parsing from the DB query, but yes, it is much better to avoid changes in the structure that will affect the data itself.

idoshamun: We also decided to not compare timestamps at the moment so this shouldn't be a problem

idoshamun (approved):
LGTM! 🎉

sshanzel:
Hi @rebelchris would love to hear your comments on the changes and additional tests 🚀

rebelchris (approved):
Nice work on those tests!

Love it ❤️

PR #793, opened 2021-11-11
refactor: parse data received from client

sshanzel:
Instead of FE worrying about the actual date format to send, whether it should have milliseconds or not, I have thought the BE must be the one to decide that better.

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -360,7 +360,7 @@ export const resolvers: IResolvers<any, Context> = {
         .set({ hidden: true })
         .where('"postId" = :postId', { postId })
         .andWhere(`date_trunc('second', "timestamp"::timestamp) = :param`, {
-          param: timestamp,
+          param: new Date(timestamp).toJSON().split('.')[0],
         })
         .andWhere('"userId" = :userId', { userId: ctx.userId })
         .execute(),


idoshamun (changes requested):
Great suggestion!
You might want to add a test case for this.

sshanzel:
Totally makes sense. Let me pull one up now.

__tests__/users.ts
```diff
@@ -803,4 +803,28 @@ describe('mutation hideReadHistory', () => {
...
+
+    const res = await client.mutate({
+      mutation: MUTATION,
+      variables: {
+        postId: 'p1',
+        timestamp: createdAtDifferentMS.toISOString(),
+      },
+    });
+
+    expect(res.errors).toBeFalsy();
```
idoshamun, line 828: I don't see that you check here that it is actually deleted.
You check only that the were no errors

idoshamun (changes requested):

__tests__/users.ts
sshanzel: I have now included two entities in each test so we can better see in the snapshot that one of the data changed due to successful modification. I initially thought it would not actually update the DB and just mock the request, but I was wrong.

sshanzel:
Aaaand, the tests turned green! 🎉  🟢

idoshamun (approved):
Woohoo!!

PR #794, opened 2021-11-12
feat: time zone

rebelchris:
Added timezone support in the API.
Also modified the graphQL queries to be able to pass timezones.

I had to make a change in the DevCard user is being managed, this new way is faster because it uses the API database instead of the gateway.

Related changes in [Gateway](https://github.com/dailydotdev/daily-gateway/pull/380) and [Apps](https://github.com/dailydotdev/apps/pull/690)

DD-212


sshanzel (approved):
LGTM! 🎉

src/common/devcard.ts
```diff
@@ -65,12 +66,12 @@ export async function getDevCardData(
   userId: string,
   con: Connection,
 ): Promise<DevCardData> {
-  const [user, articlesRead, tags, sourcesLogos, rank] = await Promise.all([
-    fetchUser(userId),
+  const user = await con.getRepository(User).findOne(userId);
```
idoshamun, line 69: If `user` is not defined, I think this function should fail

src/schema/users.ts
```diff
@@ -88,6 +88,10 @@ export const typeDefs = gql`
     Hashnode handle of the user
     """
     hashnode: String
+    """
+    Time zone of the user
+    """
+    timezone: String
```
idoshamun, line 94: I think it should be private and not accessible to everyone

src/schema/users.ts
```diff
@@ -147,16 +151,21 @@ export const typeDefs = gql`
...
     """
     Get a heatmap of reads per day in a given time frame.
     """
-    userReadHistory(id: ID!, after: String!, before: String!): [ReadHistory]
+    userReadHistory(
+      id: ID!
+      timezone: String
+      after: String!
+      before: String!
+    ): [ReadHistory]
```
idoshamun, line 168: I think the timezone here should be queried from the database according to the user id instead of receiving it in the request. This also makes sure the right timezone is used

idoshamun (changes requested):
Some design changes are required before a thorough review can be made

src/common/devcard.ts
rebelchris: Just checking, but this function can never be called if not authorized right?
Because all the sub functions also use the user.

I can add a hard return fail though

idoshamun: You are right. You can use `findOneOrFail`, this function should do it for you. Might named a bit different

rebelchris:
@idoshamun Refactored, frontend changes also coming up soon.
Was able to remove all timezone sending from the frontend.

The only one using it, is the internal graphQL query.

idoshamun (approved):
Looks great!

PR #795, opened 2021-11-12
refactor: alerts rank column

sshanzel:
DD-259

Status to check if the rank achievement popup should be displayed or not.

This will be passed through the gateway as well and be part of the boot query. The current implementation in `daily-gateway` should be able to cover the handling of data for additional properties like this.


src/entity/Alerts.ts
```diff
@@ -8,6 +8,12 @@ export class Alerts {
 
   @Column({ type: 'bool', default: true })
   filter: boolean;
+
+  @Column({ type: 'bool', default: true })
```
idoshamun, line 12: Why is this `true` by default?

src/workers/cdc.ts
```diff
@@ -150,13 +150,18 @@ const onUserChange = async (
...
+) => Object.entries(after).some(([key, value]) => before[key] !== value);
+
 const onAlertsChange = async (
   con: Connection,
   logger: Logger,
   data: ChangeMessage<Alerts>,
 ): Promise<void> => {
   if (data.payload.op === 'u') {
-    if (data.payload.before?.filter !== data.payload.after.filter) {
+    if (anyHasChanged(data.payload.after, data.payload.before)) {
```
idoshamun, line 164: If this is an update operation, it means that something has changed so this is pretty redundant

idoshamun (changes requested):
I'm not sure about this implementation like when this should be set to true or false.
In this PR you only introduce the rank property in the alert entity but the logic to how it actually works is missing.
Before merging it we need to address it.

src/workers/cdc.ts
sshanzel: I was thinking to save some resources since when any column has triggered an update even the new value is the same as the value before, this part will still be triggered. Just to safeguard the usage since our update mechanism for this table is very general but we can definitely remove this comparison since updates from this table are not very frequent.

idoshamun: An update event will not be triggered if the values are the same even if there was an update statement. The safeguard is not needed here :)

sshanzel: Ah, I see. Thought it would run the same as how it would on local test cases where it did execute.

Though if that's the case, then we should just remove this extra safeguard check.

src/entity/Alerts.ts
```diff
@@ -8,6 +8,12 @@ export class Alerts {
...
+  rank: boolean;
 }
 
-export const ALERTS_DEFAULT: Omit<Alerts, 'userId'> = { filter: true };
+export const ALERTS_DEFAULT: Omit<Alerts, 'userId'> = {
+  filter: true,
+  rank: true,
```
rebelchris, line 18: Also wondering about the defaults.
In the app we set the default filter to false, should these not also be false?

rebelchris (commented):
Just one question from my side.

src/entity/Alerts.ts
sshanzel: This is the source of truth - for better UX we don't want to turn it on in UI then suddenly false after fetching the actual values (it would flick), and I believe we have decided this to be `true` as default based on your suggestion back then 😅

sshanzel (commented):
Woop! Woop! This should now be ready for review.

@idoshamun as discussed we will be using the timestamp to identify whether to display the rank achievement popup or not. Though after going through it, we can't set `null` (or value 0) to never show the popup indefinitely, since we won't have any default value on migration. With that, I will be using the `null` value to check for pre-existing preferences in the UI's browser storage for this matter.

We will just have to manually set an unimaginable date where it will never be reached, for users who opted out of the popup.

rebelchris (approved):
Nice! This looks great Lee 🎉

idoshamun (approved):
Great! 💪

PR #797, opened 2021-11-16
fix: fix duplicate authors for featured comments

rebelchris:
Fixed the duplicate authors for featured comments.

Screenshot 2021-11-16 at 11 46 18
Screenshot 2021-11-16 at 11 46 24
Screenshot 2021-11-16 at 11 45 37
Screenshot 2021-11-16 at 11 45 48

Note: We need the " quotes hence the force backticks.

DD-307 #done

### DIFF
--- a/__tests__/cron/__snapshots__/updateFeaturedComments.ts.snap
+++ b/__tests__/cron/__snapshots__/updateFeaturedComments.ts.snap
@@ -31,7 +31,7 @@ Array [
 
 exports[`should update featured comments 2`] = `
 Array [
-  "c3",
   "c2",
+  "c3",
 ]
 `;
--- a/__tests__/cron/__snapshots__/updateFeaturedComments.ts.snap
+++ b/__tests__/cron/__snapshots__/updateFeaturedComments.ts.snap
@@ -26,6 +26,10 @@ Array [
     "featured": false,
     "id": "c6",
   },
+  Comment {
+    "featured": false,
+    "id": "c7",
+  },
 ]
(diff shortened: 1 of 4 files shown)


sshanzel (approved):
It is always so nice to see single-line fixes!

idoshamun:
I thought it would be better to fix it at the root cause:
https://github.com/dailydotdev/daily-api/blob/master/src/cron/updateFeaturedComments.ts

What do you think?

rebelchris:
@idoshamun Good point, I can fix it there for future reference, but guess we either have to manually run a fix query for legacy comments or keep this current fix as well?

What would you prefer?

idoshamun:
@rebelchris we can keep the existing situation as is or create a migration file that fixes it. Whatever you think is a better solution

rebelchris:
Hmm I think this fallback adds a nice net so would suggest we leave it now.

idoshamun (changes requested):
I suggest adding a test case for `updateFeaturedComments` to test this fix

idoshamun (changes requested):
I suggest adding a test case for `updateFeaturedComments` to test this fix

rebelchris:
@idoshamun Good point!

Modified the test to check for this.

- Added `c7` which is the same user as `c2`.
- Ran the old version see result below (as expected)
- Now the new version this test succeeds

Screenshot 2021-11-17 at 09 53 31

idoshamun (approved):
LGTM!

PR #798, opened 2021-11-16
feat: added the show top sites query to the API side

rebelchris:
Introduced the API side of the show top sites query.

Related changes in [Gateway](https://github.com/dailydotdev/daily-gateway/pull/381) and [apps](https://github.com/dailydotdev/apps/pull/697).

DD-252


idoshamun (changes requested):
I think the new property should go to the `Settings` entity.
This seems like a better fit than the `User` entity.

rebelchris:
@idoshamun Oh I see now we have it there already, but the toggle should still go on the user profile?
Or should the whole toggle move to the settings?

idoshamun:
@rebelchris it used to be in the settings panel, I think we should keep it there

PR #800, opened 2021-11-19
feat: delete account

rebelchris:
DD-59

**Related changes:**
[Gateway](https://github.com/dailydotdev/daily-gateway/pull/384) and [Apps](https://github.com/dailydotdev/apps/pull/701)

**Description:**
This introduces the delete account function.
We decided this will cleanup any trace of the user.

The api receives these from the CDC so it can cleanup in the background.
Cleans up the following:
- User
- Alerts
- Bookmarks
- Bookmark Lists
- Comment and replies to main comment from user
- Comment upvotes
- DevCard
- Feed
- Hidden Post
- Post Report
- Settings
- Source Display
- Source Request
- Upvote
- View
- Post -> Cleans author from post

**Introduced test:**
Added `worker/deleteUser`.
This test has the full delete flow.

Meaning before every test we introduce ALL elements for a user.
This way we ensure to test if there are no conflicts when removing items.

Then we test for all the above mentioned tables.


src/workers/deleteUser.ts
```diff
@@ -0,0 +1,91 @@
...
+      await con.getRepository(HiddenPost).delete({ userId: data.user.id });
+      await con.getRepository(PostReport).delete({ userId: data.user.id });
+      await con.getRepository(Settings).delete({ userId: data.user.id });
+      await con.getRepository(SourceDisplay).delete({ userId: data.user.id });
+      await con.getRepository(SourceRequest).delete({ userId: data.user.id });
+      await con.getRepository(Upvote).delete({ userId: data.user.id });
+      await con
+        .getRepository(Post)
+        .update({ authorId: data.user.id }, { authorId: null });
+      await con.getRepository(User).delete(data.user.id);
```
idoshamun, line 66: It's better to run all these queries in a single transaction. This way it's either success or failure (all or nothing). Right now some queries may succeed and others may fail

src/workers/deleteUser.ts
```diff
@@ -0,0 +1,91 @@
...
+        {
+          userId: data.user.id,
+          messageId: message.messageId,
+          err,
+        },
+        'failed to delete user',
+      );
+      if (err.name === 'QueryFailedError') {
+        return;
+      }
```
idoshamun, line 85: I think we can drop this case. This should fail. I usually use this when there are some constraints that need to be ignored like uniqueness.

idoshamun (changes requested):
Great! A few minor changes

src/workers/deleteUser.ts
rebelchris: Makes sense, I've changed it to a transaction looking forward to hear if that is the right way of doing it.

rebelchris:
@idoshamun Would you mind checking if I used the single transaction correctly?

sshanzel (approved):
Can't imagine the time taken for writing all these tests. But in the end, all of those are well worth it.

LGTM! 🚀

idoshamun (approved):
Awesome!

PR #805, opened 2021-11-26
refactor: user reading rank include last read time

sshanzel:
DD-259

As part of whether to display the achievement popup or not, we decided to compare the rank last seen to last article read by the user.


rebelchris (approved):
This looks great! 👏

No comments from my side

src/common/users.ts
```diff
@@ -102,12 +104,14 @@ export const getUserReadingRank = async (
       `count(*) filter(where "timestamp" at time zone '${timezone}' >= date_trunc('day', ${now}))`,
       'today',
     )
+    .addSelect('MAX(timestamp)', 'lastReadTime')
```
idoshamun, line 107: Thanks to this we no longer need the previous select `today`. We used it to calculate the `readToday` prop. But now we can use the `lastReadTime` for it. Just remember to convert to the right time zone when comparing days.

idoshamun (changes requested):
The PR looks great I have a suggestion for improvement

src/common/users.ts
sshanzel: Thank you for noting this! I have now applied the changes 🚀

src/common/date.ts
```diff
@@ -0,0 +1,16 @@
+import { isEqual } from 'date-fns';
+
+export const isDateOnlyEqual = (left: Date, right: Date): boolean => {
```
idoshamun, line 3: There's a a `date-fns` function called `isSameDay` I believe this is what you're looking for

src/common/users.ts
```diff
@@ -101,10 +102,9 @@ export const getUserReadingRank = async (
       'lastWeek',
     )
     .addSelect(
-      `count(*) filter(where "timestamp" at time zone '${timezone}' >= date_trunc('day', ${now}))`,
-      'today',
+      `MAX(date_trunc('second', "timestamp" at time zone '${timezone}'))`,
```
idoshamun, line 105: Why do you apply `date_trunc` it will mess the also the client code.
We need to return the full date time

idoshamun (changes requested):
Please check my comments :)

src/common/date.ts
sshanzel: Oh, nice. Yes, I thought they didn't have this common requirement.

src/common/users.ts
sshanzel: I was thinking to remove the milliseconds since there are differences between the saved value (6 digits) vs the parsed value on js (3 digits). But after realizing the case, I agree, this shouldn't be necessary.

__tests__/__snapshots__/users.ts.snap
```diff
@@ -218,6 +219,7 @@ Object {
 exports[`query userReadingRank should return the reading rank 1`] = `
 Object {
   "currentRank": 2,
+  "lastReadTime": "2021-12-03T00:00:00.000Z",
```
idoshamun, line 222: This is going to fail tomorrow probably. We need to use fixed dates. What do you think?

idoshamun (changes requested):
Last comment!

__tests__/__snapshots__/users.ts.snap
sshanzel: I think should run smoothly since our expectation is `expect.anyhing`. I think this is due to the reason we compare the date today every time to see if it now matches the content of the query.

https://github.com/dailydotdev/daily-api/blob/0d0a998646ec08900a5623ca9f0363bbc48182b4/__tests__/users.ts#L270-L272

sshanzel: And for the `lastReadTime`, we use the values in our `fixtures` for post which has static values.

idoshamun: `lastReadTime` is not generated from the fixtures. It is calculated by the views we push to the database during the test. These views have relative timestamps because we want to test the ranking functionality. So I think you cannot test the `lastReadTime` as you do now. You have to compare it to the value not using snapshots but using the variables we define in the tests. Let me know if you want to talk about it

sshanzel: Now, I understand. After checking the actual tests, this needs to have a separate test case. Thank you so much for elaborating very clearly 💎

sshanzel:
@idoshamun thank you so much for your very sharp eyes and clearly elaborating the case! 💪

idoshamun (approved):
Woohoo! This is super!

PR #806, opened 2021-11-26
feat: markdown comments

rebelchris:
Added markdown support for the comment sections.

We have the user input and edit markdown, but on save this will get parsed to HTML.
I also added a highlighter and a basic highlighting theme so codeblocks are styled.

It's essential to run the temporary bin script, this script will parse existing comments to html format.
Subnote: It also converts links as `http://google.com` to a full link.

DD-210


bin/commentMarkdownToHTML.ts
```diff
@@ -0,0 +1,38 @@
...
+
+(async (): Promise<void> => {
+  const con = await createOrGetConnection();
+
+  const resStream = await con
+    .createQueryBuilder()
+    .select('id, content')
+    .from(Comment, 'c')
+    .where('c."content_html" is NULL')
+    .limit(50000)
```
idoshamun, line 19: If you use stream I think it's not necessary to limit

src/common/markdown.ts
```diff
@@ -0,0 +1,18 @@
...
+  html: true,
+  linkify: true,
+  typographer: true,
+  highlight: function (str, lang) {
+    if (lang && hljs.getLanguage(lang)) {
+      try {
+        return hljs.highlight(str, { language: lang }).value;
+      } catch (__) {}
+    }
+    return '';
```
idoshamun, line 14: Should we return the empty string or the original string `str`?

src/common/markdown.ts
```diff
@@ -0,0 +1,18 @@
+import MarkdownIt from 'markdown-it';
+import hljs from 'highlight.js';
+
+const markdown: MarkdownIt = MarkdownIt({
```
idoshamun, line 4: You can add `export` before the `const` instead of adding a new line at the bottom

src/schema/comments.ts
```diff
@@ -383,6 +390,7 @@ export const resolvers: IResolvers<any, Context> = {
             postId,
             userId: ctx.userId,
             content,
+            content_html: markdown.render(content),
```
idoshamun, line 393: I think we can use TypeORM listeners for better coverage:
https://typeorm.io/#/listeners-and-subscribers/beforeinsert
We can use before insert and before update

idoshamun (changes requested):
A few suggestions for improvement

src/common/markdown.ts
rebelchris: Yep! Missed that one

src/schema/comments.ts
rebelchris: Oh this is really cool! Will adjust to this indeed 👏

rebelchris:
@idoshamun thanks for that amazing suggestion.
There seems to be some weird rules for the listeners as they only work on `save` and not on `update` for example.

So also had to wrap the create in a actual create to trigger it.
See discussion over here for more information:
https://github.com/typeorm/typeorm/issues/2036

__tests__/workers/cdc.ts
```diff
@@ -282,6 +282,7 @@ describe('comment', () => {
     featured: false,
     createdAt: 0,
     lastUpdatedAt: 0,
+    setContentHTML: null,
```
idoshamun, line 285: Why do you do that? I think it's better to fix the ChangeObject type to ignore functions in the given type

src/entity/Comment.ts
```diff
@@ -60,4 +69,10 @@ export class Comment {
     onDelete: 'CASCADE',
   })
   parent: Promise<Comment>;
+
+  @BeforeInsert()
+  @BeforeUpdate()
+  setContentHTML(): void {
+    this.content_html = markdown.render(this.content);
```
idoshamun, line 76: I just noticed `content_html` is not camelCase. Please fix it according to our convention

idoshamun (changes requested):
Looks great!

rebelchris:
@idoshamun It should be ready, but let me know if you spot anything else :D

idoshamun (approved):
Woohoo!!! Ship it! 🚢

PR #809, opened 2021-11-29
fix: markdown links target and rel

rebelchris:
Make sure markdown links get target blank and rel attribute.

Note: When pushing life, we have to re-seed the comments.
- Remove all existing content_html and run the helper again.

### DIFF
--- a/src/common/markdown.ts
+++ b/src/common/markdown.ts
@@ -21,20 +21,18 @@ const defaultRender =
     return self.renderToken(tokens, idx, options);
   };
 
+const setTokenAttribute = (tokens, attribute, attributeValue) => {
+  const attributeIndex = tokens.attrIndex('attribute');
+  if (attributeIndex < 0) {
+    tokens.attrPush([attribute, attributeValue]);
+  } else {
+    tokens.attrs[attributeIndex][1] = attributeValue;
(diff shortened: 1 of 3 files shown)


src/common/markdown.ts
```diff
@@ -14,3 +14,27 @@ export const markdown: MarkdownIt = MarkdownIt({
...
+    return self.renderToken(tokens, idx, options);
+  };
+
+markdown.renderer.rules.link_open = function (tokens, idx, options, env, self) {
+  const targetIndex = tokens[idx].attrIndex('target');
+  if (targetIndex < 0) {
+    tokens[idx].attrPush(['target', '_blank']);
+  } else {
+    tokens[idx].attrs[targetIndex][1] = '_blank';
+  }
```
idoshamun, line 30: I see you use this structure twice. Can you turn it into a function, please?
Something like `setTokenAttribute`.
It will make it much more readable as well.

idoshamun (changes requested):
Let's add a test to check this use case as well

rebelchris:
@idoshamun Refactored the function, and changed the markdown test to also check for links :)

idoshamun (approved):
Great!

PR #812, opened 2021-12-02
feat: sidebar toggle mode

rebelchris:
This is the related API change for the sidebar toggle mode.
Related to frontend changes: https://github.com/dailydotdev/apps/pull/721


src/schema/settings.ts
```diff
@@ -123,6 +130,11 @@ export const typeDefs = gql`
     Whether to open articles on new tab
     """
     openNewTab: Boolean
+
+    """
+    Whether to show the sidebar open
+    """
+    openSidebar: Boolean!
```
sshanzel, line 137: This shouldn't be required just like the others.

```suggestion
    openSidebar: Boolean
```

sshanzel (changes requested):
Just this one thing and update the snapshots then this should be good to go 🚀

idoshamun (approved):
Looking good!
I love the fact that we don't have to change anything on the query once we introduce a new column.

sshanzel (approved):
LGTM! 🚢

PR #819, opened 2021-12-07
feat: user most read tags query

sshanzel:
DD-36

Inline with the newest addition in our Profile Tooltip. This has the API for getting most-read tags by the user.

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -1,3 +1,4 @@
+import { getMostReadTags } from './../common/devcard';
 import { GraphORMBuilder } from '../graphorm/graphorm';
 import { Connection, ConnectionArguments } from 'graphql-relay';
 import {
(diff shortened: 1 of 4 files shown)


src/schema/users.ts
```diff
@@ -245,6 +260,15 @@ export const resolvers: IResolvers<any, Context> = {
...
       }
     },
+    userMostReadTags: async (
+      _,
+      { id }: { id: string },
+      ctx: Context,
+    ): Promise<GQLMostReadTag[]> => {
+      const user = await ctx.con.getRepository(User).findOneOrFail(id);
+
+      return getMostReadTags(ctx.con, user.id, { limit: 5 });
```
rebelchris, line 270: Would we not want the frontend to specify the limit? 
Just wondering for other implementations.

Answer might also be: We'll deal with it when needed :)

rebelchris (approved):
Nice work Lee!
Lovely you where able to reuse what we have for the DevCard

idoshamun (approved):
Wonderful! 😁

src/schema/users.ts
sshanzel: That's a very good point. I agree we can probably deal with it by the time comes and add a max limit since we usually want to get small amount of "most" data only

PR #826, opened 2021-12-13
feat: experiments capability for advanced settings

sshanzel:


src/common/feedGenerator.ts
```diff
@@ -77,10 +79,26 @@ export const feedToFilters = async (
...
+          id: parseInt(key),
+          defaultEnabledState: experimentSettings.value[key],
+        }));
+
+        settings.forEach((experiment) => {
+          subQuery = subQuery.orWhere(
+            `(adv.id = :id AND COALESCE(fas.enabled, :enabled, adv.defaultEnabledState) = false)`,
+            { id: experiment.id, enabled: experiment.defaultEnabledState },
+          );
+        });
```
idoshamun, line 99: This can be refactored into a single reduce instead of a map and forEach.
The reduce will iterate over the keys of the experiments settings and will return the updated sub query

src/common/feedGenerator.ts
```diff
@@ -77,10 +79,26 @@ export const feedToFilters = async (
...
             { feedId },
           )
-          .where('COALESCE(fas.enabled, adv.defaultEnabledState) = false')
-          .getQuery();
+          .where('COALESCE(fas.enabled, adv.defaultEnabledState) = false');
+
+        const experimentSettings = features?.advanced_settings_default_values;
+        if (!experimentSettings?.enabled) {
+          return `s.advancedSettings && array(${subQuery.getQuery()})`;
+        }
```
idoshamun, line 87: I suggest changing this control flow. Instead of returning earlier as you did, we can update the subquery only if the experiment settings are enabled and then we will have a single return statement.

src/personalizedFeed.ts
```diff
@@ -23,7 +24,8 @@ async function fetchTinybirdFeed(
     url += `&user_id=${userId}`;
   }
   if (feedId) {
-    const filters = await feedToFilters(con, feedId);
+    const features = await fetchUserFeatures(feedId);
+    const filters = await feedToFilters(con, feedId, features);
```
idoshamun, line 28: I think there's a way to run these two in parallel if we do the merge between them on the server-side instead of the database side as you do now.
So from the database, we can fetch the defaults and the user settings and if the advanced settings defaults are enabled we can override. It will help us cut a few ms probably.
What do you think?

src/personalizedFeed.ts
```diff
@@ -23,7 +24,8 @@ async function fetchTinybirdFeed(
     url += `&user_id=${userId}`;
   }
   if (feedId) {
-    const filters = await feedToFilters(con, feedId);
+    const features = await fetchUserFeatures(feedId);
```
idoshamun, line 27:
```suggestion
    const features = await fetchUserFeatures(userId);
```

src/personalizedFeed.ts
idoshamun: Another suggestion would be to add the fetching of the flags into the feed filters function.
This will make sure that whenever we use this function we get the same results and keep it consistent.

idoshamun (changes requested):
I have some design concerns that I shared.
I also think we're missing some tests.
But overall that's a great work and I can't wait to merge

src/common/feedGenerator.ts
sshanzel: That's a good way to do it. I agree it will make the result much more predictable.

src/personalizedFeed.ts
sshanzel: Hmm, that's a very interesting idea. Though for us to merge, we will have to again run the same query for pulling all the sources to exclude. I think the Query will have to be called twice but I agree that it will be able to run in parallel which is also very fast and not wait for anything. Really interesting.

Initially, I applied the fetching of features to the feed filters but I realized our anonymous feed goes through here as well. It became inapplicable for some cases and just had to inject only if necessary.

idoshamun: Oh you are right! Then forget what I mentioned about making it parallel.

As for injecting the parameters, I see what you mean. So please make sure this parameter is a required parameter and not optional so we will inject it everywhere.
Because I didn't see you updated the GraphQL query that returns the defaults for the advanced settings

sshanzel: But it should be optional right? Since anonymous feed doesn't have a user id and it is the user id that we use to fetch our features (flagsmith) whether it is enabled for their account or not.

sshanzel:
Thank you for the review! Yes, I still have to work on its test cases. I just pushed some initial code, then I will also apply the overriding for returning the user's feed settings list

src/personalizedFeed.ts
idoshamun: We have a user id for every user even the not logged-in ones. We should use it for the anonymous feed as well.

sshanzel: Oh, cool. Then we would really have to make it necessary and be inside the feedToFilters function since this will be now used across all cases.

idoshamun (changes requested):
Please reconsider adding fetching the flags inside the `feedToFilters` for better consistency.
I know that we talked about it but now I don't see any reason why not to do it inside.
Please add a test that set multiple advanced settings to see that you handle it correctly. Each flag sets to a different value.
The GraphQL query `advancedSettings` doesn't use the feature flags to determine the default values. I think this must be fixed.

Have you tested it locally and checked that everything works?

sshanzel:
That makes total sense. Let's keep it in one place to make things much simpler. I'd also add a few more cases for the test as you have mentioned.

Yes, I mocked the expected values from the flagsmith and it worked. Then I also extracted the actual query and tested it locally and the result provided was accurate.

sshanzel:
Oh, and yeah, the settings should also be updated. I was so deep in making the tests turn green and I accidentally have left it out 😅

src/common/feedGenerator.ts
```diff
@@ -56,31 +57,52 @@ export const whereKeyword = (
...
+  const feature = features?.advanced_settings_default_values;
+  const excludedSettings = advancedSettings.filter((adv) => {
+    const hasNoSettings = userSettings[adv.id] === undefined;
+    const conditions = [
+      userSettings[adv.id],
+      feature?.value[adv.id],
+      !(hasNoSettings && adv.defaultEnabledState === false),
+    ];
+
+    return conditions.some((value) => value === false);
```
rebelchris, line 86: Is this the correct assumption?

- default: false
- feed: false
- user: true

The whole thing will return false?
Or does that not matter for this part?

src/common/feedGenerator.ts
```diff
@@ -56,31 +57,52 @@ export const whereKeyword = (
...
+  ]);
+  const userSettings = feedAdvancedSettings.reduce(
+    (obj, settings) => ({
+      ...obj,
+      [settings.advancedSettingsId]: settings.enabled,
+    }),
+    {},
+  );
+  const feature = features?.advanced_settings_default_values;
+  const excludedSettings = advancedSettings.filter((adv) => {
```
rebelchris, line 78: Just a global note:
Finding this whole function super hard to read, is there any way we can abstract it a bit? 

It's super hard to see what the 3 values are, and what we actually return from here.

rebelchris (changes requested):
Wow, amazing work!
Just a small question and a general readability remark.

👏👏👏

src/common/feedGenerator.ts
sshanzel: That should be the expected result, if there is any false it should automatically be included in the list.

src/common/feedGenerator.ts
sshanzel: If it was a bit hard to read for someone then it can be for the others as well. Let me find a way on how we can make this much simpler to read. Thanks for the note!

src/common/feedGenerator.ts
rebelchris: Oke great 👏

rebelchris (approved):
Thanks for that! Looks great 👏

sshanzel:
I have now made the necessary changes. By looking at the tests and snapshots, there are different scenarios for both default as true and false then overriding when there is no existing value for the user's personal settings.

Thank you both for the awesome reviews!

src/common/feedGenerator.ts
```diff
@@ -56,31 +57,53 @@ export const whereKeyword = (
   return `EXISTS${query}`;
 };
 
+export const getExcludedAdvancedSettings = async (
```
idoshamun, line 60: I think your previous implementation was better.
Now you introduce more database queries with no potential performance gain because we still have to wait for the features request.
Why did you introduce this change?

idoshamun (changes requested):
One step closer.
I saw that the snapshots for the `advancedSettings` query were updated but I don't see any change in the query in this PR. Did I miss it?

src/common/feedGenerator.ts
sshanzel: I realized an edge case with the previous implementation when the scenario has the `defaultEnabledState` as false and overriding it to true. In the previous first condition, it would meet the requirement which when applying the additional where conditions it may not meet the additional requirements but since we use `OR` it still became valid.

The condition has to be strict (based on priorities) and check each advanced settings available which led me to merge through the backend.

sshanzel:
Oh, man. I was so deep into the feed. Let me pull up one query that grabs the settings as well.

src/common/feedGenerator.ts
idoshamun: Ok, let's let it be for now. Once we finish the experiment we may think of other ways :)

idoshamun (approved):
Woohoo!

PR #830, opened 2021-12-14
fix: missing headers for fetching features

sshanzel:
The function used by the changes provided:
```
const authorizedHeaders = (userId: string): { [key: string]: string } => ({
  authorization: `Service ${process.env.GATEWAY_SECRET}`,
  'user-id': userId,
  'logged-in': 'true',
});
```


idoshamun (approved):
Let's merge it!

PR #832, opened 2021-12-16
fix: renamed openSidebar to sidebarExpanded

rebelchris:
After discussing this with @sshanzel we decided to be more explicit about this setting.
What is does is either show the expanded sidebar or the minimised function.

Hence we renamed `openSidebar` to `sidebarExpanded`


src/schema/settings.ts
```diff
@@ -82,7 +82,7 @@ export const typeDefs = /* GraphQL */ `
     """
     Whether to show the sidebar open
```
sshanzel, line 83:
```suggestion
    Whether to show the sidebar in expanded form
```

src/schema/settings.ts
```diff
@@ -134,7 +134,7 @@ export const typeDefs = /* GraphQL */ `
...
```
sshanzel, line 135:
```suggestion
    Whether to show the sidebar in expanded form
```

sshanzel (approved):
Wow, thought you were just gonna apply an alias in the FE but you brought it to the next level 💯 

All looks good to me, just adjusted the definitions as well.

idoshamun (approved):
Make sure to resolve Lee's suggestion first but it looks good!

PR #834, opened 2021-12-17
feat: publishing user settings to other services

sshanzel:
DD-337

After further checking, it seems below properties are now used anymore:
- appInsaneMode - default = true
- enableCardAnimations - default = true

We have decided to remove these ones but keep a stub value for compatibility purposes.

~~**Question:** has anyone experienced having the below tests fail even though they are exactly the same assertion of the `alerts` feature?
Though I have manually checked by logging if it passes through the desired functions and it displays the log messages, but the assertion still fails.~~


sshanzel:
@rebelchris thanks for taking the time of having the debug session! Finally, the tests have turned green 🟢

src/common/pubsub.ts
```diff
@@ -133,10 +134,35 @@ export const notifyUserReputationUpdated = async (
...
+];
+
+export const notifySettingsUpdated = (
+  log: EventLogger,
+  settings: ChangeObject<Settings>,
+): Promise<void> =>
+  publishEvent(
+    log,
+    settingsUpdatedTopic,
+    removeSomeProps(settings, excludedSettingsProps),
```
idoshamun, line 164: Why do you remove the `userId` and `updatedAt` props from the event?

src/common/pubsub.ts
```diff
@@ -133,10 +134,35 @@ export const notifyUserReputationUpdated = async (
...
+
+    return { ...result, [key]: obj[key] };
+  }, {});
+
+export const notifyAlertsUpdated = (
   log: EventLogger,
   alerts: ChangeObject<Alerts>,
-): Promise<void> => publishEvent(log, alertsFilterUpdatedTopic, alerts);
+): Promise<void> =>
+  publishEvent(log, alertsUpdatedTopic, removeSomeProps(alerts, ['userId']));
```
idoshamun, line 150: Why do you remove `userId`? It's bad practice removing props of existing topics. It might break the subscribers.

src/workers/cdc.ts
```diff
@@ -161,6 +164,35 @@ const onAlertsChange = async (
...
+  _: Connection,
+  logger: FastifyLoggerInstance,
+  data: ChangeMessage<Settings>,
+): Promise<void> => {
+  if (data.payload.op === 'u') {
+    const hasChanged = hasAnyChanged(
+      data.payload.before,
+      data.payload.after,
+      excludedSettingsProps,
+    );
```
idoshamun, line 186: I think this is a bit redundant because practically `userId` will never be updated and `updatedAt` will be updated only when another field is being updated. So you can just fire the event on any update operation

idoshamun (changes requested):
Please apply some changes but great catch with add the settings table to the allow list and enabling full replica

src/common/pubsub.ts
sshanzel: I was thinking that we always embed the `userId` in our message, but I agree it is not a good idea to reshape existing messages.

src/common/pubsub.ts
sshanzel: Was thinking it won't be too much of use for the other end. But I realized it should be the received end's concern if they will not process some of the data. Let me fix this one.

src/common/pubsub.ts
idoshamun: And also `userId` is not embedded anywhere else if you don't provide it in the payload

idoshamun (approved):
LGTM! 🚢

src/workers/cdc.ts
sshanzel: Oh, yes. I just checked. My mistake.

PR #843, opened 2021-12-31
feat: introducing a new alert for the my feed message

rebelchris:
This PR introduces the needed changes for the API to handle the new `myFeed` alert.
This alert has three states:

`default` (also the default state) means the user has already filtered stuff and is applicable to most people
`manual` this state should happen once the user click the `Create` button for the My Feed (Awaiting another PR to implement this)
`NULL` The user clicked on the cross button to hide this alert message.

Gateway changes: https://github.com/dailydotdev/daily-gateway/pull/403
Frontend changes: https://github.com/dailydotdev/apps/pull/786


src/entity/Alerts.ts
```diff
@@ -11,9 +13,13 @@ export class Alerts {
 
   @Column({ type: 'timestamp without time zone', default: null })
   rankLastSeen: Date | null;
+
+  @Column({ type: 'enum', enum: ['default', 'manual'], default: 'default' })
```
idoshamun, line 17: According to my experience enums can be very strict and changing them requires some hassle.
My suggestion is to go with a string or number.
We can define a TypeScript enum that translates into a number (like every enum) and we will keep the number in the database. Whatever you want I just want it to be dynamic enough.

idoshamun (changes requested):
This looks great but I have a few comments:
* The name `default` doesn't describe well the state, I think. Because actually this default state can be applied only on existing users so far. But once we release the "my feed" feature every new user will be assigned to `manual`. Meaning that in a year from now probably more users will go through the `manual` state than the `default`. So I think a better naming is required here.
* I don't see the logic that turns this new alert into the `manual` state when the user is created and they don't have a customized feed.
* We need to set the correct value for existing users in the migration file. If they have a customized feed, it should be `default`, otherwise `null`.
* I think that we should the alert default value to `manual` so when a new user is created it will automatically set to the correct value. And only a subset of users should be assigned to what is now called `default`.

What do you think?

rebelchris:
@idoshamun Thanks for these questions.

Regarding the name, agreed. At first I had `auto` because it was "auto" added by us. Not sure if that's better?

The manual logic can only be added once we introduce the actual `create` button, which is not built yet.

The idea behind the `default` as the default value is that it is a fail safe.
We have no way of detecting this on a action so setting it to `NULL` might actually cause glitches.

Since we will be adding the `manual` on user action more people will actually follow this flow.

But maybe you are right, we could set `manual` as the default, and run a migration for all current users to set it to `default` (or whatever the name is going to be)

src/entity/Alerts.ts
rebelchris: Ok that makes total sense, will work on that change

idoshamun:
> Regarding the name, agreed. At first I had `auto` because it was "auto" added by us. Not sure if that's better?

Maybe something like "migrated", "moved"? And following the same logic "manual" can turn into "created".

> The manual logic can only be added once we introduce the actual `create` button, which is not built yet.

Why? At this point, we can return whatever value we want in the new alert and the app will just ignore it because it isn't aware of its existence. My suggestion is to have this decision tree in the app:
Is filter alert on? Show the create my feed widget.
No, is my feed alert set? Show the my feed button along with the relevant balloon.
No? Just show my feed button with nothing else.

In this case even if by default we will set the myfeed alert to `created` (or `manual`), the balloon will be visible only once the process is done. And then once the user dismisses the balloon it will be set to null.
What do you think?

rebelchris:
@idoshamun In this new scenario where we change the "default" set to `manual` it makes sense indeed!
In this current PR however it was not possible to do it like that.
Let me work on these changes and we'll take it from there.

I think your flow describes is indeed what we'll go for :D

sshanzel:
I know we've discussed this before, but I think instead of using a string, we should instead use a bool since the only missing case we can't cover with the `alert.filter` is for the newly created feeds. What do you think of just creating `newlyFiltered` `bool`. With this, we can easily catch the cases.
- if `alert` = `true` - check if `newlyFiltered` = `true` then show message for newly filtered else the migration message.
- Then the case for creating new feed would be if it is anonymous user of if `hasAnyFilter` = `false`.
- alert = false nothing to display at all.

idoshamun:
I think this introduces a complex decision logic that it's harder to follow.
I'd stick to a simpler solution, I don't mind losing a few more bits for the string.
Thanks for the suggestion though :)

rebelchris:
@sshanzel and @idoshamun 

I've reworked this PR to switch the behaviour, also updated the DR states here:
https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/126189571/DD-354+My+Feed

Decided to keep this value as a string so we can introduce new states later on without having to worry about it.

The new flow:
default value = `created` meaning the user created their feed after we launch the my feed.
added a custom script we should run once we go live, this will set everyone who has a filter already on to `migrated`.

bin/migratedMyFeedAlert.ts
```diff
@@ -0,0 +1,35 @@
+import '../src/config';
```
idoshamun, line 1: I think it's better to turn this functionality into a single upsert statement in SQL and run it as part of the migration. This would be the seamless solution for this.

idoshamun (changes requested):
By the far the simplest solution. I do have one comment below that will make it even better.

rebelchris:
@idoshamun Great suggestion making it even simpler.

As you can see when running the migration:
people with filter = TRUE get the default `created`
people who have filter = FALSE get `migrated`

Screenshot 2022-01-03 at 14 34 23

idoshamun (approved):
💪

sshanzel (approved):
LGTM! 🚀

PR #846, opened 2022-01-05
feat: add new route to add posts

idoshamun:
The new route will be used by our new Gatekeeper service to push new content to the database.
Currently, adding new articles is supported only through Pub/Sub and the Gatekeeper requires a synchronous solution so Pub/Sub is not suited for this.

### DIFF
--- a/src/entity/Post.ts
+++ b/src/entity/Post.ts
@@ -2,15 +2,20 @@ import {
   Column,
   Connection,
   Entity,
+  EntityManager,
+  In,
   Index,
   ManyToOne,
   OneToMany,
   PrimaryColumn,
 } from 'typeorm';
+import shortid from 'shortid';
+import * as he from 'he';
 import { PostTag } from './PostTag';
 import { Source } from './Source';
 import { User } from './User';
 import { PostKeyword } from './PostKeyword';
+import { Keyword } from './Keyword';
 
 export type TocItem = { text: string; id?: string; children?: TocItem[] };
 export type Toc = TocItem[];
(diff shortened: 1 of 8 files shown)


src/entity/Post.ts
```diff
@@ -179,3 +184,238 @@ export const getAuthorPostStats = (
...
+      'url = :url or url = :canonicalUrl or "canonicalUrl" = :url or "canonicalUrl" = :canonicalUrl',
+      { url: data.url, canonicalUrl: data.canonicalUrl },
+    )
+    .getRawOne();
+  if (p) {
+    return 'exists';
+  }
+  if (data.creatorTwitter === '@NewGenDeveloper') {
+    return 'author banned';
+  }
```
rebelchris, line 248: Is this now a weird way of checking for banned authors? 
Perhaps we could keep track of this via an array definition so it will be easier to maintain in the future?

src/entity/Post.ts
```diff
@@ -179,3 +184,238 @@ export const getAuthorPostStats = (
...
+  keywords?: string[],
+): Promise<{ mergedKeywords: string[]; allowedKeywords: string[] }> => {
+  if (keywords?.length) {
+    const synonymKeywords = await entityManager.getRepository(Keyword).find({
+      where: {
+        status: 'synonym',
+        value: In(keywords),
+      },
+    });
+    const mergedKeywords = Array.from(
```
rebelchris, line 280: In the apps we refactor something similar to more named version.
So the Array.from new Set could be something like uniqueRecords.

Not urgent, but would maybe be good to generalize these things along our code bases.
https://github.com/dailydotdev/apps/blob/209ad77c5801c8475bfb0c38cfafc4c683acade1/packages/eslint-rules/lib/rules/no-custom-color.js#L44

rebelchris (approved):
Nice work, just some small non urgent things I've noticed.

src/entity/Post.ts
idoshamun: This was a temporary fix for someone who kept spamming our feed with low quality.
But I agree for now I moved it into an array and later we will move it to a database

src/entity/Post.ts
idoshamun: Great idea!

PR #848, opened 2022-01-07
feat: :sparkles: Added a new hook to post to slack on every comment

rebelchris:
I've added a new slack hook to act on every comment.
For now this will post in the same channel but with out avo green color.

I haven't been able to test end-to-end because I don't know the new webhook url.

The test and local CDC all go well though.

DD-326 #done


idoshamun (changes requested):
The best practice here is not to use CDC but rather create a new subscription that only sends the slack notification. The reason is that we want to optimally run exactly one side-effect for every message we handle. Otherwise, the retry mechanism can be a mess. Imagine you sent the slack notification here and then for some reason something failed down the line. It will cause Pub/Sub to send again the same message and basically flood our Slack channel. This is something called distributed transaction in distributed systems. The best way to handle it, is to bypass the need for such transactions by introducing multiple subscriptions that listens to the same domain event. In our case the domain event is pos commented. The CDC topic is not a domain event but rather something technical that we use to generate domain events out of it.

rebelchris:
@idoshamun Thanks, that makes sense, I'll refactor to it'w own subscription.
I did see the report webhook runs over the CDC, that's still fine then?

rebelchris:
@idoshamun Refactored to two new subscriptions one for the commentComment and one for postComment.
It seems you cannot re-use the same subscription on the same topic, hence I made two.

Looking forward to hearing what you think.

infra/workers.ts
```diff
@@ -58,6 +58,10 @@ export const workers: Worker[] = [
     topic: 'comment-commented',
     subscription: 'comment-commented-author-mail',
   },
+  {
+    topic: 'post-commented',
```
idoshamun, line 62: 👀 

```suggestion
    topic: 'comment-commented',
```

infra/workers.ts
```diff
@@ -58,6 +58,10 @@ export const workers: Worker[] = [
...
+    subscription: 'comment-comment-slack-message',
```
idoshamun, line 63:
```suggestion
    subscription: 'comment-commented-slack-message',
```

infra/workers.ts
```diff
@@ -86,6 +90,10 @@ export const workers: Worker[] = [
     topic: 'post-commented',
     subscription: 'post-commented-redis',
   },
+  {
+    topic: 'post-commented',
+    subscription: 'post-comment-slack-message',
```
idoshamun, line 95:
```suggestion
    subscription: 'post-commented-slack-message',
```

src/workers/commentCommentSlackMessage.ts
```diff
@@ -0,0 +1,48 @@
...
+import { notifyNewComment } from '../common';
+
+interface Data {
+  userId: string;
+  childCommentId: string;
+  postId: string;
+}
+
+const worker: Worker = {
+  subscription: 'comment-comment-slack-message',
```
idoshamun, line 12:
```suggestion
  subscription: 'comment-commented-slack-message',
```

src/workers/postCommentSlackMessage.ts
```diff
@@ -0,0 +1,48 @@
...
+import { notifyNewComment } from '../common';
+
+interface Data {
+  userId: string;
+  commentId: string;
+  postId: string;
+}
+
+const worker: Worker = {
+  subscription: 'post-comment-slack-message',
```
idoshamun, line 12:
```suggestion
  subscription: 'post-commented-slack-message',
```

idoshamun (changes requested):
Other than naming concerns I'm speechless.
Way to go :)

idoshamun (approved):
Looks great!

sshanzel (approved):
So cool to see we can now confidently say the CDC pipeline works. Well done! 🚢

PR #853, opened 2022-01-13
fix: anonymous feed support for logged in users

idoshamun:
Feed v2 introduced a bug where if the user is logged-in then it will automatically fetch the feed with the filters. This PR fixes the issue and adds support for anonymous feed for logged-in users.


rebelchris (approved):
Nice one

PR #860, opened 2022-01-20
refactor: custom links as part of settings entity

sshanzel:
DD-85 #done

Custom links column to store user provided shortcut links.


rebelchris (approved):
Looks nice.

I would just suggest to add a test for the following scenario.

1 valid and 1 invalid link

This makes sure the the combi also fails and arrays are accepted.

sshanzel:
@rebelchris good point! I have now added the test case for having both valid and invalid links in the same mutation. Thank you for the awesome review!

src/common/links.ts
```diff
@@ -1,6 +1,14 @@
 export const getDiscussionLink = (postId: string): string =>
   `${process.env.COMMENTS_PREFIX}/posts/${postId}`;
 
+const isRegexValidURL = (url: string) => {
```
idoshamun, line 4: Why do we need both `isValidHttpUrl` and `isRegexValidUrl`?

src/schema/settings.ts
```diff
@@ -160,33 +178,56 @@ export const typeDefs = /* GraphQL */ `
...
+};
+
+const updateUserSettings = async (
+  data: GQLUpdateSettingsInput,
+  ctx: Context,
+) => {
+  const repo = ctx.getRepository(Settings);
+  const settings = await getOrCreateSettings(ctx);
+
+  return repo.save(repo.merge(settings, data));
```
idoshamun, line 199: It's better to use a transaction here I think

src/schema/settings.ts
```diff
@@ -160,33 +178,56 @@ export const typeDefs = /* GraphQL */ `
...
+    updateCustomLinks: async (
+      _,
+      { links }: { links: string[] },
+      ctx,
     ): Promise<GQLSettings> => {
-      const repo = ctx.getRepository(Settings);
-      const settings = await repo.findOne(ctx.userId);
-      if (!settings) {
-        return repo.save(repo.merge(repo.create(data), { userId: ctx.userId }));
+      if (!links || !links.length) {
```
idoshamun, line 215:
```suggestion
      if (!links?.length) {
```

idoshamun (changes requested):
Why don't we use the existing mutation to update the settings? What's unique in the custom links use case?
One more thing we have to address, every time we add a new setting/alert we need to add it in here as well: 
https://github.com/dailydotdev/daily-gateway/blob/master/src/integration.js#L32
We might want to create dedicated routes that immediately return all the fields, something like our compatibility routes but outside the compat directory because the routes there are legacy.

src/common/links.ts
sshanzel: These are two ways to cover the validity of URLs. The `isValidHttpUrl` is the simplest way to validate but it fails in the case of `http://abc` which is not an actual valid URL. That case is caught by the regex.

I am using both just to be sure all edge-cases are covered

sshanzel:
@idoshamun thank you for the awesome review! The reason behind creating a separate mutation is to validate the URLs sent by the user so we won't have to validate it when we want to update just a single property in the settings. But it makes sense if we want to run the validation every time when utilizing the existing settings mutation.

sshanzel:
We could probably utilize the existing fastify route for settings since it already returns all the props needed.

src/common/links.ts
idoshamun: This is kind of confusing. Future devs will not be able to understand it by name. Also, I wouldn't go into too much hassle of validating these URLs for this purpose.

sshanzel: Cool. Then let me remove the validation through regex.

idoshamun:
> @idoshamun thank you for the awesome review! The reason behind creating a separate mutation is to validate the URLs sent by the user so we won't have to validate it when we want to update just a single property in the settings. 

We can validate the URLs only if we receive them from the client, otherwise, it's not necessary. Also, it should be pretty fast so I wouldn't worry about it.

> Also, we set the links to null if the param is empty or undefined and if the user just updates a single property it would automatically set the links to null even when unnecessary.

Instead of relying on empty or undefined input, we can rely on an empty array. This is a much more explicit way to nullify something. I do think we can stick to one mutation and there's no need for another one.

> We could probably utilize the existing fastify route for settings since it already returns all the props needed.

The existing one is part of the compatibility suite and kind of boxed and should not be touched, otherwise, we may break old versions. Or maybe we can take it out of the compatibility suite and put it into the regularly maintained routes. But we have to check if someone uses this endpoint or not. Please let me know how you'd like to proceed and I can do this check.

idoshamun (approved):
Looking good!

PR #863, opened 2022-01-26
fix: timezone was not actively used for the reading history overview

rebelchris:
We introduced the timezone for all reading history queries, but not the actual overview.
This fix makes sure we convert the timezone to the user's set timezone.

DD-497 #done


sshanzel (approved):
🙌

idoshamun (approved):
Perfect!

PR #864, opened 2022-01-27
fix: reading history was failing on users with no timezone

rebelchris:
In the way I was loading the user before it would fail if the user didn't have a timezone set.
Meaning they wouldn't be able to load their reading history.

After fixing this issue it also seems that the removing started working locally again.
Will have to double check on merge.

DD-498 #done


sshanzel (approved):
💪

idoshamun (approved):

PR #865, opened 2022-01-27
fix: fix for the hiding of the read history

rebelchris:
Since we introduced the timezone in a local timezone for that specific user we needed to also send the database timezone as we use this to hide the view.

See related changes on frontend here:

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -36,6 +36,7 @@ export interface GQLUser {
 export interface GQLView {
   post: Post;
   timestamp: Date;
+  timestamp_db: Date;
 }
 
 type CommentStats = { numComments: number; numCommentUpvotes: number };
@@ -142,6 +143,7 @@ export const typeDefs = /* GraphQL */ `
 
   type ReadingHistory {
     timestamp: DateTime!
+    timestamp_db: DateTime!
     post: Post!
   }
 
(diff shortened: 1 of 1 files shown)


src/schema/users.ts
```diff
@@ -36,6 +36,7 @@ export interface GQLUser {
 export interface GQLView {
   post: Post;
   timestamp: Date;
+  timestamp_db: Date;
```
idoshamun, line 39: Please use camelCase

src/schema/users.ts
```diff
@@ -142,6 +143,7 @@ export const typeDefs = /* GraphQL */ `
 
   type ReadingHistory {
     timestamp: DateTime!
+    timestamp_db: DateTime!
```
idoshamun, line 146: Please use camelCase

idoshamun (changes requested):

idoshamun:
Please also make sure to add relevant tests / update existing ones

idoshamun:
@rebelchris but you want to make sure the new field returns the right value, don't you?

idoshamun (approved):

rebelchris:
@idoshamun Good point, I don't think we test anything against timezone as it would change over time with DST etc.
Do we want to introduce test regarding to that fact?

idoshamun:
I think it's better to have something, just in case

rebelchris:
@idoshamun Added specific tests for the timezoned users (and one for non)
Let me know what you think

idoshamun (approved):
Looks great!

PR #867, opened 2022-01-28
refactor: removal of advanced settings experiment

sshanzel:
DD-500 #done

Most of the time when we finished an experiment, we don't necessarily dispose of the additional block of codes that integrates the features flags.

In this case, it impacts the performance is at would request to the gateway's cache for the features flags even when it is not needed anymore since the experiment has finished.


rebelchris (approved):
Nice work Lee 👍

idoshamun (changes requested):
I want to keep the support in the code. So it can be either using an environment variable that I can turn on/off or even just comment out the relevant sections. But I do want to have this code available in case we want to introduce again

sshanzel:
I am also in with using an env variable - what do you think about a query parameter? Named `enableSettingsExperiment`?

idoshamun:
@sshanzel sounds like a proper naming :)

src/common/feedGenerator.ts
```diff
@@ -113,32 +113,65 @@ export const getExcludedAdvancedSettings = async (
...
+          'fas',
+          'adv.id = fas."advancedSettingsId" AND fas."feedId" = :feedId',
+          { feedId },
+        )
+        .where('COALESCE(fas.enabled, adv."defaultEnabledState") = false')
+        .getQuery();
+
+      return `s."advancedSettings" && array(${subQuery})`;
+    });
+  }
```
idoshamun, line 150: I think it's better to use the same implementation but in one case take into consideration the flags and on the other not. It will make it more consistent and predictable. Easier to maintain as well

idoshamun (changes requested):
I thought we agreed that `enableSettingsExperiment` will be an environment variable not a query parameter. Meaning a machine environment variable. By default it will be off unless I set this environment variable on.
What do you think? I don't want to rely on the client for this

src/common/feedGenerator.ts
idoshamun: I know that your original concern was performance but for now I prefer simplicity over performance. If we see that it still remains as a bottleneck we will fix it :)

idoshamun (approved):
Slick!

PR #868, opened 2022-01-30
feat: db seeds for category and advanced settings

sshanzel:
DD-262 #done

To improve the DevEx, we wanted to provide all the data that might be required to push through development.

Have tested and validated locally by running `db:migrate:reset` then `db:seed:import` - No error occurred and the values in the DB have reflected.


idoshamun (approved):
That's a great addition! Thanks a lot

PR #871, opened 2022-01-31
feat: public bookmarks with rss support

idoshamun:
This PR adds support in sharing bookmarks using RSS.
It adds a new GraphQL query for receiving the sharing settings, `bookmarksSharing`. A mutation for manipulating the settings `setBookmarksSharing`. And revives the old RSS functionality that we had for the premium users with a slight change.

DD-515 #done
DD-516 #done
DD-518 #done


rebelchris (approved):
Not much to add, really well done.
Do we want to keep the commented out tests?

idoshamun:
@rebelchris I can remove them if it makes more sense. Thought I'd leave them because this time I actually used some of it. But I can do either. Anyway I'll check the tests tomorrow and fix it

PR #872, opened 2022-02-01
refactor: user reading rank with tags

sshanzel:
DD-503 #done

@idoshamun thank you for the query first of all. I just now refactored the condition to match the requirements in this specific query.


__tests__/users.ts
```diff
@@ -330,6 +354,74 @@ describe('query userReadingRank', () => {
...
+    });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.userReadingRank).toMatchSnapshot({
+      readToday: expect.anything(),
+      lastReadTime: expect.anything(),
+    });
+    const { tags } = res.data.userReadingRank;
+    expect(tags.length).toEqual(8);
+    const sum = tags.reduce((total, { readingDays }) => total + readingDays, 0);
+    expect(sum).toEqual(12);
```
idoshamun, line 422: Why this is needed if we match a snapshot?

src/common/users.ts
```diff
@@ -97,16 +104,50 @@ const rankFromProgress = (progress: number) => {
...
+  con: Connection,
+  { userId, timezone = 'utc', limit = 8 }: ReadingDaysArgs,
+) => {
+  const now = `timezone('${timezone}', now())`;
+  return con.query(
+    `
+    with filtered_view as (
+      select *, CAST(v."timestamp" at time zone '${timezone}' AS DATE) as day
+      from "view" v
+      where "userId" = $1 and v."timestamp" at time zone '${timezone}' >= date_trunc('week', ${now})
```
idoshamun, line 123: I think it's better to support a custom time frame here. This way we can reuse it for other purposes as well

idoshamun (changes requested):
Please make sure to use the new implementation of the tags for the DevCard as well.
Can you update it as part of this PR, please?

src/common/users.ts
rebelchris: Agreed would be better to support the time frame, so we can use it on the profile page as well.

rebelchris (approved):
Would love to see the tags based on a variable time frame.
Otherwise nothing to add, really well done 👀

rebelchris (changes requested):
Would love to see the tags based on a variable time frame.
Otherwise nothing to add, really well done 👀

__tests__/users.ts
sshanzel: Ah, yes. The snapshot overkills it. Let me keep the computation to make it easier to understand than looking at the snapshots.

sshanzel:
Thank you both for the reviews! The function now accepts the start and end dates to have better handling of time frames.

rebelchris (approved):
Looks good to me 🥳

src/common/users.ts
```diff
@@ -121,16 +169,37 @@ export const getUserReadingRank = async (
...
   return {
-    lastReadTime: res.lastReadTime,
+    lastReadTime: lastReadTime,
     currentRank: rankThisWeek > rankLastWeek ? rankThisWeek : rankLastWeek,
-    progressThisWeek: res.thisWeek,
+    progressThisWeek: thisWeek,
     rankLastWeek,
     rankThisWeek,
-    readToday: isSameDay(res.lastReadTime, new Date()),
+    readToday: isSameDay(lastReadTime, new Date()),
```
idoshamun, line 202: You can take the `now` definition outside of the function scope and utilize here as well :)

src/common/users.ts
```diff
@@ -121,16 +169,37 @@ export const getUserReadingRank = async (
...
+
+  const [{ thisWeek, lastWeek, lastReadTime }, tags] = await Promise.all([
+    req.getRawOne<ReadingRankQueryResult>(),
+    getReadingTags(),
+  ]);
+  const rankThisWeek = rankFromProgress(thisWeek);
+  const rankLastWeek = rankFromProgress(lastWeek);
   return {
-    lastReadTime: res.lastReadTime,
+    lastReadTime: lastReadTime,
```
idoshamun, line 197:
```suggestion
    lastReadTime,
```

src/schema/users.ts
```diff
@@ -169,12 +184,20 @@ export const typeDefs = /* GraphQL */ `
...
     """
     userMostReadTags(id: ID!): [MostReadTag]
     """
+    Get the reading top tags of the user
+    """
+    userReadingTopTags(
+      id: ID!
+      before: String!
+      after: String!
+    ): [TagsReadingStatus]
```
idoshamun, line 199: What is this new query and what's the difference compared to `userMostReadTags`?

idoshamun (changes requested):
The PR looks great! I love how you approached it.
A few comments

src/schema/users.ts
sshanzel: This has the percentage - should we keep them in one query?

src/common/users.ts
```diff
@@ -97,16 +104,57 @@ const rankFromProgress = (progress: number) => {
   return 0;
 };
 
+type DateRange = { start: string; end: string };
```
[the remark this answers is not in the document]
sshanzel: In our existing implementation of a date range - I noticed we utilized string type for the date params - to keep them consistent as the 3 of them will have the same parameters sent in one batched query.

src/schema/users.ts
idoshamun: Yes, please. Otherwise, it might be confusing

src/common/users.ts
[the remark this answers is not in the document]
idoshamun: Oh, ok. Makes sense :)

src/common/devcard.ts
```diff
@@ -12,24 +12,24 @@ export interface MostReadTag {
...
-    .innerJoin(PostKeyword, 'pk', 'v."postId" = pk."postId"')
-    .where('v."userId" = :id', { id: userId })
-    .andWhere(`pk.status = 'allow'`)
-    .andWhere(`pk.keyword != 'general-programming'`)
-    .groupBy('pk.keyword')
-    .orderBy('2', 'DESC')
-    .limit(limit)
-    .getRawMany();
+): Promise<MostReadTag[]> => {
+  const start = new Date(0).toISOString();
```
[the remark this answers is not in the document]
sshanzel: Ah, yes, the specific condition.Let me work that up

sshanzel:
@idoshamun - have removed the unnecessary query now and modified the shape of the return type. Thank you for the review!

src/common/devcard.ts
```diff
@@ -76,10 +70,12 @@ export async function getDevCardData(
   userId: string,
   con: Connection,
 ): Promise<DevCardData> {
+  const start = new Date(START_OF_DEVCARD).toISOString();
```
idoshamun, line 73: This is not what I meant actually. I meant that we should subtract a year from now and use it as the start date.
Because by default when you go to the profile, it shows the information of last year (365 days). So I want it to be aligned with the profile.

src/schema/users.ts
```diff
@@ -335,7 +330,7 @@ export const resolvers: IResolvers<any, Context> = {
                       from "view"
                       join "user" on "user".id = view."userId"
                       where "userId" = $1
-                        and "timestamp" >= '2020-12-14'
+                        and "timestamp" >= '${START_OF_DEVCARD}'
```
idoshamun, line 333: Not needed

src/schema/users.ts
```diff
@@ -296,24 +295,20 @@ export const resolvers: IResolvers<any, Context> = {
...
-        dateRange: { start: after, end: before },
-      });
-    },
     userMostReadTags: async (
       _,
-      { id }: { id: string },
+      { id, before, after, limit = 5 }: ReadingHistyoryArgs,
       ctx: Context,
     ): Promise<GQLMostReadTag[]> => {
+      const start = after ?? new Date(START_OF_DEVCARD).toISOString();
```
idoshamun, line 303: If after is not defined we can indeed use `Date(0)`

idoshamun (changes requested):

src/common/devcard.ts
sshanzel: Ah, I see. Cool. Thanks for the clarification, my bad. Let me work it quickly!

src/schema/users.ts
```diff
@@ -330,7 +330,7 @@ export const resolvers: IResolvers<any, Context> = {
                       from "view"
                       join "user" on "user".id = view."userId"
                       where "userId" = $1
-                        and "timestamp" >= '${START_OF_DEVCARD}'
+                        and "timestamp" >= '2020-12-14'
```
idoshamun, line 333: This is not needed anymore

idoshamun (changes requested):
One comment and we can merge. You probably missed it in my last review

src/schema/users.ts
sshanzel: Oh, cool. Let me remove it.

sshanzel:
Thank you! Thought the refactor was the only thing not needed. I have completely removed the condition now.

idoshamun (approved):
Woohoo! Let's ship

PR #873, opened 2022-02-01
feat: add opt out from weekly goal setting

idoshamun:
Add a new user setting to opt out from the weekly goal.

DD-505 #done


rebelchris (approved):
Nice work 🥳

PR #874, opened 2022-02-02
refactor: tags limit on reading rank

sshanzel:
The `userReadingRank` now accepts a `limit` param for the number of tags to return but the default value is 6.


rebelchris (approved):
Nice one

PR #877, opened 2022-02-07
fix: banned or deleted post should not be in history

sshanzel:
DD-528

This fixes the banned and deleted posts being shown in the user's history.


idoshamun (changes requested):
Looks great! The only comment is that banned posts should be kept in the reading history. We need to remove only the deleted ones

sshanzel:
Woop! Woop! Updated the query now. Thank you! 🙇  @idoshamun

rebelchris (approved):
Nice one!

idoshamun (approved):
Yes!!!

PR #883, opened 2022-02-14
fix: use the correct timezone for calculating tags

idoshamun:
Before, we use the functions `startOfWeek` and `endOfWeek` to calculate the date range for the user tags reading rank.
These functions don't take into account the user's timezone.
This fix uses postgres timezone function before truncating to week.

DD-547 #done

### DIFF
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -1,5 +1,5 @@
 import { IFlags } from 'flagsmith-nodejs';
-import { endOfWeek, isSameDay, startOfWeek } from 'date-fns';
+import { isSameDay } from 'date-fns';
 import fetch from 'node-fetch';
 import { Connection } from 'typeorm';
 import { View } from '../entity';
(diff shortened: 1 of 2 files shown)


src/common/users.ts
```diff
@@ -113,35 +113,44 @@ export interface ReadingDaysArgs {
...
+      on u."id" = v."userId"
+
+      where u."id" = $1
+        and "timestamp" >= date_trunc('week'
+          , timezone(${timezone}
+          , $2))
+        and "timestamp"
+          < date_trunc('week'
+          , timezone(${timezone}
+          , $3)) + interval '7 days'
```
rebelchris, line 137: Does this not break the year function on the profile page?

src/common/users.ts
```diff
@@ -180,10 +189,10 @@ export const getUserReadingRank = async (
       return Promise.resolve(null);
     }
 
-    const start = new Date(startOfWeek(now).getTime()).toISOString();
-    const end = new Date(endOfWeek(now).getTime()).toISOString();
+    const start = now.toISOString();
+    const end = start;
```
rebelchris, line 193: Since these are now the same we might want to change the naming to `timezone` I think this will become very confusing in the future.
What do you think?

Subnote about the above one if it doesn't break the profile implementation.

rebelchris (changes requested):
Some questions around the profile use-case

src/common/users.ts
idoshamun: When I did the manual QA I mentioned that it's already broken because it doesn't take into account the full week. So I thought I fixed that now but I was confused with the reading rank. It's not relevant for the tags. I'll have to look for another solution.
What bothers me actually the most is that we have several ways in the same function to calculate the date range. I'm not a fan and we might want to address it in the near future as it can cause misalignments between the different queries

rebelchris: Good point we should actually combine these date based queries into one generic function that takes the timezone into account.

PR #889, opened 2022-02-16
fix: In some cases the timezone was actually negative

rebelchris:
Let's take the following data as an example:
Screenshot 2022-02-16 at 14 20 51

In the current live version (double timezone) we actually reset the timezone back to UTC
Screenshot 2022-02-16 at 14 22 16
 
When only using one timezone conversion it shows the time in that users zone:
Screenshot 2022-02-16 at 14 21 59

Let me know what you think.
I left the weekly rank, as this is based on the users week, which might start on a different day.

DD-550 #done

### DIFF
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -154,7 +154,7 @@ export const getUserReadingRank = async (
   version = 1,
   limit = 6,
 ): Promise<ReadingRank> => {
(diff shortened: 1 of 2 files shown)


idoshamun (approved):
I don't want to see this timezone issues anymore 😓

PR #890, opened 2022-02-16
feat: mention user in comments

sshanzel:
DD-469

The only missing piece now is the background worker which we will whitelist the new table in our CDC so whenever a new entry is created, we will send out an email to properly notify the user.

Regarding the user mention, after checking out available plugins for `markdown-it`, I've seen there isn't much to choose from, I found one but it has a different implementation in the frontend, then I also found something simple that could've been used. Even if we try to use the latter, we will have to do customization on its renderer to apply a custom attribute for our tooltip to utilize and since customization will have to happen, I have decided to generate the link tag by ourselves as it was fairly simple.

I utilized the 2nd object parameter of the `render` method mentioned in the docs as it can be used to pass additional data.
https://markdown-it.github.io/markdown-it/#MarkdownIt.render

Key features:
- [x] query to fetch recommended users.
- [x] store data for mentioned users
- [x] markdown to apply element on the mentioned user as part of the content
- [x] notify mentioned users using a background worker
- [x] when a user changes username, all comments should be updated where the user was mentioned


sshanzel:
Testing GitHub's mention @Rebelchris

src/common/markdown.ts
```diff
@@ -15,6 +16,9 @@ export const markdown: MarkdownIt = MarkdownIt({
   },
 });
 
+export const getMentionLink = (username: string): string =>
+  `<a href="${WEBAPP_URL}/${username}" target="_blank" data-mention-username="${username}">${username}</a>`;
```
idoshamun, line 20: I think we have a function that returns the user profile link. If not please create one

idoshamun: Also, I'm not sure about the target blank. Please advise with Nimrod.

src/config.ts
```diff
@@ -2,5 +2,8 @@ import dotenv from 'dotenv';
 
 const env = process.env.NODE_ENV || 'development';
 
+export const WEBAPP_URL =
+  env === 'development' ? 'http://localhost:5002' : 'https://app.daily.dev';
```
idoshamun, line 6: This already defined somewhere I think

src/entity/Comment.ts
```diff
@@ -70,9 +70,13 @@ export class Comment {
   })
   parent: Promise<Comment>;
 
+  mentions?: string[];
```
idoshamun, line 73: We agreed on implementing it as another table rather than arrays as far as I remember.
This way it can be easier to track when someone changes their username.
What do you think?

src/entity/CommentMention.ts
```diff
@@ -0,0 +1,17 @@
+import { Comment } from './Comment';
+import { Entity, ManyToOne, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class CommentMention {
+  @PrimaryColumn()
+  commentId: string;
+
+  @PrimaryColumn()
+  mentionedUserId: string;
```
idoshamun, line 10: This should be also a foreign key

src/schema/comments.ts
```diff
@@ -136,6 +150,12 @@ export const typeDefs = /* GraphQL */ `
     edges: [CommentUpvoteEdge!]!
   }
 
+  type MentionUser {
```
idoshamun, line 153: Why do you define a new type instead of using the existing user type?

src/schema/comments.ts
```diff
@@ -195,6 +215,15 @@ export const typeDefs = /* GraphQL */ `
       """
       first: Int
     ): CommentConnection!
+
+    """
+    Recommend users to mention in the comments
+    """
+    recommendedMentions(
+      postId: String!
+      name: String
```
idoshamun, line 224: This naming makes more sense, given it's a search query in a way

```suggestion
      query: String
```

src/schema/comments.ts
```diff
@@ -284,6 +313,86 @@ export interface GQLUserCommentsArgs extends ConnectionArguments {
...
+  { limit = 5, name }: RecentMentionsProps,
+): Promise<GQLMentionUser[]> => {
+  let query = con
+    .getRepository(CommentMention)
+    .createQueryBuilder('cm')
+    .select('u.name, u.username, u.image')
+    .innerJoin(Comment, 'c', 'cm."commentId" = c.id')
+    .innerJoin(User, 'u', 'u.id = cm."mentionedUserId"')
+    .where('c."userId" = :userId', { userId })
+    .limit(limit);
```
idoshamun, line 385: What users should this query return?

src/schema/comments.ts
```diff
@@ -374,6 +483,73 @@ export const resolvers: IResolvers<any, Context> = {
...
       );
     },
+    recommendedMentions: async (
+      _,
+      { postId, name, limit = 5 }: GQLMentionUserArgs,
+      ctx,
+    ): Promise<GQLMentionUser[]> => {
+      if (name) {
+        const recent = await getRecentMentions(ctx.con, ctx.userId, {
+          name,
```
idoshamun, line 493:
```suggestion
          name, limit
```

src/schema/comments.ts
```diff
@@ -374,6 +483,73 @@ export const resolvers: IResolvers<any, Context> = {
...
+
+      if (missing === 0) {
+        return recommendations;
+      }
+
+      const recent = await getRecentMentions(ctx.con, ctx.userId, {
+        limit: missing,
+      });
+
+      return recommendations.concat(recent);
```
idoshamun, line 551: This needs more abstractions. Quite challenging to read. Especially because of all the inline queries

src/schema/comments.ts
```diff
@@ -389,17 +565,25 @@ export const resolvers: IResolvers<any, Context> = {
...
+            content,
+          });
+          createdComment.mentions = mentions.map((mention) => mention.username);
+          const comment = await entityManager
+            .getRepository(Comment)
+            .save(createdComment);
           await entityManager
             .getRepository(Post)
             .increment({ id: postId }, 'comments', 1);
+          await saveCommentMentions(entityManager, comment.id, mentions);
```
idoshamun, line 586: I see that this looks quite similar to `commentOnComment`. Maybe we can reuse some of the stuff?

idoshamun (changes requested):
That's a lot of work! Way to go 🎉 
My initial comments

src/entity/Comment.ts
sshanzel: Yes. This property is non-existent in the db. It is just a placeholder for when the comment is created or updated, the handler for generating the markdown has a reference to those should be mentioned.

idoshamun: I think this can lead to confusion when using this entity. Some developers might forget to set this property before saving the object and then it can be invalid. We don't have any failsafe mechanism to make sure other devs use this property correctly

sshanzel: That makes total sense. In that case, we will have to take out the function inside the comment entity from instead automatically trigger generating the markdown to manually do it. With that, I will create a generic function to utilize in both cases (insert, update).

src/schema/comments.ts
sshanzel: Oh, right. We can just use the `User` type.

src/schema/comments.ts
sshanzel: It is the users that were recently mentioned by the requesting user.

src/common/markdown.ts
sshanzel: Let me create one as I searched and none came out.

With the target, in the spec, it was mentioned to target blank but after checking with Nimrod a few minutes ago, he wanted it to be on the same tab. Let me apply the changes.

src/config.ts
sshanzel: We only have one for the gateway, for the webapp, it has not been defined yet after searching through the repo.

idoshamun: See here:
https://github.com/dailydotdev/daily-api/blob/32e4367fb2405f8204171b368d5953c7b2618b85/src/common/links.ts#L2

It's the same link. Please consolidate :)

idoshamun:
@sshanzel let me know when it's ready for another round :)

sshanzel:
@idoshamun surely! I am just verifying the email sending so it's gonna be one full review. Thank you for asking! 🚢

sshanzel:
The code itself should now be ready for review, I am now just still testing locally if the email would proceed. Unfortunately, my trials have not yet pushed any email.

@idoshamun @rebelchris

infra/application.properties
```diff
@@ -7,7 +7,7 @@ debezium.source.database.user=%database_user%
 debezium.source.database.password=%database_pass%
 debezium.source.database.dbname=%database_dbname%
 debezium.source.database.server.name=api
-debezium.source.table.include.list=public.comment,public.comment_upvote,public.source_request,public.upvote,public.post,public.user,public.post_report,public.alerts,public.source_feed,public.settings
+debezium.source.table.include.list=public.comment,public.comment_upvote,public.comment_mention,public.source_request,public.upvote,public.post,public.user,public.post_report,public.alerts,public.source_feed,public.settings
```
idoshamun, line 10: I think we have another file for the local debezium setup. You may want to update it there as well

idoshamun (changes requested):
Great progress on this!
We still have major gaps to close. My suggestion for complying perfectly with GraphQL is to split the query to two parts. First part is to collect the user ids that we want for the recommendations than to have a simple GraphORM call that will fetch the users' info where the id is in the array mentioned above. This way you don't have to worry about selecting the right fields.

src/schema/comments.ts
```diff
@@ -284,6 +311,121 @@ export interface GQLUserCommentsArgs extends ConnectionArguments {
...
+  return transaction
+    .createQueryBuilder()
+    .insert()
+    .into(CommentMention)
+    .values(mentions)
+    .orIgnore()
+    .execute();
+};
+
+const saveComment = async (
```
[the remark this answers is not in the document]
sshanzel: I was thinking if we delete that, then the user re-mentions the user, we will be sending out two emails. What do you think?

src/common/users.ts
```diff
@@ -114,6 +125,94 @@ export interface ReadingDaysArgs {
...
+): Promise<MentionUser[]> => {
+  let queryBuilder = con
+    .getRepository(CommentMention)
+    .createQueryBuilder('cm')
+    .select('DISTINCT u.name, u.username, u.image')
+    .innerJoin(Comment, 'c', 'cm."commentId" = c.id')
+    .innerJoin(DbUser, 'u', 'u.id = cm."mentionedUserId"')
+    .where('c."userId" = :userId', { userId })
+    .andWhere('cm."mentionedUserId" != :userId', { userId })
+    .limit(limit);
```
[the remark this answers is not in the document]
sshanzel: I agree, that would make it much easier in terms of querying.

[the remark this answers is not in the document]
sshanzel: That makes total sense, let me check if I can do the inner join even if I already finished writing the where clause since there can be a condition where we can to search for the user's query.

src/schema/comments.ts
```diff
@@ -284,6 +311,121 @@ export interface GQLUserCommentsArgs extends ConnectionArguments {
...
+) =>
+  con
+    .getRepository(Comment)
+    .createQueryBuilder('c')
+    .select('c.*')
+    .innerJoin(CommentMention, 'cm', 'cm."commentId" = c.id')
+    .where('cm."mentionedUserId" = :userId', { userId })
+    .getRawMany<Comment>();
+
+export const updateMentions = async (
```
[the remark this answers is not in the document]
sshanzel: I agree, let me create a ticket for this.

src/schema/comments.ts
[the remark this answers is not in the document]
idoshamun: Ok, you're right!

src/entity/CommentMention.ts
```diff
@@ -0,0 +1,33 @@
+import { Entity, ManyToOne, PrimaryColumn } from 'typeorm';
+import { User } from './User';
+import { Comment } from './Comment';
+
+@Entity()
+export class CommentMention {
+  @PrimaryColumn()
```
idoshamun, line 7: Please check the exact column type of the `id` property in the `Comment` entity

src/entity/CommentMention.ts
```diff
@@ -0,0 +1,33 @@
...
+  commentId: string;
+
+  @PrimaryColumn()
```
idoshamun, line 10: Please check the exact column type of the `id` property in the `User` entity

src/entity/CommentMention.ts
```diff
@@ -0,0 +1,33 @@
...
+  commentByUserId: string;
+
+  @PrimaryColumn()
```
idoshamun, line 13: Please check the exact column type of the `id` property in the `User` entity

src/schema/comments.ts
```diff
@@ -284,6 +303,139 @@ export interface GQLUserCommentsArgs extends ConnectionArguments {
...
+) => {
+  const ids = users.map((user) => user.id);
+  transaction
+    .createQueryBuilder()
+    .delete()
+    .from(CommentMention)
+    .where({ commentId })
+    .andWhere({ commentByUserId })
+    .andWhere({ mentionedUserId: Not(In(ids)) })
+    .execute();
```
idoshamun, line 355: Actually I agree with what you mentioned in the comments that it might better to not delete from the table

src/schema/comments.ts
```diff
@@ -374,6 +526,21 @@ export const resolvers: IResolvers<any, Context> = {
...
+    ): Promise<User[]> => {
+      const { con, userId } = ctx;
+      const ids = await (query
+        ? recommendUsersByQuery(con, userId, { query, limit })
+        : recommendUsersToMention(con, postId, userId, { limit }));
+
+      return con.getRepository(User).findByIds(
+        ids.filter((id) => id !== userId),
+        { order: { name: 'ASC' }, take: limit },
+      );
```
idoshamun, line 542: This should be a `GraphORM` statement so it will return the exact fields needed by the GQL query

idoshamun (changes requested):
Another thing that worries me is the update username scenario.
What happens if a user is mentioned in like 1k comments and then they update their username.
Let's talk it over when you get the chance and decide how to proceed

src/schema/comments.ts
idoshamun: Exactly. We need to use the `query` function like we do for feeds or anything else

src/workers/commentMentionEmail.ts
```diff
@@ -0,0 +1,53 @@
...
+  const mentioned = await con
+    .getRepository(User)
+    .findOne(commentMention.mentionedUserId);
+  const [first_name] = mentioned.name.split(' ');
+  const content = getContent(comment.content, mentioned.username);
+
+  await sendEmail({
+    ...baseNotificationEmailData,
+    to: mentioned.email,
+    templateId: 'd-6949e2e50def4c6698900032973d469b',
```
rebelchris, line 40: We might want to find a way to make this fetch from the config or something?
I'm thinking what happens if we want to change template, but maybe it's overkill.
Let me know what you guys think.

rebelchris (approved):
Wow Lee, really well done on this beast 🙌
I have some small questions really, can't find anything major I would change.

sshanzel:
@idoshamun should now be ready for review including the background workers for updating mentions when the user has updated username. Though locally, I'm not sure why it is not publishing a message on my `daily-local-lee` for this new topic.

src/common/pubsub.ts
```diff
@@ -124,6 +126,30 @@ export const notifyCommentFeatured = async (
     commentId,
   });
 
+export const notifyUsernameChanged = async (
```
idoshamun, line 129: I'm not sure if it's better to publish this message here or from the gateway.
Because originally the gateway is in charge of the user entity.
What do you think?

src/schema/comments.ts
```diff
@@ -387,31 +370,23 @@ const saveComment = async (
...
-      .map((word) => (word === `@${user.username}` ? `@${newUsername}` : word))
+      .map((word) => (word === `@${oldUsername}` ? `@${newUsername}` : word))
       .join(' ');
+    comment.content = content;
 
-    return { ...comment, content };
+    return comment;
   });
 
   return Promise.all(updated.map((comment) => saveComment(con, comment)));
```
idoshamun, line 392: We need to use a transaction object here instead of a direct connection

src/workers/commentMentionEmail.ts
```diff
@@ -6,6 +7,18 @@ import { baseNotificationEmailData, sendEmail, truncatePost } from '../common';
 import { Connection } from 'typeorm';
 import { getPostPermalink } from '../schema/posts';
 
+const getContent = (comment: string, username: string) => {
```
idoshamun, line 10: Can you give this function a more explicit name? I'm not sure what you are trying to do

src/workers/usernameChanged.ts
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      const commentMentions = await con
+        .getRepository(CommentMention)
+        .find({ mentionedUserId: userId });
+      const commentIds = commentMentions.map((mention) => mention.commentId);
+      const slices = commentIds.length / limit + 1;
+      const batches = [];
+      for (let slice = 0; slice < slices; slice++) {
+        const start = slice * limit;
+        const end = (slice + 1) * limit;
```
idoshamun, line 26: What happens if `end` is bigger than array size?

idoshamun (changes requested):
Almost there! Looks great :)

src/common/pubsub.ts
sshanzel: Yes, was actually thinking of that initially, but I've thought of ensuring first the entity has changed already, so I made it watch from the API's DB. What do you think?

src/workers/usernameChanged.ts
sshanzel: Tried it in a console yesterday and it doesn't result in an error but fetches what's left instead.

src/common/pubsub.ts
idoshamun: We can do the same in the gateway, look for the changes there. I think it's better in the current architecture

sshanzel:
@idoshamun the failing test is not related to the changes applied, so feel free to review the PR. I'll look into why this one is failing on some cases and succeeds after a few re-runs.

src/schema/users.ts
```diff
@@ -238,6 +239,11 @@ export const typeDefs = /* GraphQL */ `
       query: String!
     ): SearchReadingHistorySuggestionsResults!
 
+    """
+    Get user's info
+    """
+    userInfo(id: ID!): User
```
idoshamun, line 245: Why do we need this new query?

idoshamun (commented):
One question before we proceed

src/schema/users.ts
sshanzel: This is so we can do a batch request on our tooltip data requirements instead of making two subsequent requests.

src/schema/users.ts
```diff
@@ -375,8 +375,7 @@ export const resolvers: IResolvers<any, Context> = {
           : null,
       };
     },
-    userInfo: async (_, { id }: ReadingRankArgs): Promise<GQLUser> =>
-      fetchUserById(id),
+    user: (_, { id }: ReadingRankArgs): Promise<GQLUser> => fetchUserById(id),
```
idoshamun, line 378: I think this is better

```suggestion
    user: (_, { id }:{id: string): Promise<GQLUser> => fetchUserById(id),
```

idoshamun (approved):
Non blocking suggestion to improve typing.
Once merged make sure the comments still work in production.
This is quite a change right there.

sshanzel:
The license/cla is stucked. I'll force merge as the test and build have already passed.

PR #891, opened 2022-02-18
fix: timezone issues + expanded test cases

rebelchris:
We had to revert the test cases in order to not break the application for a user.
However the issue was still remaining where a user could have 8 reading days.

And on the other side the userReadingHistory was negative 1 for certain timezones.

For this specific task I've added the specific issue as a test case + another issue found (days being exactly ON the timezone millisecond)

Note:
When we merge this we should run the query to test the user's issue to 100% validate it.

Query as it sits in this PR:

```text
SELECT
	count(distinct date_trunc('day', "timestamp"::timestamptz at time zone timezone) at time zone timezone) filter(where "timestamp"::timestamptz at time zone timezone BETWEEN (date_trunc('week', timezone(timezone, now()) - interval '7 days') at time zone timezone) AND (date_trunc('week', timezone(timezone, now())) at time zone timezone))
FROM
	VIEW
	JOIN "user" ON "user".id = view. "userId"
WHERE
	"userId" = 'i6nUrr2osa7TZhJP69HWz'
```

DD-550 #done


rebelchris:
Screenshot 2022-02-18 at 14 45 57

Going to leave this here, we need to determine if 00:00:00 should be on the current day or not and work from there.

rebelchris:
@idoshamun Let me know what you think 💪

idoshamun (approved):
Perfect!

PR #892, opened 2022-02-23
feat: exclude deleted views

rebelchris:
I've added a `ActiveView` viewEntity 
using this entity will automatically not load deleted views, so we don't have to exclude them by hand.

DD-551 #done

### DIFF
--- a/__tests__/setup.ts
+++ b/__tests__/setup.ts
@@ -4,12 +4,11 @@ import { createOrGetConnection } from '../src/db';
 import { redisClient, redisPubSub } from '../src/redis';
 
 let con: Connection;
-const excludeEntities = ['ActiveView'];
 
 const cleanDatabase = async (): Promise<void> => {
   for (const entity of con.entityMetadatas) {
-    if (excludeEntities.includes(entity.name)) continue;
     const repository = con.getRepository(entity.name);
+    if (repository.metadata.tableType === 'view') continue;
     await repository.query(`DELETE FROM "${entity.tableName}";`);
 
     for (const column of entity.primaryColumns) {
--- a/__tests__/setup.ts
+++ b/__tests__/setup.ts
@@ -4,9 +4,11 @@ import { createOrGetConnection } from '../src/db';
 import { redisClient, redisPubSub } from '../src/redis';
 
 let con: Connection;
+const excludeEntities = ['ActiveView'];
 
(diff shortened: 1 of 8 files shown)


__tests__/setup.ts
```diff
@@ -4,9 +4,11 @@ import { createOrGetConnection } from '../src/db';
 import { redisClient, redisPubSub } from '../src/redis';
 
 let con: Connection;
+const excludeEntities = ['ActiveView'];
 
 const cleanDatabase = async (): Promise<void> => {
   for (const entity of con.entityMetadatas) {
+    if (excludeEntities.includes(entity.name)) continue;
```
idoshamun, line 11: Not a blocker but we should consider checking in the metadata of the entity if it's a view instead of keeping a list of entities to exclude. What do you think?

src/entity/ActiveView.ts
```diff
@@ -0,0 +1,29 @@
...
+import { View } from './View';
+
+@ViewEntity({
+  expression: (connection: Connection) =>
+    connection
+      .createQueryBuilder()
+      .select('view.userId', 'userId')
+      .addSelect('view.postId', 'postId')
+      .addSelect('view.timestamp', 'timestamp')
+      .addSelect('view.hidden', 'hidden')
```
idoshamun, line 12: Maybe we can select `view.*` to make it more simple?

idoshamun (changes requested):
I love it that we introduce views. It makes everything much easier.

__tests__/setup.ts
rebelchris: Great suggestion, refactoring that.

__tests__/setup.ts
```diff
@@ -4,12 +4,11 @@ import { createOrGetConnection } from '../src/db';
 import { redisClient, redisPubSub } from '../src/redis';
 
 let con: Connection;
-const excludeEntities = ['ActiveView'];
 
 const cleanDatabase = async (): Promise<void> => {
   for (const entity of con.entityMetadatas) {
-    if (excludeEntities.includes(entity.name)) continue;
     const repository = con.getRepository(entity.name);
+    if (repository.metadata.tableType === 'view') continue;
```
idoshamun, line 11: Nice!

idoshamun (approved):
Love it! 💪

PR #896, opened 2022-03-04
fix: duplicated and missing attributes

sshanzel:
DD-567 #done

Have tested all variance manually and none is resulting in an error now. The issue was coming from duplicate attributes.


sshanzel:
This should fix the open issue here: https://github.com/dailydotdev/daily/issues/491

idoshamun (approved):
Wow! Crazy fix

PR #898, opened 2022-03-07
feat: companion

rebelchris:
Introducing the changes needed for the companion app

- optOutCompanion setting
- getPostByCanonicalUrl query

dd-458


src/index.ts
```diff
@@ -179,6 +179,9 @@ export default async function app(
             extendKey: trackingExtendKey,
           },
           banner: true,
+          postCanonical: {
```
idoshamun, line 182: Nice!

src/schema/posts.ts
```diff
@@ -304,6 +304,16 @@ export const typeDefs = /* GraphQL */ `
       id: ID
     ): Post!
 
+    """
+    Get post by canonical URL
+    """
+    postCanonical(
```
idoshamun, line 310: I would suggest this name which is a bit more explicit in my opinion.
I also think we should compare both canonical and the original URL to the given parameter.
What do you think?

```suggestion
    postByUrl(
```

src/schema/posts.ts
```diff
@@ -457,6 +467,24 @@ export const resolvers: IResolvers<any, Context> = {
       }
       throw new NotFoundError('Post not found');
     },
+    postCanonical: async (
+      source,
+      { canonicalUrl }: { id: string; canonicalUrl: string },
+      ctx: Context,
+      info,
+    ) => {
+      const res = await graphorm.query(ctx, info, (builder) => ({
```
idoshamun, line 476: Please add a limit 1 to this query should be faster (a bit).

idoshamun (approved):
Looking good! A few suggestions for improvement and we can merge 😬

src/schema/posts.ts
rebelchris: Good point, my idea was that the canonical would be leading anyway, but happy to include a URL or Canonical as the query.

rebelchris:
@idoshamun Switched to postByUrl, added both test cases for URL and Canonical URL.

idoshamun (approved):
Awesome!

PR #899, opened 2022-03-09
fix: missing character and avoid email spamming in user mention

sshanzel:
This also includes a fix for recommending users without a username as part of the post commented users.

### DIFF
--- a/src/workers/commentMentionEmail.ts
+++ b/src/workers/commentMentionEmail.ts
@@ -32,14 +32,11 @@ export const sendEmailToMentionedUser = async (
   }
 
   if (comment.parentId !== null) {
-    const comments = await con
+    const commentOnSameThread = await con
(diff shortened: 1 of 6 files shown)


src/workers/commentMentionEmail.ts
```diff
@@ -21,17 +21,31 @@ const removeFirstWordIfMention = (comment: string, username: string) => {
...
+    return;
+  }
+
+  if (comment.parentId !== null) {
+    const comments = await con
+      .getRepository(Comment)
+      .find({ parentId: comment.parentId });
+    const commentedTheSameThread = comments.some(
+      (comment) => comment.userId === mentionedUserId,
+    );
```
idoshamun, line 40: You can this logic to the query

idoshamun (approved):

PR #900, opened 2022-03-11
fix: restrict mention on null username

sshanzel:
Since the name was the one used for searching the LIKE fails on username even being null.

This should explicitly restrict the result for users without usernames.

### DIFF
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -179,7 +179,7 @@ export const getUserIdsByNameOrUsername = async (
     .getRepository(DbUser)
     .createQueryBuilder()
     .select('id')
-    .where('(name ILIKE :name OR username ILIKE :name)', {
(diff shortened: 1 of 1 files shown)


idoshamun (approved):
Please check the performance of this after we merge.
We might need to index every user with right indexes to improve performance and reduce database overhead

sshanzel:
Sure, I'll keep records of performance on some of our densely populated pages (posts with a lot of comments). Then see how things turn out once everyone knows mentioning is available.

sshanzel:
The amount of time it takes is increasing when the query also increases.

1. 369.95ms
2. 235.08ms
3. 242.37ms
4. 309.22ms
5. 436.61ms

idoshamun:
We will have to address it.
I have a feeling that's going to be a bottleneck once we release.
We can address it after the launch, it's not a deal-breaker in my opinion.

PR #902, opened 2022-03-15
feat: query for comments preview

sshanzel:
DD-468 #done

When generating the preview, we will be just utilizing our existing function to provide the same result than re-creating the overrides in the frontend.

I have also removed the flaky test which was deemed unnecessary. Reference: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1646829007663149?thread_ts=1646818642.801639&cid=C02E2C3C13R

### DIFF
--- a/__tests__/__snapshots__/comments.ts.snap
+++ b/__tests__/__snapshots__/comments.ts.snap
@@ -92,7 +92,7 @@ Array [
 ]
 `;
 
-exports[`query commentPreview should return author and previously mentioned users if query is empty 1`] = `
+exports[`query commentPreview should return markdown equivalent of the content 1`] = `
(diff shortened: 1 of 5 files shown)


rebelchris (approved):
Look good
Perhaps we should add some extended test markdown, including a mention for instance, this way we can quickly failtest if we accidentally change something in the future.

sshanzel:
@rebelchris yes, totally agree. Though I think we will need additional tests on cases where we override a specific renderer. Like as you mentioned, in the user mentions. For that case, I think we have an existing one already.

idoshamun (approved):
I love the idea of generating the preview in the server side! 😇

PR #906, opened 2022-03-23
fix: comment mention email link

sshanzel:


src/schema/posts.ts
```diff
@@ -449,6 +449,9 @@ export const reportReasons = new Map([
 export const getPostPermalink = (post: Pick<GQLPost, 'shortId'>): string =>
   `${process.env.URL_PREFIX}/r/${post.shortId}`;
 
+export const getPostPageLink = (post: Pick<GQLPost, 'shortId'>): string =>
```
idoshamun, line 452: I think we have it already somewhere in the codebase

idoshamun (changes requested):
Please check my comment :)

src/schema/posts.ts
rebelchris: commentsPermalink I think from the top of my head

sshanzel: You're right, Chris! Thank you for the reference!

idoshamun (approved):
Ok Ok, let's roll! 🚢

rebelchris (commented):
Nice! 💖

PR #915, opened 2022-04-06
refactor: disable graphql subscriptions

idoshamun:
Disable GraphQL subscriptions due to the memory and CPU overload which caused performance regression.
We can evaluate separating subscriptions to a new deployable unit


rebelchris (approved):

sshanzel (approved):
Oohhh. I very much like having a dedicated services for this.

PR #918, opened 2022-04-11
refactor: reputation rules

sshanzel:
DD-532

A new entity was introduced to contain all the events that have happened whenever there's a change with the user's reputation. The only thing missing now is how we will approach the migration of new reputation rules.


src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+  ManyToOne,
+  PrimaryColumn,
+  Column,
+  BeforeInsert,
+  Index,
+} from 'typeorm';
+import { User } from './User';
+
+export enum ReputationReason {
+  PostUpvote = 'post_upvote',
```
idoshamun, line 12:
```suggestion
  PostUpvoted = 'post_upvoted',
```

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+  PostBanned = 'post_banned',
+  CommentUpvote = 'comment_upvote',
```
idoshamun, line 14:
```suggestion
  CommentUpvoted = 'comment_upvoted',
```

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+  @PrimaryColumn({ length: 36 })
+  grantToId: string;
+
+  @PrimaryColumn({ length: 36 })
+  targetId: string;
+
+  @PrimaryColumn({ length: 36 })
+  reason: ReputationReason;
+
+  @PrimaryColumn({ length: 36, type: 'varchar' })
```
idoshamun, line 51: `varchar` is the default type

```suggestion
  @PrimaryColumn({ length: 36 })
```

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+  [ReputationReason.PostReportConfirmed]: 100,
+  [ReputationReason.SourceRequestApproved]: 200,
+};
+
+@Entity()
+export class ReputationEvent {
+  @PrimaryColumn({ default: () => 'now()' })
+  timestamp: Date;
+
+  @PrimaryColumn({ length: 36, default: '' })
```
idoshamun, line 38: Why this is the default? Default should be null

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+  [ReputationReason.PostUpvote]: 10,
+  [ReputationReason.PostBanned]: -100,
+  [ReputationReason.CommentUpvote]: 50,
+  [ReputationReason.PostReportConfirmed]: 100,
+  [ReputationReason.SourceRequestApproved]: 200,
+};
+
+@Entity()
+export class ReputationEvent {
+  @PrimaryColumn({ default: () => 'now()' })
```
idoshamun, line 35: Thinking about it again, not sure `timestamp` needs to be part of the primary key

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+  @PrimaryColumn({ length: 36 })
+  targetId: string;
+
+  @PrimaryColumn({ length: 36 })
+  reason: ReputationReason;
+
+  @PrimaryColumn({ length: 36, type: 'varchar' })
+  targetType: ReputationType;
+
+  @Column({ type: 'int' })
```
idoshamun, line 54: TypeORM knows to extract type automatically

```suggestion
  @Column()
```

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  grantByUser?: Promise<User>;
+
+  @ManyToOne(() => User, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  grantToUser?: Promise<User>;
```
idoshamun, line 67: This is not nullable

```suggestion
  grantToUser: Promise<User>;
```

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+const reputationReasonAmount: Record<ReputationReason, number> = {
+  [ReputationReason.PostUpvote]: 10,
+  [ReputationReason.PostBanned]: -100,
+  [ReputationReason.CommentUpvote]: 50,
+  [ReputationReason.PostReportConfirmed]: 100,
+  [ReputationReason.SourceRequestApproved]: 200,
+};
+
+@Entity()
+export class ReputationEvent {
```
idoshamun, line 34: Make sure the length 36 is aligned with the actual primary key from the different tables, because I'm not sure about it

src/workers/commentUpvoteCanceledRep.ts
```diff
@@ -14,7 +19,16 @@ const worker: Worker = {
     try {
       const comment = await con.getRepository(Comment).findOne(data.commentId);
       if (comment.userId !== data.userId) {
-        await increaseReputation(con, logger, comment.userId, -1);
+        const repo = con.getRepository(ReputationEvent);
+        const event = await repo.findOne({
```
idoshamun, line 23: We don't need `findOne` and then `delete`. We can just send a single delete request

src/workers/postBannedRep.ts
```diff
@@ -12,16 +17,33 @@ const worker: Worker = {
...
-        logger,
-        reports.map(({ userId }) => userId),
-        1,
+      const reports = await con.getRepository(PostReport).find({ postId: id });
+      const userIds = reports.map(({ userId }) => userId);
+      const repo = con.getRepository(ReputationEvent);
+      const events = userIds.map((userId) =>
+        repo.create({
+          grantToId: userId,
+          grantById: '',
```
idoshamun, line 28: This should be null

src/workers/postBannedRep.ts
```diff
@@ -12,16 +17,33 @@ const worker: Worker = {
   subscription: 'post-banned-rep',
   handler: async (message, con, logger): Promise<void> => {
     const data: Data = messageToJson(message);
+    const { id, authorId } = data.post;
```
idoshamun, line 20: All the different database calls should be in a transaction

src/workers/postUpvoteCanceledRep.ts
```diff
@@ -14,7 +19,16 @@ const worker: Worker = {
     try {
       const post = await con.getRepository(Post).findOne(data.postId);
       if (post?.authorId && post?.authorId !== data.userId) {
-        await increaseReputation(con, logger, post.authorId, -1);
+        const repo = con.getRepository(ReputationEvent);
+        const event = await repo.findOne({
```
idoshamun, line 23: See the comments in the comment upvote worker

idoshamun (changes requested):
I think it's better to have the actual change of reputation in the CDC part when a new reputation even is added or existing one is deleted. Also I think when you add a new reputation event, it's better to make sure that on conflict it ignores the insert rather than fail.

rebelchris (commented):
Just have one question around the actual reputation values

src/entity/ReputationEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+  Source = 'source',
+}
+
+const reputationReasonAmount: Record<ReputationReason, number> = {
+  [ReputationReason.PostUpvote]: 10,
+  [ReputationReason.PostBanned]: -100,
+  [ReputationReason.CommentUpvote]: 50,
+  [ReputationReason.PostReportConfirmed]: 100,
+  [ReputationReason.SourceRequestApproved]: 200,
+};
```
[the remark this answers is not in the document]
rebelchris: Resolved on Slack :)

src/entity/ReputationEvent.ts
sshanzel: This was considered a primary key and it is not possible to make it a null column. We can either make it a null `Column` or a `PrimaryColumn` with an empty string as a value.

idoshamun:
Copying my response on Slack here just in case:
I see it makes sense. But let's add it though. The primary key will make sure we don't present any duplicates and also we can run the calculation just in case to make sure it's aligned. Also, one thing I recall is missing in the PR is the minimum threshold required for granting the reputation as described below
https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/137789441/Reputation

src/entity/ReputationEvent.ts
sshanzel: Yes. User id has a max length of 36.

idoshamun: What about the rest? (post id, comment id, source id)

sshanzel: All 3 of them didn't specify any (default length, probably).

sshanzel:
Note: I have removed the subscriptions for workers relating to the featured comment but didn't remove the feature itself completely.

idoshamun (changes requested):
Not there yet, it requires a bit more work

src/entity/ReputationEvent.ts
```diff
@@ -23,22 +23,23 @@ export enum ReputationType {
...
+);
+
 @Entity()
 export class ReputationEvent {
-  @PrimaryColumn({ default: () => 'now()' })
-  timestamp: Date;
-
-  @PrimaryColumn({ length: 36, default: '' })
-  grantById: string;
+  @Index({ unique: true })
```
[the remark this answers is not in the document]
sshanzel: I mean primary constraint is just a combination of `unique` and `not null` - so are we not gonna make this a primary column anymore since it might limit the capacity of the table?

Let me revert this back to default an empty string as discussed.

idoshamun:
Another thing I now realized that is missing, capping minimum reputation to zero. Now that people can lose their reputation we have to make sure they can't have a negative reputation

idoshamun:
And also initial reputation should be 10

src/entity/ReputationEvent.ts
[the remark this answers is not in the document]
sshanzel: @idoshamun just wanted to confirm if we would still push with `PrimaryColumn` for the `grantById` since under the hood it is just a combination of a `unique` constraint with `not null` that might limit the table.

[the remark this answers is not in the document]
idoshamun: Composite primary keys like we have here force uniqueness on the combination, not on every column that makes up the composite key. In our case this is exactly what we want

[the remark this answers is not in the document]
sshanzel: Ah, cool, I think that could also be the case for the `unique` constraint why it has allowed multiple `null` values initially. Let me apply the `PrimaryColumn` and the supposed empty string.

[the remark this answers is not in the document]
idoshamun: Please also check the other comments I submitted outside the review

sshanzel:
@idoshamun thank you for the review! I initially thought of removing the cron jobs since there will be no reward for them anymore - with that, I have reverted the said change.

Regarding the change in the gateway - that will be part of the deployment phase right?

Note: the failing test is a known issue that happens on Monday and Friday (have logged a ticket on it already).

idoshamun:
> I initially thought of removing the cron jobs since there will be no reward for them anymore - with that, I have reverted the said change.

There's no reward that's true but we should be aware of breaking existing features or at least mention that this change is going to break them.

> Regarding the change in the gateway - that will be part of the deployment phase right?

Yes, but it's better to have the PR ready

src/workers/cdc.ts
```diff
@@ -295,6 +299,40 @@ const onFeedChange = async (
...
+    if (amount > 0) {
+      await increaseReputation(con, logger, entity.grantToId, amount);
+    } else {
+      const user = await con.getRepository(User).findOne(entity.grantToId);
+      const result = getNegativeAmount(amount, user);
+      await increaseReputation(con, logger, entity.grantToId, result);
+    }
+  } else if (data.payload.op === 'd') {
+    const entity = data.payload.before;
+    await increaseReputation(con, logger, entity.grantToId, -entity.amount);
```
idoshamun, line 332: I think a single function to handle the change in reputation is the right way to go. It's quite confusing to have `increaseReputation` which doesn't handle all the different edge cases. Other developers might get confused.
Also, I believe there's no need for JavaScript here. You can do everything in a single SQL query with more simple logic. You can use the math function `max` in SQL to make sure the new reputation is not below 0

src/workers/commentUpvotedRep.ts
```diff
@@ -11,35 +16,49 @@ const worker: Worker = {
...
+          .getRepository(User)
+          .findOne(data.userId);
+
+        if (comment.userId === grantBy.id) {
+          return;
+        }
+
+        if (grantBy.reputation < REPUTATION_THRESHOLD) {
+          return;
+        }
```
idoshamun, line 41: You can merge it together

idoshamun (changes requested):
My comments are inside. Also you missed changing the default reputation as I requested earlier. You can find it here:
https://github.com/dailydotdev/daily-api/blob/master/src/entity/User.ts#L31

__tests__/workers/cdc.ts
```diff
@@ -879,3 +967,70 @@ describe('source feed', () => {
     ]);
   });
 });
+
+describe('reputation event', () => {
```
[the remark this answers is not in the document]
sshanzel: oh, this is redundant now. I shall remove the one I made initially.

sshanzel:
> There's no reward that's true but we should be aware of breaking existing features or at least mention that this change is going to break them.

Totally agree. Though when the cron jobs are stopped - it shouldn't break anything, right? Since existing featured comments will still be intact, it just won't look for a new comment to feature? (just out of curiosity, if we happen to stop some cron jobs in the future).

Thank you for mentioning the use of a single SQL command for the reputation part, it is so much cleaner now - though I have used `greatest` since `max` wasn't working as part of the value to update.

idoshamun (approved):
Yes!! 💪

PR #926, opened 2022-04-18
fix: post import seeds for development

sshanzel:
Since we used to have posts having empty `canonicalUrl` property - our seeds have those as well. Now that we have enforced a unique index, our seeds are now invalid.

If you fresh pull `daily-api` having a fresh DB, importing the seeds would result in below error:
Screen Shot 2022-04-18 at 10 01 25 PM

Steps taken to update:
1. Went back to the commit before enforcing a unique index.
2. Run `npm run db:migrate:reset` then `npm run db:seed:import`
3. Checkout master (latest commit).
4. Run `npm run db:migrate:latest` to run the latest migrations including the updating of posts' canonical URLs.
5. Run `npm run db:seed:export`.
6. Cherry-pick the only necessary files (`Post` and `PostKeyword`).

Have verified locally that importing seeds now work as intended.


idoshamun:
Thanks for doing this!

PR #929, opened 2022-04-20
fix: remove HTML option from markdown parser

rebelchris:
We had the HTML option enabled in the markdown parser, allowing all HTML to be valid.
We removed this so the rendered only renders markdown to HTML, and converts HTML into plain strings.

We also added a test case to validate this change, and tested it locally.

DD-595 #done


sshanzel (approved):
Well done on this fix!

idoshamun (approved):
Sweet!

PR #931, opened 2022-04-21
Dd 458 companion expanded

rebelchris:
Introducing a new setting for the companion state.
This actually defaults to `NULL` as we want to AB test which state works best.

We will only run the AB on people who have this as NULL, and this will be handled in the gateway boot query.
Everyone who was true/false as a value already made some changes and should not be reflected by the test case.

Related to: DD-458


sshanzel (approved):
Nice! This is probably a good thing to remember in future developments for boolean properties to allow A/B testing.

idoshamun (approved):
💪

PR #933, opened 2022-04-29
refactor: settings entity - additional property (autoDismissNotifications)

sshanzel:
As part of our accessibility compliance, we will have to allow the user to decide whether they want to automatically dismiss notifications or not.
- The property is known as `autoDismissNotifications`
- Default value is true. The false value is intended for users that might not be able to react quickly enough and click the action button inside our notifs.


rebelchris (approved):
Nice one! ⛴

idoshamun (approved):
Awesome :)

PR #935, opened 2022-05-09
feat: submissions entity and its workers

sshanzel:


idoshamun:
@sshanzel I see that it's still in draft but you asked for my review. Should I proceed with the review?

sshanzel:
@idoshamun oh, I think you automatically got tagged due to being a code owner. Please refrain from reviewing. Once done with the tests and some concerns I found with graphorm, will tag you for a full review. Thank you! 🙇

rebelchris (changes requested):

src/schema/feeds.ts
```diff
@@ -885,9 +885,22 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
-        }),
+        builder
+          .addSelect(
+            `count(*) filter (where ${alias}."scoutId" = '${author}')`,
+            'isScout',
+          )
+          .addSelect(
+            `count(*) filter (where ${alias}."authorId" = '${author}')`,
+            'isAuthor',
+          )
```
sshanzel, line 896: Not sure if we must have the additional selects when we can check if the user is the author or scout if the relevant id property has value.

rebelchris: It could be for a different user, and either scout or author.
Not sure what the alternative is here.

sshanzel:
There are things waiting to be changed but this can be reviewed by now, I think.
The advanced settings inserting a new community link was not added - we can also do it after the deployment since we still don't have the title and description for it.

cc @rebelchris @idoshamun

infra/index.ts
```diff
@@ -39,6 +39,9 @@ const debeziumTopic = new gcp.pubsub.Topic('debezium-topic', {
   'alerts-updated',
   'settings-updated',
   'update-comments',
+  'post-scout-matched',
+  'submission-changed',
+  'community-link-submitted',
```
idoshamun, line 44: There's some misalignment with the names here. They don't follow the same convention

src/entity/Post.ts
```diff
@@ -109,6 +110,15 @@ export class Post {
   @Index('IDX_post_comments')
   comments: number;
 
+  @Column({ length: 36, nullable: true })
+  scoutId: string | null;
```
idoshamun, line 114: We need to index this field

src/entity/Post.ts
```diff
@@ -391,6 +417,24 @@ const addPostAndKeywordsToDb = async (
...
+
+  return con.transaction(async (entityManager) => {
+    try {
+      const submissionId = await rejectSubmission(entityManager, data);
+      return { status: 'ok', submissionId };
+    } catch (error) {
+      throw error;
+    }
+  });
+};
```
idoshamun, line 436: I think this can go to the private routes if we use it only there.
`addNewPost` is here for compatibility reasons

src/entity/ReputationEvent.ts
```diff
@@ -23,6 +24,7 @@ export enum ReputationType {
 
 const reputationReasonAmount: Record<ReputationReason, number> = {
   [ReputationReason.PostUpvoted]: 10,
+  [ReputationReason.PostSubmissionApproved]: 100, // actual value is still TBD
```
idoshamun, line 27: We are not rewarding reputation for submitting a link as far as I remember

src/entity/Submission.ts
```diff
@@ -0,0 +1,85 @@
...
+  entityManager: EntityManager,
+  id: string,
+): Promise<string> => {
+  if (!id) {
+    return null;
+  }
+
+  const submission = await entityManager
+    .getRepository(Submission)
+    .findOneOrFail(id);
```
idoshamun, line 72: I think you want to use `findOne` otherwise it will throw an error in case no such submission

src/entity/Submission.ts
```diff
@@ -0,0 +1,85 @@
...
+    {
+      status: SubmissionStatus.Rejected,
+      reason: data.reason,
+    },
+  );
+
+  return data.submissionId;
+};
+
+export const validateSubmission = async (
```
idoshamun, line 62: The name doesn't indicate that there's going to be a side-effect which is to approve the submission

src/graphorm/graphorm.ts
```diff
@@ -95,9 +96,21 @@ export class GraphORM {
...
+    const inverseRelation =
+      inverseRelations.length === 1
+        ? inverseRelations[0]
+        : inverseRelations.find((rel) => {
+            const fk = rel.foreignKeys[0].columnNames[0];
+            // the issue came from when we reference a table twice in the same entity (ex. Author and Source) - always using the first element of the array won't work
+            // strip the id part to compare the reference as it is a common convention on ORMs (ex. on EF Core ORM)
+            // ex. when we're looking for the "author" - it is highliy likely the relevant FK is named "authorId" - so we remove the "Id" part for comparison
+            // else we will have to force utilizing the metadata that would come from the "JoinColumn" provided by typeorm but have to explicitly define it
+            return fk.substring(0, fk.length - 2) === field;
```
idoshamun, line 111: Nice catch! I think that at some point we will have to optimize graphorm. The runtime logic is growing and basically many of it can be done on boot

src/schema/submissions.ts
```diff
@@ -0,0 +1,113 @@
...
+    hasAccess: Boolean!
+    limit: Int!
+    todaySubmissionsCount: Int!
+  }
+
+  extend type Query {
+    """
+    Information regarding the access of submitting community links
+    """
+    submissionsAndLimit: SubmissionAvailability!
```
idoshamun, line 32: I think this name makes more sense as this is the name of the object as well

```suggestion
    submissionAvailability: SubmissionAvailability!
```

src/schema/submissions.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+  extend type Mutation {
+    """
+    Submit an article to surface on users feed
+    """
+    submitArticle(url: String!): EmptyResponse @auth
```
idoshamun, line 38: We might want to return the submissions object even though the UI will do nothing with it. What do you think?
It might be future proof

src/schema/submissions.ts
```diff
@@ -0,0 +1,113 @@
...
+    ): Promise<GQLEmptyResponse> => {
+      if (!isValidHttpUrl(url)) {
+        throw new ValidationError('Invalid URL!');
+      }
+
+      const postRepo = ctx.con.getRepository(Post);
+      const existingPost = await postRepo
+        .createQueryBuilder('post')
+        .where('url = :url or "canonicalUrl" = :url', { url })
+        .andWhere('deleted = false')
```
idoshamun, line 84: This is some what tricky. Because if there's a deleted post with the same url or canonical url you will not be able to add it to the db

src/schema/submissions.ts
```diff
@@ -0,0 +1,113 @@
...
+      const submissions = await repo.find({
+        where: `date_trunc('day', "createdAt")::timestamptz ${atTimezone} = date_trunc('day', ${nowTimezone})::timestamptz`,
+      });
+      const hasAccess = true; // subject for change on how we define the access
+
+      return { limit, todaySubmissionsCount: submissions.length, hasAccess };
+    },
+  },
+  Mutation: {
+    submitArticle: async (
```
idoshamun, line 71: In case the url already exists we want to return the post itself.
So we need to return either a post or a submission object.
We can create a parent object and use a parent object something like `{result: 'succeed' | 'exists', post?: Post, submission?: Submission}` (just an example)

src/workers/cdc.ts
```diff
@@ -312,6 +325,25 @@ const onReputationEventChange = async (
...
+) => {
+  const entity = data.payload.after;
+  if (data.payload.op === 'c') {
+    await notifySubmissionCreated(logger, {
+      sourceId: 'TBD',
+      url: entity.url,
+      submissionId: entity.id,
+    });
+  } else if (data.payload.op === 'u') {
+    if (entity.status !== SubmissionStatus.Accepted) {
```
idoshamun, line 341: Why?

src/workers/postScoutMatchedMail.ts
```diff
@@ -0,0 +1,56 @@
...
+} from '../common';
+
+interface Data {
+  postId: string;
+  scoutId: string;
+}
+
+// TODO:: Update the dynamic template data once available
+
+const worker: Worker = {
```
idoshamun, line 23: We need to send the email only when the post is approved and added to the DB not when it is matched

src/workers/postScoutMatchedRep.ts
```diff
@@ -0,0 +1,62 @@
...
+} from '../entity/ReputationEvent';
+import { messageToJson, Worker } from './worker';
+import { Post } from '../entity';
+
+interface Data {
+  scoutId: string;
+  postId: string;
+}
+
+const worker: Worker = {
```
idoshamun, line 15: There's no such reputation rule

src/workers/postScoutMatchedSlack.ts
```diff
@@ -0,0 +1,49 @@
...
+  handler: async (message, _, logger): Promise<void> => {
+    const data: Data = messageToJson(message);
+    const { post } = data;
+    try {
+      if (!post.scoutId) {
+        return;
+      }
+
+      await webhook.send({
+        text: 'Post submission was approved!',
```
idoshamun, line 21:
```suggestion
        text: 'New community link!',
```

idoshamun (changes requested):
Very impressive PR, I have some comments though.
A big portion that is missing is the reputation rules. We should apply the same reputation points that we apply for authors but also for scouts. This means upvotes, ban, etc

src/schema/submissions.ts
rebelchris: We are returning the post below, in this case as the simple ValidationError as we parse the full post for now.
Is your suggestion more around a more object approach?

src/workers/postScoutMatchedMail.ts
sshanzel: This gets triggered when a new post is created in the DB. Do you want it to be moved when the submission status has changed to approved?

https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/190939137/WT-4+Community+Links#API%2FGateway-changes

src/workers/cdc.ts
sshanzel: We used to send the mail for when the post is created. Though after the review, it seems to be better off when the submission changed to accepted, this condition should now be omitted.

src/schema/submissions.ts
sshanzel: Can't agree more! Let's do it! :ship

src/schema/submissions.ts
sshanzel: Should we remove the deleted condition so it won't have to go through the process before realizing it existed and deleted?

src/workers/postScoutMatchedMail.ts
idoshamun: Nope. My bad

src/schema/submissions.ts
idoshamun: Yes. I suggest that in both cases the status code will be 200 and the difference will be the object. I find it easier for the client-side to handle known issues and parse the payload

src/schema/submissions.ts
rebelchris: Changed the way we handle deleted posts :)

src/schema/submissions.ts
rebelchris: Done, added some different flows for both the submission and the newPost routes.

src/schema/posts.ts
```diff
@@ -264,6 +274,21 @@ export const typeDefs = /* GraphQL */ `
...
+
+    """
+    Whether the user is the scout
+    """
+    isScout: Int
+
+    """
+    Whether post is deleted
+    """
+    deleted: Boolean
```
idoshamun, line 291: Why did we add this?
I'm not sure this should be publicly available. If post is deleted we can just return not found can't we?

src/schema/submissions.ts
```diff
@@ -0,0 +1,137 @@
...
+        return { limit, hasAccess: false, todaySubmissionsCount: 0 };
+      }
+
+      const timezone = user.timezone ?? 'utc';
+      const nowTimezone = `timezone('${timezone}', now())`;
+      const atTimezone = `at time zone '${timezone}'`;
+      const submissions = await repo.find({
+        where: `date_trunc('day', "createdAt")::timestamptz ${atTimezone} = date_trunc('day', ${nowTimezone})::timestamptz`,
+      });
+      const hasAccess = true; // subject for change on how we define the access
```
idoshamun, line 96: I think we have the definition in the spec already. Please confirm

src/schema/posts.ts
idoshamun: But I didn't see where we plan to use it. We can return the proper to the client without the actual post element

idoshamun: You can add the `deleted` field to the required columns in this case. It's a workaround but that's good enough for now

sshanzel:
@idoshamun  after running through the spec and resolving the comments, I think this now meets the requirements (still missing the email ids tho).

For the URL standardization, I think that was tackled in this PR: https://github.com/dailydotdev/daily-api/pull/946

idoshamun:
@sshanzel I think #946 is a temporary fix, we had a much broader solution for standardizing the URLs. This needs to be addressed in a separate issue before we release the community links. @rebelchris what do you think?

__tests__/submissions.ts
```diff
@@ -0,0 +1,223 @@
...
+      process.env.SCOUT_SUBMISSION_LIMIT || DEFAULT_SUBMISSION_LIMIT,
+    );
+    expect(res.errors).toBeFalsy();
+    expect(res.data.submissionAvailability.limit).toEqual(limit);
+    expect(res.data.submissionAvailability.hasAccess).toEqual(true);
+    expect(res.data.submissionAvailability.todaySubmissionsCount).toEqual(1);
+  });
+});
+
+describe('mutation submitArticle', () => {
```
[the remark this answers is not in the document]
idoshamun: We're still missing a test to check if the user has access to this feature

__tests__/workers/cdc.ts
```diff
@@ -975,3 +1005,54 @@ describe('reputation event', () => {
...
+      mockChangeMessage<ObjectType>({
+        after,
+        before: null,
+        op: 'c',
+        table: 'submission',
+      }),
+    );
+    expect(notifySubmissionCreated).toBeCalledTimes(1);
+    expect(mocked(notifySubmissionCreated).mock.calls[0].slice(1)).toEqual([
+      { url: after.url, submissionId: after.id, sourceId: 'TBD' },
```
idoshamun, line 1034: `sourceId` should be fixed because we agreed to add the migration with the new source

src/common/pubsub.ts
```diff
@@ -246,3 +267,28 @@ export const notifySourceFeedRemoved = async (
...
     sourceId,
   });
+
+export const notifySubmissionChanged = async (
+  log: EventLogger,
+  submission: ChangeObject<Submission>,
+): Promise<void> => {
+  if (submission.status === SubmissionStatus.Rejected) {
+    publishEvent(log, communityLinkRejectedTopic, submission);
+  }
```
idoshamun, line 277: This should not include any business logic. The business logic should be in the calling function

rebelchris:
Agreed that was the start/hotfix for the companion we need to introduce a stable useage of this system and expand to everywhere (including the newPost etc)

src/schema/submissions.ts
```diff
@@ -0,0 +1,176 @@
...
+      _,
+      { url }: GQLArticleSubmission,
+      ctx,
+      info,
+    ): Promise<GQLSubmitArticleResponse> => {
+      const user = await ctx.getRepository(User).findOne({ id: ctx.userId });
+
+      if (!user || !hasSubmissionAccess(user)) {
+        return { result: 'reject', reason: 'Access denied' };
+      }
```
idoshamun, line 136: This should throw an error as we do for non-authenticated users

src/schema/submissions.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      const submissionRepo = ctx.con.getRepository(Submission);
+      const submissionsToday = await getSubmissionsToday(
+        ctx.con,
+        user.timezone,
+      );
+
+      if (submissionsToday.length >= submissionLimit) {
+        return { result: 'reject', reason: 'Submission limit reached' };
+      }
```
idoshamun, line 146: This is fine :)

src/schema/submissions.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      if (existingSubmission) {
+        return {
+          result: 'reject',
+          reason: `Article has been submitted already! Current status: ${existingSubmission.status}`,
+        };
+      }
+
+      const submission = await submissionRepo.save(
+        submissionRepo.create({ url, userId: ctx.userId }),
```
idoshamun, line 170: `create` is not needed here, you can provide directly the json object I think

src/schema/submissions.ts
```diff
@@ -0,0 +1,176 @@
...
+      );
```
idoshamun, line 171: I think if `url` is marked as unique you can merge it down to a single save statement and if there will be a unique violation an error will be thrown and you can check the error number to confirm it's a unique violation

src/workers/cdc.ts
```diff
@@ -312,6 +330,23 @@ const onReputationEventChange = async (
...
 
+const onSubmissionChange = async (
+  con: Connection,
+  logger: FastifyLoggerInstance,
+  data: ChangeMessage<Submission>,
+) => {
+  const entity = data.payload.after;
+  if (data.payload.op === 'c') {
+    await notifySubmissionCreated(logger, {
+      sourceId: 'TBD',
```
idoshamun, line 341:
```suggestion
      sourceId: 'community',
```

idoshamun (changes requested):

sshanzel:
I just saw the email having the spec for the parameters. @idoshamun kindly hold the review until I implement them and see the values reflect in our test cases. Thank you!

sshanzel:
@idoshamun mails have been set. Should be ready to be reviewed 🙌

src/workers/communityLinkAccessMail.ts
```diff
@@ -0,0 +1,38 @@
+import { templateId } from '../common/mailing';
+import { messageToJson, Worker } from './worker';
+import { fetchUser } from '../common';
+import { baseNotificationEmailData, sendEmail } from '../common';
+
+interface Data {
+  userId: string;
+}
+
+// TODO:: Update the dynamic template data once available
```
rebelchris, line 10: I guess this is done by now?

src/workers/communityLinkRejectedMail.ts
```diff
@@ -0,0 +1,51 @@
...
+import { baseNotificationEmailData, sendEmail } from '../common';
+
+interface Data {
+  url: string;
+  status: string;
+  userId: string;
+  createdAt: string;
+}
+
+// TODO:: Update the dynamic template data once available
```
rebelchris, line 15: For reference

rebelchris (approved):
Some comments still found.
Furthermore, looks like we got everything, don't forget merge is blocked until we get the standardize url's in place.

src/workers/communityLinkAccessMail.ts
sshanzel: The dangerous TODO comments 🤦
Will sort this out. Thank you!

rebelchris:
Decided to introduce standardizeURLs for submissions entries: https://github.com/dailydotdev/daily-api/pull/935/commits/7b6d8f4941cff428b9a31db863237e512a74d352

src/workers/communityLinkRejectedMail.ts
```diff
@@ -7,22 +9,27 @@ interface Data {
...
-// TODO:: Update the dynamic template data once available
-
 const worker: Worker = {
   subscription: 'community-link-rejected-mail',
   handler: async (message, con, logger): Promise<void> => {
     const data: Data = messageToJson(message);
+
+    if (data.status !== SubmissionStatus.Rejected) {
+      return;
+    }
```
idoshamun, line 22: I think it's redundant. This topic contains only rejected submissions

src/workers/cdc.ts
```diff
@@ -165,6 +172,10 @@ const onUserChange = async (
     ) {
       await notifyDevCardEligible(logger, data.payload.after.id);
     }
+
+    if (data.payload.after.reputation >= submissionAccessThreshold) {
```
idoshamun, line 176: I just noticed that this will always get published after the user reaches the threshold. There should be another check that the before value is less than the threshold. Please add tests accordingly.

idoshamun (changes requested):
One business logic comment that we need to address. Please verify again the business logic.

src/workers/cdc.ts
sshanzel: Totally agree. Now that we need a persistent value for this case. What do you think about adding a property in our `alerts` table (or anywhere else that fits) for checking if an email was sent already?

sshanzel: Ah, yes. But I've also thought of an edge case where toggling the upvote of someone's comment that would make them pass the threshold would re-send the email, again and again. Should we cover that as well? Or is it something we can pass on?

sshanzel: Ah, yes. We shouldn't send that unnecessary property. Now that I've checked SendGrid's API, and they don't have anything that would allow us to check an email attempt programmatically, the option I am thinking of now is at the user's table.

Like for DevCard, we will create `sentSubmissionEligibility`. What do you think?

sshanzel: Great idea. I think we can create an `email_events` (or `single_email_events`) table - from there we will have the user id (PK) and template id as another column. If the said template id was stored, then that means, it was sent already. What do you think?

idoshamun: First of all, don't think it has to do only with emails. It can be potentially anything. At this point, we don't add anything to the email other than the first name. So I wouldn't hassle to think about payloads. It can even be something like UserState entity. It doesn't need to store the template id. Just user id, key, and value. According to the key, you should know how to proceed. In our case, it will be emailed, but it can be any other thing as well

src/workers/cdc.ts
```diff
@@ -165,6 +173,19 @@ const onUserChange = async (
     ) {
       await notifyDevCardEligible(logger, data.payload.after.id);
     }
+
+    if (
+      data.payload.after.reputation >= submissionAccessThreshold &&
+      data.payload.before.reputation < submissionAccessThreshold
```
rebelchris, line 179:
```suggestion
      data.payload.before.reputation <= submissionAccessThreshold
```

In my test I was using a reputation of 20, threshold of 21.
Then when updating it would not accept it as my before was not matching.

Seeing we are going to implement retro-active we should consider people being on exactly 250 already.

Also what will happen if they already on reputation above the threshold?
They will never be able to pass this.

rebelchris (changes requested):
Doing single comments as I go through the tests.

src/workers/cdc.ts
sshanzel: > In my test I was using a reputation of 20, threshold of 21.
Then when updating it would not accept it as my before was not matching.

I think the current condition is correct. If the threshold is 20 and your previous reputation is already 20, then you already passed before. Only jumping from 19 to 20 should make it triggered.

> Also what will happen if they already on reputation above the threshold?

We would only send emails to those who did not have access yet and then passed afterward. I think this is a great question if we will send emails to those who already passed on launch. WDYT? @idoshamun

infra/application.properties
```diff
@@ -7,7 +7,7 @@ debezium.source.database.user=%database_user%
 debezium.source.database.password=%database_pass%
 debezium.source.database.dbname=%database_dbname%
 debezium.source.database.server.name=api
-debezium.source.table.include.list=public.comment,public.comment_upvote,public.comment_mention,public.source_request,public.upvote,public.post,public.user,public.post_report,public.alerts,public.source_feed,public.settings,public.reputation_event
+debezium.source.table.include.list=public.comment,public.comment_upvote,public.comment_mention,public.source_request,public.upvote,public.post,public.user,public.post_report,public.alerts,public.source_feed,public.settings,public.reputation_event,public.submission
```
rebelchris, line 10:
```suggestion
debezium.source.table.include.list=public.comment,public.comment_upvote,public.comment_mention,public.source_request,public.upvote,public.post,public.user,public.post_report,public.alerts,public.source_feed,public.settings,public.reputation_event,public.submission,public.user_state
```

rebelchris (changes requested):

src/workers/cdc.ts
idoshamun: I agree with @sshanzel. I think the current condition is right. We will manually send emails to those who already qualified on launch.

src/workers/cdc.ts
```diff
@@ -173,8 +174,17 @@ const onUserChange = async (
...
+    ) {
+      try {
+        await con.getRepository(UserState).insert({
+          userId: data.payload.after.id,
+          key: UserStateKey.CommunityLinkAccess,
+          value: true,
+        });
+      } catch (ex) {}
```
idoshamun, line 187: That's not a good practice to ignore every error. We want to ignore only the duplicate violation error, otherwise throw the error

idoshamun (changes requested):
One last change :)

src/workers/postScoutMatchedSlack.ts
```diff
@@ -0,0 +1,49 @@
...
+  post: ChangeObject<Post>;
+}
+
+const worker: Worker = {
+  subscription: 'post-scout-matched-slack',
+  handler: async (message, _, logger): Promise<void> => {
+    const data: Data = messageToJson(message);
+    const { post } = data;
+    try {
+      if (!post.scoutId) {
```
rebelchris, line 16: Locally getting the following error:
`"Cannot read properties of undefined (reading 'scoutId')",`

rebelchris (changes requested):

src/workers/postScoutMatchedSlack.ts
idoshamun: Have you applied the latest migration?

rebelchris (approved):
Wow Lee, testing this was quite a hell, so can only imagine what building it was like!
Well done buddy 💪

My only concern is around the creation of the topic: `community-link-submitted`.
cc: https://dailydotdev.slack.com/archives/C03C3JZPRU4/p1655210888156579

sshanzel:
@rebelchris thank you so much for the effort of manually testing each and every corner! 🙇 🟢

idoshamun (approved):
I'm missing a CDC test that checks for duplicates user states and that it is handled correctly.
You can also add test to check that other errors get thrown. I'm approving but make sure to add this test

sshanzel:
@idoshamun I just pushed the test for both cases. Have tested it locally as well and it was as expected though the CI Pipeline is having issues with memory 😭

sshanzel:
@idoshamun @rebelchris I have now increased the memory for testing phase. Let me know if you have concerns or anything related to it. Thank you!

Reference: https://support.circleci.com/hc/en-us/articles/360009208393-How-Can-I-Increase-the-Max-Memory-for-Node-

Memory increased to 6GB.

Default values:
image

__tests__/workers/cdc.ts
```diff
@@ -452,6 +452,49 @@ describe('user', () => {
...
+    try {
+      await expectSuccessfulBackground(
+        worker,
+        mockChangeMessage<ObjectType>({
+          after,
+          before: base,
+          op: 'u',
+          table: 'user',
+        }),
+      );
```
idoshamun, line 471: You should add a step here that makes sure you don't reach this step. Otherwise, the background worker may be completed successfully, and you wouldn't even know

idoshamun (approved):
Awesome just apply my fix and merge away!

PR #938, opened 2022-05-13
DD-587: internal feed generator support

vpol:


idoshamun (approved):
Exciting to try it out already

PR #940, opened 2022-05-15
DD-587: http endpoint support

vpol:


src/personalizedFeed.ts
```diff
@@ -1,16 +1,29 @@
...
 import { runInSpan } from './trace';
+import http from 'node:http';
+import https from 'node:https';
 
 interface TinybirdResponse<T> {
   data: T[];
 }
 
-const agent = new Agent({ keepAlive: true });
+const fetch_options = {
```
idoshamun, line 14: We use camelCase in the API for variables

src/personalizedFeed.ts
```diff
@@ -1,16 +1,29 @@
...
+  agent: (_parsedURL) => {
```
idoshamun, line 15: No need for the underscore prefix

idoshamun (commented):
Just some conventions. But solution is neat!

idoshamun (approved):
Perfect!

PR #945, opened 2022-05-25
refactor: alert property for add companion popup

sshanzel:
The add companion popup requires to be shown on load for logged-in users but must go back to default once closed or any action was taken inside.

### DIFF
--- a/__tests__/__snapshots__/alerts.ts.snap
+++ b/__tests__/__snapshots__/alerts.ts.snap
@@ -3,8 +3,8 @@
 exports[`mutation updateUserAlerts should create user alerts when does not exist 1`] = `
 Object {
   "updateUserAlerts": Object {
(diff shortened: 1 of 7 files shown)


rebelchris (commented):
As far as I understand existing users should only see it.
So perhaps the default value should be false, and we set it to true for everyone now?

But not sure I understood it correctly now.
CC: @idoshamun ?

sshanzel:
We always do the check for if the user is logged in we will do nothing. We won't have any trigger to make it to true so had to make it default as true and then change it to false as necessary.

idoshamun:
Didn't we decide on Slack to use local flag that we will set using a background worker instead? I think it's better than alerts eventually

sshanzel:
Oh, cool. Let me try and manage it locally. Thanks!

PR #946, opened 2022-05-25
feat: standardize urls

rebelchris:
Up until now we standardises URLs on the app directory.
As this is an ever-growing list of items we decided the API would be a better place.

For now the only site allowing querystrings is youtube.

WT-58


src/common/standardizeUrl.ts
```diff
@@ -0,0 +1,20 @@
+const excludeFromStandardization = ['youtube.com'];
```
sshanzel, line 1: I think we already have a file that contains all `url` / `link` related functions. Might be a good idea to move this all to there: https://github.com/dailydotdev/daily-api/blob/master/src/common/links.ts

sshanzel (approved):
Just had a suggestion but all looks solid to me! 🪨

PR #947, opened 2022-05-27
redis pool

vpol:


src/redis.ts
```diff
@@ -1,13 +1,14 @@
...
 export const redisOptions: RedisOptions = {
   host: process.env.REDIS_HOST,
   port: parseInt(process.env.REDIS_PORT, 10),
   password: process.env.REDIS_PASS,
 };
 
-export const redisClient: Redis.Redis = new Redis(redisOptions);
+// export const redisClient: Redis.Redis = new Redis(redisOptions);
 
 export const redisPubSub = new RedisPubSub({
```
idoshamun, line 13: Can't we migrate this as well and get rid of `ioredis`?

idoshamun (changes requested):
ts-ioredis-pool uses an old version of ioredis, should we fork it in the meantime and update the ioredis version? I see that the latest release of this package is 7 months ago

src/redis.ts
vpol: fixed. is it how you wanted it?

package.json
```diff
@@ -51,13 +53,12 @@
...
     "graphql-iso-date": "^3.6.1",
     "graphql-parse-resolve-info": "^4.12.0",
-    "graphql-redis-subscriptions": "^2.4.0",
     "graphql-relay": "^0.9.0",
     "graphql-subscriptions": "^2.0.0",
     "graphql-tools": "^4.0.8",
     "he": "^1.2.0",
     "highlight.js": "^11.3.1",
-    "ioredis": "^4.28.2",
+    "ioredis": "^5.0.5",
```
idoshamun, line 61: I think we don't need this package directly anymore. We use it indirectly through other packages, don't we?

idoshamun (changes requested):

idoshamun (changes requested):

idoshamun (approved):
Awesome!

PR #952, opened 2022-06-03
feat: add a preconnect link to redirector

omBratteng:
This will hopefully reduce the time waiting between a 200 response and when the redirect is completed, by eagerly trying to preconnect.

I have not tested this, and it is only a concept. And not sure how the code should be formatted, I only did this via GitHub edit file page 🙃


omBratteng:
Another theory I have, it would be faster to use Location headers, rather than http meta refresh for redirecting. But not sure if the html meta tag is required for some other stuff.

/cc @rebelchris

omBratteng:
Closing this, after more investigation, HTTP redirect is a more performant method, even though it could shave some milliseconds off of redirects.

Superseded by https://github.com/dailydotdev/daily-api/pull/953

idoshamun:
After some tests that I did locally, I decided to proceed with this PR as this method changes the referrer to `api.daily.dev` as expected. I believe we updated some of our packages that changed the referrer-policy defaults. This is why we couldn't see the right `document.referrer` in the previous implementation, before that it was ok.
When using 302 status, the referrer never changes to `api.daily.dev` because this is the browser policy.

Anyway, thank you very much for this! 🙏

omBratteng:
Did some local testing too, and noticed that links opened via extension did not have referer sadly when using HTTP redirect, even though it was much faster, but it's a tradeoff, but the preconnect should improve it a bit.

PR #953, opened 2022-06-03
fix: uses HTTP headers for redirecting

omBratteng:
This could potentionally save around ~150ms for each redirect, as the user does not need to load the page, wait for html to be processed and then do a refresh redirect.

### DIFF
--- a/src/routes/redirector.ts
+++ b/src/routes/redirector.ts
@@ -31,6 +31,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
         );
       }
       return res
+        .header('Referrer-Policy', 'origin, origin-when-cross-origin')
         .status(302)
         .redirect(`${post.url}${req.query.a ? `#${req.query.a}` : ''}`);
     },
--- a/src/routes/redirector.ts
+++ b/src/routes/redirector.ts
@@ -31,12 +31,8 @@ export default async function (fastify: FastifyInstance): Promise<void> {
         );
       }
       return res
-        .type('text/html')
-        .send(
-          `<html><head><meta http-equiv="refresh" content="0;URL=${post.url}${
-            req.query.a ? `#${req.query.a}` : ''
-          }"></head></html>`,
-        );
+        .status(302)
+        .redirect(`${post.url}${req.query.a ? `#${req.query.a}` : ''}`);
     },
   );
 }


omBratteng:
I've probably dug too much into performance now. But this basically shows why using html redirect has idle time, it's doing all this during the load of the API endpoint, where it has a average idle time of 150ms before doing the refresh redirect.
image
image

This PR drops HTML refresh redirects, and uses HTTP header redirect instead, which does not give any idle time between connections. The browser treats it as one network connection, therefore not triggering any html parsing of the API endpoint
image

image

idoshamun:
Hi @omBratteng,

Wow! Thanks for this so thoughtful PR.
The primary reason we use html based redirects is that the page referrer on the target website will be set.
This way the target website can see in their analytics that traffic came from daily.dev. Because we're an extension there's no page referrer by default so without the redirects there will be no way to track traffic.
If you can confirm or find a way to use http redirects and set the page referrer as well, we can merge it and get the performance boost. 🚀

omBratteng:
I see,
Yes, the extension does cause some issues. However, I don't think you're getting any page referrers now, as they header `referrer-policy: no-referrer` is currently set.
```console
❯  curl -i -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.61 Safari/537.36' https://api.daily.dev/r/S68qnmOkU
HTTP/2 200
content-security-policy: default-src 'self';base-uri 'self';block-all-mixed-content;font-src 'self' https: data:;frame-ancestors 'self';img-src 'self' data:;object-src 'none';script-src 'self';script-src-attr 'none';style-src 'self' https: 'unsafe-inline';upgrade-insecure-requests
x-dns-prefetch-control: off
expect-ct: max-age=0
x-frame-options: SAMEORIGIN
strict-transport-security: max-age=15552000; includeSubDomains
x-download-options: noopen
x-content-type-options: nosniff
x-permitted-cross-domain-policies: none
referrer-policy: no-referrer
x-xss-protection: 0
vary: Origin
access-control-allow-credentials: true
content-type: text/html
content-length: 128
date: Sun, 05 Jun 2022 23:46:34 GMT
via: 1.1 google
alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000

<html><head><meta http-equiv="refresh" content="0;URL=https://blog.pragmaticengineer.com/shipping-to-production/"></head></html>
```

It is possible PR https://github.com/dailydotdev/daily-api/pull/952 could reduce it a little bit, if the preconnect is set. Either via HTTP header or http meta, think I saw ~150ms gain, compared to a ~300ms gain with HTTP headers.

idoshamun:
Today clicks are tracked under `api.daily.dev` if you look at google analytics or any other platform.
I'll have to check it locally before I approve.

idoshamun:
I decided to proceed with https://github.com/dailydotdev/daily-api/pull/952.
Thank you very much!

PR #954, opened 2022-06-06
fix: usage of redis scan function

vpol:


src/redis.ts
```diff
@@ -12,21 +12,19 @@ export const redisPubSub = new RedisPubSub({
...
-          stream.destroy();
-          resolve();
-        }
-      });
-      stream.on('end', resolve);
-      stream.on('error', reject);
-    });
-  });
+  return ioRedisPool.execute(
+    async (client) =>
```
idoshamun, line 16: Here I'm pretty sure TS will not complain because we return a promise

```suggestion
   (client) =>
```

idoshamun (approved):
One minor thing and we're ready to go

PR #960, opened 2022-06-13
chore: modify pg connection parameters

vpol:
- increased max connection number
- increased time until idle connection is closed

Signed-off-by: Viktor Poluksht <viktor@daily.dev>


vpol:
Yes please.

PR #963, opened 2022-06-16
chore: refactor mocked to jest.mocked

rebelchris:
We where using the old implementation of `mocked` and should use `jest.mocked`.


sshanzel (approved):
Nice! Finally there will be less noise in the logs when running the tests 🚢

idoshamun (approved):
Nice!!

PR #968, opened 2022-06-17
fix: missed imported worker for granting submission access email

sshanzel:


rebelchris (approved):
Let's go :)

idoshamun:
This is a good fix, but in this same PR, I'd like to see that we comment out the line where we publish a message to this topic.
We're now testing internally so I don't want users to get email about it when they can't really access it.

idoshamun (approved):

PR #969, opened 2022-06-18
fix: user mention beside a special character

sshanzel:
I have written numerous test cases to ensure all the edge cases will be covered. The behavior of when to allow the mention of what special character it was beside, was based on Twitter. On Twitter, they allow any special character so long as it is not the at (@) sign.

First, we replace the special characters with spaces to figure out which ones are valid mentions of users.

Then while we check each word for any mention, we replace again the special characters but this time we store them in an array so we can easily reconstruct the word for which one was replaced.

If the mention was found we will generate the anchor tag with it.

### DIFF
--- a/src/common/markdown.ts
+++ b/src/common/markdown.ts
@@ -23,12 +23,6 @@ export const getMentionLink = ({ id, username }: MarkdownMention): string => {
   return `<a href="${href}" data-mention-id="${id}" data-mention-username="${username}">@${username}</a>`;
 };
 
-export const isValidUsernameChar = (char: string): boolean => {
-  const match = char.match(/^[a-z0-9_]+$/i);
-
-  return match && match[0].length === 1;
-};
-
 const defaultRender =
   markdown.renderer.rules.link_open ||
   function (tokens, idx, options, env, self) {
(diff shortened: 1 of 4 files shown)


idoshamun:
It's somewhat hard to understand the logic you describe here in your code.
Another suggestion regarding your algorithm that might ease the process is before you split the original text by spaces, you can replace all special characters with spaces and then do the split. At the end of the process, you can change the relevant spaces to the special characters. This way you will not have to deal with them in the middle of the process because you can assume they are spaces

sshanzel:
I think that can really work, the only problem we have is when we are going to map the replaced special characters to the right values after processing because the content was reconstructed and doesn't have the same length anymore. We would again have to split it by special characters to make it an array so we can track the position of the special characters it found. WDYT?

idoshamun:
I'll leave it to you to decide what's the best approach, if you want to talk about I'm available today. Either way, we need to make the implementation more readable and maintainable

sshanzel:
@idoshamun I really agreed with your comment about making it much simpler. I applied what you suggested though I had to make the replacement while processing the word to make reconstructing the replaced words much easier (mapping). WDYT with the current implementation?

src/common/markdown.ts
```diff
@@ -43,16 +43,49 @@ const defaultTextRender = markdown.renderer.rules.text;
...
+type ReplacedCharacters = string[];
+
+const getReplacedCharacters = (word: string): [string, ReplacedCharacters] => {
+  const specialCharacters = [];
+  word.split('').forEach((char) => {
+    const matched = char.match(mentionSpecialCharacters);
+    if (matched?.[0]?.length) {
+      specialCharacters.push(char);
+    }
+  });
```
idoshamun, line 57: You can use something like this: https://stackoverflow.com/a/2295681/2650104
Instead of splitting every character and doing a regex match operation

src/common/markdown.ts
```diff
@@ -43,16 +43,49 @@ const defaultTextRender = markdown.renderer.rules.text;
 
 type MarkdownMention = Pick<User, 'id' | 'username'>;
 
+export const mentionSpecialCharacters = new RegExp('[^a-zA-Z0-9_@]', 'g');
+
+type ReplacedCharacters = string[];
+
+const getReplacedCharacters = (word: string): [string, ReplacedCharacters] => {
```
idoshamun, line 50: I suggest adding here a short comment explaining about the logic of this function

idoshamun (approved):
I added a few ideas for improvement but you can merge at will :)

PR #976, opened 2022-06-24
fix: prevent duplicate rejection

rebelchris:
Currently it's possible for the gatekeeper to send duplicate request to addNewPost and rejectPost routes.
The addNewPost will simply return that the post already exists.

However with the rejection we kept updating the status to Rejected even if it already was approved or rejected.
This fix prevents duplicate resets.

WT-174 #done

### DIFF
--- a/__tests__/private.ts
+++ b/__tests__/private.ts
@@ -172,7 +172,7 @@ describe('POST /p/rejectPost', () => {
     expect(submission.status).toEqual(SubmissionStatus.Rejected);
   });
 
-  it('should not update already rejected post', async () => {
+  it('should not update already approved post', async () => {
     const uuid = randomUUID();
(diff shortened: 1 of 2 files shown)


__tests__/private.ts
```diff
@@ -172,6 +172,30 @@ describe('POST /p/rejectPost', () => {
     expect(submission.status).toEqual(SubmissionStatus.Rejected);
   });
 
+  it('should not update already rejected post', async () => {
```
idoshamun, line 175:
```suggestion
  it('should not update already approved post', async () => {
```

idoshamun (approved):
A minor change in the test name and it's ready to go!

kirillkurko (approved):

PR #980, opened 2022-06-30
feat: introducing generic error messages for community links

rebelchris:
We decided to remove string variants of error messages for community links.

As we don't have a clear plan for error codes we decided to move forward with the introduction of generic error keys.
The API manages these API keys, and we made a first start by introducing community links related keys.

Related PR:
Gatekeeper: https://github.com/dailydotdev/gatekeeper/pull/184


vpol:
I haven't noticed db migration, but there is very few rows with non-empty reason so I guess we can skip it.

vpol (approved):

PR #981, opened 2022-07-04
fix: format reject mail date

rebelchris:
The rejection date was date formatted to date causing a duplicate conversion to happen.

WT-199 #done

### DIFF
--- a/src/workers/communityLinkRejectedMail.ts
+++ b/src/workers/communityLinkRejectedMail.ts
@@ -5,7 +5,6 @@ import { fetchUser } from '../common';
 import { baseNotificationEmailData, sendEmail } from '../common';
 
 interface Data {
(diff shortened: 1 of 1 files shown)


sshanzel (commented):
I think we should also make changes in the line below - because it is the part where it converts the data.
https://github.com/dailydotdev/daily-api/blob/63444375c15352ebfb596a6530cf5f23f51be397/src/workers/communityLinkRejectedMail.ts#L25

rebelchris:
@sshanzel That one no longer needs a change.
If we receive the object as Date we simply have to parse it with new Date().

I had this before, but this way was more fail-safe for both string/date types.

sshanzel (approved):
If that's the case, then all looks good to me! 🚢

PR #982, opened 2022-07-05
fix: convert unix timestamp to actual dates

rebelchris:
After deploying the fix it still seemed to give issues on production.
Inspected the subscription payload and noticed it actually gets send as a unix timestamp. (microseconds)

Screenshot 2022-07-05 at 10 01 33

Decided to have the test reflect the same value so this won't happen in the future.

WT-200 #done


sshanzel (approved):
Nice find!

kirillkurko (approved):
Nice catch!

PR #988, opened 2022-07-10
chore: cleanup unused dev card eligible setting and associates

rebelchris:
Removed the whole devcard_eligible column from the user table
And everything associated with it, including the email we send to users when they reach this status.

Cleanup amplitude as we no longer support it.

WT-229 #done


sshanzel (approved):
It is always so nice to see codebase getting cleaned up. Thank you so much for the effort!

Just one note, do you think we should add a stub value on the returned value from our queries just so it will be compatible with any un-updated consumers?

rebelchris:
@sshanzel I thought about that, but currently 50% of the users would get `0` and the other half `50`.
The default in the featureManagement is also `50` so don't think it would really make much difference.
What do you think?

sshanzel:
Was thinking about the querying part, the un-updated client might receive an error if they asked for the property but are non-existent in the query anymore.

sshanzel:
> Was thinking about the querying part, the un-updated client might receive an error if they asked for the property but are non-existent in the query anymore.

Nevermind this comment. I realized it wasn't actually part of any query.

src/entity/User.ts
```diff
@@ -51,9 +51,6 @@ export class User {
   @Column({ length: 39, nullable: true })
   hashnode?: string;
 
-  @Column({ default: false })
-  devcardEligible: boolean;
```
idoshamun, line 55: I wouldn't recommend actually dropping this column. I'm not a fan of deleting data 😓

idoshamun (changes requested):
There's nothing like deleting code! 💪

src/entity/User.ts
rebelchris: No worries, let me put the column back and remove the migration for now :)

PR #989, opened 2022-07-11
fix: the approval link was messed up in formatting

rebelchris:
Following our other email guidelines there seems to be some formatting issue in the way we send it.
This fixes the community links approval email to contain valid links.

Also formatted the image differently on the sendgrid side so it will reflect.

WT-225 #done


idoshamun (approved):
I'm not sure what's the difference. It seems the same but different order.
I'm approving because I know you checked it

PR #993, opened 2022-07-13
refactor: author emails to send on scout

sshanzel:
We send emails to authors whenever a user has commented on their post or commented in a comment. These emails must also be sent to the scout if there are any.

The reason why the analytics was not being sent since then is, that we have a cron job that checks for post analytics to send, and if it found any, it will automatically set the property `sentAnalyticsReport` to true. Which will then trigger a CDC pubsub notification to our worker.

But then on the actual worker, we have a condition it will only send if the value is false, but the actual reason why this worker was triggered, was due to the value being true.

https://github.com/dailydotdev/daily-api/pull/993/files#diff-21447dd107efb6a2c1bc5f6b8efd19a811618dbd67a092324739f5cccb954196L22

### DIFF
--- a/src/workers/postCommentedAuthor.ts
+++ b/src/workers/postCommentedAuthor.ts
@@ -2,9 +2,10 @@ import { messageToJson, Worker } from './worker';
 import { Comment } from '../entity';
 import {
   getCommentedAuthorMailParams,
-  getCommenterAuthorScout,
+  getAuthorScout,
   hasAuthorScout,
   sendEmail,
+  fetchUser,
 } from '../common';
 
 interface Data {
@@ -29,12 +30,13 @@ const worker: Worker = {
         return;
       }
 
-      const requests = getCommenterAuthorScout(data.userId, post);
+      const requests = getAuthorScout(post, [data.userId]);
 
       if (requests.length === 0) {
         return;
       }
 
+      requests.unshift(fetchUser(data.userId));
       const [commenter, ...authorScout] = await Promise.all(requests);
 
       await Promise.all(
--- a/src/workers/postCommentedAuthor.ts
+++ b/src/workers/postCommentedAuthor.ts
@@ -21,22 +21,33 @@ const worker: Worker = {
         return;
       }
       const post = await comment.post;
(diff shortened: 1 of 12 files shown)


src/workers/postCommentedAuthor.ts
```diff
@@ -21,22 +21,40 @@ const worker: Worker = {
         return;
       }
       const post = await comment.post;
+      if (!post?.authorId && !post?.scoutId) {
```
rebelchris, line 24: Can we not generalise this somewhere?
The behaviour is twice the same in this and the commentComment?

src/workers/sendAnalyticsReportMail.ts
```diff
@@ -19,40 +19,58 @@ const worker: Worker = {
...
-          },
-        });
-        logger.info(
-          {
-            data,
-            messageId: message.messageId,
-          },
-          'analytics report sent',
-        );
+      if (!post.authorId && !post.scoutId) {
```
rebelchris, line 22: should we not add `post.sentAnalyticsReport` here? In case it's false?

rebelchris (changes requested):
Nice find on the analytics report.

Have some small remarks, but great work on this Lee.

src/workers/postCommentedAuthor.ts
sshanzel: Good point! Let me see how I can refactor it. Thank you!

src/workers/sendAnalyticsReportMail.ts
sshanzel: I agree, it would ensure we won't be sending any email if not necessary.

idoshamun (commented):
I think instead of sending two requests to sendgrid, we can send one request with multiple sets of parameters. Can you check it, please?

sshanzel:
> I think instead of sending two requests to sendgrid, we can send one request with multiple sets of parameters. Can you check it, please?

Looks like we can send an array! 🚢

rebelchris (approved):

PR #995, opened 2022-07-14
feat: source request submitted email

rebelchris:
Introducing a change from the old mailer function to include the source request mail in the API.

Small question:
I thought there was another place where we had to introduce these topics, am I mistaken?

Note:
We also might want to move 
- decline
- approve 

emails to be in the API so we can close the old mailer function

WT-251 #done


kirillkurko (approved):

sshanzel (approved):
Nice! I think you added the pubsub topic in the right place.

idoshamun:
@rebelchris, I agree that it is better to migrate the mailer functionality completely to the API.
I don't think we need to introduce a new topic here, as the mailer already subscribes to existing topics.
We need to add new subscribers to them and then delete the mailer.

rebelchris:
@idoshamun I kept the integration the same so the one file can send multiple mailers related to the source requests.
This way we keep the same topic and can remove the mailer from the daily-functions repo.

src/workers/index.ts
```diff
@@ -30,6 +30,7 @@ import usernameChanged from './usernameChanged';
 import updateComments from './updateComments';
 import deleteUser from './deleteUser';
 import cdc from './cdc';
+import sourceRequestSubmittedMail from './sourceRequestMail';
```
idoshamun, line 33:
```suggestion
import sourceRequestMail from './sourceRequestMail';
```

src/workers/index.ts
```diff
@@ -47,6 +48,7 @@ export const workers: Worker[] = [
   postScoutMatchedMail,
   communityLinkAccessMail,
   communityLinkRejectedMail,
+  sourceRequestSubmittedMail,
```
idoshamun, line 51:
```suggestion
  sourceRequestMail,
```

idoshamun (approved):
I usually don't like to introduce breaking changes to a pubsub message, but in this case it might be fine. Please make sure that no one else beside the mailer subscribes to the topic before you merge it. You can use GCP UI for that

rebelchris:
@idoshamun for your eyes, I revamped the source request reputation subscription to reflect the new flow.

idoshamun (approved):

PR #998, opened 2022-07-19
feat: adds new tags following the OneAI generated tags

chetanbasuray:
Added new tags
Added new algorithm so that tags are extended by synonyms and not overwritten as before


idoshamun (changes requested):
As for the migration, I think it's better to batch insert them in the migration instead of running a single insert per query. Is it possible?

chetanbasuray:
> As for the migration, I think it's better to batch insert them in the migration instead of running a single insert per query. Is it possible?

Checking this.

src/entity/Post.ts
```diff
@@ -314,6 +314,10 @@ const mergeKeywords = async (
         })
         .filter((keyword) => !keyword.match(/^\d+$/)),
     );
+    // with the new system, a keyword needs to be added irrespective of a synonym
+    keywords.forEach((keyword) => {
```
idoshamun, line 318: I think that we can simplify the process from line 307 to here.
What I suggest is to use the provided keywords array add to it the synonym that we get from the database, make sure everything is unique using `uniqueifyArray` and then filter using
`filter((keyword) => !keyword.match(/^\d+$/))`

So it should be roughly something like:
```
const additionalKeywords = synonymKeywords.map((synonym) => synonym.synonym)
const mergedKeywords = uniqueifyArray([...keywords, ...additionalKeywords].filter((keyword) => !keyword.match(/^\d+$/)))
```

I hope it makes sense. I didn't test the code

src/migration/1658242193484-AddNewOneAITags.ts
```diff
@@ -0,0 +1,1387 @@
...
+            ('dash','deny'),
+            ('congress','deny'),
+            ('electricity','deny'),
+            ('smashing','deny'),
+            ('lambda-function','deny'),
+            ('ui','deny') 
+            ON CONFLICT ("value") DO UPDATE SET status = EXCLUDED.status`,
+            );
+
+            await queryRunner.query(
```
idoshamun, line 653: I think this list should be `allow` not `pending`

chetanbasuray: Made changes

idoshamun:
I found another mistake in my spreadsheet file. `open-api` is a synonym of `openapi`. Please adjust the migration file, sorry for the trouble

src/entity/Post.ts
chetanbasuray: Updated

chetanbasuray:
@idoshamun I have updated the migration file with the last change you requested and also added a test to check if the keywords string is being generated as it should be and if the `keywords` table is getting updated.

idoshamun (approved):
Woohoo!!

PR #999, opened 2022-07-22
chore: update minors and patches

rebelchris:
Updated all non breaking packages

We really need to upgrade to fastify 4+, but it's dependency tree is super difficult to resolve.
So wanted to do this one first as a separate PR.


idoshamun (approved):
Thanks for doing it over and over! 🙏

PR #1001, opened 2022-07-27
feat: adds a log statement for all URLs for which more than 5 Keywords are selected

chetanbasuray:


idoshamun (changes requested):
The log statement looks great. But it's better to not create a log instance on every call. So if you can add the logger as a parameter to the function it would be great

chetanbasuray:
> The log statement looks great. But it's better to not create a log instance on every call. So if you can add the logger as a parameter to the function it would be great

Updated

src/routes/private.ts
```diff
@@ -8,14 +8,16 @@ import {
...
 
 export default async function (fastify: FastifyInstance): Promise<void> {
   fastify.post<{ Body: AddPostData }>('/newPost', async (req, res) => {
     if (!req.service) {
       return res.status(404).send();
     }
     const con = getConnection();
-    const operationResult = await addNewPost(con, req.body);
+    const logger = pino();
+    const operationResult = await addNewPost(con, req.body, logger);
```
idoshamun, line 20: The request object should already have a logger defined. No need to define a new instance.
In most cases you don't have to define an instance as a rule of thumb

idoshamun (changes requested):

src/routes/private.ts
chetanbasuray: Updated to use the logger from the req object.

idoshamun (approved):
Perfect!

PR #1003, opened 2022-07-31
feat: check user if exists by email

sshanzel:
After working with the error management of the Ory Kratos for registration, it won't be possible for us to check if the email exists until the user has filled all the required fields. Unless we will create a custom endpoint to check it there.

Kratos would only allow us to check if the identity exists if we have filled all the other traits, but then if we have filled it (random values) and it is not found, it will proceed to create an account.

### DIFF
--- a/__tests__/__snapshots__/users.ts.snap
+++ b/__tests__/__snapshots__/users.ts.snap
@@ -308,6 +308,14 @@ Object {
 }
 `;
 
+exports[`query userByEmail should return user info with name, username, and image 1`] = `
+Object {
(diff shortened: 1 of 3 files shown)


idoshamun:
This endpoint opens us to security concerns such as user enumeration. I highly advise that we will utilize Kratos and see if there's any flow that can work for this case rather than add new endpoints

sshanzel:
cc @rebelchris @vpol

rebelchris:
Also think the email check should be done inside kratos.
We should only have a username check on the API, but where possible only accessible for Kratos?

PR #1008, opened 2022-08-03
fix: non exposed comment thread emails

rebelchris:
We accidentally exposed multiple recipients by using the `to` field.
Now we introduce a personalisation option to send singular emails to each individual subscribed to the thread.


idoshamun:
@rebelchris way to go! Can you please validate that we didn't do this mistake in other places as well?
Also have you checked that it works?

omBratteng:
Awesome, nice to see fix implemented so quickly from reporting it.

rebelchris:
@idoshamun Had a look at the to: parameter and looks like this was the only one that had it in place.
Unless we still have external apps using the mailers.

idoshamun (approved):

PR #1009, opened 2022-08-03
feat: introducing whoami dedicated route

rebelchris:
Introducing a whoami call that the gateway can use to retrieve user profile from the API.

WT-296


src/common/users.ts
```diff
@@ -18,6 +18,7 @@ export interface User {
   email: string;
   name: string;
   image: string;
+  infoConfirmed: boolean;
```
kirillkurko, line 21: Maybe it can be just `confirmed` or `verified`, what do you think?

kirillkurko (approved):

src/common/users.ts
idoshamun: We must keep it `infoConfirmed` for backwards compatibility

idoshamun (approved):

rebelchris:
@idoshamun I just went over all gatewayUser fixtures and changed them to be loaded from 1 origin.
This is a bit more maintable for when we add/remove fields later on.

Also this one seems like we can already merge, doesn't introduce anything breaking for the current version.

idoshamun (approved):
I agree, this shouldn't cause any breaking changes. We can safely merge

PR #1013, opened 2022-08-11
fix: user replication

idoshamun:
* Parse correctly boolean fields (`infoConfirmed` and `acceptedMarketing`)
* Add missing column `referral`
* Increase username length as a bonus :P

WT-317


rebelchris (approved):
With this referral we have to implement it in user creation as well? 👀

idoshamun:
@rebelchris you're right! Although it's legacy I think we should not break it for now

PR #1016, opened 2022-08-15
feat: updating account profile

sshanzel:
Since the source of truth is now for the profile information, we will be having an API to allow updating it.

Question:
  - Should we now remove the subscription on the gateway for watching changes in the user?
  - Would we want to instead create a worker to send a post request to the Kratos server? (current one sends a post request in the FE)

### DIFF
--- a/src/entity/User.ts
+++ b/src/entity/User.ts
@@ -42,23 +42,25 @@ export class User {
   reputation: number;
 
   @Column({ length: 39, nullable: true })
-  @Index()
+  @Index('users_username_unique', { unique: true })
   username?: string;
 
   @Column({ type: 'text', nullable: true })
   bio?: string;
 
   @Column({ length: 15, nullable: true })
-  @Index()
+  @Index('users_twitter_unique', { unique: true })
   twitter?: string;
 
   @Column({ length: 39, nullable: true })
+  @Index('users_github_unique', { unique: true })
   github?: string;
 
   @Column({ type: 'text', nullable: true })
   portfolio?: string;
 
   @Column({ length: 39, nullable: true })
+  @Index('users_hashnode_unique', { unique: true })
   hashnode?: string;
 
   @Column({ default: false })
--- a/src/migration/1660640181400-UniqueSocialHandles.ts
+++ b/src/migration/1660640181400-UniqueSocialHandles.ts
@@ -5,6 +5,12 @@ export class UniqueSocialHandles1660640181400 implements MigrationInterface {
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
+      `DROP INDEX "public"."IDX_b67337b7f8aa8406e936c2ff75"`,
+    );
+    await queryRunner.query(
+      `DROP INDEX "public"."IDX_d7820fad49c99eb918d92519e2"`,
+    );
+    await queryRunner.query(
       `CREATE UNIQUE INDEX "users_username_unique" ON "user" ("username") `,
     );
     await queryRunner.query(
@@ -23,5 +29,11 @@ export class UniqueSocialHandles1660640181400 implements MigrationInterface {
     await queryRunner.query(`DROP INDEX "public"."users_github_unique"`);
     await queryRunner.query(`DROP INDEX "public"."users_twitter_unique"`);
     await queryRunner.query(`DROP INDEX "public"."users_username_unique"`);
+    await queryRunner.query(
+      `CREATE INDEX "IDX_d7820fad49c99eb918d92519e2" ON "user" ("twitter") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_b67337b7f8aa8406e936c2ff75" ON "user" ("username") `,
+    );
   }
 }
--- a/src/migration/1660640181400-UniqueSocialHandles.ts
+++ b/src/migration/1660640181400-UniqueSocialHandles.ts
@@ -1,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UniqueSocialHandles1660640181400 implements MigrationInterface {
+  name = 'UniqueSocialHandles1660640181400';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX "users_username_unique" ON "user" ("username") `,
+    );
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX "users_twitter_unique" ON "user" ("twitter") `,
+    );
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX "users_github_unique" ON "user" ("github") `,
+    );
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX "users_hashnode_unique" ON "user" ("hashnode") `,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "public"."users_hashnode_unique"`);
+    await queryRunner.query(`DROP INDEX "public"."users_github_unique"`);
+    await queryRunner.query(`DROP INDEX "public"."users_twitter_unique"`);
+    await queryRunner.query(`DROP INDEX "public"."users_username_unique"`);
+  }
+}
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -53,7 +53,7 @@ interface GQLUserParameters {
 export interface GQLUser {
   id: string;
   name: string;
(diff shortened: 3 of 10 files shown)


idoshamun:
At the moment we should not remove anything, just add new endpoints until we make the change

src/schema/users.ts
```diff
@@ -27,11 +31,29 @@ import { getSearchQuery } from './common';
...
+interface GQLUserParameters {
+  data: GQLUpdateUserInput;
+  upload: FileUpload;
+}
 
 export interface GQLUser {
   id: string;
   name: string;
-  image: string;
+  image?: string;
```
idoshamun, line 56: Why is this optional all of a sudden?

src/schema/users.ts
```diff
@@ -567,6 +648,44 @@ export const resolvers: IResolvers<any, Context> = {
...
+    updateUserProfile: async (
+      _,
+      { data }: GQLUserParameters,
+      ctx,
+    ): Promise<GQLUser> => {
+      const repo = ctx.con.getRepository(User);
+      const user = await repo.findOne({ id: ctx.userId });
+
+      if (!user) {
+        throw new AuthenticationError(
```
idoshamun, line 660: Let's just throw a forbidden error (or whatever is thrown when the user is not authenticated) with no explanation in this case. This will prevent leaking our database in extreme cases

src/schema/users.ts
```diff
@@ -567,6 +648,44 @@ export const resolvers: IResolvers<any, Context> = {
...
+      if (data.username && user.username != data.username) {
+        const takenUsername = await repo.findOne({ username: data.username });
+        if (takenUsername) {
+          throw new ValidationError(
+            JSON.stringify({ username: 'Username is already taken' }),
+          );
+        }
+      }
+
+      const links = ['twitter', 'github', 'hashnode', 'portfolio'];
```
idoshamun, line 674: I think that twitter and github are not stored as links.
It's better to copy the validation logic from the gateway

src/schema/users.ts
```diff
@@ -567,6 +648,44 @@ export const resolvers: IResolvers<any, Context> = {
         )}.png?r=${randomStr}`,
       };
     },
+    updateUserProfile: async (
```
idoshamun, line 651: I couldn't find where you handle the file upload of the image

idoshamun (changes requested):

src/schema/users.ts
sshanzel: In our actual User entity, this is optional. We don't handle the fallback in our response.

src/schema/users.ts
sshanzel: Oh, cool. It is `UNAUTHENTICATED` which I think is Authentication Error. Let me remove the error message.

src/schema/users.ts
sshanzel: Ah, thanks for the tip. Let me check that one.

sshanzel: Ah, thanks for the tip. Let me check that one. Though we use a library for validating the body, should we install it here as well?

idoshamun: No but we can use the same logic I guess

src/schema/users.ts
sshanzel: Kindly check if the recently pushed commit for file upload fits. Thank you!

src/entity/User.ts
```diff
@@ -42,23 +42,25 @@ export class User {
...
-  @Index()
+  @Index('users_username_unique', { unique: true })
   username?: string;
 
   @Column({ type: 'text', nullable: true })
   bio?: string;
 
   @Column({ length: 15, nullable: true })
-  @Index()
+  @Index('users_twitter_unique', { unique: true })
```
idoshamun, line 52: Why are you changing it to unique?
If so, we need to validate if there can be two users with twitter as null. Because null sometimes is also treated as unique. Can you verify please?

src/migration/1660640181400-UniqueSocialHandles.ts
```diff
@@ -0,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UniqueSocialHandles1660640181400 implements MigrationInterface {
+  name = 'UniqueSocialHandles1660640181400';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
idoshamun, line 6: I see that we had already index on `username` and `twitter`, but I don't see where we get rid of the old indexes here

src/schema/users.ts
```diff
@@ -567,6 +649,91 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const repo = ctx.con.getRepository(User);
+      const user = await repo.findOne({ id: ctx.userId });
+
+      if (!user) {
+        throw new AuthenticationError(
+          JSON.stringify({ auth: 'User not found' }),
+        );
+      }
+
+      const isUsernameValid = new RegExp(/^@?(\w){1,15}$/).test(data.username);
```
idoshamun, line 666: I extended it to 39 chars as well in my previous commit

```suggestion
      const isUsernameValid = new RegExp(/^@?(\w){1,39}$/).test(data.username);
```

src/schema/users.ts
```diff
@@ -567,6 +649,91 @@ export const resolvers: IResolvers<any, Context> = {
...
+        throw new ValidationError(
+          JSON.stringify({ twitter: 'twitter is invalid' }),
+        );
+      }
+
+      if (data.hashnode && !isHashnodeValid) {
+        throw new ValidationError(
+          JSON.stringify({ hashnode: 'hashnode is invalid' }),
+        );
+      }
```
idoshamun, line 695: Can we perhaps turn some of it into a reusable function or take it outside this function?
It feels like it adds some mess

src/schema/users.ts
```diff
@@ -567,6 +649,91 @@ export const resolvers: IResolvers<any, Context> = {
...
+
+      const avatar = upload
+        ? await uploadAvatar(user.id, upload.createReadStream())
+        : user.image;
+
+      try {
+        const result = await ctx.con
+          .getRepository(User)
+          .save({ ...user, ...data, avatar });
```
idoshamun, line 704: Make sure it works

```suggestion
          .save({ ...user, ...data, image: avatar });
```

idoshamun (changes requested):
Almost there...

src/entity/User.ts
sshanzel: In our previous validation in the gateway, we leveraged the uniqueness of the column.
Yes, I've read in postgresql docs being null in multiple rows is allowed. Sample test below:
image

src/migration/1660640181400-UniqueSocialHandles.ts
sshanzel: Oh, good catch. The migration result is a bit mess, I had to pick the new ones manually.

src/schema/users.ts
sshanzel: The ValidationError doesn't accept object but string only. So we have to stringify it.

sshanzel:
@idoshamun let me know what you think. Thank you for the review! 🚢

src/schema/users.ts
sshanzel: Yes. The only thing it gives back now is "must supply api_key" - which I think is available in prod environment. Or do we have to set it manually somewhere? Would love to be able to test it locally.

src/schema/users.ts
```diff
@@ -658,50 +664,28 @@ export const resolvers: IResolvers<any, Context> = {
...
-        );
+      const regexParams: ValidateRegex[] = [
+        ['username', data.username, new RegExp(/^@?(\w){1,39}$/)],
+        ['github', data.github, new RegExp(/^@?([\w-]){1,39}$/i)],
+        ['twitter', data.twitter, new RegExp(/^@?(\w){1,15}$/)],
+        ['hashnode', data.hashnode, new RegExp(/^@?([\w-]){1,39}$/i)],
+      ];
+      const regexResult = validateRegex(regexParams);
+      if (Object.keys(regexResult).length) {
+        throw new ValidationError(JSON.stringify(regexResult));
```
idoshamun, line 678: We just have to make sure the client knows how to parse this error and present it

idoshamun (approved):
One comment regarding the integration aspect. Also make sure to validate the cloudinary upload with the key i sent you privately.
Other than that, we're ready to roll. Make sure to monitor the migration once you merge

sshanzel:
@idoshamun have confirmed locally the upload works. And the PR dedicated to profile updates in FE is familiar with the error message it will be receiving. I'll monitor the deployment until it ticks all the checkboxes.

PR #1017, opened 2022-08-16
build: garden support

idoshamun:
Add garden project and module for the service. It spins up a kubernetes environment with all the dependencies for the service.
I created two new dockerfiles one for development which will be used for hot reload and the other for production builds using garden. We may want to consider using the multistage for real production workloads as well

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,5 @@
 .idea/
 .garden/
-build/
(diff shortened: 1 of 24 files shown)


idoshamun:
Mistakenly I didn't open it as a draft. But it's still a draft 😓

PR #1018, opened 2022-08-17
feat: modify user endpoint

rebelchris:
With regards to the frontend changes we no longer need to query the Gateway for user info.
This PR ensures we fetch users from the API directly (leveraging the new graphQL for it)

WT-298

Related to: https://github.com/dailydotdev/apps/pull/1141


src/schema/users.ts
```diff
@@ -154,9 +160,13 @@ export const typeDefs = /* GraphQL */ `
     """
     infoConfirmed: Boolean
```
idoshamun, line 161: I know that it's not part of this PR and I might have missed that, but this should be provided only to the same logged-in user. For example, if Lee is trying to get Chris's `infoConfirmed` field it should be null for him. Only if Chris tries to get Chris's `infoConfirmed` it should have the right value

src/schema/users.ts
```diff
@@ -491,7 +501,26 @@ export const resolvers: IResolvers<any, Context> = {
...
+        builder.queryBuilder = builder.queryBuilder
+          .andWhere(
+            `("${builder.alias}"."id" = :id OR "${builder.alias}"."username" = :id)`,
+            { id },
+          )
+          .limit(1);
+        return builder;
+      });
+      if (!res[0]) {
+        throw new ForbiddenError('user not found');
```
idoshamun, line 520: Do we throw Forbidden or not found usually?

src/common/users.ts
```diff
@@ -45,17 +45,6 @@ export const fetchUser = async (userId: string): Promise<User | null> => {
...
-    method: 'GET',
-    headers: authorizedHeaders(userId),
-  });
-  if (res.status !== 200) {
-    return null;
-  }
-  return res.json();
-};
-
 export const fetchUserInfo = async (userId: string): Promise<UserInfo> => {
```
idoshamun, line 48: `fetchUserInfo` is used mostly in the worker and needs to be updated.
`fetchUserRoles` I'm not sure where we used it, but it's worth checking.

idoshamun (commented):
I think we should expand the scope of this PR, please see my comments.
What you implemented so far looks great

src/schema/users.ts
rebelchris: We changed it, this is the general user call.
In the graphql adapter we made it forbidden error for the whoami call.

src/common/users.ts
rebelchris: Alright, let me work on those in this PR

src/schema/users.ts
idoshamun: In the whoami it makes sense because you are not logged in so you should receive forbidden if there's an issue.
But here the user doesn't exist, so it's better I think to throw NotFound

rebelchris:
@idoshamun Sorry, became quite big because the wide range of test cases that mocked the Gateway 😅
Addressed all issues mentioned.

As stated on Slack, we should consider making some fields required on a DB level at once stage.

src/common/users.ts
```diff
@@ -34,42 +29,12 @@ const authorizedHeaders = (userId: string): { [key: string]: string } => ({
...
-  const res = await fetch(`${process.env.GATEWAY_URL}/v1/users/me/roles`, {
-    method: 'GET',
-    headers: authorizedHeaders(userId),
-  });
-  return res.json();
+export const fetchUser = async (
+  userId: string,
+  con: Connection,
+): Promise<User | null> => {
+  const user = await con.getRepository(DbUser).findOneOrFail({ id: userId });
```
idoshamun, line 36: I see that if the user doesn't exist we simply return null not failing

src/entity/User.ts
```diff
@@ -88,6 +89,14 @@ export class User {
 
   @OneToMany(() => DevCard, (devcard) => devcard.user, { lazy: true })
   devCards: Promise<DevCard[]>;
+
+  permalink: string;
+  @AfterLoad()
+  setComputed() {
```
idoshamun, line 95: Nice!!

src/entity/User.ts
```diff
@@ -18,13 +19,13 @@ export class User {
   id: string;
 
   @Column({ type: 'text', nullable: true })
-  name?: string;
+  name: string;
```
idoshamun, line 22: We need to remember to generate a migration file afterwards

src/workers/communityLinkAccessMail.ts
```diff
@@ -25,6 +25,7 @@ const worker: Worker = {
           data.userId,
       );
     } catch (err) {
+      console.log(err);
```
idoshamun, line 28: Not needed

idoshamun (approved):
Some comments, overall looks great!

idoshamun (approved):

PR #1019, opened 2022-08-22
feat: user mailing list

sshanzel:
Migrated the workers from the gateway.

### DIFF
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -15,6 +15,14 @@ export const workers: Worker[] = [
     subscription: 'user-updated-api',
   },
   {
+    topic: 'user-updated',
+    subscription: 'user-updated-mailing',
+  },
+  {
+    topic: 'user-updated',
+    subscription: 'user-deleted-mailing',
+  },
+  {
     topic: 'user-deleted',
     subscription: 'user-deleted-api',
   },
--- a/src/common/mailing.ts
+++ b/src/common/mailing.ts
@@ -1,4 +1,5 @@
 import sgMail from '@sendgrid/mail';
+import client from '@sendgrid/client';
 import { MailDataRequired } from '@sendgrid/helpers/classes/mail';
 import { Comment, Post } from '../entity';
 import { getDiscussionLink } from './links';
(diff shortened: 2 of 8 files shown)


rebelchris (approved):
Nice work on this 🙏

src/common/mailing.ts
```diff
@@ -91,3 +92,95 @@ export const getCommentedAuthorMailParams = (
     },
   };
 };
+
+// taken from sendgrid library itself
+type HttpMethod =
```
sshanzel, line 97: The type-safety of sendgrid shouts when sending the parameters without casting the method to this type.

.infra/workers.ts
```diff
@@ -19,6 +19,14 @@ export const workers: Worker[] = [
...
     subscription: 'user-updated-api',
   },
+  {
+    topic: 'user-updated',
+    subscription: 'user-updated-mailing',
+  },
+  {
+    topic: 'user-updated',
+    subscription: 'user-deleted-mailing',
+  },
```
idoshamun, line 29: I think these subs already exist in the gateway. Am I right?
If so, Pulumi will not be able to create them. We might need to do some manual work here. Let's think about it later

idoshamun (changes requested):
I think we have to approach this PR a bit differently.
If we still rely on the events we receive from the gateway, we still rely on the gateway itself.
So I suggest publishing new events based on changes in the users table of the API and then update the mailing list accordingly.
What do you think? cc @rebelchris

sshanzel:
> I think we have to approach this PR a bit differently.
If we still rely on the events we receive from the gateway, we still rely on the gateway itself.
So I suggest publishing new events based on changes in the user's table of the API and then update the mailing list accordingly.

I actually initially thought of publishing an event here in the API (but was missing in this PR), but now that we have kept the user change event in the gateway, and kept the changes minimal, we should not need an additional worker for the moment. WDYT?

idoshamun:
Let's wait for the planning meeting and we decide :)

.infra/workers.ts
```diff
@@ -19,6 +19,14 @@ export const workers: Worker[] = [
     topic: 'user-updated',
     subscription: 'user-updated-api',
   },
+  {
+    topic: 'user-updated',
+    subscription: 'user-updated-mailing',
+  },
+  {
+    topic: 'user-updated',
```
rebelchris, line 27: This should be user-deleted topic

rebelchris (approved):

PR #1020, opened 2022-08-24
feat: API user deletion

rebelchris:
Introduces API user deletion.
Since we have foreign key constraints on userId's we have to follow the existing deletion flow.

We also need to support existing deletion from the Gateway side of things.
Thus added a Kratos flow to not duplicate run queries when the user is already deleted.

WT-333

### DIFF
--- a/src/workers/cdc.ts
+++ b/src/workers/cdc.ts
@@ -41,6 +41,7 @@ import {
   notifySubmissionCreated,
   notifySubmissionGrantedAccess,
   NotificationReason,
+  notifyUserDeleted,
 } from '../common';
 import { ChangeMessage } from '../types';
 import { Connection } from 'typeorm';
@@ -384,6 +385,9 @@ const onUserStateChange = async (
     if (data.payload.after.key === UserStateKey.CommunityLinkAccess) {
       await notifySubmissionGrantedAccess(logger, data.payload.after.userId);
     }
+  }
+  if (data.payload.op === 'd') {
+    await notifyUserDeleted(logger, data.payload.before.userId, true);
   }
 };
 
--- a/src/workers/deleteUser.ts
+++ b/src/workers/deleteUser.ts
@@ -17,6 +17,8 @@ import { Settings } from '../entity/Settings';
 import { SourceDisplay } from '../entity/SourceDisplay';
 import { SourceRequest } from '../entity/SourceRequest';
 import { Upvote } from '../entity/Upvote';
+import { FastifyLoggerInstance } from 'fastify';
+import { Connection } from 'typeorm';
 
 interface UserData {
   id: string;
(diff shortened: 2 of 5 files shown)


src/workers/cdc.ts
```diff
@@ -385,6 +386,9 @@ const onUserStateChange = async (
       await notifySubmissionGrantedAccess(logger, data.payload.after.userId);
     }
   }
+  if (data.payload.op === 'd') {
+    await notifyUserDeleted(logger, data.payload.before.userId, true);
```
idoshamun, line 390: At the moment this will cause this message to be fired two times.
The current application version sends the request to the gateway which deletes the account on its behalf, then triggers a `user-deleted` message, API picks it up and delete, API sends another message.
Are there any side effects we have to be aware of?

rebelchris: Hmm you are right the gateway retriggers now.
I tried to make it work with both systems, not quite sure how we can achieve that way to work.

Alternatively we can create a new topic?
Kratos could then subscribe to both versions?

idoshamun: As long as there are no significant side effects, it's ok. Can you check for potential side effects, please?

rebelchris: The only side effect I could think about is that it can cause the subscription to keep re-trying as it will already be deleted, but unsure how we can safely test it out.

I don't see any other side effects, but it's a hard one to try from the current local setup.
Might have to setup the new system for it.

src/workers/deleteUser.ts
```diff
@@ -35,63 +37,67 @@ interface UserData {
...
+      await entityManager.getRepository(Feed).delete({ userId });
+      await entityManager.getRepository(HiddenPost).delete({ userId });
+      await entityManager.getRepository(PostReport).delete({ userId });
+      await entityManager.getRepository(Settings).delete({ userId });
+      await entityManager.getRepository(SourceDisplay).delete({ userId });
+      await entityManager.getRepository(SourceRequest).delete({ userId });
+      await entityManager.getRepository(Upvote).delete({ userId });
+      await entityManager
+        .getRepository(Post)
+        .update({ authorId: userId }, { authorId: null });
```
idoshamun, line 67: We need to another statement for scoutId

src/workers/cdc.ts
idoshamun: I don't see anything that should fail in case the user doesn't exist. It should go smoothly. Just in case, we can add a test case for this scenario

src/workers/deleteUser.ts
rebelchris: Oh dang, well spotted! 🤯

src/workers/cdc.ts
rebelchris: Will add a test for it 🙏

rebelchris:
@idoshamun Added the scout cleanup and tried adding test cases that test both mutation and the bg worker.
No weird errors are thrown and the results are always good.

idoshamun (approved):

PR #1021, opened 2022-08-24
feat: add private route for changing email

rebelchris:
Added private route for changing the users email.
Basically only Kratos will be able to call this.

@sshanzel No other call can modify the email right?

WT-368 #done


sshanzel (approved):
> @sshanzel No other call can modify the email right?

Yep, there should be none. It should only be the kratos to decide whether it is finalized by being verified. Though I'm not sure if it should be an API or a worker. Either it updates the data and it is a protected route.

src/entity/User.ts
```diff
@@ -135,6 +136,45 @@ const checkUsernameAndEmail = async (
...
+          data,
+          userId: data.id,
+          error,
+        },
+        'failed to create user profile',
+      );
+
+      // Unique
+      if (error?.code === '23505') {
+        return { status: 'failed', reason: 'USER_DOESNT_EXISTS' };
```
idoshamun, line 170:
```suggestion
        return { status: 'failed', reason: 'USER_DOESNT_EXIST' };
```

src/entity/User.ts
```diff
@@ -135,6 +136,45 @@ const checkUsernameAndEmail = async (
...
+        {
+          data,
+          userId: data.id,
+          error,
+        },
+        'failed to create user profile',
+      );
+
+      // Unique
+      if (error?.code === '23505') {
```
idoshamun, line 169: This error will be thrown only when you try to update the email to an already existing email.
If the user doesn't exist nothing will be thrown

src/routes/private.ts
```diff
@@ -67,6 +69,18 @@ export default async function (fastify: FastifyInstance): Promise<void> {
     const operationResult = await addNewUser(con, req.body, req.log);
     return res.status(200).send(operationResult);
   });
+  fastify.post<{ Body: UpdateUserEmailData }>(
+    '/updateUserEmail',
+    async (req, res) => {
+      if (!req.service) {
+        return res.status(401).send();
```
idoshamun, line 76: We usually return `404` so an attacker will not know this endpoint exists even
```suggestion
        return res.status(404).send();
```

idoshamun (changes requested):

rebelchris:
@idoshamun Totally missed that I didn't finish the user not found query.
Also updated the requests to reflect a 404 instead of 401

src/entity/User.ts
```diff
@@ -150,9 +150,16 @@ export const updateUserEmail = async (
 
   return con.transaction(async (entityManager) => {
     try {
-      await entityManager
+      const res = await entityManager
         .getRepository(User)
         .update({ id: data.id }, { email: data.email });
+      if (res.affected === 0) {
```
idoshamun, line 156: Should we care about this edge case even?

idoshamun (approved):

src/entity/User.ts
rebelchris: I guess it can't harm to know if it ever happens, most likely won't ever be hit, but still

PR #1022, opened 2022-08-28
fix: minor issues for the new auth

idoshamun:
* Image field can be empty in some cases
* Add logging in case of an unexpected error in the `injectGraphQL` function
* Add `reputation` field to the `whoami` call

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -56,7 +56,7 @@ interface GQLUserParameters {
 export interface GQLUser {
   id: string;
   name: string;
-  image: string;
+  image?: string;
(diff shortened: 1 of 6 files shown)


src/schema/users.ts
```diff
@@ -571,26 +571,20 @@ export const resolvers: IResolvers<any, Context> = {
 
       return ctx.con.query(
         `
-        select ${rankColumn} as "rank",
```
idoshamun, line 574: I haven't changed anything here, it's probably auto-ident of intellij

rebelchris (approved):
Looks good!

PR #1023, opened 2022-08-29
fix: for adding users from the Gateway call

rebelchris:
A fix for making certain fields un-required and allowing optional fields from the gateway users.

WT-370

### DIFF
--- a/src/entity/User.ts
+++ b/src/entity/User.ts
@@ -118,7 +118,7 @@ type AddNewUserResult =
   | { status: 'failed'; reason: UserFailErrorKeys; error?: Error };
 
 const checkRequiredFields = (data: AddUserData): boolean => {
-  return !!(data && data.id && data.name && data.image);
+  return !!(data && data.id && data.image);
 };
 
 const checkUsernameAndEmail = async (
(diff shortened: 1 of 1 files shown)


src/entity/User.ts
```diff
@@ -183,6 +185,7 @@ export const addNewUser = async (
   logger: FastifyLoggerInstance,
 ): Promise<AddNewUserResult> => {
   if (!checkRequiredFields(data)) {
+    console.log(data);
```
idoshamun, line 188: if it's a proper log, you should use the `logger` object. Otherwise please remove it

src/entity/User.ts
```diff
@@ -101,22 +101,24 @@ export class User {
...
 const checkRequiredFields = (data: AddUserData): boolean => {
-  return !!(
-    data &&
-    data.id &&
-    data.name &&
-    data.image &&
-    data.username &&
-    data.email
-  );
+  return !!(data && data.id && data.name && data.image);
```
idoshamun, line 121: `image` and `name` can also potentially be empty

idoshamun (approved):
Please see my comments before proceeding

src/entity/User.ts
rebelchris: Image should always have a fallback set as far as I can see.
Name is true, i'll remove that.

idoshamun: Could be, but we can't rely on that. It's better to allow null and set the fallback before return using GQL resolvers or something just in case. What do you think?

PR #1025, opened 2022-08-30
Wt 370 update gateway user

rebelchris:
Introducing method for updating users from the Gateway.
Decided to use the existing graphQL layer, but check if it's a service call *(from gateway)

Related to WT-370


src/entity/User.ts
```diff
@@ -185,6 +185,7 @@ export const addNewUser = async (
   logger: FastifyLoggerInstance,
 ): Promise<AddNewUserResult> => {
   if (!checkRequiredFields(data)) {
+    logger.info({ message: 'Missing fields', data });
```
idoshamun, line 188:
```suggestion
    logger.info({ data }, 'missing fields when adding new user');
```

idoshamun (approved):

PR #1030, opened 2022-08-31
feat: move username changed to API

rebelchris:
Introduce username changed in API as this is now source of truth.

Related to: https://github.com/dailydotdev/daily-gateway/pull/521


sshanzel (approved):
LGTM 🚢

idoshamun (changes requested):
There's a missing test

src/workers/cdc.ts
```diff
@@ -195,6 +196,14 @@ const onUserChange = async (
         }
       }
     }
+    if (data.payload.before.username !== data.payload.after.username) {
```
idoshamun, line 199: To make sure we submit this event only when the info is confirmed. Otherwise on every registration we will dispatch this event

```suggestion
    if (data.payload.before.infoConfirmed && data.payload.before.username !== data.payload.after.username) {
```

PR #1035, opened 2022-09-06
refactor: update user info confirmed

sshanzel:


rebelchris (approved):
Nice work 🙏

src/schema/users.ts
```diff
@@ -765,9 +765,11 @@ export const resolvers: IResolvers<any, Context> = {
           : user.image;
 
       try {
-        const result = await ctx.con
-          .getRepository(User)
-          .save({ ...user, ...data, image: avatar });
+        const updatedUser = { ...user, ...data, image: avatar };
+        if (!user.infoConfirmed && updatedUser.email && updatedUser.username) {
```
idoshamun, line 769:
```suggestion
        if (!user.infoConfirmed && updatedUser.email && updatedUser.username && updatedUser.name) {
```

idoshamun (approved):
You were missing the name field as well

src/schema/users.ts
sshanzel: This one is always required in our auth so it will always have a value, so I did not include it. But I agree, being strict is always safer.

PR #1043, opened 2022-09-14
refactor: adding new user can have true infoConfirmed

sshanzel:
The info confirmed is a criterion and when the user registered with complete info, it should automatically be true.


idoshamun (commented):
Can you also add a test to check if `infoConfirmed` is false when it needs to be?

sshanzel:
> Can you also add a test to check if infoConfirmed is false when it needs to be?

Good point. Added it now.

idoshamun (approved):

PR #1047, opened 2022-09-20
feat: clean zombie users 🧟

idoshamun:
Zombie users are users that have `infoConfirmed = false` and are older than one hour. A new cron job is added to remove zombie users from the database.

WT-486#done


rebelchris (approved):
This way also fired the CDC right, so kratos get's cleaned?

idoshamun:
@rebelchris yes. That's the magic of CDC. Everything that happens on the database triggers a pubsub message

PR #1053, opened 2022-10-03
fix: ignore the github handle if it already exists when adding new user

idoshamun:
When a user signs up with GitHub we prepopulate the GitHub handle. This may yield a unique key violation which we want to prevent.

WT-621 #done

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,5 +7,5 @@ dotenv.config({ path: '.env' });
 
 export const fallbackImages = {
   avatar:
-    'https://daily-now-res.cloudinary.com/image/upload/t_logo,f_auto/v1664367305/placeholders/placeholder3',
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -7,5 +7,5 @@ dotenv.config({ path: '.env' });
 
 export const fallbackImages = {
   avatar:
-    'https://daily-now-res.cloudinary.com/image/upload/t_logo,f_auto/v1664367305/placeholders/placeholder3',
+    'https://daily-now-res.cloudinary.com/image/upload/f_auto/v1664367305/placeholders/placeholder3',
```
rebelchris, line 10: This also needs to happen in the webapp though for the email flow, would you mind doing it there as well? 🙏

rebelchris (approved):
Nice one! 

Just one small request so we can cross close the issue

src/config.ts
idoshamun: Already did in my other PR :)

PR #1056, opened 2022-10-07
fix: convert usernames to lowercase

rebelchris:
Converting the search for usernames as well as the adding of users to lowercase usernames.

WT-644 #done


idoshamun (commented):
Is there any way to make sure we set the username to lowercase in a TypeORM hook perhaps? A solution like this might be more resilient to human errors

rebelchris:
@idoshamun Yeah agreed, it's was a pressure issue at the time of creating to get this out before launch so didn't really take all options into account.

Can relook at it today.

idoshamun:
@rebelchris no pressure, it can be post-launch as well as it existed for a while

rebelchris:
@idoshamun I went with the TypeORM transform for setting the usernames.
For looking up usernames I decided to keep the manual flow for now, as we don't want to convert existing ones.

As soon as we retroactively fixed usernames we can opt to remove this explicit search.

Love to hear what you think.

idoshamun:
The only problem I see now is that on update, the user will be able to set their username however they want. 
The PR doesn't lowercase on update

rebelchris:
@idoshamun I added some changes for the mentions to work.
Also added more test cases including the username changes which seems to all work.

The transform should catch both create and update.

src/schema/users.ts
```diff
@@ -770,6 +770,7 @@ export const resolvers: IResolvers<any, Context> = {
 
       try {
         const updatedUser = { ...user, ...data, image: avatar };
+        updatedUser.username = updatedUser.username;
```
idoshamun, line 773: What's this?

idoshamun (approved):
Please see my comment before merging

rebelchris:
@idoshamun Before merging we actually need to address what we'll do with the existing people who have duplicate usernames.

idoshamun:
@rebelchris let's discuss today during 1:1

src/routes/private.ts
```diff
@@ -95,9 +95,15 @@ export default async function (fastify: FastifyInstance): Promise<void> {
       }
 
       const con = await createOrGetConnection();
+
       const user = await con
         .getRepository(User)
-        .findOneBy({ username: search });
+        .createQueryBuilder()
+        .select('id')
+        .where('LOWER(username) = LOWER(:search)', {
```
el, line 103: why downt we update the db instead of lowercasing in transaction

src/entity/User.ts
```diff
@@ -135,7 +149,7 @@ const checkUsernameAndEmail = async (
     .getRepository(User)
     .createQueryBuilder()
     .select('id')
-    .where('email = :email or username = :username', {
```
el, line 138: I think we should do a lowercase while saving the user (if we are not doing that already)

el (approved):

bhumit070:
is this issue dones ? or is it still open?

idoshamun:
@bhumit070 it's done on a different PR. Thanks for checking :)

PR #1057, opened 2022-10-25
refactor: importing keyword using .save

sshanzel:
Whenever we reset and migrate the `latest` from scratch, it seems the keyword table is automatically populated. I even deleted the `Keyword.json` file to ensure the values are not coming from it, but after further investigation, one of the migration files have hard-coded importing Keyword.
Logs from db:migrate:latest:
image

Now, whenever I run the import command on a fresh DB, it always fails at the `Keyword` part (due to duplicate PK as said from the other logs), which makes me remove the first few imports manually. To avoid to having doing manual work, for the `Keyword` import, I made it use `.save` so it won't bother existing (just update) keywords but still fill it with the missing ones.

Might just be difficult if new ones will have to experience the same, so I made the changes.

Performance-wise, `.insert` will always be faster, but running `.save` on just `Keyword` doesn't feel like the command became worse.


idoshamun:
We added this migration when we did a massive change to our keyword system.
Maybe the best solution would be to simply not include the keyword as part of the import step?
Have you tried that?

sshanzel:
> Have you tried that?

Yep, it seems to proceed without an issue. I just thought there could be missing some keywords in the migration that can be found in the file. I could be wrong tho, of course.

Let me know if we would prefer to just remove it, then I shall make the changes from the enum.

idoshamun:
I think we can remove it for now :)
Please don't merge anything just yet anyway

sshanzel:
> Please don't merge anything just yet anyway

Done! I'll just keep this on standby until merging is clear. This doesn't affect anything in production anyway.

el (approved):

rebelchris (approved):

PR #1058, opened 2022-10-26
build: update to node 18, fastify 4, mercurius 11, graphql 16

idoshamun:
As part of my work on the crons, I noticed that we have a lot of outdated dependencies, which results in high vulnerabilities.
I took the time to update everything before I proceed.
Please note that I opted for legacy peer dependencies due to a conflict with one of our packages.
I opened a PR to fix that on the relevant package: https://github.com/graphile/graphile-engine/pull/814
Once the PR above is merged we can go back to modern peer deps.
Node 18 is required for the new command arguments parsing capabilities it provides. I need it for cron


rebelchris (approved):
thanks!

This was kind of on hold because Fastify had some peer deps that could not be met, seems that's mostly sorted now 🥳

Did you upgrade everything to v4 format, as there are some side things to look at:
https://www.fastify.io/docs/latest/Guides/Migration-Guide-V4/

Lastly, perhaps also good to eventually upgrade everything to node 18? (I can add backlog items for it and do it on my next update?)

idoshamun:
@rebelchris agree about node 18. No rush though.
The pain point with fastify v4 was typescript. But I managed to fix that. The rest should be ok, I skimmed through the guide and made sure tests pass

PR #1062, opened 2022-10-26
ci: run crons via k8s instead of pubsub

idoshamun:
This change introduces a centralized cli that runs a specific sub-application. It's useful to separate crons from background worker and choose more explicitly what app to run


rebelchris (approved):
Wow nice, also a lot more scaleable like this.
It doesn't introduce a complete new cluster right, just inside the container a different execution? (for my own reference)

idoshamun:
The way k8s cronjob works is that you define a `CronJob` entity and when it the time comes it triggers a new container to run the command specified. So whatever happens while running this cron will not affect our background worker. And you also get retries out of the box.

.infra/crons.ts
```diff
@@ -1,5 +1,4 @@
 interface CronPubSub {
```
rebelchris, line 1: The interface could be renamed for the sake of being correct then.

PR #1063, opened 2022-10-27
fix: handle invalid uuid for rss

idoshamun:
Currently when an invalid uuid is provided the server returns 500. This is fixed now.


rebelchris (approved):
Nice find and fix 🥳

PR #1073, opened 2022-11-16
fix: only hide banned post for community source

rebelchris:
Only removed banned posts from community source feed.

WT-769 #done


idoshamun (approved):
Thanks for the quick response!

PR #1075, opened 2022-11-21
fix: settings create conflict on same transaction

sshanzel:
~~WIP: still spooling up my local env to test the changes.~~

Running 2 save commands in the same transaction seems to result in conflict with one another even though we used upsert. Probably the ORM has thought the same command should use the insert as the first save is deferred due to being in the same transaction.


idoshamun (changes requested):
Please see my comment

rebelchris (approved):
Seems valid, I do still think it's slightly weird it got introduced by the frontend meaning we are doing something differently there as well.

idoshamun (approved):
Nice! Thanks for that

PR #1078, opened 2022-11-27
feat: add authentication support for graphql subscriptions

idoshamun:
Cookies are not provided when using WebSocket in the extension. Hence, our authentication doesn't work in GraphQL Subscriptions from the extension. 
This PR adds another step in the connection to check the token provided by the client explicitly. 
This is required for the notifications.

Apps PR: https://github.com/dailydotdev/apps/pull/1394

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -1,5 +1,5 @@
 import { DataSource, EntitySchema, ObjectType, Repository } from 'typeorm';
-import { FastifyRequest, FastifyLoggerInstance } from 'fastify';
+import { FastifyRequest, FastifyBaseLogger } from 'fastify';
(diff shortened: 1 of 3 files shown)


src/Context.ts
```diff
@@ -35,7 +35,7 @@ export class Context {
     return this.req.roles ?? [];
   }
 
-  get log(): FastifyLoggerInstance {
+  get log(): FastifyBaseLogger {
```
rebelchris, line 38: Is this the new version for it, should we change it in multiple places?

rebelchris (approved):
Looks great 🙏

src/Context.ts
idoshamun: `FastifyLoggerInstance` is deprecated. We should migrate, but didn't want to turn this PR into a migration 😬

PR #1079, opened 2022-11-28
feat: create notification schema

idoshamun:
Add 3 new tables notification, notification avatar and notification attachments. Add GraphQL types accordingly and GraphORM proper sorting for attachments and avatars when querying for notification

WT-774 #done


src/entity/Notification.ts
```diff
@@ -1,18 +1,81 @@
...
+  OneToMany,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import { User } from './User';
+import { NotificationAvatar } from './NotificationAvatar';
+import { NotificationAttachment } from './NotificationAttachment';
+
+export type NotificationType =
+  | 'community_picks_failed'
+  | 'community_picks_success'
```
rebelchris, line 15: Should this not be `community_picks_succeeded` to be inline with correct tense?

src/entity/NotificationAttachment.ts
```diff
@@ -0,0 +1,29 @@
+import { Column, Entity, Index, ManyToOne, PrimaryColumn } from 'typeorm';
+import { Notification } from './Notification';
+
+export type NotificationAttachmentType = 'post';
```
rebelchris, line 4: I'm actually thinking now, would we not want to opt to store the referenced post as well?
In the case of the post being deleted for instance?

src/entity/NotificationAvatar.ts
```diff
@@ -0,0 +1,33 @@
+import { Column, Entity, Index, ManyToOne, PrimaryColumn } from 'typeorm';
+import { Notification } from './Notification';
+
+export type NotificationAvatarType = 'source' | 'user';
```
rebelchris, line 4: Same question here, for instance with spammers who get deleted and perhaps use inappropriate images?

rebelchris (approved):
Happy with the code, just have some additional questions around how we safe things.

src/entity/Notification.ts
idoshamun: Copied from the DR. I'll fix that!

src/entity/NotificationAttachment.ts
idoshamun: Good point!

src/entity/NotificationAvatar.ts
idoshamun: Here we already store the `referenceId`

PR #1082, opened 2022-11-30
feat: boot endpoint to retrieve number of unread notifications

rebelchris:
Added a new route endpoint and graphQL query to retrieve the number of unread notifications for a user.
This endpoint will be called by the boot endpoint if no redis cache is found (gateway PR incoming)

WT-784

### DIFF
--- a/.infra/application.properties
+++ b/.infra/application.properties
@@ -16,4 +16,4 @@ debezium.transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
 debezium.transforms.Reroute.topic.regex=(.*)
 debezium.transforms.Reroute.topic.replacement=%topic%
 debezium.sink.type=pubsub
-quarkus.log.level=TRACE
+quarkus.log.level=ERROR
--- a/src/schema/notifications.ts
+++ b/src/schema/notifications.ts
@@ -1,7 +1,8 @@
 import { IResolvers } from '@graphql-tools/utils';
 import { traceResolvers } from './trace';
 import { Context } from '../Context';
-import { Banner } from '../entity';
+import { Banner, Notification } from '../entity';
+import { ConnectionArguments } from 'graphql-relay';
 
 interface GQLBanner {
   timestamp: Date;
@@ -127,6 +128,10 @@ export const typeDefs = /* GraphQL */ `
 
   extend type Query {
     """
+    Get the active notification count for a user
+    """
+    notificationCount: Int @auth
+    """
     Get a banner to show, if any
     """
     banner(
@@ -141,6 +146,17 @@ export const typeDefs = /* GraphQL */ `
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const resolvers: IResolvers<any, Context> = traceResolvers({
   Query: {
+    notificationCount: async (
+      source,
+      args: ConnectionArguments,
+      ctx,
+    ): Promise<number> =>
+      ctx.getRepository(Notification).count({
+        where: {
+          userId: ctx.userId,
+          public: true,
+        },
+      }),
     banner: async (
       source,
       { lastSeen }: { lastSeen: Date },
--- a/.infra/Pulumi.adhoc.yaml
+++ b/.infra/Pulumi.adhoc.yaml
@@ -25,6 +25,6 @@ config:
     typeormUser: postgres
     tz: UTC
     urlPrefix: http://api.local.com
-  api:image: api-image:tilt-a8bfab3ac1d9fecb
+  api:image: api-image:tilt-71bca7fd406d9a1c
   api:k8s:
     namespace: local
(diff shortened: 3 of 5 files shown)


.infra/application.properties
```diff
@@ -16,4 +16,4 @@ debezium.transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
 debezium.transforms.Reroute.topic.regex=(.*)
 debezium.transforms.Reroute.topic.replacement=%topic%
 debezium.sink.type=pubsub
-quarkus.log.level=TRACE
+quarkus.log.level=ERROR
```
idoshamun, line 19: Please revert this change. It will affect production and we currently need it to be trace

__tests__/notifications.ts
```diff
@@ -1,26 +1,118 @@
...
+
+beforeEach(async () => {
+  loggedUser = null;
 });
 
 afterAll(() => disposeGraphQLTesting(state));
 
+describe('notifications route', () => {
+  it('should return not found when not authorized', async () => {
+    await authorizeRequest(request(app.server).get('/notifications')).expect(
```
idoshamun, line 39: This will probably fail the gateway. Should we perhaps return a default value there? Or on the contrary support 401 on the gateway side

src/schema/notifications.ts
```diff
@@ -126,6 +128,10 @@ export const typeDefs = /* GraphQL */ `
   }
 
   extend type Query {
+    """
+    Get the active notification count for a user
+    """
+    notificationCount: Int @auth
```
idoshamun, line 134: I think it's a better name

```suggestion
    unreadNotificationsCount: Int @auth
```

src/workers/cdc.ts
```diff
@@ -225,6 +227,23 @@ const onAlertsChange = async (
   }
 };
 
+const onNotificationsChange = async (
```
idoshamun, line 230: We discussed today this is will probably spam the system so it's better to send the message on the `readNotifications` mutation

idoshamun (changes requested):
Please see my comments

src/workers/cdc.ts
sshanzel: I agree. I think this should be covered [in this section](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/378667009/PD-39+Notification+System#Reset-all-notifications-to-%E2%80%9CRead%E2%80%9D) (Reset all notifications to “Read” section) of the DR. Though for having a plus +1 in the unread, we can probably keep the `create` event here.

src/schema/notifications.ts
```diff
@@ -144,6 +145,10 @@ export const typeDefs = /* GraphQL */ `
   }
 
   extend type Query {
+    """
+    Get the active notification count for a user
+    """
+    unreadNotificationsCount: Int @auth
```
sshanzel, line 151: Nice. Thank you for adding this, I missed this in the DR.

sshanzel (commented):
Nothing more I can add than what was already mentioned. Amazing work! ✨

src/workers/cdc.ts
rebelchris: Updated to create and readNotifications mutation

src/common/pubsub.ts
```diff
@@ -192,6 +193,12 @@ export const notifySettingsUpdated = (
   settings: ChangeObject<Settings>,
 ): Promise<void> => publishEvent(log, settingsUpdatedTopic, settings);
 
+export const notifyNotificationsUpdated = (
```
idoshamun, line 196: Please fix the function name

src/workers/cdc.ts
```diff
@@ -225,6 +227,25 @@ const onAlertsChange = async (
...
+  if (['c'].includes(data.payload.op)) {
+    const unreadNotificationsCount = await con
+      .getRepository(Notification)
+      .count({
+        where: {
+          userId: data.payload.after.userId,
+          public: true,
+          readAt: IsNull(),
+        },
+      });
```
idoshamun, line 244: Perhaps it will be better to abstract this functionality as we use it in the graphql query as well

src/workers/cdc.ts
```diff
@@ -225,6 +227,25 @@ const onAlertsChange = async (
   }
 };
 
+const onNotificationsChange = async (
+  con: DataSource,
+  logger: FastifyLoggerInstance,
+  data: ChangeMessage<Notification>,
+): Promise<void> => {
+  if (['c'].includes(data.payload.op)) {
```
idoshamun, line 235: I find this version more readable

```suggestion
  if (data.payload.op === 'c') {
```

idoshamun (approved):
Some minor comments but looks good

src/workers/cdc.ts
rebelchris: Ah good idea, we didn't have this before but let me refactor it 👀

src/workers/cdc.ts
rebelchris: Makes sense for the singular one, will refactor

PR #1084, opened 2022-12-02
feat: readNotifications mutation

idoshamun:
New mutation to set all unread notifications of the user as read.

WT-776 #done


rebelchris (approved):
Nice looks good, we can hook the pubsub into this then 👍

PR #1087, opened 2022-12-07
feat: generate notifications

idoshamun:
Generate all supported notifications using background workers.
This PR introduces new concepts:
* `NotificationWorker` - Lightweight variation of our good old `Worker` that its sole purpose is to generate notifications. It returns an array of notification types to generate along with their relevant context (parameters).
* `NotificationBuilder` - To make it easy to build new notification types, the builder has pre-built functions for common use cases, and everything is chainable to make it more readable.
* `NotificationBaseContext` - Used to store all the parameters a single notification needs in order to be generated completely. This type has inherited types that makes it flexible to provide only the needed parameters per type.

The `generateNotificationMap` object holds a map that using the `NotificationBuilder` it constructs the notification object that later on can be stored.

The storage mechanism is exactly-once, meaning that if for some reason pub/sub decided to replay the message (which can happen), we will still save it only once and there will be no notification spam. 

It's a huge PR, I suggest starting with the notifications folder as it contains all the fundamental stuff and then you can review the new workers.

WT-786 #done

### DIFF
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -130,4 +130,61 @@ export const workers: Worker[] = [
     subscription: 'api-cdc',
     args: { enableMessageOrdering: true },
   },
+  {
+    topic: 'api.v1.new-comment-mention',
+    subscription: 'comment-mention-mail',
+  },
+  // Notifications
+  {
+    topic: 'community-link-rejected',
+    subscription: 'api.community-picks-failed-notification',
+  },
+  {
+    topic: 'post-scout-matched',
+    subscription: 'api.community-picks-succeeded-notification',
+  },
+  {
+    topic: 'community-link-access',
+    subscription: 'api.community-picks-granted-notification',
+  },
+  {
+    topic: 'post-author-matched',
+    subscription: 'api.article-picked-notification',
+  },
+  {
+    topic: 'post-commented',
+    subscription: 'api.article-new-comment-notification.post-commented',
+  },
+  {
+    topic: 'comment-commented',
+    subscription: 'api.article-new-comment-notification.comment-commented',
+  },
+  {
+    topic: 'post-upvoted',
+    subscription: 'api.article-upvote-milestone-notification',
+  },
+  {
+    topic: 'send-analytics-report',
+    subscription: 'api.article-analytics-notification',
+  },
+  {
+    topic: 'post-banned-or-removed',
+    subscription: 'api.article-report-approved-notification',
+  },
+  {
+    topic: 'pub-request',
+    subscription: 'api.source-request-notification',
+  },
+  {
+    topic: 'api.v1.new-comment-mention',
+    subscription: 'api.comment-mention-notification',
+  },
+  {
+    topic: 'comment-commented',
+    subscription: 'api.comment-reply-notification',
+  },
+  {
+    topic: 'comment-upvoted',
+    subscription: 'api.comment-upvote-milestone-notification',
+  },
 ];
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -132,7 +132,7 @@ export const workers: Worker[] = [
   },
   {
     topic: 'api.v1.new-comment-mention',
-    subscription: 'comment-mention-mail',
+    subscription: 'api.comment-mention-mail',
   },
   // Notifications
   {
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -1,0 +1,125 @@
+import { NotificationType } from '../entity';
+import { NotificationBuilder } from './builder';
+import { NotificationIcon } from './icons';
+import { scoutArticleLink } from '../common';
+import {
+  NotificationBaseContext,
+  NotificationCommentContext,
+  NotificationCommenterContext,
+  NotificationPostContext,
+  NotificationSourceContext,
+  NotificationSourceRequestContext,
+  NotificationSubmissionContext,
+  NotificationUpvotersContext,
+} from './types';
+
+export const generateNotificationMap: Record<
+  NotificationType,
+  (
+    builder: NotificationBuilder,
+    ctx: NotificationBaseContext,
+  ) => NotificationBuilder
+> = {
+  community_picks_failed: (builder, ctx: NotificationSubmissionContext) =>
+    builder.systemNotification().referenceSubmission(ctx.submission),
+  community_picks_succeeded: (builder, ctx: NotificationPostContext) =>
+    builder
+      .referencePost(ctx.post)
+      .icon(NotificationIcon.CommunityPicks)
+      .title(
+        `<b>Community picks:</b> An article you Scouted was accepted and is now <span class="text-theme-color-cabbage">live</span> on the daily.dev feed!`,
+      )
+      .targetPost(ctx.post)
+      .attachmentPost(ctx.post),
+  community_picks_granted: (builder) =>
+    builder
+      .referenceSystem()
+      .icon(NotificationIcon.DailyDev)
+      .title(
+        `<b>Community picks:</b> You have earned enough reputation to <span class="text-theme-color-cabbage">Scout and submit</span> articles.`,
+      )
+      .description(`<u>Submit your first article now!</u>`)
+      //TODO: validate with web team that this is going to be supported!
+      .targetUrl(scoutArticleLink),
+  article_picked: (builder, ctx: NotificationPostContext) =>
+    builder
+      .referencePost(ctx.post)
+      .icon(NotificationIcon.DailyDev)
+      .title(
+        `Congratulations! <b>Your article</b> got <span class="text-theme-color-cabbage">listed</span> on the daily.dev feed!`,
+      )
+      .targetPost(ctx.post)
+      .attachmentPost(ctx.post),
+  article_new_comment: (builder, ctx: NotificationCommenterContext) =>
+    builder
+      .referenceComment(ctx.comment)
+      .icon(NotificationIcon.Comment)
+      .title(
+        `<b>${ctx.commenter.name}</b> posted a <span class="text-theme-color-blueCheese">comment</span> on your article.`,
+      )
+      .descriptionComment(ctx.comment)
+      .targetPost(ctx.post, ctx.comment)
+      .avatarManyUsers([ctx.commenter]),
+  article_upvote_milestone: (
+    builder,
+    ctx: NotificationPostContext & NotificationUpvotersContext,
+  ) =>
+    builder
+      .referencePost(ctx.post)
+      .upvotes(ctx.upvotes, ctx.upvoters)
+      .title(
+        `<b>You rock!</b> Your article <span class="text-theme-color-avocado">earned ${ctx.upvotes} upvotes!</span>`,
+      )
+      .targetPost(ctx.post)
+      .attachmentPost(ctx.post),
+  article_report_approved: (builder, ctx: NotificationPostContext) =>
+    builder.referencePost(ctx.post).systemNotification(),
+  article_analytics: (builder, ctx: NotificationPostContext) =>
+    builder.referencePost(ctx.post).systemNotification(),
+  source_approved: (
+    builder,
+    ctx: NotificationSourceRequestContext & NotificationSourceContext,
+  ) =>
+    builder
+      .referenceSourceRequest(ctx.sourceRequest)
+      .icon(NotificationIcon.DailyDev)
+      .title(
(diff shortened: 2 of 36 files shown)


src/workers/notifications/index.ts
```diff
@@ -0,0 +1,55 @@
...
+import articleAnalytics from './articleAnalytics';
+import sourceRequest from './sourceRequest';
+import commentMention from './commentMention';
+import commentReply from './commentReply';
+import commentUpvoteMilestone from './commentUpvoteMilestone';
+
+function notificationWorkerToWorker(worker: NotificationWorker): Worker {
+  return {
+    ...worker,
+    handler: async (message, con, logger) => {
```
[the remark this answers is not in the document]
rebelchris: Ah, so this would execute after the original handler, so does it become one bg worker?

src/workers/notifications/communityPicksSucceeded.ts
```diff
@@ -0,0 +1,26 @@
...
+import { NotificationWorker } from './worker';
+
+interface Data {
+  postId: string;
+  scoutId: string;
+}
+
+const worker: NotificationWorker = {
+  subscription: 'api.community-picks-succeeded-notification',
+  handler: async (message, con) => {
```
rebelchris, line 13: In this example are you not afraid it adds a lot of overhead?
As this doesn't particularly do anything expect re-sending the data?

Code wise happy with how it looks, but wondering if executing directly in the first workers doesn't help with load?

src/workers/notifications/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      const args = await worker.handler(message, con, logger);
+      if (!args) {
+        return;
+      }
+      const bundles = args.map(({ type, ctx }) =>
+        generateNotification(type, ctx),
```
rebelchris, line 30: The ctx is always singular right?

src/workers/notifications/commentReply.ts
```diff
@@ -0,0 +1,55 @@
...
+          post.scoutId ?? '',
+        ],
+      })
+      .getRawMany();
+    const ctx: Omit<NotificationCommenterContext, 'userId'> = {
+      post,
+      comment,
+      commenter,
+    };
+    return [parent.userId, ...threadFollowers.map(({ userId }) => userId)].map(
```
rebelchris, line 46: Sweet! 👀

src/workers/notifications/articleUpvoteMilestone.ts
```diff
@@ -0,0 +1,48 @@
...
+      'userId'
+    > = {
+      post,
+      upvoters,
+      upvotes: post.upvotes,
+    };
+    return users.map((userId) => ({
+      type: 'article_upvote_milestone',
+      ctx: { ...ctx, userId },
+    }));
```
rebelchris, line 44: this one I have some questions about.

Will this not float already shown upvoters in certain cases?
so upvote #5 = create with 5, which would be fine.

Then with 6th it would simply return the same no? (with 1 difference)

Mainly raising this how it would handle on the notification centre with a lot of upvotes on the same article.

src/workers/commentMentionMail.ts
```diff
@@ -64,3 +66,17 @@ export const sendEmailToMentionedUser = async (
   });
   logger.info('comment mention email sent to: ' + commenter.id);
 };
+
+interface Data {
+  commentMention: ChangeObject<CommentMention>;
+}
+
+const worker: Worker = {
+  subscription: 'comment-mention-mail',
```
rebelchris, line 75: Just raising a question: Should we not opt to retroactively fix the subscription names?

.infra/workers.ts
```diff
@@ -130,4 +130,61 @@ export const workers: Worker[] = [
     subscription: 'api-cdc',
     args: { enableMessageOrdering: true },
   },
+  {
+    topic: 'api.v1.new-comment-mention',
+    subscription: 'comment-mention-mail',
+  },
+  // Notifications
+  {
+    topic: 'community-link-rejected',
```
rebelchris, line 139: Would you not want to retro fix these namings?

rebelchris (changes requested):
Wow! 

Love to see how you tackled this and how dynamic it's setup.
Have some generic questions and remarks and some blockers due to missing pieces.
But overall happy to see how this is coming along 🥳

src/notifications/generate.ts
```diff
@@ -0,0 +1,125 @@
...
+  ) => NotificationBuilder
+> = {
+  community_picks_failed: (builder, ctx: NotificationSubmissionContext) =>
+    builder.systemNotification().referenceSubmission(ctx.submission),
+  community_picks_succeeded: (builder, ctx: NotificationPostContext) =>
+    builder
+      .referencePost(ctx.post)
+      .icon(NotificationIcon.CommunityPicks)
+      .title(
+        `<b>Community picks:</b> An article you Scouted was accepted and is now <span class="text-theme-color-cabbage">live</span> on the daily.dev feed!`,
```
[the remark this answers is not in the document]
idoshamun: Awesome idea. This has to be a function actually that receives context, but I get your point. I'll do that

src/notifications/generate.ts
```diff
@@ -0,0 +1,125 @@
...
+  community_picks_granted: (builder) =>
+    builder
+      .referenceSystem()
+      .icon(NotificationIcon.DailyDev)
+      .title(
+        `<b>Community picks:</b> You have earned enough reputation to <span class="text-theme-color-cabbage">Scout and submit</span> articles.`,
+      )
+      .description(`<u>Submit your first article now!</u>`)
+      //TODO: validate with web team that this is going to be supported!
+      .targetUrl(scoutArticleLink),
```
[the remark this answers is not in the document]
idoshamun: Yes! I'll fix that

src/notifications/index.ts
```diff
@@ -0,0 +1,74 @@
...
+export async function storeNotificationBundle(
+  entityManager: EntityManager,
+  bundles: NotificationBundle[],
+): Promise<void> {
+  const { identifiers } = await entityManager
+    .createQueryBuilder()
+    .insert()
+    .into(Notification)
+    .values(bundles.map(({ notification }) => notification))
+    .orIgnore()
```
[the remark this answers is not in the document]
idoshamun: Not really, this can casually happen due to retries of pub/sub or a person upvoting and removing the upvote and so on

src/workers/notifications/utils.ts
```diff
@@ -0,0 +1,42 @@
+import { NotificationHandlerReturn } from './worker';
+import { Comment, Post } from '../../entity';
+import { NotificationCommenterContext } from '../../notifications';
+import { DataSource } from 'typeorm';
+
+export const uniquePostOwners = (post: Post, exclude?: string): string[] =>
+  [...new Set([post.scoutId, post.authorId])].filter(
+    (userId) => userId && userId !== exclude,
+  );
```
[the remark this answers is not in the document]
idoshamun: We currently use it in the emails but it's going to get revamped due to the notifications so I wouldn't touch it

src/workers/notifications/index.ts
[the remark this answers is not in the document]
idoshamun: This will be executed by the background worker. Think of it as a background worker for a specific use case of generating notifications. It has very structured workflow so I thought it may be nice to have it

src/workers/notifications/communityPicksSucceeded.ts
idoshamun: An alternative could be to add it directly to the code where we add new post I guess? Or have you thought about something else?

src/workers/notifications/index.ts
idoshamun: What do you mean?

src/workers/notifications/articleUpvoteMilestone.ts
idoshamun: These are the upvote milestones: `export const UPVOTE_MILESTONES = [
  1, 2, 3, 4, 5, 10, 20, 50, 100, 200, 500, 1000, 1250, 1500, 5000, 7500, 10000,
];`
We show up to 5 latest upvoters. Does that make sense?

src/workers/commentMentionMail.ts
idoshamun: I'll fix it

.infra/workers.ts
idoshamun: These topics exist already so changing their names is out of scope

src/notifications/generate.ts
```diff
@@ -0,0 +1,125 @@
...
+      .title(
+        `<b>Community picks:</b> An article you Scouted was accepted and is now <span class="text-theme-color-cabbage">live</span> on the daily.dev feed!`,
+      )
+      .targetPost(ctx.post)
+      .attachmentPost(ctx.post),
+  community_picks_granted: (builder) =>
+    builder
+      .referenceSystem()
+      .icon(NotificationIcon.DailyDev)
+      .title(
```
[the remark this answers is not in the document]
rebelchris: Alright, guess we don't have to worry about handling internal url's yet.

src/workers/notifications/index.ts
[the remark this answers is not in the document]
rebelchris: Yeah agreed, it's a nice flow and quite dynamic. 🥳

src/workers/notifications/communityPicksSucceeded.ts
rebelchris: Hmm more referring to the notification worker, that could execute directly, but see your thinking so let's leave this as is :)

src/workers/notifications/index.ts
rebelchris: Regarding this, It showed clearly down the line of my review, forgot to remove this.

src/workers/notifications/articleUpvoteMilestone.ts
rebelchris: I do think so, but to clarify for the sake of example let's say I have 1 article and all notifications are upvotes on this article, I would see the following notifications in my screen right?

[ upvoted by user 6, 7, 8, 9, 10 ]
[ upvoted by user 1, 2, 3, 4, 5 ]
[ upvoted by user 1, 2, 3, 4 ]
[ upvoted by user 1, 2, 3 ]
[ upvoted by user 1, 2 ]
[ upvoted by user 1 ]

Have we considered any threshold for when it gets invoked or is it just realtime upvotes?

idoshamun: Do you mean like think about throttling it or so?

rebelchris: Yeah mainly the ones from 1-5 seems a bit redundant if they would happen right after each other? (so maybe there could be a x minute processing time for certain ones?

rebelchris (approved):
Just one generic question otherwise seems good!
(And we still have the open question for product with the icons that might impact)

src/workers/notifications/articleUpvoteMilestone.ts
idoshamun: I think we should check if it's possible to throttle emails and push from the provider side of things. From notification center point of view it kinda makes sense to have them I think. We can also remove the 2,4 milestones. So it will be 1,3,5. Let me know what you think

src/notifications/generate.ts
```diff
@@ -0,0 +1,133 @@
...
+    builder
+      .referencePost(ctx.post)
+      .icon(NotificationIcon.CommunityPicks)
+      .targetPost(ctx.post)
+      .attachmentPost(ctx.post),
+  community_picks_granted: (builder) =>
+    builder
+      .referenceSystem()
+      .icon(NotificationIcon.DailyDev)
+      .description(`<u>Submit your first article now!</u>`)
```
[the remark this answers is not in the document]
idoshamun: Yes, look here. Actually it's the only notification that uses it with hard coded text. The rest usually just show the comment preview
image

idoshamun:
@rebelchris regarding the icon, I'll ask design to be consistent and add icons to everything. This will only be introduced in squads so we have enough time to give them feedback

src/workers/notifications/articleUpvoteMilestone.ts
rebelchris: Yeah that makes a lot more sense considering how we query, so let's go for those steps.

rebelchris (approved):
Besides the steps we are good to go 🚢

PR #1088, opened 2022-12-08
fix: ensure name and username are required

rebelchris:
We didn't actually check if the username and name are null, this introduces that check for both fields.

WT-807 #done 

Related:
https://github.com/dailydotdev/apps/pull/1416


idoshamun (approved):
Please add test cases before merging

idoshamun (approved):

PR #1089, opened 2022-12-08
feat: new notification subscription

idoshamun:
* Update unread notification count on new notification in a dedicated worker.
* Create a worker to publish message to Redis for the new GraphQL Subscription.
* Add GraphQL Subscription that clients can listen to new notifications.
* Debezium settings changed to ignore changes on the `readAt` column of the notifications table. Otherwise it may spam our Pub/Sub

WT-783 #done


src/schema/notifications.ts
```diff
@@ -254,4 +262,26 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       return { _: true };
     },
   },
+  Subscription: {
```
rebelchris, line 265: This would be for the realtime capabilities, is it not something OneSignal already comes with perhaps that we can tap into?

__tests__/workers/unreadNotificationCount.ts
```diff
@@ -0,0 +1,33 @@
...
+
+beforeAll(async () => {
+  con = await createOrGetConnection();
+});
+
+beforeEach(async () => {
+  await saveFixtures(con, User, [usersFixture[0]]);
+});
+
+//TODO: once WT-786 is merged add more test cases
```
rebelchris, line 24: Will block for now.

__tests__/workers/cdc.ts
```diff
@@ -1126,3 +1129,35 @@ describe('submission', () => {
...
+    userId: '1',
+    type: 'community_picks_granted',
+    title: 'hello',
+    targetUrl: 'target',
+    icon: 'icon',
+    public: true,
+    createdAt: Date.now(),
+  };
+
+  it('should notify new notification', async () => {
```
rebelchris, line 1147: Perhaps we can also add a case where we update the readAt and it should not trigger, just for future proofing.

.infra/workers.ts
```diff
@@ -130,4 +130,12 @@ export const workers: Worker[] = [
     subscription: 'api-cdc',
     args: { enableMessageOrdering: true },
   },
+  {
+    topic: 'api.v1.new-notification',
+    subscription: 'api.unread-notification-count',
+  },
+  {
+    topic: 'api.v1.new-notification',
+    subscription: 'api.new-notification-redis',
```
rebelchris, line 139: Question around naming, should we opt for the redis part here?
We removed it from the above one as it's technically also a redis subscription on one side.

rebelchris (changes requested):
Looks great, have some implementation questions.

src/schema/notifications.ts
idoshamun: When we worked on the DR we discussed the different alternatives and agreed that since we have graphql subscriptions already it will be the easiest and most flexible to use. Also, OneSignal doesn't provide real-time directly as far as I'm aware.

__tests__/workers/cdc.ts
idoshamun: This requires debezium in the test environment and we currently don't support it

.infra/workers.ts
idoshamun: I followed the existing subscriptions that we have. I can change it for something else if you want.
`api.new-notification-real-time`, or `api.new-notification-subscriptions`, feel free to suggest

src/schema/notifications.ts
```diff
@@ -254,4 +262,26 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+    //TODO: decide on a proper data model
```
rebelchris, line 266: Can we remove this?

rebelchris (approved):
One small remark, otherwise good to go 🥳

src/schema/notifications.ts
idoshamun: Of course! Silly me🤦‍♂️

PR #1091, opened 2022-12-15
feat: send push notifications

idoshamun:
Send push notifications via OneSignal on a new notification. 
No tests to check that as it's more integration and manual testing we will have to apply here.

WT-782 #done


rebelchris (approved):
Wohoo where can I sign up 👀

idoshamun:
@rebelchris I'll give you my bank account on Slack. Will sign you up for private beta 😂

PR #1094, opened 2022-12-16
feat: notification email preference

sshanzel:
As discussed, to keep things simple, we will include the notification email preference in the user's table than creating a new one.

We will then create a separate table once we decide to implement a more granular control of the notification preference.

### DIFF
--- a/__tests__/__snapshots__/users.ts.snap
+++ b/__tests__/__snapshots__/users.ts.snap
@@ -48,6 +48,7 @@ Object {
   "image": "https://daily.dev/ido.jpg",
   "infoConfirmed": false,
   "name": "Ido",
+  "notificationEmail": true,
   "permalink": "http://localhost:5002/a1",
   "timezone": "Asia/Manila",
   "twitter": null,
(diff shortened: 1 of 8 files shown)


idoshamun (changes requested):
Looks perfect! One change I'd like to ask is to respect the new field in the email sending worker. So if it's false we will not send any notification email

rebelchris (approved):
Nothing besides Ido's comment

el (approved):
LGTM, maybe we can improve the naming of the field from `notificationEmail` to `notifyByEmail` but not necessary

sshanzel:
@idoshamun I have also added a test to ensure the worker won't send any email in case the user preferred not to receive them.

sshanzel:
> LGTM, maybe we can improve the naming of the field from `notificationEmail` to `notifyByEmail` but not a blocker

@el good point. We have a plan in the future to make a separate table for the notifications preferences, we can probably apply the suggestion in the next phase. WDYT?

PR #1102, opened 2022-12-22
feat: do not send push notif to connected users

idoshamun:
Use Redis to keep a list of connected users to our gql subscription. Using this list we decide whether to send a push notif or not.

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -227,7 +227,7 @@ const [apps] = deployApplicationSuite(
           ],
     },
     debezium: {
-      version: '1.9',
+      version: isAdhocEnv ? '2.0' : '1.9',
       topicName: debeziumTopicName,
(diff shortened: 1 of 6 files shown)


.infra/index.ts
```diff
@@ -227,7 +227,7 @@ const [apps] = deployApplicationSuite(
           ],
     },
     debezium: {
-      version: '1.9',
+      version: isAdhocEnv ? '2.0' : '1.9',
```
sshanzel, line 230: Not sure if this was intended.

sshanzel (approved):
Looking super tight! 🚢 🚢 🚢

.infra/index.ts
idoshamun: Yes, it's a fix for local env

PR #1111, opened 2022-12-25
feat: squads notifications

idoshamun:
Add notifications for:
* New post added to squad
* Squad post was viewed
* New member joined the squad

Blocker: Still waiting for emails

WT-834 #done
WT-836 #done
WT-837 #done

PR #1132, opened 2023-01-08
fix: local environment

sshanzel:
This fixes the issues below:
- Importing seeds fails due to the env variable missing values. I have values for it both on `.env` and `.env.development` but it somehow failed. To be on the safe side and have a smoother experience, I applied a fallback value.
- Importing Source fails due to the missing `handle` property.
- Feed is empty and have to manually run a command to update the `post.type` for posts to be available.

After doing manual fixes, I extracted the new data in this PR by running `npm run db:seed:export`


rebelchris:
@sshanzel i think this is all fixed in Idos notifications branch already

sshanzel:
> @sshanzel i think this is all fixed in Idos notifications branch already

Unfortunately, I am still having the issue.

Steps to reproduce:
1. Pull the main branch to be up-to-date.
2. Execute `npm run db:migrate:reset`.
3. Execute `npm run db:seed:import`.

Results to below image:
image

idoshamun:
I already applied the relevant fixes here: https://github.com/dailydotdev/daily-api/pull/1111
Unfortunately, it was blocked more than expected. Hopefully, I can resolve everything this week

PR #1135, opened 2023-01-12
fix: temporary upvote fix for inventory issue

rebelchris:
Temporary set to 0 upvotes because of inventory issue

### DIFF
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -941,11 +941,12 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
         alias,
       ) => {
         const interval = [7, 30, 365].find((num) => num === period) ?? 7;
+        // !TODO: Temporary switched from 10 upvotes to 0 because of Dec 2022 inventory issue
         return builder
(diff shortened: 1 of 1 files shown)


src/schema/feeds.ts
```diff
@@ -941,11 +941,12 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
         alias,
       ) => {
         const interval = [7, 30, 365].find((num) => num === period) ?? 7;
+        // !TODO: Temporary switched from 10 upvotes to 0 because of Dec 2022 inventory issue
         return builder
           .andWhere(
             `${alias}."createdAt" > now() - interval '${interval} days'`,
           )
-          .andWhere(`${alias}."upvotes" >= 10`)
+          .andWhere(`${alias}."upvotes" >= 0`)
```
idoshamun, line 949: Just so we will not return posts with zero upvotes
```suggestion
          .andWhere(`${alias}."upvotes" > 0`)
```

idoshamun (approved):

PR #1136, opened 2023-01-12
feat: add members count field to source

idoshamun:
Add new field that returns the number of members in the squad.

Also fixed an issue with creation of squad, now they are set to private by default.


rebelchris (approved):
Nice work

PR #1137, opened 2023-01-12
feat: delete squad mutation

rebelchris:
Added delete squad option

WT-842

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -276,7 +276,7 @@ export const typeDefs = /* GraphQL */ `
     ): Source! @auth
 
     """
-    Deletes a inactive squad
+    Deletes a squad
     """
     deleteSource(
       """
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -304,6 +304,10 @@ export const canAccessSource = async (
   source: Source,
   permission = SourcePermissions.View,
 ): Promise<boolean> => {
+  if (permission === SourcePermissions.View && !source.private) {
+    return true;
+  }
+
   if (ctx.userId) {
     const member = await ctx.con.getRepository(SourceMember).findOneBy({
       userId: ctx.userId,
@@ -311,11 +315,6 @@ export const canAccessSource = async (
     });
 
     switch (permission) {
-      case SourcePermissions.View:
-        if (!source.private) {
-          return true;
-        }
-        break;
       case SourcePermissions.Post:
         if (source.type !== 'squad') {
           return false;
@@ -331,6 +330,7 @@ export const canAccessSource = async (
       return true;
     }
   }
+
   return false;
 };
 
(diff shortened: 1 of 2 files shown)


src/schema/sources.ts
```diff
@@ -273,6 +274,16 @@ export const typeDefs = /* GraphQL */ `
       """
       token: String
     ): Source! @auth
+
+    """
+    Deletes a inactive squad
+    """
+    removeSource(
```
idoshamun, line 281: Remove sounds less strict than delete

```suggestion
    deleteSource(
```

src/schema/sources.ts
```diff
@@ -575,6 +586,25 @@ export const resolvers: IResolvers<any, Context> = {
         throw err;
       }
     },
+    removeSource: async (
```
rebelchris, line 589:
```suggestion
    deleteSource: async (
```

src/schema/sources.ts
```diff
@@ -575,6 +586,25 @@ export const resolvers: IResolvers<any, Context> = {
...
+      ctx,
+    ): Promise<GQLEmptyResponse> => {
+      const sourceMember = await ctx.con
+        .getRepository(SourceMember)
+        .findOneByOrFail({ sourceId, userId: ctx.userId });
+      if (sourceMember.role !== SourceMemberRoles.Owner) {
+        throw new ForbiddenError(
+          'Access denied! You do not have permission for this action!',
+        );
+      }
```
idoshamun, line 601: I created a function called `ensureSourcePermissions`.
https://github.com/dailydotdev/daily-api/blob/a044d864444e2dde848a3e5a99b0c59541d7493e/src/schema/sources.ts#L319
I suggest extending its functionality to support the delete operation to make permissions check consolidated.

src/schema/sources.ts
```diff
@@ -575,6 +586,25 @@ export const resolvers: IResolvers<any, Context> = {
...
+      await ctx.con.getRepository(Source).delete({
+        id: sourceId,
+        active: false,
```
idoshamun, line 604: Why do you check for active false?

rebelchris: Because at time of writing you could only delete inactive sources. (1 member)
With the new addition this is not needed.

src/schema/sources.ts
rebelchris: @idoshamun Oh thanks for that, modified, let me know what you think.

src/schema/sources.ts
```diff
@@ -325,7 +342,10 @@ export const ensureSourcePermissions = async (
     const source = await ctx.con
       .getRepository(Source)
       .findOneByOrFail([{ id: sourceId }, { handle: sourceId }]);
-    if (await canAccessSource(ctx, source, permission)) {
+    const sourceMember = await ctx.con
```
idoshamun, line 345: Can you unify it with the query above using `Promise.all` just to make it concurrent

idoshamun, line 345: I see we already fetch the relevant source member inside `canAccessSource` so maybe instead of fetching it here we can move the condition from the switch case to inside the if. We also have to make sure the source is a squad just like we do to post

src/schema/sources.ts
```diff
@@ -273,6 +274,16 @@ export const typeDefs = /* GraphQL */ `
       """
       token: String
     ): Source! @auth
+
+    """
+    Deletes a inactive squad
```
idoshamun, line 279:
```suggestion
    Deletes a squad
```

idoshamun (approved):
Yes! Looks great

PR #1140, opened 2023-01-15
feat: current member

rebelchris:
Added a function to retrieve the current logged in users role for the requested squad.


src/schema/sources.ts
```diff
@@ -273,6 +279,16 @@ export const typeDefs = /* GraphQL */ `
       """
       token: String
     ): Source! @auth
+
+    """
+    Removes the logged-in user as a member from the source
+    """
+    leaveSource(
```
idoshamun, line 286: Did you branch out from the leave source branch, perhaps?

src/schema/sources.ts
```diff
@@ -117,6 +118,11 @@ export const typeDefs = /* GraphQL */ `
     Number of members in the source
     """
     membersCount: Int!
+
+    """
+    Logged Member role
+    """
+    currentMember: String
```
idoshamun, line 125: I think it would be more consistent to return a `SourceMember` object here. What do you think?

rebelchris: Hmm we could, but only really care about the role for that specific squad.
What other props would you be interested in?

idoshamun: Right now it's only this, but for future compatibility it's better to use the object.
We might want to present join date to the squad or notifications settings in the future and so on. Using the source member as an object here will give you support in all of them out of the box, whereas using only role may become legacy.

rebelchris:
@idoshamun Added a relationalfield instead

__tests__/sources.ts
```diff
@@ -186,6 +186,37 @@ query SourceByFeed($data: String!) {
   });
 });
 
+describe('query source current member', () => {
```
idoshamun, line 189: Can you please add a test of an anonymous user and user which is not a source member?

src/schema/sources.ts
```diff
@@ -118,6 +119,11 @@ export const typeDefs = /* GraphQL */ `
     Number of members in the source
     """
     membersCount: Int!
+
+    """
+    Logged-in member object
+    """
+    currentMember: SourceMember!
```
idoshamun, line 126: It can be null

```suggestion
    currentMember: SourceMember
```

__tests__/sources.ts
rebelchris: It should fail right? as it's trying to query role which is required on a member?

idoshamun: I'd say return null. We wouldn't want the whole query to fail in this case.
It's a valid use case for public squads for example

rebelchris: Yep agreed, modified the query and test cases now.

idoshamun (approved):
Looks great!

PR #1142, opened 2023-01-15
fix: add contacts to sendgrid

idoshamun:
For the past few months, there was an issue with adding registered users to SendGrid contact lists. 
I added a new worker that subscribes to a new user event and if the info is confirmed, it will add it to SendGrid. 
Before that, we listened only for updates, which caused many users not to be added in the first place. 
Furthermore, when we checked for updates, the condition we had there didn't support the case where the only change is info confirmed. The condition always checked for email change


rebelchris (approved):
Looks great, thanks for the urgent fix.

PR #1143, opened 2023-01-15
fix: case insensitive source search

sshanzel:
- Fixes case-insensitive search (can be seen through the additional test - and existing ones still work as intended).
- Since our dedicated page for squads can be accessed through `id` or `handle` - the invitation link page can be built to use the handle as the parameter for joining the squad. Hence, we refactored the source getter to utilize the generic function we have.


rebelchris (approved):
Looks great, thanks Lee

idoshamun:
The use of `ILIKE` on user provided input is very dangerous and can lead to sql injection. The easy fix here would be to lowercase everything client side as on server we lowercase everything on save.
How does it sound?

sshanzel:
> The use of ILIKE on user provided input is very dangerous and can lead to sql injection. The easy fix here would be to lowercase everything client side as on server we lowercase everything on save.
How does it sound?

I think that's a great idea since there will be no uppercase on our DB side. Though I checked from the repo it is safe from SQL injection. Should we still switch the parameters from the Frontend?

https://github.com/typeorm/typeorm/issues/7784#issuecomment-866516667

idoshamun:
That's great to know thanks for sharing! But ILike is also more expensive than a simple equal check I think. Also we wouldn't want to use it for IDs only for the handles. All in all I strongly think we should cover it on the client side

sshanzel:
> That's great to know thanks for sharing! But ILike is also more expensive than a simple equal check I think. Also we wouldn't want to use it for IDs only for the handles. All in all I strongly think we should cover it on the client side

Cool. Let's do everything in the Frontend then. Let me revert the changes relating to usage of ILike.

rebelchris:
Good point @idoshamun its definitely cheaper to do it client side regardless 🤙

sshanzel:
@idoshamun I have now reverted the usage of `ILike` and kept the fix for `joinSource` where it can pass `handle` or `id`. Thanks for the awesome review! 🚢

PR #1147, opened 2023-01-23
feat: check handle outside of permission

sshanzel:
To validly check whether the `handle` is available or not, we must check it outside of the user's permission. To consider the security of the information, we will only return a boolean value instead of the whole `Source` data itself as commonly done in other queries.

Relevant discussion: https://dailydotdev.slack.com/archives/C03QZBF46JC/p1674486326108479?thread_ts=1674402848.080819&cid=C03QZBF46JC

### DIFF
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -279,28 +279,42 @@ query Source($id: ID!) {
   });
 });
 
-describe('query sourceHandleTaken', () => {
+describe('query sourceHandleExists', () => {
   const QUERY = `
-    query SourceHandleTaken($handle: String!) {
-      sourceHandleTaken(handle: $handle) 
+    query SourceHandleExists($handle: String!) {
+      sourceHandleExists(handle: $handle) 
     }
   `;
 
   const updateHandle = (handle = 'a') =>
     con.getRepository(Source).update({ id: 'a' }, { handle, private: true });
 
+  it('should not authorize when user is not logged in', () =>
+    testQueryErrorCode(
+      client,
+      { query: QUERY, variables: { handle: 'a' } },
+      'UNAUTHENTICATED',
+    ));
+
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Looks good.
Do we use anything similar yet?

Was mainly wondering about the positive annotation compared to fail like one.
(taken vs exists or free) but very open to whatever we have now.

src/schema/sources.ts
```diff
@@ -239,6 +239,11 @@ export const typeDefs = /* GraphQL */ `
     Get source member by referral token
     """
     sourceMemberByToken(token: String!): SourceMember!
+
+    """
+    Check if source handle is already taken
+    """
+    sourceHandleTaken(handle: String!): Boolean!
```
idoshamun, line 246: This needs to be an authenticated query

__tests__/sources.ts
```diff
@@ -279,6 +279,31 @@ query Source($id: ID!) {
   });
 });
 
+describe('query sourceHandleTaken', () => {
```
idoshamun, line 282: Please add test case to validate that the an error is thrown if not authenticated

src/schema/sources.ts
```diff
@@ -472,6 +477,11 @@ export const resolvers: IResolvers<any, Context> = {
       await ensureSourcePermissions(ctx, id);
       return getSourceById(ctx, info, id);
     },
+    sourceHandleTaken: async (_, { handle }: { handle: string }, ctx) => {
```
idoshamun, line 480: I think we need to lowercase the handle right?

sshanzel: I agree, even though we lowercase it in FE, it would be better to ensure it is lowercased here. Thanks!

idoshamun (approved):

PR #1149, opened 2023-01-24
feat: new feature table + squad access notif

rebelchris:
Added a new feature table.
Added a new flow:

-> Source member added -> Add feature access for squads
-> Feature added -> Send notification (currently only squads supported)

## Todo:
[ ] Wait for email template + test

WT-924

Related PRs:
https://github.com/dailydotdev/streams/pull/6
https://github.com/dailydotdev/apps/pull/1511
https://github.com/dailydotdev/daily-gateway/pull/573


src/workers/newNotificationMail.ts
```diff
@@ -53,6 +53,7 @@ const notificationToTemplateId: Record<NotificationType, string> = {
   squad_new_comment: 'd-587c6c6fd1554fdf98e79b435b082f9e',
   squad_reply: 'd-cbb2de40b61840c38d3aa21028af0c68',
   squad_post_viewed: 'd-dc0eb578886c4f84a7dcc25515c7b6a4',
+  squad_access: 'missing',
```
rebelchris, line 56: Still awaiting this

src/workers/newNotificationMail.ts
```diff
@@ -428,6 +429,9 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
       user_image: author.image,
     };
   },
+  squad_access: async () => {
+    return {};
+  },
```
rebelchris, line 434: Need to fill when we get the template

src/workers/notifications/featureAccessNotification.ts
```diff
@@ -0,0 +1,24 @@
...
+  feature: ChangeObject<Feature>;
+}
+
+const worker: NotificationWorker = {
+  subscription: 'api.feature-access-notification',
+  handler: async (message) => {
+    const { feature: changeFeature }: Data = messageToJson(message);
+
+    // We only support notifications for squad access
+    if (changeFeature.feature !== 'squad') {
```
rebelchris, line 16: Potentially in future we can add a switch or state here?

src/workers/squadFeatureAccess.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    if (hasAccess) {
+      console.log('found one', hasAccess);
+      return;
+    }
+
+    console.log('should insert!');
+
+    await con.getRepository(Feature).insert({
+      feature: 'squad',
```
sshanzel, line 23: I think we can store this in an `enum`.

sshanzel (commented):
Looking great for initial review! Thanks for the initiative on making this! 🚢 🚢 🚢

src/workers/squadFeatureAccess.ts
rebelchris: Ah yes, actually did it on the FE, so let's pull it inline! 🙏

src/routes/boot.ts
```diff
@@ -25,6 +26,11 @@ const excludeProperties = <T, K extends keyof T>(
   return obj;
 };
 
+const getFeatures = async (
```
idoshamun, line 29: Should we keep the same object schema as flagsmith?

src/workers/squadFeatureAccess.ts
```diff
@@ -0,0 +1,27 @@
...
+      feature: FeatureType.Squad,
+      userId: member.userId,
+    };
+    const hasAccess = await con.getRepository(Feature).findOneBy(variables);
+
+    if (hasAccess) {
+      return;
+    }
+
+    await con.getRepository(Feature).insert(variables);
```
idoshamun, line 23: I'd like to suggest doing only the insert with try and catch. If the error thrown is due to duplicate we just ignore it otherwise we throw the error. Will turn 2 queries into one :)

idoshamun (changes requested):
I have some comments, also blocking for the email

src/routes/boot.ts
rebelchris: Hmm, we can do that as well, but we technically are not building a full blown feature management system now.
Do you think it won't be overkill?

src/workers/squadFeatureAccess.ts
rebelchris: Ah good point! 🙏

src/routes/boot.ts
idoshamun: I'm suggesting it so we can merge it easily with flagsmith flags so the client will not care of who owns the flags actually. It's just the schema so I don't see it as an overkill for now

rebelchris:
@idoshamun Would love to hear your feedback on the changes

src/routes/boot.ts
```diff
@@ -25,6 +27,17 @@ const excludeProperties = <T, K extends keyof T>(
...
+  con: DataSource,
+  userId: string,
+): Promise<IFlags[]> => {
```
idoshamun, line 33: The return object is not an array but an object where the key is the name of the feature

src/routes/boot.ts
```diff
@@ -74,13 +88,15 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
         squads,
+        features,
       });
     }
     return res.send({
       alerts: ALERTS_DEFAULT,
       settings: SETTINGS_DEFAULT,
       notifications: { unreadNotificationsCount: 0 },
       squads: [],
+      features: [],
```
idoshamun, line 99: Should be an empty object not array
```suggestion
      features: {},
```

src/workers/squadFeatureAccess.ts
```diff
@@ -0,0 +1,36 @@
...
+      logger.error(
+        {
+          member,
+          messageId: message.messageId,
+          err,
+        },
+        'failed to give user squad feature access',
+      );
+      // Query failed or status is duplicate
+      if (err.name === 'QueryFailedError' || err.code === 187) {
```
idoshamun, line 28: If query failed I think we should throw the error. We should skip only in case of duplicate.
Also I think we defined an enum to deal with postgres error codes

rebelchris: This was copied from existing implementation, but i'll take a look at the enums.
Agreed on the QueryFailed

rebelchris: Ah we only got them for TypeORM errors:
```
export enum TypeOrmError {
  FOREIGN_KEY = '23503',
  DUPLICATE_ENTRY = '23505',
}
```

idoshamun: We haven't migrated fully for the error code enums. It's a relatively new thing

idoshamun: @rebelchris this should do it. Please use them. It should be the same code

rebelchris:
@idoshamun Revamped and also modified the frontend now to use a combined flags, thanks for that suggestion.

src/workers/newNotificationMail.ts
rebelchris: @idoshamun Added the template 🙏

src/workers/squadFeatureAccess.ts
```diff
@@ -0,0 +1,38 @@
...
+  subscription: 'api.squad-feature-access',
+  handler: async (message, con, logger): Promise<void> => {
+    const { sourceMember: member }: Data = messageToJson(message);
+    try {
+      await con.getRepository(Feature).insert({
+        feature: FeatureType.Squad,
+        userId: member.userId,
+      });
+    } catch (err) {
+      console.log(err.code);
```
idoshamun, line 20: Not needed

src/workers/squadFeatureAccess.ts
```diff
@@ -0,0 +1,38 @@
...
+      logger.error(
```
idoshamun, line 21: This needs to be logged after the if because if it already exists so there's no really an error

idoshamun (approved):
Excellent! Two minor comments before you merge but ready to go

PR #1150, opened 2023-01-24
refactor: user mention inside a squad 

sshanzel:
- Recommended users to mention must be part of the squad if the `squadId` parameter is passed.
- Posting a comment must check if the source is a `Squad` then filter the mentions to be only users inside the Squad itself.
- Preview comment should work similarly as posting a comment but the `squadId` will be from a parameter.

Relevant test cases for all 3 of the items above were created.


rebelchris (changes requested):
Amazing!
Love this PR, had some minor suggestions.

src/schema/comments.ts
```diff
@@ -309,6 +327,20 @@ export interface GQLUserCommentsArgs extends ConnectionArguments {
...
+const getSquadMembersFromIds = (
+  con: DataSource | EntityManager,
+  sourceId: string,
+  ids: string[],
+): Promise<SourceMember[]> =>
+  con
+    .getRepository(SourceMember)
+    .createQueryBuilder('sm')
+    .select('sm."userId"')
+    .innerJoin(Source, 's', 's.id = sm."sourceId"')
```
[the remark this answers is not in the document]
sshanzel: Ah, this was due to the possibility that the parameter for `sourceId` is actually the handle, but as you mentioned, we control it and we can send the actual id. I will reduce this to a plain SquadMember call first thing.

src/entity/Source.ts
```diff
@@ -17,6 +17,11 @@ export const COMMUNITY_PICKS_SOURCE = 'community';
 export const SQUAD_IMAGE_PLACEHOLDER =
   'https://daily-now-res.cloudinary.com/image/upload/v1672041320/squads/squad_placeholder.jpg';
 
+export enum SourceType {
+  Machine = 'machine',
+  Squad = 'squad',
+}
```
rebelchris, line 23: You can actually also set this as the type for `type` it just means we need to adjust it everywhere, but it's worthwhile doing so.

src/schema/comments.ts
```diff
@@ -523,14 +561,28 @@ export const resolvers: IResolvers<any, Context> = {
...
+          ? recommendUsersByQuery(con, userId, { query, limit })
+          : recommendUsersToMention(con, postId, userId, { limit }));
+
+        if (!sourceId) {
+          return userIds;
+        }
+
+        const members: SourceMember[] = await con
+          .getRepository(SourceMember)
+          .findBy({ userId: In(userIds), sourceId });
```
rebelchris, line 580: Is the `In` needed?
Does it not limit the mentions, by only recommended people in the first place?

Maybe it doesn't though.

rebelchris (changes requested):
Two more questions, but liking this approach.

src/entity/Source.ts
sshanzel: I have thought of it but I was afraid it might change the data type to enum (as postgresql supports it), so I kept the original values and types. WDYT?

src/schema/comments.ts
sshanzel: This is already the recommended people. We are just filtering it.

src/entity/Source.ts
rebelchris: Hmm I've done it for Features now, @idoshamun Do you know if this negatively impacts?
I personally prefer enums, but if it's bad we should not do it.

idoshamun: You can use the typescript enum for the variable type and make sure typeorm keeps treating it as text. In this case it has no impact

sshanzel: Alrighty. Will carefully do the changes with this one.

sshanzel:
@rebelchris @idoshamun I have now updated all the strings to utilize the newly introduced enum. I also tried to generate a migration to see if it would notice any difference between the current data type, and nothing resulted in any required change.

rebelchris (approved):
Nice! 

Looks amazing Lee, thanks for the extra effort.

sshanzel:
I've also checked the functionality of mention on a regular (public) Source feed. Works as intended (includes everyone).

src/schema/comments.ts
```diff
@@ -333,9 +364,24 @@ const getMentions = async (
...
+  const source = await con.getRepository(Source).findOneBy({ id: sourceId });
+
+  if (!source.private) {
+    return users;
+  }
+
+  const ids = users.map(({ id }) => id);
+  const members = await getMemberIds(con, sourceId, ids);
+
+  return users.filter(({ id }) => members.includes(id));
```
idoshamun, line 384: It's better to do this within the database. select members that are in the provided array. Otherwise there could be squads with hundreds of members

sshanzel: We are doing that already in the `getMemberIds`

src/schema/comments.ts
```diff
@@ -659,9 +739,13 @@ export const resolvers: IResolvers<any, Context> = {
         if (comment.userId !== ctx.userId) {
           throw new ForbiddenError("Cannot edit someone else's comment");
         }
+        const post = await comment.post;
+        const source = await post.source;
```
idoshamun, line 743:
```suggestion
        const [post, source] = await Promise.all([comment.post, comment.source]);
```

sshanzel: Oh, wow. Didn't know this one. Amazing!

src/schema/comments.ts
idoshamun: Oh I see, but still it's not needed as it can be basically turned into one sql statement that returns the users already filtered

sshanzel: This only happens when you have `sourceId` in your parameters. If none, it is just the user id. Though we can definitely combine them and delete the `getMemberIds` function.

idoshamun: Yes. I think this needs to be fixed

src/schema/comments.ts
sshanzel: I tried this but it failed since the comment doesn't have the property of `source` - we will have to make it a sequence.

sshanzel:
@idoshamun instead of doing the decision at the query schema call, I refactored it to happen on the functions we use to generate the actual mentions (talking about recommended users to mention while typing). Then I applied the single query to fetch users related to squad members. Tests have turned green, let me know what you think! 🚢

idoshamun (approved):
Looks great! We'll have to carefully test it as well

src/common/users.ts
```diff
@@ -119,6 +120,12 @@ export const getRecentMentionsIds = async (
     .where({ commentByUserId: userId })
     .andWhere({ mentionedUserId: Not(userId) });
 
+  if (sourceId) {
+    queryBuilder = queryBuilder
+      .innerJoin(SourceMember, 'sm', 'sm."userId" = cm."mentionedUserId"')
+      .andWhere('sm."sourceId" = :sourceId', { sourceId });
+  }
```
rebelchris, line 127: This is quite nice, cool addition 🙏

rebelchris: The only future proofing is to abstract it, but really not high prio for now.

rebelchris (approved):
Really nice addition by querying the source right away.

PR #1155, opened 2023-01-29
fix: refactor referral to relationship field

rebelchris:
Migrated referral column to be a relationship to a user.
Also modified the `newUser` route to handle legacy input + new input.
And added 2 test cases for these scenario's

WT-988

### DIFF
--- a/src/entity/User.ts
+++ b/src/entity/User.ts
@@ -5,6 +5,7 @@ import {
   Entity,
   EntityManager,
   Index,
+  ManyToOne,
   OneToMany,
   PrimaryColumn,
 } from 'typeorm';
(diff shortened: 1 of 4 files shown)


src/entity/User.ts
```diff
@@ -267,7 +276,7 @@ export const addNewUser = async (
         profileConfirmed: data.profileConfirmed,
         infoConfirmed: isInfoConfirmed(data),
         createdAt: data.createdAt,
-        referral: data.referral,
+        referralId: data.referralId,
```
idoshamun, line 279: Should it not be?

```suggestion
        referralId: data.referralId || data.referral,
```

src/entity/User.ts
```diff
@@ -118,12 +126,13 @@ export type AddUserData = Pick<
...
   | 'github'
   | 'twitter'
-  | 'referral'
+  | 'referralId'
   | 'infoConfirmed'
   | 'profileConfirmed'
   | 'acceptedMarketing'
   | 'timezone'
 >;
+export type AddUserDataPost = { referral: string } & AddUserData;
```
idoshamun, line 135: `{ referral: string }` this part is no longer applicable right?

idoshamun (approved):

src/entity/User.ts
rebelchris: Yep, it's just added there to retroactively support existing implementations.
We merge referral to referralId if it exists.

rebelchris: It's a new type only for the post option

idoshamun:
Ok let's do it :)

PR #1160, opened 2023-01-31
refactor: activate squad instantly when a member joins

idoshamun:
Instead of using CDC for squad activations, we use transaction to make sure it's an atomic operation.


sshanzel (approved):
Nice! Thanks for the fix!

PR #1162, opened 2023-02-02
fix: only allow public posts in ws

rebelchris:
Modified the shared post retrieval to evaluate wether to even publish to `ws` stream.
The one test case is generic as it won't even get to that part, the newly introduced test is to simply ensure the functions behaviour over time.

WT-983 #done

### DIFF
--- a/__tests__/workers/postCommentedRedis.ts
+++ b/__tests__/workers/postCommentedRedis.ts
@@ -1,11 +1,12 @@
 import { expectSuccessfulBackground, saveFixtures } from '../helpers';
 import worker from '../../src/workers/postCommentedRedis';
-import { ArticlePost, Source } from '../../src/entity';
+import { ArticlePost, Post, Source } from '../../src/entity';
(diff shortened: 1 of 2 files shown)


sshanzel (commented):
Code-wise, this looks good to me.

Are posts in a private Squad considered a private post?

~~If yes, how would we send updates to the post in "real-time" since we have also disabled refetching of data in the frontend.~~ Edit: fixed.

rebelchris:
After discussion we should actually modify this to publish the event, but if post is private only send it if user is part of the source.

idoshamun:
@rebelchris just highlighting that this is not a blocker

capJavert (approved):
Looks good to me.

idoshamun (approved):

omBratteng:
Was this supposed to be merged? The task in jira is marked as done 😬

rebelchris:
Nah its missing some nuance we put it on hold, probably can just delete at this point and redo 🫠

PR #1163, opened 2023-02-03
fix: validate handle based on regex before checking availability

sshanzel:
When we check for the availability of the handle, we must also check if the given parameter is actually valid based on our rules.

This would also allow us to keep the form validations in the same step rather than doing it when the user is finishing the Squad they are trying to create.

### DIFF
--- a/src/common/object.ts
+++ b/src/common/object.ts
@@ -22,10 +22,8 @@ type Value = string;
 type IsRequired = boolean;
 export type ValidateRegex = [Key, Value, RegExp, IsRequired?];
 
-export const validateRegex = (
-  params: ValidateRegex[],
-): Record<string, string> =>
-  params.reduce((result, [key, value, regex, isRequired]) => {
+export const validateRegex = (params: ValidateRegex[]): void => {
+  const result = params.reduce((result, [key, value, regex, isRequired]) => {
     if (isNullOrUndefined(value)) {
       return isRequired ? { ...result, [key]: `${key} is required!` } : result;
     }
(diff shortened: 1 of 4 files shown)


src/common/object.ts
```diff
@@ -36,3 +37,11 @@ export const validateRegex = (
...
 export const handleRegex = new RegExp(/^@?([\w-]){1,39}$/i);
 export const descriptionRegex = new RegExp(/^(.){1,250}$/);
+
+export const validateRegexOrFail = (params: ValidateRegex[]): void => {
+  const regexResult = validateRegex(params);
+
+  if (Object.keys(regexResult).length) {
+    throw new ValidationError(JSON.stringify(regexResult));
+  }
+};
```
rebelchris, line 47: We already have a function for this.
I'd suggest we either use that or revamp the other to this?

You can find it as `validateRegex`

rebelchris (changes requested):
Just one concern about duplication and future maintenance.

src/common/object.ts
sshanzel: Yes, this is it. It is using `validateRegex`. This just abstracted the throwing of error since we automatically do that.

sshanzel: This is like the typeorm's function `findOne` and `findOneOrFail`.

rebelchris: Yes I understand, but this is not exactly doing what those 2 other implementations do, so maybe we should convert those into this function 🙏

sshanzel: Cool. Let me update this existing function to automatically throw since we do that anyways.

rebelchris (approved):
Amazing!
And thanks for thinking about the test case! 🙏

idoshamun (approved):

el (approved):
great work ❤️

PR #1164, opened 2023-02-03
fix: option to return only public posts

rebelchris:
Option to return only public posts in reading history list.


idoshamun:
@rebelchris can you add tests please?

rebelchris:
@idoshamun Added public test case as well now.

el (approved):
Looks great

idoshamun (approved):

PR #1165, opened 2023-02-06
feat: add createdAt to source table

rebelchris:
Add createdAt column to source table.

WT-1029 #done

### DIFF
--- a/src/entity/Source.ts
+++ b/src/entity/Source.ts
@@ -34,7 +34,6 @@ export class Source {
   type: SourceType;
 
   @Column({ default: () => 'now()' })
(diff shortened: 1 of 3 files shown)


src/entity/Source.ts
```diff
@@ -33,6 +33,10 @@ export class Source {
   @Column({ default: SourceType.Machine })
   type: SourceType;
 
+  @Column({ default: () => 'now()' })
+  @Index('IDX_source_createdAt', { synchronize: false })
```
idoshamun, line 37: We don't need the index

idoshamun (approved):

PR #1167, opened 2023-02-06
fix: double notification for authors mentioned

sshanzel:
When the author or scout was mentioned, they receive two notifications, one from being an author/scout of the post that is subscribed to the post's activity and one when you get mentioned.

The condition of early return should fix and not send a notification when not needed.

### DIFF
--- a/src/workers/notifications/commentMention.ts
+++ b/src/workers/notifications/commentMention.ts
@@ -24,6 +24,10 @@ const worker: NotificationWorker = {
     if (!postCtx) {
       return;
     }
+    const authors = [postCtx.post.authorId, postCtx.post.scoutId];
+    if (authors.includes(data.commentMention.mentionedUserId)) {
+      return;
+    }
     const commenter = await comment.user;
     const ctx: NotificationCommenterContext = {
       ...postCtx,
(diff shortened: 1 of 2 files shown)


src/workers/notifications/commentMention.ts
```diff
@@ -24,6 +24,10 @@ const worker: NotificationWorker = {
     if (!postCtx) {
       return;
     }
+    const authors = [postCtx.post.authorId, postCtx.post.scoutId];
+    if (authors.includes(data.commentMention.mentionedUserId)) {
+      return;
+    }
```
rebelchris, line 30: Can't we simply use `new set[author, scout]`?
Meaning we could unique the array no?

rebelchris (changes requested):

idoshamun (approved):

src/workers/notifications/commentMention.ts
```diff
@@ -24,6 +24,10 @@ const worker: NotificationWorker = {
     if (!postCtx) {
       return;
     }
+    const authors = [...new Set([postCtx.post.authorId, postCtx.post.scoutId])];
+    if (authors.includes(data.commentMention.mentionedUserId)) {
```
rebelchris, line 28: Do you still need this with the set?

rebelchris (approved):

src/workers/notifications/commentMention.ts
sshanzel: It was not being recognized as an array, but I found a workaround by using `.has` API 🤯

PR #1170, opened 2023-02-08
refactor: removal of view notification on squad post

sshanzel:
Removal of view notification on view post.

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -35,11 +35,9 @@ import { postsFixture, postTagsFixture } from './fixture/post';
 import { Roles } from '../src/roles';
 import { DataSource, DeepPartial } from 'typeorm';
 import createOrGetConnection from '../src/db';
-import { notifyView } from '../src/common';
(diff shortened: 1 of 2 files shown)


idoshamun:
@sshanzel thanks for taking care but unfortunately this is the wrong implementation. Removing this pubsub message can impact reading history and more. We need to remove the notification worker that listens to this event and create notification

sshanzel:
> @sshanzel thanks for taking care but unfortunately this is the wrong implementation. Removing this pubsub message can impact reading history and more. We need to remove the notification worker that listens to this event and create notification

Ah, right. I forgot there could be more subscribers to this topic. Let me update quickly.

sshanzel:
We have two options, current changes (totally deleting the worker) or just removing the subscription. I feel recovering the deleted worker back is easy if ever, so I went for totally deleting the worker. WDYT? @idoshamun

idoshamun (approved):
Awesome!

PR #1171, opened 2023-02-09
fix: order of the squad members

el:
SCR-20230209-im5

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -561,6 +561,7 @@ export const resolvers: IResolvers<any, Context> = {
             .andWhere(`${builder.alias}."sourceId" = :source`, {
               source: args.sourceId,
             })
+            .addOrderBy(`${builder.alias}."role"`, 'DESC')
             .addOrderBy(`${builder.alias}."createdAt"`, 'DESC');
 
           builder.queryBuilder.limit(page.limit);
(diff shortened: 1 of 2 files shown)


src/schema/sources.ts
```diff
@@ -561,6 +561,7 @@ export const resolvers: IResolvers<any, Context> = {
             .andWhere(`${builder.alias}."sourceId" = :source`, {
               source: args.sourceId,
             })
+            .addOrderBy(`${builder.alias}."role"`, 'DESC')
```
sshanzel, line 564: Should we add select commands to manually set the value for each role to get the desired order? Something like `IF role = 'member' THEN 1 ELSE 10` (since we don't have a table in DB to have the ordering using a column to reference).

Not sure if having additional order by criterion will affect the pagination (first time seeing it) since it is currently based on a Date property. I think it should not, but just for visibility cc @idoshamun

idoshamun: I think for now it's fine. But we should definitely keep it in mind once we introduce more roles

idoshamun:
@el thanks for fixing this. Please fix the tests so we can approve the PR :)

rebelchris (approved):
Nice work, glad to see a API PR from you 🚀

As for Lee's comment, great suggestion and something we indeed might tackle once we need that role.
For now this is good.

sshanzel (approved):
Nice work! Since the change was about the order of the response, by updating the snapshots the test should turn green 🟢

el:
@idoshamun fixed 🙏

idoshamun (approved):

PR #1173, opened 2023-02-09
feat: introducing changelog flag for boot

rebelchris:
Added a new alert flag for lastChangelog, FE will be responsible for updating it through existing channels.
Added a new worker to set a redis cached value of our last changelog date.
Boot can then check wether this date is more up to date, and set Changelog to true.

WT-1054 #done

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -89,7 +89,10 @@ export default async function (fastify: FastifyInstance): Promise<void> {
         getRedisObject(REDIS_CHANGELOG_KEY),
       ]);
       return res.send({
-        alerts: excludeProperties(alerts, ['userId']),
+        alerts: {
+          ...excludeProperties(alerts, ['userId']),
+          changelog: alerts.lastChangelog < new Date(lastChangelog),
+        },
         settings: excludeProperties(settings, [
           'userId',
           'updatedAt',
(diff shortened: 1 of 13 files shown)


src/routes/boot.ts
```diff
@@ -89,6 +98,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
         notifications: { unreadNotificationsCount },
         squads,
         features,
+        changelog: alerts.lastChangelog < new Date(lastChangelog),
```
idoshamun, line 101: I forgot we have an alerts object. Should we put this inside the alerts object? Makes more sense I guess?

idoshamun:
@rebelchris let me know when tests are green so I can have another look :)

src/routes/boot.ts
rebelchris: I added the actual changelog date as the user object.
The evaluation always needs to happen seperate, not sure it makes sense to sync that to the database?

idoshamun: We don't need to sync that to the db. I mean something like:
```
alerts: {
  ...alerts,
  lastChangelog: alerts.lastChangelog < new Date(lastChangelog),
}
```
Client doesn't really need the timestamp itself just the true/false.

idoshamun:
@rebelchris I don't remember but do we have a mutation to update the user's alert?

src/routes/boot.ts
rebelchris: Ah right, yeah we can do something like that.
The update will always happen as a date anyway.

rebelchris: Hmm, but userAlerts is actually a datetime in the database, then we're mixing up types which might not be great.
Maybe we can simply return alerts.changelog seperate?

idoshamun: Ok, np 😁

src/routes/boot.ts
```diff
@@ -92,7 +104,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
       });
     }
     return res.send({
-      alerts: ALERTS_DEFAULT,
+      alerts: { ...ALERTS_DEFAULT, lastChangelog: false },
```
idoshamun, line 107: This should be changelog I guess?

rebelchris: Ah yes will change that 🙌

rebelchris: Fixed

idoshamun (approved):

PR #1176, opened 2023-02-16
fix: do not send notification when you mentioned parent author

sshanzel:
Since we have a dedicated worker for sending a notification when you replied to a comment, we should not send another notification when you mentioned the user you replied to.


src/workers/notifications/commentMention.ts
```diff
@@ -28,6 +28,10 @@ const worker: NotificationWorker = {
     if (authors.has(data.commentMention.mentionedUserId)) {
       return;
     }
+    const parent = comment.parentId && (await comment.parent);
+    if (data.commentMention.mentionedUserId === parent?.userId) {
```
idoshamun, line 32: I think all the commenters in the same thread receive a notification not just the parent comment.
So we need to check for that. I suggest looking at the "replied to" notification and see the logic there.

idoshamun (approved):

PR #1182, opened 2023-02-20
fix: time feed annonymous posts

rebelchris:
Add exclusion for time based anonymous feed.

@idoshamun I'm not 100% sure, but it seems the regular `feed` actually already takes sources into account right?

WT-1088 #done


idoshamun (approved):
This is a good as an hotfix but let's log to add support for private posts in the chronological feed. We can deal with it when we get more time

PR #1185, opened 2023-02-22
refactor: alert entity to include state of squad tour viewed

sshanzel:
Introduced a new alert property to identify whether to display the squad tour and be synced across devices.

### DIFF
--- a/src/entity/Alerts.ts
+++ b/src/entity/Alerts.ts
@@ -19,7 +19,7 @@ export class Alerts {
   companionHelper: boolean;
 
   @Column({ type: 'bool', default: true })
-  showSquadTour: boolean;
+  squadTour: boolean;
 
   @Column({ type: 'timestamp without time zone', default: () => 'now()' })
   lastChangelog: Date | null;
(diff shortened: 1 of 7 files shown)


src/entity/Alerts.ts
```diff
@@ -18,6 +18,9 @@ export class Alerts {
   @Column({ type: 'bool', default: true })
   companionHelper: boolean;
 
+  @Column({ type: 'bool', default: true })
+  showSquadTour: boolean;
```
rebelchris, line 22: Actually looking at the other we don't prefix `show` so maybe `squadTour` is enough?

rebelchris (approved):
Just one remark on naming

src/entity/Alerts.ts
idoshamun: Agree with this convention. Let's fix to `squadTour`

sshanzel: Good point. Let me refactor it :ship

sshanzel:
Have updated the property now.

@idoshamun should be ready for your review now 👀

rebelchris (approved):
Looks good!

idoshamun (approved):

PR #1186, opened 2023-02-22
fix: prevent notification for users who are part of squad

rebelchris:
Prevent notification for squad access for users who are already in a squad

WT-1095 #done


idoshamun:
@rebelchris let's add a test case please

rebelchris:
@idoshamun Added 2 test cases for each scenario

idoshamun (approved):
Awesome!

PR #1188, opened 2023-02-23
feat: introduction of squad moderator role

sshanzel:
Introduction of a moderator role. This includes the proper ordering of the source members.

Note: the privileges of a moderator will be worked on a different PR.

### DIFF
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -411,6 +411,7 @@ query SourceMembers($id: ID!) {
     }
     edges {
       node {
+        role
         user { id }
         source { id }
       }
@@ -430,6 +431,27 @@ query SourceMembers($id: ID!) {
     const res = await client.query(QUERY, { variables: { id: 'a' } });
     expect(res.errors).toBeFalsy();
     expect(res.data).toMatchSnapshot();
+  });
+
+  it('should return source members and order by their role', async () => {
+    const noModRes = await client.query(QUERY, { variables: { id: 'a' } });
+    expect(noModRes.errors).toBeFalsy();
+    const [noModFirst, noModSecond] = noModRes.data.sourceMembers.edges;
+    expect(noModFirst.node.role).toEqual(SourceMemberRoles.Owner);
+    expect(noModSecond.node.role).toEqual(SourceMemberRoles.Member);
+
+    await con
+      .getRepository(SourceMember)
+      .update(
+        { userId: '3' },
+        { role: SourceMemberRoles.Moderator, sourceId: 'a' },
+      );
+
+    const res = await client.query(QUERY, { variables: { id: 'a' } });
+    expect(res.errors).toBeFalsy();
+    const [first, second] = res.data.sourceMembers.edges;
+    expect(first.node.role).toEqual(SourceMemberRoles.Owner);
+    expect(second.node.role).toEqual(SourceMemberRoles.Moderator);
   });
 
   it('should return source members of private source when user is a member', async () => {
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -6,6 +6,7 @@ import { Context } from '../Context';
 import {
   createSharePost,
   generateMemberToken,
+  roleRank,
   Source,
   SourceFeed,
   SourceMember,
@@ -558,10 +559,19 @@ export const resolvers: IResolvers<any, Context> = {
           membershipsPageGenerator.nodeToCursor(page, args, node, index),
         (builder) => {
           builder.queryBuilder
+            .addSelect(
+              `
+                CASE
+                  WHEN ${builder.alias}.role = '${SourceMemberRoles.Owner}' THEN ${roleRank.owner}
+                  WHEN ${builder.alias}.role = '${SourceMemberRoles.Moderator}' THEN ${roleRank.moderator}
+                ELSE 0 END AS "roleRank"
+                
+              `,
+            )
             .andWhere(`${builder.alias}."sourceId" = :source`, {
               source: args.sourceId,
             })
-            .addOrderBy(`${builder.alias}."role"`, 'DESC')
+            .addOrderBy('"roleRank"', 'DESC')
             .addOrderBy(`${builder.alias}."createdAt"`, 'DESC');
 
           builder.queryBuilder.limit(page.limit);
--- a/__tests__/__snapshots__/sources.ts.snap
+++ b/__tests__/__snapshots__/sources.ts.snap
@@ -178,6 +178,7 @@ Object {
     "edges": Array [
       Object {
         "node": Object {
+          "role": "owner",
           "source": Object {
             "id": "a",
           },
@@ -188,6 +189,7 @@ Object {
       },
       Object {
         "node": Object {
+          "role": "member",
           "source": Object {
             "id": "a",
           },
@@ -211,6 +213,7 @@ Object {
     "edges": Array [
       Object {
         "node": Object {
+          "role": "owner",
           "source": Object {
             "id": "a",
           },
@@ -221,6 +224,7 @@ Object {
       },
       Object {
         "node": Object {
+          "role": "member",
           "source": Object {
             "id": "a",
           },
--- a/src/entity/SourceMember.ts
+++ b/src/entity/SourceMember.ts
@@ -8,8 +8,15 @@ const randomBytesAsync = promisify(randomBytes);
 
 export enum SourceMemberRoles {
   Owner = 'owner',
+  Moderator = 'moderator',
   Member = 'member',
 }
+
+export const roleRank: Record<SourceMemberRoles, number> = {
+  owner: 10,
+  moderator: 1,
+  member: 0,
+};
 
 @Entity()
 export class SourceMember {


__tests__/sources.ts
```diff
@@ -432,6 +433,27 @@ query SourceMembers($id: ID!) {
     expect(res.data).toMatchSnapshot();
   });
 
+  it('should return source members and order by their role', async () => {
```
idoshamun, line 436: Can you make sure in this test that the following order holds: Owner, Moderator, Member

idoshamun (changes requested):
Please adjust the tests :)

__tests__/sources.ts
sshanzel: Oh, right. I did not include the `Member` on the test. Let me add it quickly.

src/schema/sources.ts
```diff
@@ -822,4 +836,7 @@ export const resolvers: IResolvers<any, Context> = {
   Source: {
     permalink: (source: GQLSource): string => getSourceLink(source),
   },
+  SourceMember: {
+    roleRank: (member: GQLSourceMember) => roleRank[member.role],
+  },
```
sshanzel, line 841: I have introduced this to make it easier to decide in the Frontend whether to show the context menu option of "Remove member".

rebelchris: This would return a string right?
The type you described states `Int`.

Do we need to know the string value though?

rebelchris (changes requested):
One question, otherwise happy!

src/schema/sources.ts
sshanzel: No, this would return an `Int` as shown in the snapshot.

rebelchris (approved):
Looks good! 🥳

src/schema/sources.ts
idoshamun: It doesn't feel right to me to "leak" this information and rely on the frontend.
I'd suggest having a field `canRemoveMember` and the frontend will use this.
Or an alternative would be to return an array of allowed actions. This might even be a more flexible solution for future actions

idoshamun (changes requested):
See my comment

sshanzel:
@idoshamun I have now updated the query to return the property of whether the user has the privilege to remove the user.

With regards to the complete access of the user, I think it would be the best-case scenario though I believe it is very complex and must be tackled in a separate PR.

idoshamun:
@sshanzel I don't want to add a new field and later on deprecate it. It will just cause tech debt.
I suggest returning an array of possible actions for the member instead of having a field per action.
@rebelchris wdyt?

rebelchris:
@idoshamun / @sshanzel Yeah agreed, it's actually better to return actions a user is able to execute.
Important to note they should only be exposed to the user themselves, not any public endpoints requesting user info.

sshanzel:
@rebelchris @idoshamun I have included the permissions on the `currentMember` property as it is very fitting to that context. This is only returned when you are a member of the squad. All the changes related to this should be under this commit: https://github.com/dailydotdev/daily-api/pull/1188/commits/009720b5f15c6f184b65822240ba99a6f450130d

src/schema/sources.ts
```diff
@@ -818,8 +905,42 @@ export const resolvers: IResolvers<any, Context> = {
...
+        ]);
+
+        const memberRank = roleRank[member.role];
+        const loggedUserRank = roleRank[loggedUser.role];
+
+        if (memberRank >= loggedUserRank) {
+          throw new ForbiddenError(
+            `You don't have the required permission for this action!`,
+          );
+        }
```
rebelchris, line 932: Do we want to leave this as is?
Optional could be using the same rank function

src/schema/sources.ts
```diff
@@ -347,13 +374,74 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
 
 export enum SourcePermissions {
-  View,
-  Post,
-  Leave,
-  Delete,
-  Edit,
+  View = 'view',
+  Post = 'post',
+  PostDelete = 'post_delete',
```
rebelchris, line 379: I think there is also a option where members can delete their own posts, how will we tackle this?

src/schema/sources.ts
```diff
@@ -347,13 +374,74 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+  RemoveMember = 'remove_member',
```
rebelchris, line 380: Do you want to evaluate mods as well? 
Or is the proposed FE check for owners enough?

rebelchris (commented):
Mainly some comments for now.

src/graphorm/index.ts
```diff
@@ -176,7 +178,15 @@ const obj = new GraphORM({
               .where(`sm."userId" = :userId`, { userId: ctx.userId })
               .andWhere(`sm."sourceId" = "${parentAlias}".id`),
         },
-        transform: nullIfNotLoggedIn,
+        transform: (value: SourceMember, ctx: Context, parent: Source) => {
```
idoshamun, line 181: This is not the right place to add this field. This field is part of the `SourceMember` object and there are various ways to fetch it. The more accurate and future proof solution would be to add a custom column to the `SourceMember` entity and if the requested user and the logged in are not the same we can return null or empty array

src/schema/sources.ts
sshanzel: We will just make the necessary adjustments in the `hasPermissionCheck` function once implementation comes. Then it should automatically be in the rightful permissions array.

sshanzel: Or do you mean that we do that here, now?

rebelchris:
@sshanzel Thanks for the work on this, but in our meeting this morning we discussed we would rather write it down in the confluence page before building.

Can you please make sure that happens before we continue the PR?

sshanzel:
> @sshanzel Thanks for the work on this, but in our meeting this morning we discussed we would rather write it down in the confluence page before building.

I believe I have already written down the decision we had in the meeting. Not sure what else to write.

src/schema/sources.ts
sshanzel: Good point. Let me extract this as a generic function.

src/graphorm/index.ts
idoshamun: As a rule of thumb, if you added the field to the `SourceMember` object in graphql schema, it needs to be add to graphorm to the same entity. In this case you added it only to a field in the `Source` object. It can create discrepancies and unexpected behavior in other queries. This is not future proof as the client might decide at some point to use a different query to get the permissions but it will not be set properly

sshanzel: Ohh. Now I understand. Do you think we should introduce a separate query for this instead? Since the content here is not really found in the database. Then we can do query batching for these permissions.

sshanzel: Ohh. Now I understand. Thank you for explaining. Do you think we should introduce a separate query for this instead? Since the content here is not really found in the database. Then we can do query batching for these permissions.

idoshamun: @sshanzel we can use a similar implementation to custom fields using JS as can be seen here:
https://github.com/dailydotdev/daily-api/pull/1188/files/009720b5f15c6f184b65822240ba99a6f450130d#diff-928042d056cac827c70794ef59920facfe32f158ebf447ff7d2826fcad5195aeL143

src/schema/posts.ts
```diff
@@ -740,6 +753,38 @@ export const resolvers: IResolvers<any, Context> = {
...
+
+        if (post.authorId !== ctx.userId) {
+          const sourceId = post.sourceId;
+          const memberRepo = manager.getRepository(SourceMember);
+          const [member, loggedUser] = await Promise.all([
+            memberRepo.findOneByOrFail({ sourceId, userId: post.authorId }),
+            memberRepo.findOneByOrFail({ sourceId, userId: ctx.userId }),
+          ]);
+          hasGreaterAccessCheck(loggedUser, member);
+        }
```
rebelchris, line 781: Should this not all be inside the ensureSourcePermissions?

Maybe actually say 

if ctx.user is NOT the post owner, then evaluate if he is a mod or admin.

sshanzel: I'm not sure we should. Some checks can vary from implementation to implementation like on the removal of members, we didn't have to check if the passed parameter is the logged-in user. Might make the function bloated with numerous conditions.

> if ctx.user is NOT the post owner, then evaluate if he is a mod or admin.

I think this is exactly the condition in line 773. Then the 780 evaluates if the user is a mod or admin and checks which has greater access privilege.

rebelchris: But this particular section only evaluates the `postdelete` anyway, so not sure what it has to do with removal of members.

I just don't see where we actually have the `PostDelete` function?
That whole section seems to be not doing anything really.

sshanzel: > But this particular section only evaluates the postdelete anyway, so not sure what it has to do with removal of members.

What I mean it has a similar process where it checks the privilege and their only difference is checking for the author which I think would add more complexity to the function if we pass another object which will only be used once.

> I just don't see where we actually have the PostDelete function?

The default value returned on the function is `true`, we have removed the `PostDelete` section since it is also now available for members. We will be using that enum value in the frontend to display the option of deleting the post. It could be not used here, but it will be utilized by the frontend.

rebelchris: Let's discuss this on Gather rather, think we are meaning the same thing but not aligned on the implementation.

src/schema/sources.ts
```diff
@@ -347,56 +375,131 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+        return false;
+      }
+      break;
+    case SourcePermissions.PostDelete:
+      if (!canPostDeleteRoles.includes(member.role)) {
+        return false;
+      }
+      break;
+    case SourcePermissions.RemoveMember:
+      if (!canRemoveMemberRoles.includes(member.role)) {
```
rebelchris, line 432: Just a thought should we maybe not include `hasGreaterAccessCheck` here?
Then all access happens in one switch?

rebelchris (changes requested):
Amazing, this is starting to look super clean and slick!
Thanks for the hard work Lee, it's not a easy one to crack.

Have some small comments.

src/schema/sources.ts
```diff
@@ -347,56 +375,131 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+
+const canRemoveMemberRoles = [
+  SourceMemberRoles.Owner,
+  SourceMemberRoles.Moderator,
+];
+
+const canPostDeleteRoles = [
+  SourceMemberRoles.Owner,
+  SourceMemberRoles.Moderator,
+];
```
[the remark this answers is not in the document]
sshanzel: Really agree with that. Let me change the one for members so it will be easier to manage if there will be numerous actions under the same context.

Wdyt of below examples:
- member remove
- post delete

So when we can add directly it will be:
- member add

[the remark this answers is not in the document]
rebelchris: That makes sense, let's do that 🙏

src/graphorm/index.ts
```diff
@@ -143,6 +144,15 @@ const obj = new GraphORM({
       public: {
         transform: (value: boolean): boolean => !value,
       },
+      permissions: {
+        transform: (_, ctx: Context, source: GQLSource) => {
+          if (!ctx.userId || !source.currentMember) {
```
idoshamun, line 149: We discussed that this field should be in `SourceMember` not in `Source` please adjust

src/schema/posts.ts
```diff
@@ -740,6 +751,28 @@ export const resolvers: IResolvers<any, Context> = {
       }
       return { _: true };
     },
+    deleteSharedPost: async (
```
idoshamun, line 754: We already have a mutation for deleting a post. I'd suggest streamlining everything to the existing function.
Everything is a post but depending on a type we have a different logic, I don't want to separate the mutations

src/schema/posts.ts
```diff
@@ -824,7 +857,9 @@ export const resolvers: IResolvers<any, Context> = {
     ): Promise<GQLPost> => {
       const post = await ctx.con.getRepository(Post).findOneByOrFail({ id });
       await Promise.all([
-        ensureSourcePermissions(ctx, sourceId, SourcePermissions.Post),
+        ensureSourcePermissions(ctx, sourceId, {
```
idoshamun, line 860: Why do we call twice the same function? Can we maybe reduce the number of calls?

src/schema/sources.ts
```diff
@@ -126,6 +132,11 @@ export const typeDefs = /* GraphQL */ `
     Logged-in member object
     """
     currentMember: SourceMember
+
+    """
+    User squad permissions
+    """
+    permissions: [String]
```
idoshamun, line 139: Why it's here? It should be in `SourceMember`

src/schema/sources.ts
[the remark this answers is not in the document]
idoshamun: Instead of having multiple arrays per action, I suggest having an array of actions per role.
Otherwise there will be more work to add new action or role.

src/schema/sources.ts
```diff
@@ -557,11 +659,18 @@ export const resolvers: IResolvers<any, Context> = {
         (node, index) =>
           membershipsPageGenerator.nodeToCursor(page, args, node, index),
         (builder) => {
+          const roleRankQuery = `
```
idoshamun, line 662: I suggest using the same way to calculate the role rank everywhere. Sometimes you use SQL and sometimes you use the array. Please stick to the latter.

src/graphorm/index.ts
```diff
@@ -143,6 +144,15 @@ const obj = new GraphORM({
...
+            return null;
+          }
+
+          return getUserPermissions(source, source.currentMember);
+        },
+      },
```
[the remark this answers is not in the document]
rebelchris: Ah I see, I thought we had a regular function as well to streamline the null response.
But guess the proposed change is ok then.

idoshamun (changes requested):
I'd like you to focus on the below:
* You introduce many ways to do the same thing which can be hard to maintain and create issues (e.g. role rank with query vs js, array of actions per role vs array of roles per action)
* In the previous review round we agreed that permissions will be part of the `SourceMember` and you moved it now to `Source`.
My suggestion is to try to have another look on everything and see how you can minimize and reuse the logic everywhere.
I'm here if you want to discuss 🙏

sshanzel:
> You introduce many ways to do the same thing which can be hard to maintain and create issues (e.g. role rank with query vs js, array of actions per role vs array of roles per action)

This is the reason why I wanted to have everything in DB as it is inevitable to have the query pulling the information from JS with regard to role rank to be used for sorting (then later on used for actions as well). The DB query has no reference on which has greater access privilege since it resides in JS (roleRank).

> In the previous review round we agreed that permissions will be part of the SourceMember and you moved it now to Source.

Oh, right, but it wasn't supposed to be at `currentMember` property and the property won't be used in the list since the key part is the currently logged-in member's permission which can be pulled with the source too. Apologies, I think I was not able to explain it myself towards the end. But this is so we won't have to sift through the list as it is paginated and the logged-in user could be not part of the available pages yet.

> I'm here if you want to discuss 🙏

Thank you so much for the support. Let me apply yours and Chris' comments then if I hit a bumpy road, I will send out a message 🙇

idoshamun:
> This is the reason why I wanted to have everything in DB as it is inevitable to have the query pulling the information from JS with regard to role rank to be used for sorting (then later on used for actions as well). The DB query has no reference on which has greater access privilege since it resides in JS (roleRank).

So my suggestion is to add a `roleRank` using the custom select to graphorm so we can order by it. This way you will not need the graphql resolver also. To make streamlined the query can loop over the array of ranks and generate itself using that. This way if we introduce a new role rank the query will automatically be updated and it will not require human intervention

src/schema/posts.ts
sshanzel: The `deletePost` requires a direct `moderator` role from a user. 
Schema:
```
deletePost(
      """
      Id of the post to delete
      """
      id: ID
    ): EmptyResponse @auth(requires: [MODERATOR])
```

Not sure what we will try to reuse from the mutation below since it is just a direct call to the repository:
```
deletePost: async (
      source,
      { id }: { id: string },
      ctx: Context,
      await ctx.getRepository(Post).update({ id }, { deleted: true });
      return { _: true };
    },
```

src/schema/sources.ts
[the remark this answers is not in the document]
sshanzel: That would mean we will no longer use the switch case since we will explicitly define their access. Let me update the current implementation quickly.

src/schema/posts.ts
idoshamun: It's not about reusing the logic, it's about making it easy for the client. Client knows it needs to delete post so there is a single mutation for it. Also from maintainer point of view, we know there's only one endpoint to delete a post

sshanzel: I just think endpoints that require higher access levels should be separate for security's sake. But we can merge them if necessary.

What do you think of removing the `requires: [MODERATOR]` and then checking it at the JS level when the user is not a moderator and the post is an article?

src/schema/posts.ts
sshanzel: Oh. Merge issues. Let me remove the other one.

sshanzel: It looks to be from existing implementations. Let me check which one we can remove.

src/schema/posts.ts
idoshamun: Yes makes sense :)

src/schema/sources.ts
sshanzel: This is so that when you load the Squad page, the member viewing the feed would already have all the information regarding whether they are able to do certain actions.

Or do you want to include it in the `SourceMember` then it would be depending on the client whether they want to include the property in the `currentMember`?

src/schema/posts.ts
```diff
@@ -740,6 +751,28 @@ export const resolvers: IResolvers<any, Context> = {
...
+      { id }: { id: string },
+      ctx: Context,
+    ): Promise<GQLEmptyResponse> => {
+      await ctx.con.transaction(async (manager) => {
+        const repo = manager.getRepository(SharePost);
+        const post = await repo.findOneBy({ id });
+
+        if (!post?.sharedPostId) return;
+
+        if (post.authorId !== ctx.userId) {
```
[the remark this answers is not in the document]
sshanzel: That is also checked inside the `ensureSourcePermission` when the user is not a member, they are not allowed to do any action anymore.

src/schema/sources.ts
sshanzel: Good point. We can probably do that, so all checks will happen in a single place.

src/schema/posts.ts
```diff
@@ -741,11 +742,29 @@ export const resolvers: IResolvers<any, Context> = {
...
-      await ctx.getRepository(Post).update({ id }, { deleted: true });
+      if (ctx.roles.includes(Roles.Moderator)) {
+        await ctx.getRepository(Post).update({ id }, { deleted: true });
+        return { _: true };
+      }
+
+      await ctx.con.transaction(async (manager) => {
+        const repo = manager.getRepository(Post);
+        const post = await repo.findOneBy({ id });
+        if (post.type === PostType.Share) {
```
idoshamun, line 757: I'm not sure if it's the right condition or whether we should check if it's a squad post. Let's keep it for now like this

src/schema/sources.ts
```diff
@@ -347,56 +374,146 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+  MemberRemove = 'member_remove',
+  ModeratorAdd = 'moderator_add',
+  ModeratorRemove = 'moderator_remove',
+  InviteDisable = 'invite_disable',
+  Leave = 'leave',
+  Delete = 'delete',
+  Edit = 'edit',
 }
 
+export const roleSourcePermissions: Record<
```
idoshamun, line 390: Suggestion for improvement concat permissions of lower rank and add only the difference for higher ranks.
For example: Moderator is member + PostDelete, MemberRemove etc.
Might be easier to manage

sshanzel: Great idea. I like this one.

src/schema/sources.ts
```diff
@@ -347,56 +374,146 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+    hasGreaterAccessCheck(member, validateRankAgainst);
+  }
+
+  const rolePermissions =
+    roleSourcePermissions[member.role] ?? roleSourcePermissions.member;
+
+  return rolePermissions.includes(permission);
+};
+
+export const getUserPermissions = (
```
idoshamun, line 470: What does this function do exactly? If I understand correctly you can just return the right array of permissions based on the role

src/schema/sources.ts
```diff
@@ -347,56 +374,146 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+  validateRankAgainstId: string;
+}
+
+const hasPermissionCheck = (
+  source: Source | GQLSource,
+  member: BaseSourceMember,
+  permission: SourcePermissions,
+  validateRankAgainst?: BaseSourceMember,
+) => {
+  if (
```
idoshamun, line 453: Can you explain this condition please?

src/schema/sources.ts
```diff
@@ -347,56 +374,146 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+    [SourcePermissions.Post, SourcePermissions.Leave].includes(permission)
+  ) {
+    return false;
+  }
+
+  if (validateRankAgainst) {
+    hasGreaterAccessCheck(member, validateRankAgainst);
+  }
+
+  const rolePermissions =
```
idoshamun, line 464: I don't think we should fall back to member permissions in this case

src/schema/sources.ts
sshanzel: I think I forgot to remove this when I migrated to the new process. My bad. Will clear this one out.

src/schema/sources.ts
```diff
@@ -419,13 +536,13 @@ const validateSquadData = ({
 export const ensureSourcePermissions = async (
   ctx: Context,
   sourceId: string | undefined,
-  permission = SourcePermissions.View,
+  props?: Partial<SourcePermissionsProps>,
```
idoshamun, line 539: Instead of merging both props into one variable, I suggest separating them as in most cases the permission will be provided and the validate against will not. So this will make it a bit shorter to call.

src/schema/sources.ts
sshanzel: This was to support the previous implementation. This is to validate making a post and leaving a source that only happens on a Squad.

https://github.com/dailydotdev/daily-api/pull/1188/files/b94ea967f8264516644d8bbc8a267c36c2b5ed65#diff-95123b05e21b34d3705f1b3748fe295fa17db5caeeb22238a294368729175a07L378-L385

image

src/schema/sources.ts
```diff
@@ -818,8 +947,27 @@ export const resolvers: IResolvers<any, Context> = {
...
+      _,
+      { sourceId, memberId }: RemoveMemberArgs,
+      ctx,
+    ): Promise<GQLEmptyResponse> => {
+      await ensureSourcePermissions(ctx, sourceId, {
+        validateRankAgainstId: memberId,
+        permission: SourcePermissions.MemberRemove,
+      });
+
+      await ctx
```
idoshamun, line 960: I'm not sure that's the right implementation. Product expects to have some sort of blocked users list. Suggesting to scope it out and work separately on this. @rebelchris

src/schema/sources.ts
idoshamun: I see, so let's be more explicit about this and change it to if it's the squad leader they can't leave. Very simple condition

sshanzel: This was also to support avoiding making a post on a non-squad source. So I merged them. But I can definitely separate the two.

idoshamun: Now with the arrays, these edge cases don't exist. You need to be a member to post, otherwise you will not have the permission. But please add or validate the relevant tests to make sure they don't exist indeed

sshanzel: Got it. Will check the test cases if there are any existing checks and create one if there are none.

src/schema/sources.ts
sshanzel: It was optional previously and wanted to keep supporting it but we can definitely make this required and update existing implementations of the omitted third parameter.

src/schema/sources.ts
```diff
@@ -818,8 +911,29 @@ export const resolvers: IResolvers<any, Context> = {
...
+        .delete({ sourceId, userId: memberId });
+
+      return { _: true };
+    },
   }),
   Source: {
     permalink: (source: GQLSource): string => getSourceLink(source),
   },
+  SourceMember: {
+    roleRank: (member: SourceMember) => roleRank[member.role],
```
idoshamun, line 937: Not needed due to graphorm

src/schema/sources.ts
idoshamun: Please make sure it's not a function that runs on every access. We can do it only one time

src/schema/sources.ts
sshanzel: @rebelchris let me know if you want this removed and wait until the "banned" user takes place or just keep it as it is.

src/schema/sources.ts
sshanzel: The snapshot is failing, due to the property always returning null.

idoshamun: Can you try and remove the transform part of this field in graphorm? Maybe this is the issue

sshanzel: I tried it but no luck. Also tried adding the query on the custom relation like below:
```
currentMember: {
        relation: {
          isMany: false,
          customRelation(ctx, parentAlias, childAlias, qb): QueryBuilder {
            return qb
              .select('*')
              .addSelect(this.SourceMember.fields.roleRank.select, '"roleRank"') // added this one
              .from(SourceMember, 'sm')
              .where(`sm."userId" = :userId`, { userId: ctx.userId })
              .andWhere(`sm."sourceId" = "${parentAlias}".id`);
          },
        },
```

idoshamun: I'll try to see if I can fix that. Sec please

src/graphorm/index.ts
```diff
@@ -171,17 +174,40 @@ const obj = new GraphORM({
...
+        transform: (_, ctx: Context, member: SourceMember) => {
+          if (!ctx.userId || member.userId !== ctx.userId) {
+            return null;
+          }
+
+          return (
+            roleSourcePermissions[member.role] ?? roleSourcePermissions.member
+          );
+        },
+      },
```
rebelchris, line 197: Do we need this one as a query?
It seems we can determine it with the roleRank already right?

src/schema/posts.ts
```diff
@@ -741,11 +742,31 @@ export const resolvers: IResolvers<any, Context> = {
...
-      source,
+      _,
       { id }: { id: string },
       ctx: Context,
     ): Promise<GQLEmptyResponse> => {
-      await ctx.getRepository(Post).update({ id }, { deleted: true });
+      if (ctx.roles.includes(Roles.Moderator)) {
+        await ctx.getRepository(Post).update({ id }, { deleted: true });
+        return { _: true };
+      }
```
rebelchris, line 752: Why do we manually check a role here, should that not be handled by ensureSourcePermissions?

src/schema/sources.ts
```diff
@@ -347,56 +374,117 @@ const sourceToGQL = (source: Source): GQLSource => ({
...
+  SourcePermissions.Edit,
+];
+const ownerPermissions = [
+  ...moderatorPermissions,
+  SourcePermissions.PostLimit,
+  SourcePermissions.ModeratorAdd,
+  SourcePermissions.ModeratorRemove,
+  SourcePermissions.InviteDisable,
+  SourcePermissions.Delete,
+].filter((permission) => permission !== SourcePermissions.Leave);
```
rebelchris, line 408: Why is this filter needed, did we not say owners can simply leave for now? (not sure though)

src/schema/sources.ts
rebelchris: @sshanzel I know we just discussed it, but let's rather remove it for now to avoid confusion later on.
Rather we re-implement it again.

Sorry for the mixed feedback 😅

rebelchris (changes requested):
Some small remarks, but wow this looks great!

src/schema/sources.ts
```diff
@@ -159,6 +165,14 @@ export const typeDefs = /* GraphQL */ `
...
     """
     referralToken: String!
+    """
+    Numerical representation of the user's role
+    """
+    roleRank: Int
+    """
+    User squad permissions
+    """
+    permissions: [String]
```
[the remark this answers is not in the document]
sshanzel: It is necessary, but this is only returned for the `currentMember`. We won't return other user's permissions list.

src/schema/sources.ts
sshanzel: Yes, they either delete the Squad or if in the future, pass ownership.

src/schema/posts.ts
sshanzel: This is for an app-wide Moderator role. Like banning any article, etc.

src/graphorm/index.ts
sshanzel: With role rank, we can only determine the level of access privilege, but whether you have permission to actually do certain actions, it is based on this array.

src/graphorm/index.ts
```diff
@@ -171,17 +174,40 @@ const obj = new GraphORM({
           isMany: false,
           customRelation: (ctx, parentAlias, childAlias, qb): QueryBuilder =>
             qb
-              .select('*')
-              .from(SourceMember, 'sm')
-              .where(`sm."userId" = :userId`, { userId: ctx.userId })
-              .andWhere(`sm."sourceId" = "${parentAlias}".id`),
+              .addSelect('"userId"')
```
idoshamun, line 177: Please don't add a select here, if `userId` is always needed add it as a required column in `SourceMember`

src/schema/sources.ts
sshanzel: I have removed the function in the schema **but** I commented out the test cases, I think the scenarios written would be a good starting point for writing the feature as the intent would almost likely be the same. Let me know what you think @rebelchris @idoshamun

idoshamun (changes requested):
One last comment and we're there

src/graphorm/index.ts
sshanzel: Cool. Let me add it as required 🚢

sshanzel: Worked like a magic! 🧙

src/schema/sources.ts
rebelchris: But why do we filter the permission then, we simply allow deletion now right?

rebelchris: Yes that's supposed to happen as far as I know:
https://dailydotdev.slack.com/archives/C03QZBF46JC/p1676535540610609?thread_ts=1676530981.133169&cid=C03QZBF46JC

rebelchris (approved):
Looks great :)
The edge-case is not that important for now.

idoshamun (approved):

PR #1189, opened 2023-02-23
fix: changelog should be correctly set for users

capJavert:
~~We found a bug where changelog was shown to the anonymous users. Think the best way to handle this is on the API side directly and if `lastChangelog` is `null` frontend will not show anything.~~

In the end issue was with micro seconds timestamp not being parsed correctly from change events.

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -194,7 +194,6 @@ const loggedInBoot = async (
     return handleNonExistentUser(con, req, res, middleware);
   }
   const accessToken = await setAuthCookie(req, res, userId, roles);
-  const lastChangelogMs = +lastChangelog / 1000;
 
   return {
     user: {
@@ -213,7 +212,7 @@ const loggedInBoot = async (
     alerts: {
       ...excludeProperties(alerts, ['userId']),
       // read only, used in frontend to decide if changelog post should be fetched
-      changelog: alerts.lastChangelog < new Date(lastChangelogMs),
+      changelog: alerts.lastChangelog < new Date(lastChangelog),
     },
     settings: excludeProperties(settings, [
       'userId',
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -212,6 +212,7 @@ const loggedInBoot = async (
     flags: adjustFlagsToUser(flags, user),
     alerts: {
       ...excludeProperties(alerts, ['userId']),
+      // read only, used in frontend to decide if changelog post should be fetched
       changelog: alerts.lastChangelog < new Date(lastChangelogMs),
     },
     settings: excludeProperties(settings, [
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -194,6 +194,8 @@ const loggedInBoot = async (
     return handleNonExistentUser(con, req, res, middleware);
   }
   const accessToken = await setAuthCookie(req, res, userId, roles);
+  const lastChangelogMs = +lastChangelog / 1000;
+
   return {
     user: {
       ...excludeProperties(user, [
@@ -210,7 +212,7 @@ const loggedInBoot = async (
     flags: adjustFlagsToUser(flags, user),
     alerts: {
       ...excludeProperties(alerts, ['userId']),
-      changelog: alerts.lastChangelog < new Date(lastChangelog),
+      changelog: alerts.lastChangelog < new Date(lastChangelogMs),
     },
     settings: excludeProperties(settings, [
       'userId',
--- a/src/workers/postChangelogAdded.ts
+++ b/src/workers/postChangelogAdded.ts
@@ -19,7 +19,12 @@ const worker: Worker = {
       return;
     }
     try {
-      await setRedisObject(REDIS_CHANGELOG_KEY, post.createdAt);
+      const timestampMs = post.createdAt / 1000; // createdAt comes as μs here
+
+      await setRedisObject(
+        REDIS_CHANGELOG_KEY,
+        new Date(timestampMs).toISOString(),
+      );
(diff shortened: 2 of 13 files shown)


idoshamun (approved):

capJavert:
@rebelchris looking at the failed tests I am just now realising we have two properties `lastChangelog` and `changelog`. I used `lastChangelog` to check date since last chagnelog post and show popup.

Obviously there is no need because `changelog` already abides by that logic. I just did not know it existed since in the beginning we were talking about only `lastChangelog` field.

idoshamun:
@capJavert I think it will be better to remove `lastChangelog` from the boot endpoint. It creates confusion and not used

rebelchris:
Why is it not used though?
Based on what date do we evaluate now?

I designed it to do all heavy calculations on API side wether to show it or not.

We might have changed implementation but think we need a date that get updated per user right?

idoshamun:
@rebelchris we don't use it on the client side, I mean. The server uses it but there's no reason to return it back to the client

capJavert:
@rebelchris @idoshamun while refactoring frontend to use `changelog` property I realised that because `changelog` is just a computed field on boot we can not use automatic updates via mutations. This is due to updateAlerts mutation from `AlertContext` not returning `changelog` so we need to update boot state value manually inside `BootDataProvider` or trigger boot refetch which is less performant since we are doing another request.

The solution to this (if we are keeping only `changelog` property) would be to somehow return computed field `changelog` from `UpdateAlertsInput`.

src/routes/boot.ts
```diff
@@ -210,7 +212,7 @@ const loggedInBoot = async (
     flags: adjustFlagsToUser(flags, user),
     alerts: {
       ...excludeProperties(alerts, ['userId']),
-      changelog: alerts.lastChangelog < new Date(lastChangelog),
+      changelog: alerts.lastChangelog < new Date(lastChangelogMs),
```
capJavert, line 215: From what I can see locally, value returned from `getRedisObject` is string representation of `μs` timestamp that was stored inside.

`new Date` was returning `InvalidDate` and `changelog` was always `false`.

idoshamun: I don't think Redis has a Date object even. Maybe it's something in the day we store the value.
We can save the date as an iso string to make sure we always read it the right way. Wdyt?

capJavert: It actually stores ISO strings with `setRedisObject(REDIS_CHANGELOG_KEY, post.createdAt)` but not sure where it gets converted to timestamp string, kinda weird.

capJavert: I think I found it, inside the worker `post` is of type `ChangeObject`, which changes `Date` to `number`, so this is where the sting timestamp comes from.

src/workers/postChangelogAdded.ts
```diff
@@ -19,7 +19,12 @@ const worker: Worker = {
       return;
     }
     try {
-      await setRedisObject(REDIS_CHANGELOG_KEY, post.createdAt);
+      const timestampMs = post.createdAt / 1000; // createdAt comes as μs here
```
capJavert, line 22: @idoshamun from what I can see locally `messageToJson` returns `μs` timestamp. Could you verify this from your end?

idoshamun: Yes. I'll check

src/routes/boot.ts
```diff
@@ -210,6 +211,7 @@ const loggedInBoot = async (
     flags: adjustFlagsToUser(flags, user),
     alerts: {
       ...excludeProperties(alerts, ['userId']),
```
idoshamun, line 213: Should we exclude also the `lastChangelog` property here?

src/workers/postChangelogAdded.ts
idoshamun: Ok I know what's the problem. It's Debezium not pubsub. When Debezium publishes a change event, it transforms all dates to microseconds. We can see that `ChangeObject` type replaces all dates with numbers. I didn't know it uses micro, I thought it's millis.
To confirm my suggestion is to try and add a new post to the `daily_updates` source, which should trigger the whole process and see what gets saved in the end.

@rebelchris if this is true (which probably it is), we have to check whether we use dates for other cdc related processing. If so, this issue occur there as well probably.

src/routes/boot.ts
capJavert: @idoshamun we decided that we do need `lastChangelog`, so we are leaving both. This is because updating boot is not really easy currently on the frontend (without invalidation, which is expensive). 

If we would leave only `changelog` we would need to have a computed property on graphql read/update alerts query which is not the best pattern we think - since `changelog` currently only exists in boot. 

So to sumarize we use:
- we use `changelog` to quickly know on webapp startup if we need to fetch the latest changelog article
- we then use `lastChangelog` for updating and hiding it after the update

FYI @rebelchris

idoshamun (approved):

src/workers/postChangelogAdded.ts
capJavert: Yep, I confirmed locally by inserting new post into the database that timestamp is in fact μs, so this should be ok.

image

idoshamun: Let's merge it then!

rebelchris: @idoshamun Just to close the loop.
I checked and we don't rely on it for any other evaluations based on Debezium values.
Only community picks uses dates, but they fetch them from the DB directly.

idoshamun: @rebelchris thanks for letting me know and checking that! 🙏

PR #1191, opened 2023-02-24
feat: removing member from a squad privilege

sshanzel:
Removal of Squad member privilege. The access should be as follow:

Owner > Moderator > Member.

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -63,6 +63,11 @@ export interface GQLSourceMember {
   referralToken: string;
 }
 
+interface RemoveMemberArgs {
+  sourceId: string;
+  memberId: string;
+}
+
 export const typeDefs = /* GraphQL */ `
   """
   Source to discover posts from (usually blogs)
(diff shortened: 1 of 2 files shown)


src/schema/sources.ts
```diff
@@ -841,6 +870,40 @@ export const resolvers: IResolvers<any, Context> = {
...
+        const repo = manager.getRepository(SourceMember);
+        const [member, loggedUser] = await Promise.all([
+          repo.findOneByOrFail({ sourceId, userId: memberId }),
+          repo.findOneByOrFail({ sourceId, userId: ctx.userId }),
+        ]);
+
+        const memberRank = roleRank[member.role];
+        const loggedUserRank = roleRank[loggedUser.role];
+
+        if (memberRank >= loggedUserRank) {
```
rebelchris, line 893: Would this mean technically `members` have some kind of access?
(I know we don't have a lower role now)

rebelchris (approved):
Nice one Lee!
Just one remark for members themselves.

src/schema/sources.ts
sshanzel: The most basic ones, which are the viewing of Squad, posting, etc. - this is just to make it easy to identify which has a higher hierarchy since we now have more than 2 roles.

rebelchris: Understood, just meant to say, technically now users have access to this function.
Not a problem now, but might be if we introduce another role, but maybe that's fine enough for now 🙏

sshanzel: Ah. Yes. I misunderstood at first. Unfortunately, we can't use the built-in directives provided by GraphQL as we use the `role` for users differently. So yeah, this is exposed to our users that's why I have written numerous and strict test cases.

idoshamun (approved):
Looks great!

PR #1192, opened 2023-02-27
refactor: shared post banned/deleted author rep unchanged

sshanzel:
When a shared post was banned/deleted - this should not affect the author of the shared post's reputation.

### DIFF
--- a/src/workers/postBannedRep.ts
+++ b/src/workers/postBannedRep.ts
@@ -5,18 +5,18 @@ import {
 } from './../entity/ReputationEvent';
 import { messageToJson, Worker } from './worker';
 import { PostReport } from '../entity/PostReport';
-import { COMMUNITY_PICKS_SOURCE, Post } from '../entity';
+import { COMMUNITY_PICKS_SOURCE, SharePost } from '../entity';
 import { ChangeObject } from '../types';
 
 interface Data {
-  post: ChangeObject<Post>;
+  post: ChangeObject<SharePost>;
 }
 
 const worker: Worker = {
   subscription: 'post-banned-rep',
   handler: async (message, con, logger): Promise<void> => {
     const data: Data = messageToJson(message);
-    const { id, authorId, scoutId } = data.post;
+    const { id, authorId, scoutId, sharedPostId } = data.post;
     try {
       await con.transaction(async (transaction) => {
         const reports = await transaction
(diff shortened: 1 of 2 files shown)


src/common/post.ts
```diff
@@ -3,7 +3,7 @@ import { Comment, Post } from '../entity';
 import { DataSource } from 'typeorm';
 
 export const defaultImage = {
-  urls: process.env.DEFAULT_IMAGE_URL.split(','),
+  urls: process.env.DEFAULT_IMAGE_URL?.split?.(',') ?? [],
```
sshanzel, line 6: From time to time, my local test are crashing due to being unable to split. Adding this to avoid the issue.

src/workers/postBannedRep.ts
```diff
@@ -32,14 +32,15 @@ const worker: Worker = {
...
-        const ownerProps = {
-          targetId: id,
-          targetType: ReputationType.Post,
-          reason: ReputationReason.PostBanned,
-        };
-
         // Skipping penalizing community picks during the beta phase
-        if (data.post.sourceId !== COMMUNITY_PICKS_SOURCE) {
+        // Skipping penalizing shared post to a Squad
+        if (data.post.sourceId !== COMMUNITY_PICKS_SOURCE && !sharedPostId) {
```
idoshamun, line 37: This is not the right way to check for shared post. Please use the `type` property in the post object

sshanzel: I have updated this now and created an enum.

rebelchris (approved):
Looks good, thanks for enum type

src/workers/postBannedRep.ts
```diff
@@ -32,14 +32,18 @@ const worker: Worker = {
...
-          targetType: ReputationType.Post,
-          reason: ReputationReason.PostBanned,
-        };
-
         // Skipping penalizing community picks during the beta phase
-        if (data.post.sourceId !== COMMUNITY_PICKS_SOURCE) {
+        // Skipping penalizing shared post
+        if (
+          data.post.sourceId !== COMMUNITY_PICKS_SOURCE &&
+          type === PostType.Share
```
idoshamun, line 39:
```suggestion
          type !== PostType.Share
```

idoshamun (changes requested):
See my comment

src/workers/postBannedRep.ts
sshanzel: Oh, right! Completely missed that! Thank you!

idoshamun (approved):

PR #1194, opened 2023-02-27
fix: do not send notification to author/scout from comment following a thread

sshanzel:
A comment notification is duplicated when the author/scout is following a thread where they will receive a "reply notificaion" when there is a new comment.

After the discussion, Sab has decided to prioritize the "reply notification" and forego the author/scout post new comment notification for these cases.

https://dailydotdev.slack.com/archives/C03QZBF46JC/p1676885204557179?thread_ts=1676878287.868739&cid=C03QZBF46JC

### DIFF
--- a/src/workers/notifications/utils.ts
+++ b/src/workers/notifications/utils.ts
@@ -4,7 +4,7 @@ import {
   NotificationCommenterContext,
   NotificationPostContext,
 } from '../../notifications';
-import { DataSource } from 'typeorm';
+import { DataSource, In } from 'typeorm';
 
 export const uniquePostOwners = (
   post: Pick<Post, 'scoutId' | 'authorId'>,
@@ -57,6 +57,7 @@ export async function articleNewCommentHandler(
 
   const isReply = !!comment.parentId;
   if (isReply && (postCtx.post.authorId || postCtx.post.scoutId)) {
+    const ids = [...new Set([postCtx.post.authorId, postCtx.post.scoutId])];
     const threadFollower = await repo
       .createQueryBuilder()
       .select('"userId"')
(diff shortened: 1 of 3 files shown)


src/workers/notifications/utils.ts
```diff
@@ -41,18 +41,50 @@ export async function articleNewCommentHandler(
...
+        userId: postCtx.post.authorId,
+      })
+      .orWhere({
+        id: comment.parentId,
+        userId: postCtx.post.scoutId,
+      })
+      .orWhere({
+        parentId: comment.parentId,
+        userId: postCtx.post.scoutId,
+      })
```
idoshamun, line 78: I suggest a simpler condition. `(id == comment.parentId || parentId == comment.parentId) && (userId == authorId || userId == scoutId)`. Also make sure to group by `userId` to make sure you don't get the same user multiple times

sshanzel: I have now updated the conditions to be much simpler. Would love to hear what you think 🚢

src/workers/notifications/utils.ts
```diff
@@ -41,18 +41,43 @@ export async function articleNewCommentHandler(
...
   }
+  const excludedUsers = [comment.userId];
+
+  const isReply = !!comment.parentId;
+  if (isReply && (postCtx.post.authorId || postCtx.post.scoutId)) {
+    const ids = [...new Set([postCtx.post.authorId, postCtx.post.scoutId])];
+    const threadFollower = await repo
+      .createQueryBuilder()
+      .select('"userId"')
+      .where({
```
idoshamun, line 64: This one is not needed right?

sshanzel: Oh, right. We can let this go. Thank you! 🙌

idoshamun (approved):

PR #1196, opened 2023-03-01
refactor: post entity updated properties

sshanzel:
Introduce 3 new properties for the post entity.
  - visible: boolean
  - visibleAt: date
  - origin: text - have set this as a private property so we won't mistakenly update the value.

### DIFF
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -19,7 +19,7 @@ export enum PostType {
 
 export enum PostOrigin {
   CommunityPicks = 'community_picks',
-  Ugc = 'ugc',
+  UserGenerated = 'user_generated',
   Crawler = 'crawler',
 }
 
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -1,6 +1,4 @@
 import {
-  BeforeInsert,
-  BeforeUpdate,
   Column,
   Entity,
   Index,
@@ -9,7 +7,7 @@ import {
   PrimaryColumn,
   TableInheritance,
 } from 'typeorm';
-import { COMMUNITY_PICKS_SOURCE, Source } from '../Source';
+import { Source } from '../Source';
 import { PostTag } from '../PostTag';
 import { PostKeyword } from '../PostKeyword';
 import { User } from '../User';
@@ -19,9 +17,9 @@ export enum PostType {
   Share = 'share',
 }
 
-enum PostOrigin {
+export enum PostOrigin {
   CommunityPicks = 'community_picks',
-  Squads = 'squads',
+  Ugc = 'ugc',
   Crawler = 'crawler',
 }
 
@@ -152,17 +150,5 @@ export class Post {
   visibleAt: Date;
 
   @Column({ default: null, type: 'text' })
-  private origin: PostOrigin;
-
-  @BeforeInsert()
-  @BeforeUpdate()
-  private setOrigin(): void {
-    if (this.sourceId === COMMUNITY_PICKS_SOURCE) {
-      this.origin = PostOrigin.CommunityPicks;
-    } else if (this.type === PostType.Share) {
-      this.origin = PostOrigin.Squads;
-    } else {
-      this.origin = PostOrigin.Crawler;
-    }
-  }
+  origin: PostOrigin;
 }
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -1,4 +1,6 @@
 import {
+  BeforeInsert,
+  BeforeUpdate,
   Column,
   Entity,
   Index,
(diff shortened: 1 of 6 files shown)


src/entity/posts/Post.ts
```diff
@@ -136,4 +144,25 @@ export class Post {
...
+  @BeforeUpdate()
+  private setOrigin(): void {
+    if (this.sourceId === COMMUNITY_PICKS_SOURCE) {
+      this.origin = PostOrigin.CommunityPicks;
+    } else if (this.type === PostType.Share) {
+      this.origin = PostOrigin.Squads;
+    } else {
+      this.origin = PostOrigin.Crawler;
+    }
+  }
```
rebelchris, line 167: I don't think we should rely on this.
Maybe it's best to have it being either set on creation or updated through the update hook (still coming)

Maybe we can say `crawler` is the default though?
@idoshamun wdyt?

rebelchris (changes requested):

src/entity/posts/Post.ts
sshanzel: Setting it as default is a good point since we have it on `else` already.

idoshamun: When thinking about it, I believe `origin` should be optional. In the case of share post, there's no origin. Or if we want to set it to something we need to call it `ugc` or something like that. We don't want to trigger Yggdrasil for share posts.
sidenote: We've had a lot of issues with `@BeforeInsert` and `@BeforeUpdate` in the past. I suggest setting the field directly in the right place instead of relying on the source

sshanzel: @rebelchris do you have any recommendations on what we should use for the `Share` type?

src/entity/posts/utils.ts
```diff
@@ -245,6 +246,9 @@ const addPostAndKeywordsToDb = async (
     summary: data.summary,
     scoutId: data.scoutId,
     private: privacy,
+    origin: data.scoutId
```
idoshamun, line 249: We should add here also `visible: true` and set `visibleAt` accordingly

src/entity/posts/utils.ts
```diff
@@ -367,6 +371,7 @@ export const createSharePost = async (
       title: commentary,
       sentAnalyticsReport: true,
       private: privacy,
+      origin: PostOrigin.Ugc,
```
idoshamun, line 374: We should add here also `visible: true` and set `visibleAt` accordingly

idoshamun (approved):

sshanzel:
@rebelchris let me know if you need further updates 🚢

src/entity/posts/Post.ts
rebelchris: What does UGC stand for in this context?

sshanzel: I think it is "User-generated content".

rebelchris: Hmm, but private links will also be UGC then.

Since we are the ones publishing the topic to yggdrasil, maybe we should be more explicit.

- PrivateLink (for new private links content) (this needs to be send)
- SharedPost (for the actual shares, these will be excluded from sending, even if they are not visible yet.

@idoshamun What do you think?

idoshamun: Private links shouldn't not be `ugc`, they should be `squads`.
Share post will get `ugc` or even null if we want. I think it might be better.

And again I'm not a fan of @BeforeInsert() and @BeforeUpdate(). I prefer to set it explicitly when needed

rebelchris: Should we not care if it's simply a sharedPost or new original private link in a squad?
We need to know the difference because for private links we will notify users where only the origin will be able to tell the difference.

idoshamun: As we agreed, upon the submission of a private link two posts will be created:
article post with the `squads` origin that will trigger yggdrasil
shared post that refers to the above with `ugc` or null (I'm fine with both) origin that shouldn't trigger yggdrasil.
Both will be invisible until yggdrasil finishes

rebelchris: Yes we have the same idea, I'm not not a fan of using UGC as technically both are user generated.
Why not be more explicit about which part is which?

My only concern is naming.

sshanzel: This is an outdated change, the latest commit is not using the `Before` decorators anymore. Either label works for me. Let me know what you think suits best.

src/entity/posts/utils.ts
```diff
@@ -220,7 +221,7 @@ const addPostAndKeywordsToDb = async (
   }
   const { private: privacy } = await entityManager
     .getRepository(Source)
-    .findOneById(data.publicationId);
+    .findOneBy({ id: data.publicationId });
```
sshanzel, line 224: Just refactored a deprecated API I saw during development.

idoshamun (approved):
Looks great to me

rebelchris (approved):
Nice work Lee 🥳

PR #1197, opened 2023-03-01
feat: unknown source

sshanzel:
Introducing new source for the private links.

Note: merging is blocked until the correct image source link is provided by the design/product team.

### DIFF
--- a/1677643070022-UnknownSource.ts
+++ b/1677643070022-UnknownSource.ts
@@ -4,9 +4,8 @@ export class UnknownSource1677643070022 implements MigrationInterface {
   name = 'UnknownSource1677643070022';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
-    // the image is still subject for change based on product team's response
     await queryRunner.query(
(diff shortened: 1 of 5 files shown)


rebelchris (approved):
Since its private maybe just default it to a placeholder image or null if we allow it.
We will render it differently on FE most likely but also happy to wait for designs answer.

sshanzel:
@rebelchris yes, I was thinking if we have a placeholder image for `source` but I couldn't remember any, so I reached out to product/design team.

sshanzel:
> @sshanzel we also need to add this source to our seed files

It would conflict with our local environment, I think. Once the migration is finished, there will already be an entity for `unknown` so once we try to run the seeds with `unknown`, it will crash. I think this should be good like the way we did for keywords (iirc).

idoshamun:
@sshanzel you're right. My bad

sshanzel:
Would love to hear what you think 🚢

idoshamun (approved):
Awesome!

rebelchris (approved):
Nice work!

PR #1199, opened 2023-03-02
fix: use the `set-cookie` header to set properly kratos cookie

idoshamun:
Instead of patching the expiration of the kratos cookie, we use the returned `set-cookie` header. This will make sure there are no discrepancies.


rebelchris (approved):
Nice find!
Guessing it's only applicable for kratos calls?

Want to ensure we don't have other system calls that might need it.

idoshamun:
@rebelchris yap. it addresses our very original way of working with kratos 😂

PR #1200, opened 2023-03-03
refactor: feed to only return visible posts

sshanzel:
With the introduction of visible property in `Post` entity, we need to ensure our feeds would only return visible posts.

Updated the fixture that our feeds tests are using and no `visible = false` in the snapshot was shown.


__tests__/cron/__snapshots__/updateTagsStr.ts.snap
```diff
@@ -26,5 +26,9 @@ Array [
     "id": "p6",
     "tagsStr": null,
   },
+  Post {
+    "id": "p7",
+    "tagsStr": null,
+  },
```
sshanzel, line 32: This is a direct call from DB in the test not related to any feed.

src/schema/posts.ts
```diff
@@ -216,6 +216,11 @@ export const typeDefs = /* GraphQL */ `
     """
     read: Boolean
 
+    """
+    Whether the post should be visible or not
+    """
+    visible: Boolean
```
idoshamun, line 222: I'm not sure we should propagate this field down to the client. Thoughts? cc @rebelchris

idoshamun:
Another idea of implementation, instead of adding a where clause in many places should we consider a new view of VisiblePost that already queries `deleted = false and visible = true` and we will use it everywhere?

rebelchris (changes requested):
Yeah think a new view entity would work really well here.
As long as we also propagate it to Tinybird.

Let me know if you want a quick catch-up on this @sshanzel

src/graphorm/graphorm.ts
```diff
@@ -293,13 +305,14 @@ export class GraphORM {
     fieldsByTypeName: { [p: string]: ResolveTree },
   ): GraphORMBuilder {
     const fields = Object.values(fieldsByTypeName);
-    const entityMetadata = ctx.con.getMetadata(
-      this.mappings?.[type]?.from || type,
-    );
+    const originalType = this.mappings?.[type]?.from ?? type;
+    const originalMetadata = ctx.con.getMetadata(originalType);
+    const tableName = originalMetadata.tableName;
```
sshanzel, line 310: @idoshamun I encountered a bug where the expected entity was not used from the query (`from` property) since the table name used for the alias was from the `fromMetadata` hence, pulling the data from the reference entity instead of the view. Let me know if you would want this implemented differently.

idoshamun: I probably missed this one. You can change it to `this.getMetadata(ctx, type)`

sshanzel: Cool. Thank you for the confirmation.

src/schema/feeds.ts
```diff
@@ -909,20 +909,17 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       (ctx, { author }: AuthorFeedArgs, builder, alias) =>
         builder
           .addSelect(
-            `count(*) filter (where ${alias}."scoutId" = '${author}')`,
+            `CASE WHEN ${alias}."scoutId" = '${author}' THEN 1 ELSE 0 END`,
             'isScout',
           )
           .addSelect(
-            `count(*) filter (where ${alias}."authorId" = '${author}')`,
+            `CASE WHEN ${alias}."authorId" = '${author}' THEN 1 ELSE 0 END`,
```
sshanzel, line 916: Had to convert it to something simpler, the View seems to include a lot of its properties without telling, so we will have to include those columns in the `group by` but there were too many of them so I have tried moving away from using group by.

src/entity/posts/ActivePost.ts
```diff
@@ -0,0 +1,131 @@
...
+    connection
+      .createQueryBuilder()
+      .select('p.*')
+      .from(Post, 'p')
+      .where('p.deleted = false')
+      .andWhere('p.visible = true')
+      .andWhere(`p."sourceId" != 'unknown'`),
+})
+export class ActivePost {
+  @ViewColumn()
```
idoshamun, line 16: Do we really need to define all these fields? I think we can get away without it

src/graphorm/graphorm.ts
idoshamun: I requested to reverted this part to whatever there's in main now. Sorry I wasn't clear before 😅

src/schema/feeds.ts
idoshamun: This is not good. The group by is here to make sure we don't get duplicates. Let's keep the query as it is and try to fix it in another way

src/schema/submissions.ts
```diff
@@ -160,15 +160,21 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
         };
       }
 
-      const existingPost = await getPostByUrl(cleanUrl, ctx, info);
+      const existingPost = await ctx
+        .getRepository(ArticlePost)
```
idoshamun, line 164: I think we should keep querying Post here

src/common/feedGenerator.ts
```diff
@@ -218,7 +218,6 @@ export const applyFeedWhere = (
   allowPrivateSources = true,
 ): SelectQueryBuilder<Post> => {
   let newBuilder = builder
-    .andWhere(`${alias}.deleted = false`)
     .andWhere(`${alias}."sourceId" != :unknown`, { unknown: UNKNOWN_SOURCE })
```
idoshamun, line 221: This is not needed any more as well right?

src/schema/feeds.ts
sshanzel: Oh, but there is no duplicate post in the post table, right? This is not an aggregated query (only became one due to the count + filter) so there's no need for us to group it. These selects were only to identify whether the user is the author and/or the scout.

src/entity/posts/ActivePost.ts
sshanzel: I agree, just not sure right now which ones are needed and had some trouble with some missing columns earlier so I stashed them all in.

src/schema/submissions.ts
sshanzel: It will not return the existing deleted post anymore since the Post entity in the graphorm will be going to use the new View entity, so it would think it was never in our DB.

src/common/feedGenerator.ts
sshanzel: Oh, right. Will do a whole cleanup for any existing implementations, tomorrow.

src/migration/1678365541578-ActivePostView.ts
```diff
@@ -0,0 +1,28 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ActivePostView1678365541578 implements MigrationInterface {
+  name = 'ActivePostView1678365541578';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE VIEW "active_post" AS SELECT p.* FROM "public"."post" "p" WHERE "p"."deleted" = false AND "p"."visible" = true AND p."sourceId" != 'unknown'`,
```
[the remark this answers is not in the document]
sshanzel: I think that's alright, we will use graphorm most likely to serve the client. Internally in the API, we can just run queries directly on the repositories as we already have direct access to all the data. One example is in the `submission.ts` changes.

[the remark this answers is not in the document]
idoshamun: But in the case I mention we use graphorm. So the post page uses the `post` query to fetch data of the given post and if it's a share post, it also asks for the referred post. In the case we ignore unknown posts, the result will be empty

src/schema/feeds.ts
idoshamun: Ok you're right!

src/schema/submissions.ts
sshanzel: Right. I think we will have to make changes to the `Post` entity to include the `url` and `canonicalUrl` since it is currently available on `ArticlePost` only. Let me update those quickly.

sshanzel: I think that is what we have here, we are querying directly to the DB by using the typeorm. That's why we had to introduce the change.

src/migration/1678365541578-ActivePostView.ts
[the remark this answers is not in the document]
sshanzel: Oh, right. We should then remove the additional checks such as the source and let the `visible` property handle whether the post should be deemed valid to be returned. Wow, nice catch right there.

src/entity/posts/ActivePost.ts
```diff
@@ -0,0 +1,19 @@
...
+@ViewEntity({
+  expression: (connection: DataSource) =>
+    connection
+      .createQueryBuilder()
+      .select('p.*')
+      .from(Post, 'p')
+      .where('p.deleted = false')
+      .andWhere('p.visible = true'),
+})
+export class ActivePost {
```
sshanzel, line 13: Looks like we only need two columns for this class! 🚢

src/common/feedGenerator.ts
sshanzel: Did clean-ups now, and just kept the source condition since as discussed, we will remove same condition from the View condition.

sshanzel:
@rebelchris would love to hear what you think! 🚢

src/schema/submissions.ts
```diff
@@ -160,15 +160,21 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
...
           return {
             result: 'rejected',
             reason: SubmissionFailErrorMessage.POST_DELETED,
           };
         }
-        return { result: 'exists', post: existingPost };
+
+        return {
+          result: 'exists',
+          post: await getPostByUrl(cleanUrl, ctx, info),
```
rebelchris, line 176: This feels a bit duplicate from the one above.
Let's maybe see if it can be uniformed one way?

rebelchris (changes requested):
Tested locally, but for instance source feed builder is not actively using this.
We need to be very careful rolling this out as it has high impact on existing feeds.

See here:
Screenshot 2023-03-10 at 13 52 10

Specific code:
Screenshot 2023-03-10 at 13 51 28

But maybe we should add test-cases with invisible posts and debug backwards?
Just for future-proofing.

src/schema/submissions.ts
sshanzel: This function is using graphorm, the change introduced above was using typeorm.

sshanzel:
> Specific code:

That is using the `feedResolver` which uses the `Post` entity in the graphorm which will use the `ActivePost` as defined. Let me check further on how this happens.

rebelchris (approved):
Nice sorted locally now.
Just want to get Ido's approval on his comment before we merge.

src/cron/exportToTinybird.ts
```diff
@@ -29,9 +29,10 @@ const cron: Cron = {
               "creatorTwitter"    AS "creator_twitter",
               "sourceId"          AS "source_id",
               "tagsStr"           AS "tags_str",
-              ("banned" or "deleted") ::int AS "banned", "type" AS "post_type",
-              "private"::int                AS "post_private"
-       FROM "post"
+              "banned"::int       AS "banned",
+              "type"              AS "post_type",
+              "private"::int      AS "post_private"
+       FROM "active_post"
```
idoshamun, line 35: This needs to use the real table

```suggestion
       FROM "post"
```

src/cron/exportToTinybird.ts
```diff
@@ -29,9 +29,10 @@ const cron: Cron = {
...
        WHERE "metadataChangedAt" > $1
```
idoshamun, line 36: Let's add here a condition that the post is visible

src/entity/posts/utils.ts
```diff
@@ -40,11 +40,10 @@ export const getAuthorPostStats = async (
     .addSelect('sum(post.views)', 'numPostViews')
     .addSelect('sum(post.upvotes)', 'numPostUpvotes')
     .addSelect('sum(post.comments)', 'numPostComments')
-    .from(Post, 'post')
+    .from(ArticlePost, 'post')
```
idoshamun, line 43: This ignores the new view. Did you mean `ActivePost`?

idoshamun (approved):

PR #1203, opened 2023-03-07
feat: introduce manual content published for private links

rebelchris:
Introduces private post mutation
Manually publishing to content requested topic

WT-1126 #done


src/common/pubsub.ts
```diff
@@ -332,3 +334,10 @@ export const notifyUserCreated = async (
 
 export const notifyFeaturesReset = async (log: EventLogger): Promise<void> =>
   publishEvent(log, featuresResetTopic, {});
+
+export type ContentRequested = Pick<ArticlePost, 'id' | 'url' | 'origin'>;
+
+export const notifyContentRequested = async (
+  log: EventLogger,
+  content: ChangeObject<ContentRequested>,
```
idoshamun, line 342: We don't need the `ChangeObject` in this case, right?

src/entity/posts/utils.ts
```diff
@@ -344,20 +348,75 @@ export const addNewPost = async (
...
+  con: DataSource,
+  logger: EventLogger,
+  sourceId: string,
+  userId: string,
+  url: string,
+  commentary: string,
+): Promise<void> => {
+  await validateCommentary(commentary);
+  const id = await generateShortId();
+  const sharedId = await generateShortId();
```
idoshamun, line 372: shareId and id should be the same

src/entity/posts/utils.ts
```diff
@@ -344,20 +348,75 @@ export const addNewPost = async (
...
+
+  return con.transaction(async (entityManager) => {
+    try {
+      await entityManager.getRepository(ArticlePost).insert({
+        id,
+        shortId: id,
+        createdAt: new Date(),
+        sourceId,
+        url,
```
idoshamun, line 381: I think we should set `canonicalUrl` as well here. Not highly important but just to make sure it's not empty

src/common/pubsub.ts
rebelchris: Ah not particular indeed, we can make it a set ContentRequested type will address this.

src/entity/posts/utils.ts
```diff
@@ -344,20 +348,75 @@ export const addNewPost = async (
...
+        visible: false,
+      });
+      await notifyContentRequested(logger, {
+        id,
+        url,
+        origin: PostOrigin.Squad,
+      });
+      return;
+    } catch (err) {
+      throw err;
```
idoshamun, line 407: Looks like there's no need for the try and catch as we simply throw the error right away

src/entity/posts/utils.ts
```diff
@@ -344,20 +348,75 @@ export const addNewPost = async (
...
+      await entityManager.getRepository(SharePost).insert({
+        id: sharedId,
+        shortId: sharedId,
+        createdAt: new Date(),
+        sourceId,
+        authorId: userId,
+        sharedPostId: id,
+        title: commentary,
+        sentAnalyticsReport: true,
+        private: true,
```
idoshamun, line 396: privacy here should be according to the source

idoshamun: I suggest unifying this creation of share post with the one below in the same function.
I'm afraid we might forget one case or the other and introduce bugs

__tests__/workers/cdc.ts
```diff
@@ -130,6 +134,21 @@ const defaultUser: ChangeObject<User> = {
   notificationEmail: true,
 };
 
+describe('content requested', () => {
```
idoshamun, line 137: This will always be true because of the way this test is designed. I would rethink it. Not sure it's needed

src/schema/posts.ts
```diff
@@ -508,6 +509,24 @@ export const typeDefs = /* GraphQL */ `
       id: ID!
     ): EmptyResponse @auth
 
+    """
+    Create private post in source
+    """
+    privatePost(
```
idoshamun, line 515: I find this name a bit confusing. Can you please come up with a new name?
It should also be phrased as an action because it's a mutation (i.e. submitExternalLink)

idoshamun (changes requested):
Please see my comments and recommendations :)

src/schema/posts.ts
rebelchris: Totally, should we then also consider renaming `sharePost`? or is that explicit enough for now.

idoshamun: Renaming it will break production but I find it explicit enough `share` is the action and `post` is the entity you share. I get the confusion though because we have a share post type as well

__tests__/workers/cdc.ts
idoshamun: Perfect!

src/schema/posts.ts
```diff
@@ -833,6 +852,25 @@ export const resolvers: IResolvers<any, Context> = {
...
+    submitExternalLink: async (
+      _,
+      {
+        sourceId,
+        url,
+        commentary,
+      }: { sourceId: string; url: string; commentary: string },
+      ctx,
+    ): Promise<GQLEmptyResponse> => {
+      await createExternalLink(
```
idoshamun, line 864: Important thing that we missed, we need to validate permissions!

rebelchris: Ah good spot!
Added the `Post` permissions now

idoshamun: Can you please add relevant tests?

rebelchris: @idoshamun Added relevant cases for permissions

rebelchris:
@idoshamun Added the check for existing articles and URL validation.

src/schema/posts.ts
```diff
@@ -833,6 +854,48 @@ export const resolvers: IResolvers<any, Context> = {
...
+        return { _: true };
+      }
+      await createExternalLink(
+        ctx.con,
+        ctx.log,
+        sourceId,
+        ctx.userId,
+        url,
+        commentary,
+      );
```
idoshamun, line 896: This needs to be in a transaction

idoshamun (approved):
Only comment is to put everything in a transaction and we're good to merge

src/schema/posts.ts
rebelchris: @idoshamun Converted it to be wrapped and have the main functions accept the manager.

src/schema/posts.ts
```diff
@@ -833,6 +854,50 @@ export const resolvers: IResolvers<any, Context> = {
...
+        const cleanUrl = standardizeURL(url);
+        if (!isValidHttpUrl(cleanUrl)) {
+          throw new ValidationError('URL is not valid');
+        }
+
+        const existingPost = await manager.getRepository(ArticlePost).findOne({
+          select: ['id'],
+          where: [
+            { url: cleanUrl, deleted: false },
+            { canonicalUrl: cleanUrl, deleted: false },
```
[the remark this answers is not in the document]
rebelchris: It will simply throw an error which we'll have to catch FE wise.
If a post is deleted for whatever reason we shouldn't allow it back (even as private) 

I'll double check with product as well.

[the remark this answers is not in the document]
sshanzel: Not sure where the error is getting thrown here. Should we catch it here in BE too? cc @idoshamun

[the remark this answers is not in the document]
rebelchris: Added the actual throw and test case for deleted posts.

sshanzel (approved):
Thank you for including the relevant test cases! 🚢 🚢 🚢

idoshamun (approved):
Looking good!

PR #1205, opened 2023-03-08
fix: post optionality and removal of tweet jobs

rebelchris:
Change post optionality:
- title is now optional
- sourceId is now required -> default `unknown` (added into the migration)
- tweeted is not gone

Also cleaned up all the tweeted related crons and workers, including package used for tweeting.

WT-1126 #done


src/migration/1678256072490-PostOptionality.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+    await queryRunner.query(
+      `INSERT INTO "public"."source" ("id", "name", "handle", "private") VALUES ('unknown', 'unknown', 'unknown', 'true') ON CONFLICT DO NOTHING`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "post" ALTER COLUMN "sourceId" SET DEFAULT 'unknown'`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "post" ALTER COLUMN "sourceId" SET NOT NULL`,
+    );
```
sshanzel, line 19: I think we already have a migration for this: https://github.com/dailydotdev/daily-api/blob/main/1677643070022-UnknownSource.ts

rebelchris: Ah I thought so, but it won't run indeed from main so I didn't see it reflect in my DB

sshanzel: Right. Feel free to delete the one in the root folder then I will close the PR itself.

sshanzel (changes requested):
Thanks for removing all the twitter-related unused codes. Kindly delete the migration file then this should be good to go 🚢

src/migration/1678256072490-PostOptionality.ts
rebelchris: Done 🙏

idoshamun (changes requested):
Please make sure to delete the workers from the worker array, the same applies to the crons. Regarding the cron, I didn't see you removed it from pulumi

rebelchris:
@idoshamun I think all of them where already removed from the worker array and pulumi when we experienced the Twitter issues.
Unless I missed something?

idoshamun:
@rebelchris you're right! 🤦‍♂️

idoshamun (approved):

PR #1207, opened 2023-03-08
fix: misplaced migration file

sshanzel:
The migration file was in the root folder. I think we will have to update the package json command to automatically save it in the `migrations` folder.


rebelchris:
Ah yes I think we both run them as: `npm run db:migrate:make src/migration/PostOptionality`

sshanzel:
Will be fixed automatically on Chris' PR.

PR #1210, opened 2023-03-09
fix: don't add article picked notification for private posts

idoshamun:
With external links we need to be careful not to send notifications for authors that can't access the post. So if the post belongs to a machine source we will not send a notification for private posts


rebelchris (approved):
Nice,

This made me realise do we also want to exclude these from the author profile calculation and list?

Maybe even the "here's how your article is doing email"

PR #1211, opened 2023-03-09
feat: introducing post updated worker

rebelchris:
@idoshamun We just need to verify my JSON expected input matches what Denis is sending.

Introduces a new post-updated subscription, can update any part of the post
Can determine if post becomes visible
Updates shared post to become visible.

WT-1127 #done

Review suggestion:
Ensure test cases are solid.

### DIFF
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -171,4 +171,8 @@ export const workers: Worker[] = [
     topic: 'features-reset',
     subscription: 'api.clear-features-cache',
   },
+  {
+    topic: 'yggdrasil.v1.content-published',
+    subscription: 'api.post-updated',
+  },
 ];
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -173,6 +173,6 @@ export const workers: Worker[] = [
   },
   {
     topic: 'yggdrasil.v1.content-published',
-    subscription: 'api.post-updated',
+    subscription: 'api.content-published',
   },
 ];
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -155,7 +155,9 @@ it('should save a new post with the relevant keywords', async () => {
     id: 'p1',
     updated_at: new Date('01-05-2023 12:00:00'),
     title: 'test',
-    keywords: ['alpine', 'a-b-testing', 'mongodb'],
+    extra: {
+      keywords: ['alpine', 'a-b-testing', 'mongodb'],
+    },
   });
   const post = await con.getRepository(ArticlePost).findOneBy({ id: 'p1' });
   expect(post.metadataChangedAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -162,6 +162,7 @@ it('should save a new post with the relevant keywords', async () => {
   expect(post.visible).toEqual(true);
   expect(post.visibleAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
   expect(post.title).toEqual('test');
+  expect(post.tagsStr).toEqual('mongodb,alpinejs,ab-testing');
   const keywords = await con.getRepository(Keyword).find({
     where: {
       value: 'alpine',
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -1,0 +1,94 @@
+import { expectSuccessfulBackground, saveFixtures } from '../helpers';
+import worker from '../../src/workers/postUpdated';
+import {
+  ArticlePost,
+  Post,
+  PostType,
+  SharePost,
+  Source,
+} from '../../src/entity';
+import { sourcesFixture } from '../fixture/source';
+import { DataSource } from 'typeorm';
+import createOrGetConnection from '../../src/db';
+
+let con: DataSource;
+
+beforeAll(async () => {
+  con = await createOrGetConnection();
+});
+
+beforeEach(async () => {
+  jest.resetAllMocks();
+  await saveFixtures(con, Source, sourcesFixture);
+  await saveFixtures(con, ArticlePost, [
+    {
+      id: 'p1',
+      shortId: 'p1',
+      url: 'http://p1.com',
+      score: 0,
+      metadataChangedAt: new Date('01-05-2020 12:00:00'),
+      sourceId: 'a',
+      visible: false,
+      createdAt: new Date('01-05-2020 12:00:00'),
+    },
+  ]);
+});
+
+const createSharedPost = async (id = 'sp1') => {
+  const post = await con.getRepository(Post).findOneBy({ id: 'p1' });
+  await con.getRepository(SharePost).save({
+    ...post,
+    id,
+    shortId: `short-${id}`,
+    sharedPostId: 'p1',
+    type: PostType.Share,
+    visible: false,
+  });
+};
+
+it('should not update if the database updated date is newer', async () => {
+  await expectSuccessfulBackground(worker, {
+    updated_at: new Date('01-05-1990 12:00:00'),
+  });
+  const post = await con.getRepository(ArticlePost).findOneBy({ id: 'p1' });
+  expect(post.metadataChangedAt).toEqual(new Date('2020-01-05T12:00:00.000Z'));
+});
+
+it('should update the post and keep it invisible if title is missing', async () => {
+  await expectSuccessfulBackground(worker, {
+    updated_at: new Date('01-05-2023 12:00:00'),
+  });
+  const post = await con.getRepository(ArticlePost).findOneBy({ id: 'p1' });
+  expect(post.metadataChangedAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
+  expect(post.visible).toEqual(false);
+});
+
+it('should update the post and make it visible if title is available', async () => {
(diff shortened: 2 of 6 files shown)


.infra/workers.ts
```diff
@@ -171,4 +171,8 @@ export const workers: Worker[] = [
     topic: 'features-reset',
     subscription: 'api.clear-features-cache',
   },
+  {
+    topic: 'yggdrasil.v1.content-published',
+    subscription: 'api.post-updated',
```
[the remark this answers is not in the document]
rebelchris: We can go with `api.content-published`?

src/workers/postUpdated.ts
```diff
@@ -0,0 +1,68 @@
+import { messageToJson, Worker } from './worker';
+import { ArticlePost, SharePost } from '../entity';
+
+interface Data extends ArticlePost {
+  updated_at: Date;
+}
+
+const worker: Worker = {
+  subscription: 'api.post-updated',
+  handler: async (message, con, logger): Promise<void> => {
```
idoshamun, line 10: This handler misses a use case for crawler and community picks where we don't prematurely generate an id that Yggdrasil can utilize. Am I right?

src/workers/postUpdated.ts
```diff
@@ -0,0 +1,68 @@
...
+            description: data?.description,
+            readTime: data?.readTime,
+            toc: data?.toc,
+            keywords: data?.keywords,
+            siteTwitter: data?.siteTwitter,
+            creatorTwitter: data?.creatorTwitter,
+            sourceId: data?.sourceId,
+            image: data?.image,
+            title: data?.title,
+            metadataChangedAt: updatedDate,
```
idoshamun, line 43: One thing that I noticed is that we change this value ourselves in this CDC in some cases. 
I think we should remove some of this cases from CDC (for example when `visible` changes) because it addresses here and it may create race conditions.

src/workers/postUpdated.ts
```diff
@@ -0,0 +1,68 @@
...
+        const becomesVisible = !databasePost?.visible && !!data?.title?.length;
+
+        await entityManager.getRepository(ArticlePost).update(
+          { id: databasePost.id },
+          {
+            summary: data?.summary,
+            description: data?.description,
+            readTime: data?.readTime,
+            toc: data?.toc,
+            keywords: data?.keywords,
```
idoshamun, line 37: There's a whole logic for the keywords update.
I suggest having a look at our existing `newPost` endpoint to make sure we don't miss anything.
Essentially, this worker will take charge after we finish the migration, so we need to ensure everything is here.

idoshamun (changes requested):
That's a good start but still misses some use cases I think

src/workers/postUpdated.ts
rebelchris: Ah good point, not sure it's high impact for now, but we can add it.
So far we have that retro use case only for later down the line.

Let me know what you think.

.infra/workers.ts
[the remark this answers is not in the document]
idoshamun: Yes that works :)

src/workers/postUpdated.ts
rebelchris: We can remove the other ones if that's what you mean.
Since those are not actually using it.

idoshamun: Tinybird exporter uses it. I'll review it now again and let you know what props can be ignored in the cdc

idoshamun: I believe we can keep only this lines:
isChanged(data.payload.before, data.payload.after, 'deleted') ||
      isChanged(data.payload.before, data.payload.after, 'banned') ||
      isChanged(data.payload.before, data.payload.after, 'tagsStr')

src/workers/postUpdated.ts
```diff
@@ -0,0 +1,68 @@
...
+            siteTwitter: data?.siteTwitter,
+            creatorTwitter: data?.creatorTwitter,
+            sourceId: data?.sourceId,
```
idoshamun, line 40: We should probably update the `private` field based on the source

src/workers/postUpdated.ts
rebelchris: @idoshamun We need to think this trough, for community picks we need to change it regardless so the post will be created same as with squads now.
For crawler I agree it's an exception, but what rules will we set since it's not "updating" then but creating a completely new one.

idoshamun: Please confirm with Viktor that we can really tackle it later. If so, that's what will do. Ask him on #engineering please

rebelchris:
@idoshamun Refactored so it uses the existing newPost flow.
Tried to re-use as much as possible, although there is still some overlap (As newPost consists of creation)

Love to hear what you think.

src/entity/posts/utils.ts
```diff
@@ -257,11 +257,21 @@ const addPostAndKeywordsToDb = async (
...
+};
+
+export const addTagsAndKeywords = async (
+  entityManager: EntityManager,
+  tags: string[],
+  mergedKeywords: string[],
+  postId: string,
+): Promise<void> => {
+  if (tags?.length) {
     await entityManager.getRepository(PostTag).insert(
```
idoshamun, line 271: By now this is no longer used and is actually deprecated. We use only the keywords

src/workers/postUpdated.ts
```diff
@@ -0,0 +1,120 @@
...
+          canonicalUrl: data?.canonicalUrl || data?.url,
+          title: data?.title && he.decode(data?.title),
+          readTime: parseReadTime(data?.readTime),
+          publishedAt: data?.publishedAt && new Date(data?.publishedAt),
+          metadataChangedAt: updatedDate,
+          visible: becomesVisible,
+          visibleAt: becomesVisible ? updatedDate : null,
+          tagsStr: allowedKeywords?.join(',') || null,
+          private: privacy,
+          sentAnalyticsReport: !data?.authorId,
```
idoshamun, line 92: If the post is private we shouldn't send analytics report

```suggestion
          sentAnalyticsReport: private || !data?.authorId,
```

__tests__/workers/postUpdated.ts
```diff
@@ -0,0 +1,107 @@
...
+  const post = await con.getRepository(ArticlePost).findOneBy({ id: 'p1' });
+  expect(post.metadataChangedAt).toEqual(new Date('2020-01-05T12:00:00.000Z'));
+});
+
+it('should not update if the post is not a squad origin', async () => {
+  await con
+    .getRepository(ArticlePost)
+    .update({ id: 'p1' }, { origin: PostOrigin.CommunityPicks });
+  await expectSuccessfulBackground(worker, {
+    updated_at: new Date('01-05-1990 12:00:00'),
```
idoshamun, line 64: An id is missing here that indicate what post to look at. And updated_at is too old compared to the metadata changed at

__tests__/workers/postUpdated.ts
```diff
@@ -0,0 +1,107 @@
...
+    shortId: `short-${id}`,
+    sharedPostId: 'p1',
+    type: PostType.Share,
+    visible: false,
+  });
+};
+
+it('should not update if the database updated date is newer', async () => {
+  await expectSuccessfulBackground(worker, {
+    updated_at: new Date('01-05-1990 12:00:00'),
```
idoshamun, line 53: No id provided here

src/workers/postUpdated.ts
```diff
@@ -0,0 +1,120 @@
...
+  keywords: string[];
+  tags: string[];
+}
+
+const worker: Worker = {
+  subscription: 'api.content-published',
+  handler: async (message, con, logger): Promise<void> => {
+    const data: Data = messageToJson(message);
+    try {
+      const { id, tags, keywords, updated_at, ...submittedData } = data;
```
idoshamun, line 26: Please add a check that id is provided before you proceed

idoshamun (changes requested):
I noticed an id is missing in the payload of all your tests. Please also add applicable tests from the new post file. For example that we set proper tags and so on.
Sorry for the tight review, it's just an important part of the app

src/entity/posts/utils.ts
rebelchris: Will remove, also will impact the existing new post then.

rebelchris:
@idoshamun I added the ID check and test case for empty ID.
Also removed the tags and added the relevant test.

It seems no other private newPost test would be applicable as we already check everything (except community picks which is not yet applicable)

__tests__/workers/postUpdated.ts
```diff
@@ -0,0 +1,174 @@
...
+    updated_at: new Date('01-05-2023 12:00:00'),
+    title: 'test',
+    keywords: ['alpine', 'a-b-testing', 'mongodb'],
+  });
+  const post = await con.getRepository(ArticlePost).findOneBy({ id: 'p1' });
+  expect(post.metadataChangedAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
+  expect(post.visible).toEqual(true);
+  expect(post.visibleAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
+  expect(post.title).toEqual('test');
+  const keywords = await con.getRepository(Keyword).find({
```
idoshamun, line 165: We are not checking completely the keywords here, let's add a more thorough check on `tagsStr`

idoshamun (changes requested):
One last comment

idoshamun (approved):

idoshamun (approved):

PR #1212, opened 2023-03-09
feat: squad post live notification

idoshamun:
New notification for external links support to notify the user when the post is live

### DIFF
--- a/src/workers/newNotificationMail.ts
+++ b/src/workers/newNotificationMail.ts
@@ -54,7 +54,7 @@ const notificationToTemplateId: Record<NotificationType, string> = {
   squad_reply: 'd-cbb2de40b61840c38d3aa21028af0c68',
   squad_post_viewed: 'd-dc0eb578886c4f84a7dcc25515c7b6a4',
   squad_access: 'd-6b3de457947b415d93d0029361edaf1d',
-  squad_post_live: 'update!',
+  squad_post_live: 'd-343845599453499d9fa5d3ffafc91514',
 };
 
 type TemplateDataFunc = (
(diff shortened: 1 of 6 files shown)


src/workers/newNotificationMail.ts
```diff
@@ -54,6 +54,7 @@ const notificationToTemplateId: Record<NotificationType, string> = {
   squad_reply: 'd-cbb2de40b61840c38d3aa21028af0c68',
   squad_post_viewed: 'd-dc0eb578886c4f84a7dcc25515c7b6a4',
   squad_access: 'd-6b3de457947b415d93d0029361edaf1d',
+  squad_post_live: 'update!',
```
rebelchris, line 57: Blocking

src/workers/notifications/postAdded.ts
```diff
@@ -61,6 +68,19 @@ const worker: NotificationWorker = {
             } as NotificationPostContext & Partial<NotificationDoneByContext>,
           }),
         );
+
+        if (post.type === PostType.Share) {
+          // squad_post_live notification
+          if (baseCtx.sharedPost?.origin === PostOrigin.Squad) {
```
rebelchris, line 74: The article post is the one that has `squad` origin though.
Shared post will always have `user_generated`

Think we need to listen to the original post, but send the linked squad post?

rebelchris: Hmm think you do have it, might be confusing to call it `sharedPost`?

rebelchris (changes requested):
Think it's alright, we just need to be aware of naming.
It can get very confusing with which one triggers the update from a reading point of view.

src/workers/notifications/postAdded.ts
idoshamun: `sharedPost` is the post that is referenced hence the article post. It's something baked in our notification context. We have `post` and `sharedPost`. Agree that it might be confusing, happy to change it to something else

rebelchris: Ah i see its a "notification shared post" not actual "SharePost" 😅
Hmm not sure what the alternative is though, but if we have some better ones I'd love to see that.

src/workers/newNotificationMail.ts
idoshamun: Fixed!

rebelchris (approved):

PR #1215, opened 2023-03-20
fix: make post creation optional for new squad

rebelchris:
We have to support backwards compatibility and step one is to make post id optional for create squad endpoint.

WT-1152


sshanzel (approved):
Way to go! 🚢

PR #1216, opened 2023-03-21
fix: make squads active by default

rebelchris:
Should only be merged once v3 is in production.
It will set squad active by default (non breaking for the frontend)

WT-1163


idoshamun (approved):

sshanzel (approved):

sshanzel:
I think we can merge this now?

PR #1217, opened 2023-03-25
fix: enforce permissions in postByUrl query

idoshamun:
* Check permissions for posts returned by `postByUrl`
* Minor perf improvement to post query for public posts


rebelchris (approved):
Good to go

PR #1218, opened 2023-03-28
refactor: mention criteria to include hyphens

sshanzel:
Allow usage of hyphens when mentioning a user.

### DIFF
--- a/__tests__/comments.ts
+++ b/__tests__/comments.ts
@@ -330,10 +330,22 @@ describe('query commentPreview', () => {
     expect(res.errors).toBeFalsy();
     expect(res.data.commentPreview).toMatchSnapshot();
 
+    const mention = '@Lee';
+    const withMention = await client.query(QUERY, {
+      variables: { content: mention },
+    });
+    expect(withMention.errors).toBeFalsy();
+    expect(withMention.data.commentPreview).toMatchSnapshot();
+  });
+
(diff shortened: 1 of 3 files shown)


__tests__/comments.ts
```diff
@@ -330,7 +330,10 @@ describe('query commentPreview', () => {
     expect(res.errors).toBeFalsy();
     expect(res.data.commentPreview).toMatchSnapshot();
 
-    const mention = '@Lee';
+    await con
+      .getRepository(User)
+      .update({ username: 'Lee' }, { username: 'Lee-' });
+    const mention = '@Lee-';
```
rebelchris, line 336: Maybe suggest to append something at the back for safety tests

rebelchris (approved):
Just a small note to improve test cases

__tests__/comments.ts
sshanzel: Good point. Let me add it as a separate one.

rebelchris (approved):
Thanks! 🎉

idoshamun (approved):
Awesome!

capJavert:
@sshanzel this will only fix new comments?

Old comments will still be broke or mention wrong person (in case of reported issue from the task)?

sshanzel:
> Old comments will still be broken or mention wrong person (in case of reported issue from the task)?

Yes, this was not intended to fix existing mentions. I will check with the produce team if they want the old ones to be fixed.

PR #1219, opened 2023-03-29
feat(squads): source members fields

capJavert:
We agreed on the web team sync to add member fields to the source entity to be able to show roles on the post feed and member listings for the squad members. 

We also agreed that it might not be best decision scaling wise because if squads have hundreds of moderators and other members this fields would become quite big. `50` limit was added to prevent any excess cases while in beta - this should catch most of the squads though since they don't have many privileged members.

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -37,16 +37,6 @@ const nullIfNotLoggedIn = <T>(value: T, ctx: Context): T | null =>
 const nullIfNotSameUser = <T>(value: T, ctx: Context, parent: User): T | null =>
   ctx.userId === parent.id ? value : null;
 
-const getMembersByRole =
-  (role: SourceMemberRoles) =>
-  (ctx: Context, alias: string, qb: QueryBuilder) =>
-    qb
-      .select('ARRAY_AGG(sm."userId")')
-      .from(SourceMember, 'sm')
-      .where(`sm."sourceId" = "${alias}".id`)
-      .andWhere(`sm.role = '${role}'`)
-      .limit(50); // limit to avoid huge arrays for members, most sources should fit into this see PR !1219 for more info
-
 const obj = new GraphORM({
   User: {
     requiredColumns: ['id', 'username'],
@@ -201,12 +191,18 @@ const obj = new GraphORM({
               .andWhere(`${childAlias}."sourceId" = "${parentAlias}".id`),
         },
       },
-      owners: {
-        select: getMembersByRole(SourceMemberRoles.Owner),
-        transform: nullIfNotLoggedIn,
-      },
-      moderators: {
-        select: getMembersByRole(SourceMemberRoles.Moderator),
+      privilegedMembers: {
+        relation: {
+          isMany: true,
+          customRelation: (ctx, parentAlias, childAlias, qb): QueryBuilder => {
+            return qb
+              .where(`${childAlias}."sourceId" = "${parentAlias}".id`)
+              .andWhere(
+                `(${childAlias}.role = '${SourceMemberRoles.Owner}' OR ${childAlias}.role = '${SourceMemberRoles.Moderator}')`,
+              )
+              .limit(50); // limit to avoid huge arrays for members, most sources should fit into this see PR !1219 for more info
+          },
+        },
         transform: nullIfNotLoggedIn,
       },
     },
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -187,6 +187,26 @@ const obj = new GraphORM({
               .andWhere(`${childAlias}."sourceId" = "${parentAlias}".id`),
         },
       },
+      owners: {
+        select: (ctx, alias, qb) => {
+          return qb
+            .select('ARRAY_AGG(sm."userId")')
+            .from(SourceMember, 'sm')
+            .where(`sm."sourceId" = "${alias}".id`)
+            .andWhere(`sm.role = 'owner'`);
+        },
+        transform: nullIfNotLoggedIn,
+      },
+      moderators: {
+        select: (ctx, alias, qb) => {
+          return qb
+            .select('ARRAY_AGG(sm."userId")')
+            .from(SourceMember, 'sm')
+            .where(`sm."sourceId" = "${alias}".id`)
+            .andWhere(`sm.role = 'moderator'`);
+        },
+        transform: nullIfNotLoggedIn,
+      },
     },
   },
   SourceMember: {
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -9,7 +9,11 @@ import {
   SourceMember,
   User,
 } from '../entity';
-import { sourceRoleRank, sourceRoleRankKeys } from '../roles';
+import {
+  SourceMemberRoles,
+  sourceRoleRank,
+  sourceRoleRankKeys,
+} from '../roles';
 
 import { Context } from '../Context';
 import { GQLBookmarkList } from '../schema/bookmarks';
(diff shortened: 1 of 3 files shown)


src/graphorm/index.ts
```diff
@@ -187,6 +187,28 @@ const obj = new GraphORM({
...
+        transform: nullIfNotLoggedIn,
+      },
+      moderators: {
+        select: (ctx, alias, qb) => {
+          return qb
+            .select('ARRAY_AGG(sm."userId")')
+            .from(SourceMember, 'sm')
+            .where(`sm."sourceId" = "${alias}".id`)
+            .andWhere(`sm.role = 'moderator'`)
+            .limit(50); // limit to avoid huge arrays, most squads should fit into this
```
sshanzel, line 208: I think we can omit the comments.
```suggestion
            .limit(50);
```

src/graphorm/index.ts
```diff
@@ -187,6 +187,28 @@ const obj = new GraphORM({
...
         },
       },
+      owners: {
+        select: (ctx, alias, qb) => {
+          return qb
+            .select('ARRAY_AGG(sm."userId")')
+            .from(SourceMember, 'sm')
+            .where(`sm."sourceId" = "${alias}".id`)
+            .andWhere(`sm.role = 'owner'`)
+            .limit(50); // limit to avoid huge arrays, most squads should fit into this
```
sshanzel, line 197:
```suggestion
            .limit(50);
```

sshanzel (approved):
Woop! Woop! Way to go! 🚢 🚢 🚢

src/graphorm/index.ts
```diff
@@ -187,6 +187,28 @@ const obj = new GraphORM({
               .andWhere(`${childAlias}."sourceId" = "${parentAlias}".id`),
         },
       },
+      owners: {
+        select: (ctx, alias, qb) => {
+          return qb
+            .select('ARRAY_AGG(sm."userId")')
+            .from(SourceMember, 'sm')
+            .where(`sm."sourceId" = "${alias}".id`)
+            .andWhere(`sm.role = 'owner'`)
```
capJavert, line 196: hardcoded `'owner'` since importing `SquadMemberRoles` enum does not resolve correctly. @sshanzel mentioned it is ongoing bundling issue with this project. (same for `'moderator'` below).

src/graphorm/index.ts
```diff
@@ -187,6 +187,28 @@ const obj = new GraphORM({
               .andWhere(`${childAlias}."sourceId" = "${parentAlias}".id`),
         },
       },
+      owners: {
+        select: (ctx, alias, qb) => {
+          return qb
```
rebelchris, line 192:
```suggestion
        select: (ctx, alias, qb) => 
         qb
```

src/graphorm/index.ts
```diff
@@ -187,6 +187,28 @@ const obj = new GraphORM({
               .andWhere(`${childAlias}."sourceId" = "${parentAlias}".id`),
         },
       },
+      owners: {
+        select: (ctx, alias, qb) => {
```
rebelchris, line 191: We can actually combine these 2 in. a function like we do for `existsByUserAndPost` so it's easier to manipulate.

Suggestion: getMembersByRole? so we can pass a role.

src/graphorm/index.ts
rebelchris: Not sure did this not work:
```
.andWhere(`sm.role = :role`, { role: SourceMemberRoles.Owner })
```

rebelchris (changes requested):
Amazing work guys.

Just a note for @capJavert API PR's always need approval from @idoshamun at this point, so they exclude the 2 review check.

Got some minor improvements I think we can make.

src/graphorm/index.ts
sshanzel: Yes, using the enum was not working. Not sure why it is saying the enum is undefined.

src/graphorm/index.ts
capJavert: Hmm, since is not visible by first why this is I though comments are useful, I guess the changes will be linked to the PR where we have some info about it. @rebelchris WDYT?

rebelchris: No strong feelings about it, comments like this can be useful as they are very specific and may cause confusion down the line.

src/graphorm/index.ts
capJavert: @rebelchris @sshanzel looks like `SourceMemberRoles` does work but for some reason when done through `variables` argument it does not work. Refactored to enums here 244f1f61a2d398dd221e1c0c1681af4b3ef10c22, feel free to try

src/graphorm/index.ts
capJavert: I would like to leave it if you don't mind. Just because it is specific beta feature. There is only one comment now since logic is extracted to `getMembersByRole` as Chris noted.

rebelchris (approved):
Omg, I love it!
So clean

@idoshamun Just need your approval on this 💜

idoshamun:
@capJavert implementation looks slick, but I have a suggestion for improvement and also an alternative for your consideration. I assume this PR is for supporting showing the "owner" and "moderator" badges in the comments. If the assumption is wrong please let me know:
* Suggestion - Instead of separating owners and mods, let's add another field that will be an array of `SourceMember` (contains both user and role already) with all the special users that require badges. 
* Alternative - My only concern with the current implementation is that it's not really flexible and probably not the ideal solution with that said I'm fine with it as an initial solution but anyway gonna pitch my alternative. Instead of manipulating the `Source` schema, we can add a badges array to the `Comment` object. This will be an enum array of badges that need to be shown alongside the author. (e.g. AUTHOR, OWNER, MODERATOR). This removes a lot of the business logic that you'll need to put in the frontend

rebelchris:
@idoshamun My two cents.

Like the badges idea, but afraid we need to add it to comment + posts and if in the future design wants to add it to user cards it's yet another entity.

As for the shared role, it doesn't look like a urgent issue, but let's do it as it makes little effort to change.

idoshamun:
@rebelchris fine by me :)

src/schema/sources.ts
```diff
@@ -126,6 +128,11 @@ export const typeDefs = /* GraphQL */ `
     Logged-in member object
     """
     currentMember: SourceMember
+
+    """
+    Privileged members
+    """
+    privilegedMembers: [SourceMember]
```
capJavert, line 135: Noticed that by exposing this someone could craft a GQL query to our API (same is for `members` field that existed before) to get `referralToken` of other users. Is this security concern? @idoshamun

capJavert: Talked to Ido, we should log this to fix in another PR.

src/graphorm/index.ts
```diff
@@ -187,6 +191,20 @@ const obj = new GraphORM({
...
         },
       },
+      privilegedMembers: {
+        relation: {
+          isMany: true,
+          customRelation: (ctx, parentAlias, childAlias, qb): QueryBuilder => {
+            return qb
+              .where(`${childAlias}."sourceId" = "${parentAlias}".id`)
+              .andWhere(
+                `(${childAlias}.role = '${SourceMemberRoles.Owner}' OR ${childAlias}.role = '${SourceMemberRoles.Moderator}')`,
```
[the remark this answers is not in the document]
capJavert: Talked to Ido gonna try to do it with `IN` statement.

src/schema/sources.ts
capJavert: Talked to Ido, we should fix this in another PR. Logged it here https://dailydotdev.atlassian.net/browse/WT-1216 FYI @rebelchris

rebelchris (changes requested):
One small type issue, otherwise good to go.
Looks neat!

rebelchris (approved):
Sweet! 🥳

idoshamun (approved):
Awesome!

PR #1220, opened 2023-03-30
feat: introduce blocked role and exclusion from squad members

rebelchris:
Added new role for blocked users.
Decided to go negative to ensure lower prio then users.

Also updated test cases to validate they no longer return blocked users.
Tested locally on FE side as well.

WT-1215 #done

### DIFF
--- a/__tests__/__snapshots__/sources.ts.snap
+++ b/__tests__/__snapshots__/sources.ts.snap
@@ -17,6 +17,33 @@ Array [
 ]
 `;
 
+exports[`members field should exclude blocked members from result 1`] = `
+Object {
+  "source": Object {
+    "id": "a",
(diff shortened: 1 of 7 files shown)


sshanzel (approved):
Looking great! Should we also have a paginated query to fetch all blocked members?

Also, was just thinking, in the GraphORM, what do you think of adding a `transform` function on the `currentMember`, where if the `roleRank` is less than 0, we can simply return null. Not necessary, but just thought of it.

rebelchris:
@sshanzel Yeah figured the retrieval is a separate task :)
As for currentMember I use it to determine the blocking of rejoining now, so would like to keep it.

idoshamun (approved):
Perfect!

PR #1221, opened 2023-03-30
fix: add check for deletion of comments by relevant role

rebelchris:
Allow deletion of comments by relevant roles.
Let me know if you guys think postDelete scope is ok, or we want a seperate scope?

WT-1183 #done

### DIFF
--- a/src/schema/comments.ts
+++ b/src/schema/comments.ts
@@ -28,7 +28,7 @@ import { GQLPost } from './posts';
 import { Roles } from '../roles';
 import { queryPaginatedByDate } from '../common/datePageGenerator';
 import { markdown, mentionSpecialCharacters } from '../common/markdown';
-import { ensureSourcePermissions } from './sources';
+import { ensureSourcePermissions, SourcePermissions } from './sources';
 import { generateShortId } from '../ids';
 import { ActivePost } from '../entity/posts/ActivePost';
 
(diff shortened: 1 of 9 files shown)


sshanzel:
> Let me know if you guys think postDelete scope is ok, or we want a seperate scope?

Oh, that is smart. If you can delete a post, then it is most likely you should be able to delete a comment as it is inside a post.

sshanzel (approved):
Well done! Just add some test cases to ensure we get the intended behavior. Thank you!

src/schema/comments.ts
```diff
@@ -731,9 +731,15 @@ export const resolvers: IResolvers<any, Context> = {
       await ctx.con.transaction(async (entityManager) => {
         const repo = entityManager.getRepository(Comment);
         const comment = await repo.findOneByOrFail({ id });
```
idoshamun, line 733: When using TypeORM, you can indicate that you are going to use the `post` relation to eagerly fetch it, instead of lazy load it later. I think you should do it to optimize. You need to use `findOneOrFail` without the `by`. And provide a `relations` array in the object

src/schema/comments.ts
```diff
@@ -731,9 +731,15 @@ export const resolvers: IResolvers<any, Context> = {
...
+        const post = await comment.post;
         if (
           comment.userId !== ctx.userId &&
-          ctx.roles.indexOf(Roles.Moderator) < 0
+          ctx.roles.indexOf(Roles.Moderator) < 0 &&
+          !(await ensureSourcePermissions(
+            ctx,
+            post.sourceId,
+            SourcePermissions.PostDelete,
```
idoshamun, line 741: I think we should add new permission for that. Wdyt?

src/schema/comments.ts
rebelchris: Ah cool yes, let me refactor it.

src/schema/comments.ts
rebelchris: Totally fine with me, was just wondering if it ever would be seperate.
But let's do it for sanity :)

PR #1222, opened 2023-03-30
feat: update role only by an owner

sshanzel:
- A new permission for updating source member roles
- A mutation to update a member role


src/schema/sources.ts
```diff
@@ -312,6 +318,26 @@ export const typeDefs = /* GraphQL */ `
       image: Upload
     ): Source! @auth
 
+    """
+    Set the source member's current role
+    """
+    updateMemberRole(
```
rebelchris, line 324: Maybe we should name it `changeMemberRole` as it does both update and demote?

src/schema/sources.ts
```diff
@@ -312,6 +318,26 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      sourceId: ID!
+
+      """
+      Member to update
+      """
+      memberId: ID!
+
+      """
+      Role to update the user to
```
rebelchris, line 336: Also would change this if we do the above

src/schema/sources.ts
```diff
@@ -846,6 +872,27 @@ export const resolvers: IResolvers<any, Context> = {
       });
       return { _: true };
     },
+    updateMemberRole: async (
```
rebelchris, line 875: And this :)

rebelchris (changes requested):
There is an edge-case you haven't covered now.
It's owner + moderator can actually remove a member (blocked role)

Would you mind adding that edge-case and some test cases for it?

src/schema/sources.ts
sshanzel: The update is applicable to demote and promote I believe. I mean, it just simply means change.

rebelchris:
@sshanzel There is no different mutation, deleting a member is simply assigning the `Blocked` role :)
that was my comment on slack when you took the task.

sshanzel:
> @sshanzel There is no different mutation, deleting a member is simply assigning the Blocked role :)

Oh, you are totally right! Let me update this to include "removal" of members.

sshanzel:
@rebelchris let me know what you think! 🚢

__tests__/__snapshots__/sources.ts.snap
```diff
@@ -213,8 +213,8 @@ Object {
         "post_delete",
         "member_remove",
         "edit",
+        "member_role_update",
         "post_limit",
-        "moderator_add",
         "moderator_remove",
```
rebelchris, line 218: Is this still needed seperate?

src/schema/sources.ts
```diff
@@ -865,6 +891,36 @@ export const resolvers: IResolvers<any, Context> = {
...
+      _,
+      { sourceId, memberId, role }: UpdateMemberRoleArgs,
+      ctx,
+    ): Promise<GQLEmptyResponse> => {
+      if (role === SourceMemberRoles.Blocked) {
```
rebelchris, line 899: We also have to think about "unblocking" basically it would be when assigning the "Member" role again, so maybe use `.includes`?

rebelchris (changes requested):
Just one more thing for unblocking members which we can tackle here.
Please also add one test case for moderators unblocking a member 🙏

__tests__/__snapshots__/sources.ts.snap
sshanzel: Oh, right. Missed deleting this one. Thanks!

rebelchris (approved):
Nice work! 🥳

idoshamun (approved):
Perfect!

sshanzel:
I checked the design and an owner seems to be allowed to block another owner. Fix should be in this commit: https://github.com/dailydotdev/daily-api/pull/1222/commits/e5cafa7b58c9b437df1e17f53eb30bb2c3c6344a

PR #1223, opened 2023-03-30
feat: squad posting permission

capJavert:
Add posting permission column for squads. By default `0` to have backward compatibility with earlier squad versions (which means members are higher roles).

Add checks for this new column when sharing posts or submitting external links in squads. 

Adjust createSquad mutation to be able to set this new column.


rebelchris (approved):
Good for this part 👍

src/schema/sources.ts
```diff
@@ -505,6 +515,31 @@ export const canAccessSource = async (
...
+): Promise<boolean> => {
+  if (squad.allowMemberPosting) {
+    return true;
+  }
+
+  const repo = ctx.getRepository(SourceMember);
+  const loggedUser = await repo.findOneByOrFail({
+    sourceId: squad.id,
+    userId: ctx.userId,
+  });
```
capJavert, line 530: If there is a way to reuse `loggedUser` to not do another query to the database I would be interested in example for that. 

I mean, there are already code paths that do this kind of thing but still curious.

src/schema/posts.ts
```diff
@@ -888,7 +890,16 @@ export const resolvers: IResolvers<any, Context> = {
...
+        const source = await ensureSourcePermissions(
+          ctx,
+          sourceId,
+          SourcePermissions.Post,
+        );
+
+        if (source.type === SourceType.Squad) {
+          await canPostToSquadCheck(ctx, source);
+        }
+
```
capJavert, line 902: I though about including this logic inside `ensureSourcePermissions` but I think that function already does enough things and we do want to explicitly check this only for squads (for now). So I think this is more verbose.

src/schema/sources.ts
idoshamun: I think it's better to get the SourceMember as a parameter, just like the source. Probably before calling this function, we also fetch it somewhere, so it's better to reuse it as mentioned

idoshamun (changes requested):
One minor comment for improvement but overall your API PRs are really slick! Way to go 🤩

src/schema/posts.ts
rebelchris: Hmm would argue that would be the better place as it already checks if you are allowed to post.
And it makes it easier down the line if we add more rules to who can post.

rebelchris (changes requested):

sshanzel:
I understand this was the implementation we discussed during the meeting, thank you for the quick work!

Though I remember Chris suggesting a property as a numerical value that we can set to easily manage more options on which ones are the allowed members to post. What do you think of it?

For example, if we want to allow Mods/Owners only, we can set the value to `1` since the member has a `0` value, and we can immediately identify that the user won't be allowed to post.

sshanzel:
I understand this was the implementation we discussed during the meeting, thank you for the quick work!

Though I remember Chris suggesting a property as a numerical value that we can set to easily manage more options on which ones are the allowed members to post. What do you think of it?

For example, if we want to allow Mods/Owners only, we can set the value to 1 since the member has a 0 value, and we can immediately identify that the user won't be allowed to post.

Would love to hear more thoughts about this. cc @idoshamun @rebelchris

rebelchris:
@sshanzel I didn't mention that, it was a discussion on the spec, but think when Ido misread that function.
This we only discussed and decided on boolean 👍

src/schema/posts.ts
capJavert: @rebelchris Issue is that it is not supposed to trigger on every route, only when sharing/creating post/link. To add this I would need to detect current path/mutation inside `ensureSourcePermissions` which I think is another piece of logic that should not be there. Does that make sense?

sshanzel:
> @sshanzel I didn't mention that, it was a discussion on the spec, but think when Ido misread that function.

Oh, cool. I must have remembered the meeting differently 😅

capJavert:
> I understand this was the implementation we discussed during the meeting, thank you for the quick work!
> 
> So sorry for bringing this now, though I remember Chris suggesting a property as a numerical value that we can set to easily manage more options on which ones are the allowed members to post. What do you think of it?
> 
> For example, if we want to allow Mods/Owners only, we can set the value to 1 since the member has a 0 value, and we can immediately identify that the user won't be allowed to post.
> 
> Would love to hear more thoughts about this. cc @idoshamun @rebelchris

I also mentioned number on the meeting but I think consensus was that we want a simple boolean. I am still up for numeric since boolean also gives problems on the frontend https://github.com/dailydotdev/apps/pull/1705 but time wise I don't know if we wanna redo everything. @rebelchris

src/schema/posts.ts
capJavert: Solved with 4822c4664aa28c883b844ca397980dd1caf41493

src/schema/sources.ts
capJavert: Solved with 4822c4664aa28c883b844ca397980dd1caf41493

src/schema/sources.ts
```diff
@@ -529,12 +553,36 @@ export const ensureSourcePermissions = async (
...
+      sourceMember,
+      permission,
+      validateRankAgainstId,
+    );
+
+    if (!canAccess) {
+      throw new ForbiddenError('Access denied!');
     }
+
+    if (
```
idoshamun, line 577: I think if the source type is not a squad this condition should also throw an error right?

src/schema/sources.ts
```diff
@@ -296,6 +296,10 @@ export const typeDefs = /* GraphQL */ `
       Commentary for the first share
       """
       commentary: String
+      """
+      Rank required for members to post
+      """
+      memberPostingRank: Int
```
capJavert, line 302: Curious if we should maybe expose this field as a string (SquadMemberRoles enum) and then just resolve integer here on the API side. Seems more bulletproof validation wise. WDYT?

capJavert: Concern is that we don't validate much so far so wanted to check first if some other reasons exist.

idoshamun (approved):

src/schema/sources.ts
```diff
@@ -438,6 +442,16 @@ const requireGreaterAccessPrivilege: Partial<
 
 type BaseSourceMember = Pick<SourceMember, 'role'>;
 
+export const hasGreaterAccess = (
```
rebelchris, line 445: What is the difference with the function below?

rebelchris (approved):
Just one comment on what seems a duplicate function

src/schema/sources.ts
capJavert: Someone could theoretically send `999` through request.

src/schema/sources.ts
capJavert: For this I guess it is not needed anymore. https://github.com/dailydotdev/daily-api/pull/1223#discussion_r1154487100 I'll remove but potentially it will raise again.

src/schema/sources.ts
capJavert: By requiring member name through string we would also be able to hide functionality of ranks from the frontend. Eg. frontend does not need to know that `owner` rank is `10`

sshanzel: That makes sense, though this mutation is not exposed publicly and is a bit extreme edge case where they will not benefit from doing (hacking the submission) after going through. For now, I think this should be fine.

sshanzel: Yes, this is exposed, but I mean **not** outside dailydev apps. We have CORS configured to only allow our known origins.

rebelchris: He means in GitHub people can simply look up the code

src/schema/sources.ts
```diff
@@ -486,25 +495,34 @@ export const canAccessSource = async (
...
+export const canPostToSquad = (
+  ctx: Context,
+  squad: SquadSource,
+  sourceMember: SourceMember,
+): boolean => {
+  if (!sourceMember) {
+    return false;
+  }
+
+  return sourceRoleRank[sourceMember.role] >= squad.memberPostingRank;
```
rebelchris, line 523: Are you sure we can't use the source permission check for this? 👀

rebelchris (changes requested):
Just one small optmization concern

src/schema/sources.ts
capJavert: Agreed in slack that this is ok for now. We might need to streamline permissions handling long term since we have multiple functions doing similar things.

src/schema/sources.ts
```diff
@@ -800,6 +841,7 @@ export const resolvers: IResolvers<any, Context> = {
         image,
         postId,
         description,
+        memberPostingRole = SourceMemberRoles.Member,
```
rebelchris, line 844: Not sure if this is dangerous if for some reason the field doesn't get send it might change from admin to member again?

rebelchris (changes requested):
One small concern

src/schema/sources.ts
capJavert: Since this is on update it will fail only if `null` is sent. By default (if nothing is sent for BC purposes it will fallback to member due to this).

capJavert: I can add a tests to cover the cases?

rebelchris: That's my concern.

Let's say someone set it to "admin only" and on a next update we only send the title for edit.
Now it will be reset to member right?

capJavert: This mutation is for create squad only right? The logic will be different for `editSquad`, this is not yet implemented, will do it in scope of another task. Added some tests for create here c9b7215cb46ec7c63b955f4c559c1e981432599a

rebelchris: Ah yes sorry, the GitHub PR showed "EditSquadArgs" above so kind of assumed it was a generic one.
But yes in that case never mind my comment.

rebelchris (approved):

PR #1224, opened 2023-03-31
feat: unblock a squad member

sshanzel:
- Unblock a member by removing the row from the table which would allow the user to re-join when they want to.
- I have also deleted the commented test cases for `removing a member` as it was already reused on the role update tests.
- Update `sourceMembers` query to accept a boolean parameter to return all blocked users only.
- Also updated the queries of tests to include `role` property to easily check snapshots.

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -414,6 +414,7 @@ const sourceToGQL = (source: Source): GQLSource => ({
 export enum SourcePermissions {
   CommentDelete = 'comment_delete',
   View = 'view',
+  ViewBlockedMembers = 'view_blocked_members',
   Post = 'post',
   PostLimit = 'post_limit',
   PostDelete = 'post_delete',
@@ -680,7 +681,12 @@ export const resolvers: IResolvers<any, Context> = {
       ctx,
       info,
     ): Promise<Connection<GQLSourceMember>> => {
-      await ensureSourcePermissions(ctx, sourceId);
+      const permission =
+        role === SourceMemberRoles.Blocked
+          ? SourcePermissions.ViewBlockedMembers
+          : SourcePermissions.View;
+
+      await ensureSourcePermissions(ctx, sourceId, permission);
       const page = membershipsPageGenerator.connArgsToPage(args);
       return graphorm.queryPaginated(
         ctx,
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -676,11 +676,11 @@ export const resolvers: IResolvers<any, Context> = {
     },
     sourceMembers: async (
       _,
-      { role, ...args }: SourceMemberArgs,
+      { role, sourceId, ...args }: SourceMemberArgs,
       ctx,
       info,
     ): Promise<Connection<GQLSourceMember>> => {
-      await ensureSourcePermissions(ctx, args.sourceId);
+      await ensureSourcePermissions(ctx, sourceId);
       const page = membershipsPageGenerator.connArgsToPage(args);
       return graphorm.queryPaginated(
         ctx,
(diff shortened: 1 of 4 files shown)


src/schema/sources.ts
```diff
@@ -344,6 +354,21 @@ export const typeDefs = /* GraphQL */ `
       role: String!
     ): EmptyResponse! @auth
 
+    """
+    Unblock a removed member with blocked role
+    """
+    unblockMember(
```
rebelchris, line 360: Since this simply removes a member from the squad, can we not leverage the existing "leaveSquad" function?

rebelchris (changes requested):
Just thinking, maybe we can actually re-use the leave query as it already exists?
It will keep things inline with what we have

src/schema/sources.ts
```diff
@@ -69,6 +69,11 @@ interface UpdateMemberRoleArgs {
   role: SourceMemberRoles;
 }
 
+interface SourceMemberArgs extends ConnectionArguments {
+  sourceId: string;
+  blockedOnly?: boolean;
```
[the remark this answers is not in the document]
sshanzel: Good idea! I will work that out.

src/schema/sources.ts
sshanzel: This makes sense, though leave refers to the logged-in user. Would love to hear more thoughts on this if can just make it accept a parameter. cc @capJavert @idoshamun

capJavert: I am in favor of dedicated function because down the line what unblocking does could change or add more things.

Logically like this, leaving squad is just one part of the unblocking flow, which makes more sense to me.

rebelchris: Alright, let's leave it for now.
How do you feel naming wise? `unblockMember` is not actually the right term in my opinion.
It totally removes someone from a squad as member.

src/schema/sources.ts
```diff
@@ -649,7 +676,7 @@ export const resolvers: IResolvers<any, Context> = {
     },
     sourceMembers: async (
       _,
-      args: ConnectionArguments & { sourceId: string },
+      { role, ...args }: SourceMemberArgs,
```
rebelchris, line 679: Why do we only extract role and not sourceId here?
I know it's not related to your change, but maybe we should extract sourceId as well?

rebelchris (approved):
Small optional change, and non blocking naming convention.

src/schema/sources.ts
sshanzel: Makes sense. Let me see if anything fails after destructuring the args and passing it without the `sourceId` to the paginator.

src/schema/sources.ts
sshanzel: Hmm. I think the intent with `unblock` is clear. But I'd love to hear suggestions on what we can replace it with.

rebelchris: I'm just not a fan of the product decision to call it unblock, as it's really not what it does.
To me the whole action is "remove member from squad", but let's leave it, it's nitpicking at this point.

sshanzel:
@idoshamun let us know what you think 🚢 🚢 🚢

src/schema/sources.ts
```diff
@@ -665,19 +692,28 @@ export const resolvers: IResolvers<any, Context> = {
...
-              } >= 0`,
-            )
+
             .addOrderBy(
               graphorm.mappings.SourceMember.fields.roleRank.select as string,
               'DESC',
             )
             .addOrderBy(`${builder.alias}."createdAt"`, 'DESC');
 
+          if (role) {
```
idoshamun, line 704: I think only the owner and mods can see blocked users right?

rebelchris (approved):
Amazing and clean 🧽

PR #1228, opened 2023-04-04
feat(squad): update posting permissions

capJavert:
Update posting permissions during `editSquad` mutation. Also expose `memberPostingRole` field.

### DIFF
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -855,8 +855,8 @@ describe('mutation createSquad', () => {
 
 describe('mutation editSquad', () => {
   const MUTATION = `
-  mutation EditSquad($sourceId: ID!, $name: String!, $handle: String!, $description: String) {
-  editSquad(sourceId: $sourceId, name: $name, handle: $handle, description: $description) {
+  mutation EditSquad($sourceId: ID!, $name: String!, $handle: String!, $description: String, $memberPostingRole: String) {
+  editSquad(sourceId: $sourceId, name: $name, handle: $handle, description: $description, memberPostingRole: $memberPostingRole) {
     id
   }
 }`;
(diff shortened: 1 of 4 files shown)


__tests__/sources.ts
```diff
@@ -952,6 +952,55 @@ describe('mutation editSquad', () => {
...
+    expect(res.errors).toBeFalsy();
+    const editSource = await con
+      .getRepository(SquadSource)
+      .findOneBy({ id: variables.sourceId });
+    expect(editSource?.memberPostingRank).toEqual(
+      sourceRoleRank[SourceMemberRoles.Moderator],
+    );
+  });
+
+  it('should leave squad memberPostingRank unchanged if not sent during edit', async () => {
```
rebelchris, line 985: Maybe also add the following scenario:

Role is set to owner only
Update only modifies "title"
Should not change role and should not throw error

src/graphorm/index.ts
```diff
@@ -210,6 +211,11 @@ const obj = new GraphORM({
         },
         transform: nullIfNotLoggedIn,
       },
+      memberPostingRole: {
```
rebelchris, line 214: Just to note with this we can remove the actual role on FE side right?

src/roles.ts
```diff
@@ -17,3 +17,12 @@ export const sourceRoleRank: Record<SourceMemberRoles, number> = {
 };
 
 export const sourceRoleRankKeys = Object.keys(sourceRoleRank);
+
+export const rankToSourceRole = Object.entries(sourceRoleRank).reduce(
```
rebelchris, line 21: Alternative something like this could maybe work?
But just a suggestion at this point.

```
const rankToSourceRole = (value) => {
  return Object.keys(sourceRoleRank).find(key => sourceRoleRank[key] === value);
}```

File: src/schema/sources.ts [hunk 4]
```diff
@@ -926,6 +937,13 @@ export const resolvers: IResolvers<any, Context> = {
       });
 
       try {
+        if (
+          typeof memberPostingRole !== 'undefined' &&
+          !sourceRoleRankKeys.includes(memberPostingRole)
+        ) {
+          throw new ValidationError('Invalid member posting role');
+        }
```
rebelchris, line 945: We should uniform this with how we handle the `createSquad` can you maybe abstract it?

src/schema/sources.ts
```diff
@@ -926,6 +937,13 @@ export const resolvers: IResolvers<any, Context> = {
       });
 
       try {
+        if (
+          typeof memberPostingRole !== 'undefined' &&
+          !sourceRoleRankKeys.includes(memberPostingRole)
```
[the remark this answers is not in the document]
idoshamun: Ok fair enough. I don't mind keeping it as it is :)

idoshamun (approved):

src/roles.ts
capJavert: Wanted to make a static map since it does not change during runtime and to mimic what we have for `sourceRoleRank`.

src/schema/sources.ts
capJavert: @rebelchris I thought about moving it to `validateSquadData`, I think it is not an issue if it just throws from there?

rebelchris: Ah yes that's totally fine as well.
As long as all fn's use this in a singular way :)

src/graphorm/index.ts
rebelchris: Yes 🙏

rebelchris (approved):
Love it 😍
think this is super neat 🙏

PR #1229, opened 2023-04-05
feat: query user's membership

sshanzel:
We display an indicator to let the our users know whether the author of the comment they are replying to won't be able to have access on their response by showing some alert. We display that alert after checking whether the user is blocked or not a member of the squad.

Changes:
- Seeing the graphorm used for fetching single-row items on numerous areas, I've decided to abstract the throwing of error when none is found.
- Introduced a query to check the user's membership.

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -580,27 +580,21 @@ const validateSquadData = ({
   return handle;
 };
 
-interface PermissionsOptionalParams {
-  validateRankAgainstId?: string;
-  primaryMemberId?: string;
-}
-
 export const ensureSourcePermissions = async (
   ctx: Context,
   sourceId: string | undefined,
   permission: SourcePermissions = SourcePermissions.View,
-  { primaryMemberId, validateRankAgainstId }: PermissionsOptionalParams = {},
+  validateRankAgainstId?: string,
 ): Promise<Source> => {
   if (sourceId) {
     const source = await ctx.con
       .getRepository(Source)
       .findOneByOrFail([{ id: sourceId }, { handle: sourceId }]);
-    const memberId = primaryMemberId ?? ctx.userId;
     const sourceMember = await ctx.con
       .getRepository(SourceMember)
-      .findOneBy({ sourceId: source.id, userId: memberId });
+      .findOneBy({ sourceId: source.id, userId: ctx.userId });
 
-    const hasAccess = await canAccessSource(
+    const canAccess = await canAccessSource(
       ctx,
       source,
       sourceMember,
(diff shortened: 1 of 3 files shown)


src/schema/sources.ts
```diff
@@ -808,6 +825,29 @@ export const resolvers: IResolvers<any, Context> = {
...
     },
+    checkUserMembership: async (
+      _,
+      { sourceId, memberId }: CheckUserMembershipArgs,
+      ctx,
+      info,
+    ): Promise<GQLSourceMember> => {
+      await ensureSourcePermissions(ctx, sourceId, SourcePermissions.View, {
+        primaryMemberId: memberId,
+      });
```
rebelchris, line 836: Why do we check this?
Technically a blocked user would have no rights to view.

My initial idea was we could simply query the user and return "false" or "true" based on their role?

rebelchris (changes requested):
I'm personally not sure we need to change this function.

The only information we need on this endpoint if true/false to see if this user is blocked correct or no longer part of the squad correct?

Cc: @idoshamun would love to get your input on this as well.

sshanzel:
> The only information we need on this endpoint if true/false to see if this user is blocked correct or no longer part of the squad correct?

We also need information on whether the role is blocked. I agree we only need true/false but I can see most definitely having a requirement in the future to get someone else's source member information. Though I could be wrong.

src/schema/sources.ts
rebelchris: Ah yes, but this seems to abuse the function.
For the specced item now a query should be more than enough no?

sshanzel: Good point. Yes, a query should be enough, just thought of making one endpoint to fetch source membership as we don't have one.

idoshamun: Agree with Chris. This is very confusing, we need to verify that the logged user has access to make this check and a simple query should be fine to check the blocked user

sshanzel:
@rebelchris I have reverted the change for the permission check, let me know if you would want to return just a boolean as it currently returns SoureMember. Thanks for the review! 🚢

src/schema/sources.ts
```diff
@@ -270,6 +275,21 @@ export const typeDefs = /* GraphQL */ `
...
+    Check whether user has membership access to a squad
+    """
+    checkUserMembership(
+      """
+      The source to check whether the user is a part of
+      """
+      sourceId: ID!
+
+      """
+      Username or the user id to check
```
rebelchris, line 288: Just for clarity this only checks on id no?

rebelchris:
@sshanzel If we use graphORM as now this is fine, but from my perspective a simple boolean and regular entityManager query would even be good enough.

That would also make the frontend side a bit cleaner.
I personally don't really see a use-case to know any other information for now, but might come, so both options are good from my side.

rebelchris (approved):

src/schema/sources.ts
sshanzel: Correct. This should be just ID. Let me refactor.

rebelchris (approved):
Looks good 🤙

PR #1231, opened 2023-04-06
fix: issue with empty email notification callback

rebelchris:
Changed the empty email callback to not fail.
Also added reference for system on the squad blocked notification.
And added test cases to ensure there is no error thrown.

WT-1229 #done

### DIFF
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -212,7 +212,10 @@ export const generateNotificationMap: Record<
       .icon(NotificationIcon.Bell)
       .objectPost(ctx.post, ctx.source, ctx.sharedPost),
   squad_blocked: (builder, ctx: NotificationSourceContext) =>
-    builder.icon(NotificationIcon.Block).referenceSource(ctx.source),
+    builder
+      .targetUrl('system')
+      .icon(NotificationIcon.Block)
+      .referenceSource(ctx.source),
   promoted_to_owner: (builder, ctx: NotificationSourceContext) =>
     builder
       .icon(NotificationIcon.Star)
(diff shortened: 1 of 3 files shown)


src/workers/newNotificationMail.ts
```diff
@@ -481,7 +481,9 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
       user_reputation: user.reputation,
     };
   },
-  squad_blocked: null,
+  squad_blocked: async () => {
+    return null;
+  },
```
capJavert, line 486: Just a question for my understanding, what is the difference here between async method and just value?

rebelchris: It's mostly the implementation expecting a function so failed on null value itself.

capJavert (approved):

src/notifications/generate.ts
```diff
@@ -213,6 +213,7 @@ export const generateNotificationMap: Record<
       .objectPost(ctx.post, ctx.source, ctx.sharedPost),
   squad_blocked: (builder, ctx: NotificationSourceContext) =>
     builder
+      .targetUrl(process.env.COMMENTS_PREFIX)
```
idoshamun, line 216: It should be added after `referenceSource` because it also sets the target url. So adding it will be overridden later.

rebelchris: Only targetSource will set it right?
As the issue was it being empty

idoshamun: My bad!

idoshamun (approved):

PR #1232, opened 2023-04-06
feat: exclude post permission for squad based on memberPostingRank

capJavert:
Since `permissions` are used on the frontend to check if user should see any action/ui element we wanted to automatically exclude `SourcePermissions.Post` permissions array based on `memberPostingRank` setting for squad. 

Also adds `currentMember.permissions` to `boot.squads` (only `SourcePermissions.Post` to keep boot size lean).

Initial discussion and alternatives can be found here https://github.com/dailydotdev/apps/pull/1723#discussion_r1159470225

Note to @idoshamun I am not sure if this kind of `select` is performant or is there a better way to do it - considering everything `graphorm` does under the hood.

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -1,4 +1,4 @@
-import { SourcePermissions, roleSourcePermissions } from './../schema/sources';
+import { getPermissionsForMember } from './../schema/sources';
 import { GraphORM, QueryBuilder } from './graphorm';
 import {
   Bookmark,
@@ -239,19 +239,7 @@ const obj = new GraphORM({
             return null;
           }
 
-          const permissions =
-            roleSourcePermissions[member.role] ?? roleSourcePermissions.member;
-          const memberRank =
-            sourceRoleRank[member.role] ??
-            sourceRoleRank[SourceMemberRoles.Member];
-
-          if (memberRank < memberPostingRank) {
-            return permissions.filter(
-              (item) => item !== SourcePermissions.Post,
-            );
-          }
-
-          return permissions;
+          return getPermissionsForMember(member, { memberPostingRank });
         },
       },
       roleRank: {
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -226,8 +226,8 @@ const obj = new GraphORM({
         select: (ctx: Context, alias: string, qb: QueryBuilder): string => {
           const query = qb
             .select('"memberPostingRank"')
-            .from(Source, 'ps')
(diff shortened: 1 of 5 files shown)


capJavert:
@rebelchris one last thing that we can maybe do in this PR is to also expose permissions from boot? We use `boot.squads` to render squads in share dialogs on the client and those do not have permissions property so we can not use `verifyPermissions` to hide them if user can't share to squad.

src/graphorm/index.ts
```diff
@@ -222,14 +223,35 @@ const obj = new GraphORM({
     requiredColumns: ['createdAt', 'userId', 'role'],
```
capJavert, line 223: I checked and `sourceId` is not needed here since we never query it but only attach a custom select over it.

src/graphorm/index.ts
```diff
@@ -222,14 +223,35 @@ const obj = new GraphORM({
...
     fields: {
       permissions: {
-        transform: (_, ctx: Context, member: SourceMember) => {
+        select: (ctx: Context, alias: string, qb: QueryBuilder): string => {
+          const query = qb
+            .select('"memberPostingRank"')
+            .from(Source, 'ps')
```
rebelchris, line 229: Not sure what ps stand for in the context

rebelchris (changes requested):
One small question otherwise happy from my side.

@idoshamun For your eyes when you get back. (Can be retroactive)

src/graphorm/index.ts
capJavert: `posting squad`, Was not sure if `source` will collide with anything else. I'll make it more verbose.

capJavert:
@rebelchris added boot implementation also, from what I can see it does not use graphorm there but i managed to connect it as we spoke.

src/routes/boot.ts
```diff
@@ -109,11 +122,34 @@ const getSquads = async (
...
+    const { role, memberPostingRank, ...restSource } = source;
+
+    const permissions = getPermissionsForMember(
+      {
+        role,
+      },
+      {
+        memberPostingRank,
+      },
+    );
```
idoshamun, line 139: Can you fix the lint here, please? It looks weird

idoshamun (approved):
One cosmetic comment but overall it looks good. As Ante wrote I have some concerns regarding the added `select` statement but we can fix it if we identify any bottleneck.
I don't have a better implementation in my head right now.

sshanzel (approved):
Got nothing to add to the comments above. Well done! 🚀

PR #1233, opened 2023-04-07
feat: filter out referral token for other users

capJavert:
Make sure `referralToken` of other users can not be queried through relations or listings. From what I talked with Ido before this should cover all cases.

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -35,8 +35,11 @@ const existsByUserAndPost =
 const nullIfNotLoggedIn = <T>(value: T, ctx: Context): T | null =>
   ctx.userId ? value : null;
 
-const nullIfNotSameUser = <T>(value: T, ctx: Context, parent: User): T | null =>
-  ctx.userId === parent.id ? value : null;
+const nullIfNotSameUser = <T>(
+  value: T,
+  ctx: Context,
+  parent: Pick<User, 'id'>,
+): T | null => (ctx.userId === parent.id ? value : null);
 
 const obj = new GraphORM({
   User: {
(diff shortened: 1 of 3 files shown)


rebelchris (commented):
But we need the referralToken for the invite oagw right?
Its not a secret or anything really

capJavert:
> But we need the referralToken for the invite oagw right? Its not a secret or anything really

You still get your own referralToken, but you should not be able to query referralTokens of other users, since this would lead to someone using other's referralToken to invite people to squad.

src/schema/sources.ts
```diff
@@ -1132,4 +1132,17 @@ export const resolvers: IResolvers<any, Context> = {
   Source: {
     permalink: (source: GQLSource): string => getSourceLink(source),
   },
+  SourceMember: {
+    referralToken: (
+      sourceMember: GQLSourceMember,
+      _,
+      ctx: Context,
+    ): string | undefined => {
+      if (!ctx.userId || ctx.userId !== sourceMember.user.id) {
```
idoshamun, line 1141: In cases where the user doesn't ask for the user field, this will not be available.
I think it's better to enforce it via graphorm as it will be a bit easier.
But less comprehensive than this implementation. I think it should cover 99% of the cases, though.

capJavert: 👍 Moved to graphorm `transform`.

One question, @idoshamun  I see we prefer returning `null` when no value from graphorm, any reasons for that vs `undefined`. I am fine with it just wanna check if it is because of something I am not aware of.

idoshamun (approved):

src/schema/sources.ts
idoshamun: @capJavert no real reason I believe

src/graphorm/index.ts
```diff
@@ -245,6 +245,15 @@ const obj = new GraphORM({
             ELSE 0 END)
           `,
       },
+      referralToken: {
+        transform: (value: string, ctx: Context, member: SourceMember) => {
```
rebelchris, line 249: Can we not simply use nullIfNotSameUser fn?

rebelchris (changes requested):
Small idea to uniform

src/graphorm/index.ts
capJavert: Did not know this one exister, for sure 👌

PR #1234, opened 2023-04-11
refactor: stop sending notifications when user is not a member anymore

sshanzel:
When a user shared a post then gets removed/blocked/left the squad. They should not receive notifications related to the Squad it came from. Removed from:
- Article new comment
- Comment new comment
- Article upvote milestone
- Comment upvote milestone.
- Post added notification

Let me know if I missed anything 🚢

### DIFF
--- a/src/workers/notifications/articleUpvoteMilestone.ts
+++ b/src/workers/notifications/articleUpvoteMilestone.ts
@@ -42,24 +42,25 @@ const worker: NotificationWorker = {
       upvoters,
       upvotes: post.upvotes,
     };
-    if (source.type !== SourceType.Squad) {
-      return users.map((userId) => ({
+
+    if (source.type === SourceType.Squad) {
+      const members = await con.getRepository(SourceMember).findBy({
+        userId: In(users),
+        sourceId: source.id,
+        role: Not(SourceMemberRoles.Blocked),
+      });
+
+      if (!members.length) {
+        return;
+      }
+
+      return members.map(({ userId }) => ({
         type: 'article_upvote_milestone',
         ctx: { ...ctx, userId },
       }));
     }
 
-    const members = await con.getRepository(SourceMember).findBy({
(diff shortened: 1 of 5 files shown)


src/workers/notifications/articleUpvoteMilestone.ts
```diff
@@ -40,7 +42,24 @@ const worker: NotificationWorker = {
...
+      userId: In(users),
+      sourceId: source.id,
+      role: Not(SourceMemberRoles.Blocked),
+    });
+
+    if (!members.length) {
+      return;
+    }
+
+    return members.map(({ userId }) => ({
```
capJavert, line 62: Maybe do all of the squad logic inside `source.type === SourceType.Squad`, since squad source is specific permission maybe better to early return from there and then on the bottom the default action stays the same. Similarly how you did it in `src/workers/notifications/commentUpvoteMilestone.ts`

src/workers/notifications/utils.ts
```diff
@@ -88,7 +98,25 @@ export async function articleNewCommentHandler(
...
+
+  const members = await con.getRepository(SourceMember).findBy({
+    userId: In(users),
+    sourceId: source.id,
+    role: Not(SourceMemberRoles.Blocked),
+  });
+
+  if (!members.length) {
+    return;
+  }
```
capJavert, line 117: Same as https://github.com/dailydotdev/daily-api/pull/1234/files#r1163771574

capJavert (approved):
Nice work 🚀 I have a minor flow style suggestion.

src/workers/notifications/articleUpvoteMilestone.ts
rebelchris: Agreed that would be cleaner.

rebelchris (approved):
Awesome :)

PR #1235, opened 2023-04-11
fix: squad shared post visibility condition

capJavert:
Re-submitting a not yet finished private link would result in making the shared post immediately visible in squad (also breaking UI since `post.sharedPost` was `undefined`).

During private link creation `visible` property of the original post is now checked and shared post visibility is set to the same value. 

Also updated `postUpdated` test to make sure all the pending squad posts are made visible once the content requirements are met through scraping (was already the case but just to be more specific).

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -1829,4 +1829,43 @@ describe('mutation submitExternalLink', () => {
     expect(sharedPost.title).toEqual('My comment');
     expect(sharedPost.visible).toEqual(true);
   });
+
+  it('should not make squad post visible if shared post is not yet ready and visible', async () => {
+    loggedUser = '1';
+    const res = await client.mutate(MUTATION, {
+      variables: {
+        ...variables,
+        url: 'http://p7.com',
+        commentary: 'Share 1',
+      },
+    });
+    expect(res.errors).toBeFalsy();
+    const articlePost = await con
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Yep thats the way!
Thanks for the fix Ante 🥳

__tests__/posts.ts
```diff
@@ -1829,4 +1829,43 @@ describe('mutation submitExternalLink', () => {
...
+      variables: {
+        ...variables,
+        url: 'http://p7.com',
+        commentary: 'Share 2',
+      },
+    });
+    expect(res2.errors).toBeFalsy();
+    const articlePost2 = await con
+      .getRepository(ArticlePost)
+      .findOneBy({ url: 'http://p7.com' });
```
sshanzel, line 1863: This could have returned the same article post from the first response. I think we need a more explicit condition to fetch, just to be sure the test is accurate.

capJavert: This is `ArticlePost`, there is a unique constraint on this entity so there can't be two. On other note, `sharedPost2` needed this so updated. Tnx! 🚀

sshanzel: Right, shared post 2 needed to have a more explicit check. Thanks for updating it! Well done! 🚢

sshanzel (approved):
All good from my end! Well done! 🚢

PR #1236, opened 2023-04-12
fix: change owner to admin naming

rebelchris:
Decided to rather fully rename everything to keep things clean.
Starting with API with a retro db update to ensure it keeps working.

All new squads will auto be `admin` role now.
We just need to update frontend to reflect it, opening PR soon.

Since we use roles and permissions it should not break existing frontend to much besides V4.

WT-1236

### DIFF
--- a/__tests__/__snapshots__/sources.ts.snap
+++ b/__tests__/__snapshots__/sources.ts.snap
@@ -25,7 +25,7 @@ Object {
       "edges": Array [
         Object {
           "node": Object {
-            "role": "owner",
+            "role": "admin",
             "source": Object {
               "id": "a",
             },
@@ -63,7 +63,7 @@ Object {
         },
         Object {
           "node": Object {
-            "role": "owner",
+            "role": "admin",
             "source": Object {
               "id": "a",
             },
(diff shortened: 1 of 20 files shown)


capJavert (commented):
Before merging API part I think we have two changes on main that potentially need updating. Files:
- https://github.com/dailydotdev/apps/blob/main/packages/shared/src/components/modals/SquadMemberModal.tsx#L95
- https://github.com/dailydotdev/apps/blob/main/packages/shared/src/components/squads/SquadHeaderMenu.tsx#L47

rebelchris:
@capJavert My idea was to potentially launch this together with V4 so there will be no extra steps needed.
Do you see any issues with that?

capJavert:
> @capJavert My idea was to potentially launch this together with V4 so there will be no extra steps needed. Do you see any issues with that?

@rebelchris Maybe from development and testing perspective, it will be hard to visually test/validate because you need to run migrations and previews will not work since API will still have old code and db data.

rebelchris:
@capJavert To be compliant with the move I added a temporary fix for the Frontend :)
https://github.com/dailydotdev/apps/pull/1741

PR #1237, opened 2023-04-12
fix: source member is no longer resolved for anonymous users

capJavert:
Issue was that `findeByOne` ignores `undefined` and `null` and then just executes query with `sourceId` eg. `SELECT * FROM source_member WHERE sourceId = :sourceId`. This would then just get first available source member from DB (most likely owner since it is created first with each squad) and if that member passed the check it would allow access.

I see multiple places where this could be issue in other parts of the code, It would be good to check in scope of another task. Logged [task](https://dailydotdev.atlassian.net/browse/WT-1254)

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -613,7 +613,7 @@ export const ensureSourcePermissions = async (
       ? await ctx.con
           .getRepository(SourceMember)
           .findOneBy({ sourceId: source.id, userId: ctx.userId })
-      : undefined;
+      : null;
 
     const canAccess = await canAccessSource(
       ctx,


rebelchris (approved):
Wow, learned something new today

capJavert:
So the issue that current tests did not catch this is that we only tested posts from sources without members. In that case this check inside `canAccessSource` was not allowing access since member was always `null`:
```js
if (!member) {
  return false;
}
```

If source (currently only squads) had members then the above invalid query would trigger and allow anonymous users to access. Added tests for anonymous and logged users. I kept the tests independent of squads since rest of the codebase for source is independent.

rebelchris (approved):
Thanks for fixing the tests

PR #1238, opened 2023-04-13
feat: submitting private link and be displayed real-time

sshanzel:
We now accept parameters for the title and image to use for the external link, but to support backwards compatibility, the said parameters are kept optional.

### DIFF
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -390,7 +390,7 @@ export const createExternalLink = async (
       sentAnalyticsReport: true,
       private: true,
       origin: PostOrigin.Squad,
-      visible: false,
+      visible: isVisible,
     });
     await createSharePost(
       entityManager,
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -361,8 +361,8 @@ const validateCommentary = async (commentary: string) => {
 
 export interface ExternalLink {
   url: string;
-  title: string;
-  image: string;
+  title?: string;
+  image?: string;
 }
 
 export const createExternalLink = async (
@@ -389,7 +389,7 @@ export const createExternalLink = async (
       sentAnalyticsReport: true,
       private: true,
       origin: PostOrigin.Squad,
-      visible: true,
+      visible: !!title,
     });
     await createSharePost(entityManager, sourceId, userId, id, commentary);
     await notifyContentRequested(logger, {
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -375,6 +375,7 @@ export const createExternalLink = async (
 ): Promise<void> => {
   await validateCommentary(commentary);
   const id = await generateShortId();
+  const isVisible = !!title;
 
   return con.transaction(async (entityManager) => {
     await entityManager.getRepository(ArticlePost).insert({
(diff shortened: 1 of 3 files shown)


src/entity/posts/utils.ts
```diff
@@ -378,6 +385,8 @@ export const createExternalLink = async (
       sourceId: UNKNOWN_SOURCE,
       url,
       canonicalUrl: url,
+      title,
+      image,
       sentAnalyticsReport: true,
       private: true,
       origin: PostOrigin.Squad,
```
rebelchris, line 392: Should we not do something with this visible as well? (Line below)

src/schema/posts.ts
```diff
@@ -880,16 +894,12 @@ export const resolvers: IResolvers<any, Context> = {
...
-        sourceId,
-        url,
-        commentary,
-      }: { sourceId: string; url: string; commentary: string },
+      { sourceId, commentary, url, title, image }: SubmitExternalLinkArgs,
       ctx,
     ): Promise<GQLEmptyResponse> => {
-      await ctx.con.transaction(async (manager) => {
-        await ensureSourcePermissions(ctx, sourceId, SourcePermissions.Post);
+      await ensureSourcePermissions(ctx, sourceId, SourcePermissions.Post);
```
rebelchris, line 900: Why did you move this outside of the transaction?

src/entity/posts/utils.ts
sshanzel: This is the article with unknown source which should still always be `visible: false`.

sshanzel:
> Also are you going to create the actual call in a separate PR?

Yes. Figured it would be difficult knowing the current setup, so decided to isolate it.

src/schema/posts.ts
sshanzel: Oh, was working on something, and ended up removing it. This got left. Though this is not utilized/needed in the transaction so nothing should be different performance-wise. Probably even better since we don't have to initialize the transaction while it is not yet needed.

src/entity/posts/utils.ts
rebelchris: That's not really correct, they also become visible when the title is available :)
They stay private at all times is what you mean.

This case would cause the error of it showing the share post, but not being able to load the actual title.

src/schema/posts.ts
rebelchris: Does query, so would prefer it to stay in the transaction to not loose scope.

src/entity/posts/utils.ts
sshanzel: Oh, in that case, my understanding was wrong. Thanks for pointing it out, let me update it.

src/schema/posts.ts
sshanzel: Oh, cool. Let's keep it inside.

rebelchris: It's still outside 👀

rebelchris (changes requested):

src/schema/posts.ts
sshanzel: Too many PRs going left and right. Forgive me this 😭

rebelchris (approved):
Nice! 🙏

sshanzel:
@capJavert would love to hear what you think 🚢 🚢 🚢

capJavert (approved):
Nice and simple 🚀

PR #1239, opened 2023-04-13
feat: external link preview

sshanzel:
A mutation to fetch the link's preview and title.

@vpol suggested we keep the URL provided in the instruction as we are doing an S2S integration.

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -1627,7 +1627,7 @@ describe('mutation submitExternalLink', () => {
   }
 }`;
 
-  const variables = {
+  const variables: Record<string, any> = {
     sourceId: 's1',
     url: 'https://daily.dev',
     commentary: 'My comment',
@@ -1659,20 +1659,14 @@ describe('mutation submitExternalLink', () => {
       'UNAUTHENTICATED',
     ));
 
-  it('should share to squad', async () => {
-    await con.getRepository(Source).insert({
-      id: UNKNOWN_SOURCE,
-      handle: UNKNOWN_SOURCE,
-      name: UNKNOWN_SOURCE,
-    });
-    loggedUser = '1';
+  const checkSharedPostExpectation = async (visible: boolean) => {
     const res = await client.mutate(MUTATION, { variables });
     expect(res.errors).toBeFalsy();
     const articlePost = await con
       .getRepository(ArticlePost)
       .findOneBy({ url: variables.url });
     expect(articlePost.url).toEqual('https://daily.dev');
-    expect(articlePost.visible).toEqual(false);
+    expect(articlePost.visible).toEqual(visible);
 
     expect(notifyContentRequested).toBeCalledTimes(1);
     expect(jest.mocked(notifyContentRequested).mock.calls[0].slice(1)).toEqual([
@@ -1684,7 +1678,28 @@ describe('mutation submitExternalLink', () => {
       .findOneBy({ sharedPostId: articlePost.id });
     expect(sharedPost.authorId).toEqual('1');
     expect(sharedPost.title).toEqual('My comment');
-    expect(sharedPost.visible).toEqual(false);
+    expect(sharedPost.visible).toEqual(visible);
+  };
+
+  it('should share to squad without title to support backwards compatibility', async () => {
+    await con.getRepository(Source).insert({
+      id: UNKNOWN_SOURCE,
+      handle: UNKNOWN_SOURCE,
+      name: UNKNOWN_SOURCE,
+    });
+    loggedUser = '1';
+    await checkSharedPostExpectation(false);
+  });
+
+  it('should share to squad and be visible automatically when title is available', async () => {
+    await con.getRepository(Source).insert({
+      id: UNKNOWN_SOURCE,
+      handle: UNKNOWN_SOURCE,
+      name: UNKNOWN_SOURCE,
+    });
+    loggedUser = '1';
+    variables.title = 'Sample external link title';
+    await checkSharedPostExpectation(true);
   });
 
   it('should share existing post to squad', async () => {
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -1621,8 +1621,8 @@ describe('mutation viewPost', () => {
 
 describe('mutation submitExternalLink', () => {
   const MUTATION = `
-  mutation SubmitExternalLink($sourceId: ID!, $url: String!, $commentary: String!) {
-  submitExternalLink(sourceId: $sourceId, url: $url, commentary: $commentary) {
+  mutation SubmitExternalLink($sourceId: ID!, $url: String!, $commentary: String!, $title: String!, $image: String!) {
+  submitExternalLink(sourceId: $sourceId, url: $url, commentary: $commentary, title: $title, image: $image) {
     _
   }
 }`;
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -38,8 +38,14 @@ import { postsFixture, postTagsFixture } from './fixture/post';
 import { Roles } from '../src/roles';
 import { DataSource, DeepPartial } from 'typeorm';
 import createOrGetConnection from '../src/db';
-import { notifyContentRequested, notifyView } from '../src/common';
+import {
+  defaultImage,
+  linkPreviewOrigin,
+  notifyContentRequested,
+  notifyView,
+} from '../src/common';
 import { randomUUID } from 'crypto';
+import nock from 'nock';
 
 jest.mock('../src/common/pubsub', () => ({
   ...(jest.requireActual('../src/common/pubsub') as Record<string, unknown>),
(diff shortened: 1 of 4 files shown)


.env
```diff
@@ -32,6 +32,7 @@ FLAGSMITH_WEBHOOK_SECRET=webhook
 
 INTERNAL_FEED=http://localhost:6000/feed.json?token=token
 
+LINK_PREVIEW_ORIGIN=http://post-scraper-one-ai-server.content.svc.cluster.local
```
rebelchris, line 35: I think we should add this as a secret pulumi key.
CC: @idoshamun not sure if that's needed?

src/common/post.ts
```diff
@@ -49,3 +50,17 @@ export const getPostCommenterIds = async (
 
 export const hasAuthorScout = (post: Post): boolean =>
   !!post?.authorId || !!post?.scoutId;
+
+export const linkPreviewOrigin = process.env.LINK_PREVIEW_ORIGIN;
+
+export const fetchLinkPreview = async (
+  url: string,
+): Promise<ExternalLinkPreview> => {
+  const res = await fetch(`${linkPreviewOrigin}/preview`, {
```
rebelchris, line 59: What happens when there is a error here?
Thinking we should maybe try/catch this in case? (+ maybe add log on the fail)

rebelchris (changes requested):
Two minor things, but looks neat.
This one would love to get Ido's thoughts on though.

src/common/post.ts
capJavert: `fetch` by default does not throw on non `200` status codes, so probably good to add `res.ok` check and then handle/log.

sshanzel: Oh, right. Let me throw an error when that happens.

src/schema/posts.ts
```diff
@@ -892,6 +913,15 @@ export const resolvers: IResolvers<any, Context> = {
...
       return { _: true };
     },
+    checkLinkPreview: async (
+      _,
+      { url, sourceId }: SubmitExternalLinkArgs,
+      ctx,
+    ): Promise<ExternalLinkPreview> => {
+      await ensureSourcePermissions(ctx, sourceId, SourcePermissions.Post);
+
+      return fetchLinkPreview(url);
```
capJavert, line 923: Would probably await `fetchLinkPreview` here and make sure only fields we expect (`ExternalLinkPreview`) are returned from the `checkLinkPreview`, if API changes in any way there could be some additional fields leaking unnecessarily.

capJavert (changes requested):
I addition to what Chris said.. some comments about response handling

src/schema/posts.ts
sshanzel: This would be bound to the GraphQL schema for the type we return. It would only allow returning those properties we define on the endpoint. So when there are additional ones, they won't be provided unnecessarily.

capJavert (approved):
🚀

.env
idoshamun: The best practice is to follow Chris's suggestion. We don't add environment variables here even if it's safe to do so. This file is relevant only for local env

src/schema/posts.ts
```diff
@@ -527,6 +534,20 @@ export const typeDefs = /* GraphQL */ `
...
     ): EmptyResponse @auth
 
+    """
+    Fetch external link's title and image preview
+    """
+    checkLinkPreview(
+      """
+      Source the user is planning to post to
+      """
+      sourceId: ID!
```
idoshamun, line 544: Why the source id is relevant here?
Potentially we can use this even for comments and other links in the app

idoshamun:
@sshanzel I think we should also add a rate limiting here based on the user id. Pretty sure we can find a package that supports it out of the box

package.json
```diff
@@ -77,6 +78,7 @@
     "pg": "^8.9.0",
     "pg-query-stream": "^4.3.0",
     "pino": "^8.10.0",
+    "rate-limiter-flexible": "^2.4.1",
```
sshanzel, line 81: Dependency of the primary package we use. Without this, build throws `ERROR: MODULE rate-limiter-flexible NOT FOUND`

Another option for rate limiting package is this: https://github.com/teamplanes/graphql-rate-limit

Though it doesn't seem to be actively maintained, last update was 2 years ago.

src/index.ts
```diff
@@ -161,17 +161,26 @@ export default async function app(
...
-                newError.message = 'Unexpected error';
-                newError.extensions = {
-                  code: 'UNEXPECTED',
-                };
+                if (
+                  error.originalError.message?.startsWith('Too many requests')
+                ) {
+                  newError.message = error.originalError.message;
+                  newError.extensions = {
+                    code: 'RATE_LIMITED',
```
sshanzel, line 169: Let me know if you would want to have a different error code to serve.

src/directive/rateLimit.ts
```diff
@@ -0,0 +1,33 @@
+import {
```
sshanzel, line 1: Inspiration came from one of their examples: https://github.com/ravangen/graphql-rate-limit/blob/main/examples/context/index.js

.env
```diff
@@ -32,6 +32,7 @@ FLAGSMITH_WEBHOOK_SECRET=webhook
 
 INTERNAL_FEED=http://localhost:6000/feed.json?token=token
 
+POST_SCRAPER_ORIGIN=http://localhost:8000
```
sshanzel, line 35: Port `8000` seems to be available. Let me know if you would want something else.

src/common/post.ts
```diff
@@ -49,3 +52,29 @@ export const getPostCommenterIds = async (
...
+export const fetchLinkPreview = async (
+  url: string,
+): Promise<ExternalLinkPreview> => {
+  if (!isValidHttpUrl(url)) {
+    throw new ValidationError('URL is not valid');
+  }
+
+  const res = await fetch(`${postScraperOrigin}/preview`, {
+    method: 'POST',
+    headers: { 'Content-Type': '' },
```
idoshamun, line 67: Should be this:

```suggestion
    headers: { 'Content-Type': 'application/json' },
```

src/common/post.ts
```diff
@@ -49,3 +52,29 @@ export const getPostCommenterIds = async (
...
+    body: JSON.stringify({ url }),
+  });
+
+  if (res.status >= 400 && res.status < 500) {
+    throw new ValidationError('Bad request!');
+  }
+
+  if (res.status >= 500) {
+    throw new ValidationError('Internal server error!');
```
idoshamun, line 76: This is not a ValidationError but an actual Error.

src/directive/rateLimit.ts
```diff
@@ -0,0 +1,61 @@
...
+  defaultKeyGenerator,
+  rateLimitDirective,
+} from 'graphql-rate-limit-directive';
+import { RateLimiterMemory } from 'rate-limiter-flexible';
+import { GraphQLError } from 'graphql';
+
+export let rateLimiter: CustomRateLimiterMemory;
+
+export class CustomRateLimiterMemory extends RateLimiterMemory {
```
idoshamun, line 10: Please switch to use the Redis setup. We already have a running redis instance and an active connection so you can reuse it and I saw the package you use supports redis as well

src/directive/rateLimit.ts
```diff
@@ -0,0 +1,61 @@
...
+  constructor(props) {
+    super(props);
+
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    rateLimiter = this;
```
idoshamun, line 15: Why are you doing this? Seems too hacky

src/directive/rateLimit.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  // for debugging purposes
+  consume(key, pointsToConsume, options) {
```
idoshamun, line 19: Can we delete it now I guess?
Or at least comment

src/common/post.ts
rebelchris: Alternative could you not check if status === 200 return the json, else return a error?
Seems more failsafe in this scenario.

src/entity/posts/utils.ts
```diff
@@ -359,10 +359,13 @@ const validateCommentary = async (commentary: string) => {
   return true;
 };
 
-export interface ExternalLink {
+export interface ExternalLinkPreview {
+  title: string;
+  image: string;
```
rebelchris, line 364: Is the image always available?

rebelchris: Ah see you make it partial anyway, wasn't clear right away

rebelchris (changes requested):
Besides Ido's comment on the rate limit file some minor questions.

src/common/post.ts
sshanzel: > This is not a ValidationError but an actual Error.

I agree, I was not able to find a similar error type we use for GraphQL that should represent Internal Server Error. Are you familiar with any?

src/directive/rateLimit.ts
sshanzel: That would be best.

src/directive/rateLimit.ts
sshanzel: I agree it is a bit hacky. I was not able to access the memory it is natively using to clear out some keys, but with using Redis, we might be able to avoid this.

src/directive/rateLimit.ts
sshanzel: I wanted to keep this so future devs would have an easy time debugging in case they stumble upon this case. Though I'm open to removing this as a whole.

src/common/post.ts
sshanzel: > Alternative could you not check if status === 200 return the json, else return a error?

Good point. We can opt for this for easier checking. Thank you for the suggestion.

sshanzel: > There is a res.ok, it would be best to use that.

Will it be still true on `res.code === 204`?

sshanzel:
We were able to remove the hacky `rateLimiter` variable 🥳 

In order to produce the rate limiter directive with redis client, it needs to fetch the connection of available ones from the redis pool first. We will need an async function to do that, but since the schema was created synchronously in a file, I had to refactor it to allow promises. @idoshamun let me know what you think with regards to the changes in this commit https://github.com/dailydotdev/daily-api/pull/1239/commits/3841c206a0df042681d6c0f19f85fc9dced0d858

src/common/post.ts
capJavert: Yes, it has `200-299` range https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

sshanzel: Oh, then I think we need to be specific with 200 since 204 is an empty response.

capJavert: Must have missed something, why would it return `204`?

sshanzel: Not really with the ins and outs of the API we are using but as a failsafe option.

capJavert: Are we not using API provided by the platform team, there should be defined valid responses?

sshanzel: We don't have yet documentation for the relevant API. We don't even have the exact 400-500 range error codes it might produce but to at least handle the situations we provided the conditions.

capJavert: If we don't know exact error codes we should return the link or throw 500 and log internally what happened. What if we start getting `401` and `403` from that API, those are not validation errors (as returned now) but auth errors. It just seems wrong to be specific in some cases and then not so specific in others when we don't actually know anything.

capJavert: Don't want to hold the discussion on technicality, maybe something to think about in the future.

sshanzel: Totally agree with that, we need to come up with something on how we should proceed for such cases. Thanks for the thoughts!

sshanzel:
Have used `ioredis@5.0.5` as it seems to be the one used at `graphql-redis-subscriptions` and `ts-ioredis-pool`. I went through the docs ioredis pool and wasn't able to find any API/process to allow the creation of an isolated connection/client. 

cc @rebelchris @idoshamun

__tests__/posts.ts
```diff
@@ -1916,3 +1923,106 @@ describe('mutation submitExternalLink', () => {
...
+    expect(res2.errors).toBeFalsy();
+    mockRequest();
+    const res3 = await client.mutate(MUTATION, { variables });
+    expect(res3.errors).toBeFalsy();
+    mockRequest();
+    const res4 = await client.mutate(MUTATION, { variables });
+    expect(res4.errors).toBeFalsy();
+    mockRequest();
+    const res5 = await client.mutate(MUTATION, { variables });
+    expect(res5.errors).toBeFalsy();
```
idoshamun, line 2004: I suggest looping instead of hard coding

package.json
```diff
@@ -77,6 +79,7 @@
...
```
idoshamun, line 82: This is not needed anymore right?

src/schema/posts.ts
```diff
@@ -527,6 +534,16 @@ export const typeDefs = /* GraphQL */ `
...
 
+    """
+    Fetch external link's title and image preview
+    """
+    checkLinkPreview(
+      """
+      URL of the external link
+      """
+      url: String!
+    ): LinkPreview @auth @rateLimit(limit: 5, duration: 60)
```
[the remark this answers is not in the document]
sshanzel: Yes, exactly. Thank you!

__tests__/posts.ts
sshanzel: Good point! Coming right up!

package.json
sshanzel: Still needed for specifying which store client to use, in our case, Redis. A direct dependency of the primary package of rate limiter too.

idoshamun (approved):

src/directive/rateLimit.ts
rebelchris: It's still active?

rebelchris (approved):
Sweet, just noticed @idoshamun comment doesn't seem resolved?

src/directive/rateLimit.ts
sshanzel: Have made a better comment now for easier understanding of its purpose.

sshanzel:
Tested backwards compatibility (locally) and all is good. Going to merge now.

PR #1240, opened 2023-04-18
feat: add memberInviteRank

capJavert:
Much like the already introduced memberPostingRank we will introduce a memberInviteRank which will default to `0` meaning everyone can invite.

Also includes updates to create/edit squad mutations with this new column.


src/graphorm/index.ts
```diff
@@ -242,6 +247,7 @@ const obj = new GraphORM({
             return null;
           }
 
+          // TODO member-invite-rank pass memberInviteRank to resolve Invite permission for member
           return getPermissionsForMember(member, { memberPostingRank });
```
capJavert, line 251: @rebelchris remember when we were adding above `select` to get `memberPostingRank` to then filter permissions array to not include `Post` permission depending on the `memberPostingRank` column.

I tried to add the same thing for `memberInviteRank` (already implemented it inside `getPermissionsForMember`) but when I try to add `memberInviteRank` column to the above `select` I get the error `error: subquery must return only one column`. Not sure what alternatives we have, tried with customRelation but it did not work same as when we tried before.

capJavert: Hacky way came to mind to use `concat("memberPostingRank", ':', "memberInviteRank")` in select and then in transform `const [memberPostingRank, memberInviteRank] = value.split(':')`. It works but I would rather find alternatives. This basically just cheats psql to think single column was returned.

rebelchris: I think it's because the :string value, you should be able to expect an array I think.
Let me know if you want to try and debug it together quickly?

capJavert: @idoshamun could you take a look at this one since we could not find the solution? Basically the issue is that we want to query two two columns from `select` in graphorm but can not due to error above.

capJavert: `array[colum1, column2]` works and it also returns native types eg. for our case `[number, number]`.

src/schema/sources.ts
```diff
@@ -706,15 +730,21 @@ const addNewSourceMember = async (
...
   if (memberRank < source.memberPostingRank) {
-    return permissions.filter((item) => item !== SourcePermissions.Post);
+    permissions = permissions.filter((item) => item !== SourcePermissions.Post);
+  }
+
+  if (memberRank < source.memberInviteRank) {
+    permissions = permissions.filter(
+      (item) => item !== SourcePermissions.Invite,
+    );
   }
```
rebelchris, line 748: Maybe reduce is better as it will not use a let?
Generally we try to avoid them as they might cause leaks down the road and are not super explicit

src/schema/sources.ts
```diff
@@ -596,6 +611,13 @@ const validateSquadData = ({
     throw new ValidationError('Invalid member posting role');
   }
 
+  if (
+    typeof memberInviteRole !== 'undefined' &&
+    !sourceRoleRankKeys.includes(memberInviteRole)
+  ) {
+    throw new ValidationError('Invalid member invite role');
+  }
```
rebelchris, line 619: Optional: Any way we can combine these 2 for future ease?

rebelchris (changes requested):
Amazing, minor suggestion and optional styling thing.

capJavert:
I have this strange fail on the tests, pretty sure I did not update that part of code, it does come from snapshots that can be tricky. https://app.circleci.com/pipelines/github/dailydotdev/daily-api/4653/workflows/5ff60546-85cd-480e-9a6b-c1f309bf4d98/jobs/17707, gotta look into that.

src/schema/sources.ts
capJavert: Would not worry about memory leaks here since value is contained inside this function. My main goal here was to do no iterations if we don't need to and I think with reduce you would at least have to double check conditions to retain this behaviour. I do wonder how you would structure this with `reduce`?

src/schema/sources.ts
capJavert: I think verbosity is ok for now so I will leave it. Let's see where this evolves and then think about abstraction then. I think in general we have to see how and with what we want to validate across all the API routes.

src/schema/sources.ts
capJavert: Chris gave another idea, implemented here 62955e1fd33862c1ca2812fb2801a4c5359a2da2

sshanzel (approved):
Looks good to me! 🥳

capJavert:
@idoshamun if you could take a look when you get the chance. 🙏

rebelchris (approved):
Great!

idoshamun (approved):
Perfect!

PR #1243, opened 2023-04-20
fix: stop sending notification on sync posting external links

sshanzel:
Don't send notifications when posting a link that is visible already (sync external link).


sshanzel:
> Awesome! Once we release the version we would want to consider removing the notification altogether but for now it's a better approach to keep backward support

I will write a ticket for this so it won't be forgotten. Thank you! 🚢

PR #1244, opened 2023-04-21
refactor: check link preview by checking db first

sshanzel:
Checking link preview should first verify in DB first if it exists. Else, we will fetch the URL preview from our post scraper service.

The `id` property would let us know that the link preview is available in our DB when used by the FE.


__tests__/posts.ts
```diff
@@ -2016,4 +2014,20 @@ describe('mutation checkLinkPreview', () => {
     expect(res.data.checkLinkPreview.title).toEqual(sampleResponse.title);
     expect(res.data.checkLinkPreview.image).toEqual(defaultImage.placeholder);
   });
+
+  it('should return post by canonical', async () => {
```
rebelchris, line 2018: Perhaps also add a case for existing post to return id?

rebelchris (changes requested):
Just one more question, with this is there any need for the old query?
If not can we remove it?

capJavert (changes requested):

sshanzel:
> Just one more question, with this is there any need for the old query?
If not can we remove it?

Haven't checked yet, but I agree, it would be better to be removed if unused.

src/schema/posts.ts
```diff
@@ -912,7 +913,19 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const standardizedUrl = standardizeURL(url);
+      const post = await ctx.con
+        .getRepository(ArticlePost)
+        .createQueryBuilder()
+        .select('id, title, image')
+        .where([{ canonicalUrl: standardizedUrl }, { url: standardizedUrl }])
+        .andWhere({ deleted: false })
+        .getRawOne();
+
+      return post ?? (await fetchLinkPreview(standardizedUrl));
```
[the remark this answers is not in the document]
sshanzel: I actually did that initially ([in this change](https://github.com/dailydotdev/daily-api/pull/1244/commits/641daf960df214de5388f08839a0dbe08847940b#diff-e3ad1f5d3fe340ded80e09a15a95f7d80be472d110fe2bdbbce21dba440b925aR927-R931)), but I thought it would be extremely rare to misunderstand the difference. WDYT?

__tests__/posts.ts
sshanzel: Good catch! I think we should add it as another `expect` statement from every test case we found the link in the DB. Let me update these two.

sshanzel: I've also added some `expect` statement on cases where we don't expect it 🚀

src/schema/posts.ts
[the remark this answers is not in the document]
idoshamun: I agree with Ante. If will make it more readable

__tests__/posts.ts
```diff
@@ -2015,5 +2014,24 @@ describe('mutation checkLinkPreview', () => {
...
+    expect(res.data.checkLinkPreview.id).toBeFalsy();
+  });
+
+  it('should return post by canonical', async () => {
+    loggedUser = '1';
+    const res = await client.mutate(MUTATION, {
+      variables: { url: 'http://p1c.com' },
+    });
+    expect(res.data.checkLinkPreview).toBeTruthy();
+    expect(res.data.checkLinkPreview.id).toBeTruthy();
```
idoshamun, line 2026: Please check that the right id is returned here

__tests__/posts.ts
```diff
@@ -2015,5 +2014,24 @@ describe('mutation checkLinkPreview', () => {
...
+  });
+
+  it('should return post by url', async () => {
+    loggedUser = '1';
+    const res = await client.mutate(MUTATION, {
+      variables: { url: 'http://p1.com' },
+    });
+    expect(res.data.checkLinkPreview).toBeTruthy();
+    expect(res.data.checkLinkPreview.id).toBeTruthy();
```
idoshamun, line 2035: Also here

idoshamun (changes requested):
Overall looks great some minor changes

sshanzel:
> Just one more question, with this is there any need for the old query?
If not can we remove it?

After checking, it seems to be still used by the `/boot/companion`.

idoshamun (approved):

rebelchris (approved):
Nice, clean fix 🙏

capJavert (approved):

PR #1245, opened 2023-04-21
feat: check permission when joining squad

capJavert:
With invite permission on squad there is a chance now that member shared a link on twitter or something and then after that admin restricted squad invites to moderators and above.

In that case since the link is out and we block the actual join request when it happens on our API.

Per comments here I refactored `errors.ts` to be `enum` instead of types https://github.com/dailydotdev/apps/pull/1761#discussion_r1173722616

### DIFF
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -5,6 +5,7 @@ import {
   initializeGraphQLTesting,
   MockContext,
   saveFixtures,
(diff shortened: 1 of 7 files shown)


idoshamun (approved):

rebelchris (approved):
Nice fix, and thanks for the retro-fix

PR #1249, opened 2023-04-26
fix: add freeform post type

rebelchris:
Added support for freeform post type.

- Re-uses image field
- New fields content and contentHtml as described in PR

Mutation is a separate ticket.

WT-1285 #done

### DIFF
--- a/src/entity/posts/FreeformPost.ts
+++ b/src/entity/posts/FreeformPost.ts
@@ -6,9 +6,9 @@ export class FreeformPost extends Post {
   @Column({ type: 'text', nullable: true })
   image?: string;
 
-  @Column({ type: 'text' })
+  @Column({ type: 'text', nullable: true })
   content: string;
 
-  @Column({ type: 'text' })
+  @Column({ type: 'text', nullable: true })
   contentHtml: string;
 }
--- a/src/entity/posts/FreeformPost.ts
+++ b/src/entity/posts/FreeformPost.ts
@@ -1,0 +1,14 @@
+import { ChildEntity, Column } from 'typeorm';
+import { Post, PostType } from './Post';
+
+@ChildEntity(PostType.Freeform)
+export class FreeformPost extends Post {
+  @Column({ type: 'text', nullable: true })
+  image?: string;
+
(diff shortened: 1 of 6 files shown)


capJavert (approved):

src/entity/posts/FreeformPost.ts
```diff
@@ -0,0 +1,14 @@
+import { ChildEntity, Column } from 'typeorm';
+import { Post, PostType } from './Post';
+
+@ChildEntity(PostType.Freeform)
+export class FreeformPost extends Post {
+  @Column({ type: 'text', nullable: true })
+  image?: string;
+
+  @Column({ type: 'text' })
+  content: string;
```
idoshamun, line 10: From what I understand from product, this is optional which makes `contentHtml` optional as well

src/migration/1682497694917-FreeformPost.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class FreeformPost1682497694917 implements MigrationInterface {
+  name = 'FreeformPost1682497694917';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "post" ADD "content" text`);
+    await queryRunner.query(`ALTER TABLE "post" ADD "contentHtml" text`);
+    await queryRunner.query(
```
idoshamun, line 9: This is not needed I guess

idoshamun (commented):
Please check my one comment

src/entity/posts/FreeformPost.ts
rebelchris: But that is the whole type right?
What else is there just a title alone?

src/migration/1682497694917-FreeformPost.ts
rebelchris: I needed it locally the view needs a update if it exists.
Unless there's an alternative way of refetching it?

idoshamun (approved):

PR #1250, opened 2023-04-26
fix: move feed endpoint call from get to post

rebelchris:
We where running out of space for the GET request with amount of params we sending.
So decided to switch to POST instead.

Note this is only verified locally on test simulations.
Needs to be tested against the actually feed to verify it doesn't break anything.

WT-1302 #done


sshanzel (approved):
Looks good to me 🥳

idoshamun (approved):

PR #1251, opened 2023-04-28
feat: action table

sshanzel:
Introduction of `action` table. The workers and the retro checks will be worked in a different PR so we can move forward with the tasks that require this information.

Thoughts: the whole schema for action is pretty small, would you rather have it in the `users` schema or continue with having its own?

### DIFF
--- a/src/entity/Action.ts
+++ b/src/entity/Action.ts
@@ -1,0 +1,19 @@
+import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
+
+export enum ActionType {
+  Notification = 'notification',
+}
+
+@Entity()
+export class Action {
+  @Index()
+  @PrimaryColumn({ type: 'text' })
+  userId: string;
+
+  @Index()
+  @PrimaryColumn({ type: 'text' })
+  type: ActionType;
+
+  @Column({ type: 'timestamp without time zone', default: () => 'now()' })
+  completedAt: Date;
+}
--- a/src/schema/actions.ts
+++ b/src/schema/actions.ts
@@ -1,0 +1,73 @@
+import { Action, ActionType } from '../entity';
+
+import { IResolvers } from '@graphql-tools/utils';
+import { traceResolvers } from './trace';
+import { Context } from '../Context';
+import { EmptyResponse } from '@google-cloud/pubsub';
+
+type GQLAction = Pick<Action, 'userId' | 'type' | 'completedAt'>;
+
+interface CompleteActionParams {
+  type: ActionType;
+}
+
+export const typeDefs = /* GraphQL */ `
+  """
+  Action that the user has completed
+  """
+  type Action {
+    """
+    The relevant user that made the action
+    """
+    userId: ID!
+
+    """
+    The type of action the user performed
+    """
+    type: String!
+
+    """
+    The date when the user completed the type of action
+    """
+    completedAt: DateTime!
+  }
+
+  extend type Mutation {
+    """
+    Complete action by the logged in user
+    """
+    completeAction(type: String!): EmptyResponse @auth
+  }
+
+  extend type Query {
+    """
+    Get the actions for user
+    """
+    actions: [Action]! @auth
+  }
+`;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const resolvers: IResolvers<any, Context> = traceResolvers({
+  Mutation: {
+    completeAction: async (
+      _,
+      { type }: CompleteActionParams,
+      { con, userId },
+    ): Promise<EmptyResponse> => {
+      await con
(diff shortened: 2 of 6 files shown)


capJavert (approved):
👍 nothing much to add.

I think separate schema for action is good just for future additions and maintainability down the road. I don't think there are any other graphql implications.

src/schema/actions.ts
```diff
@@ -0,0 +1,73 @@
...
+`;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const resolvers: IResolvers<any, Context> = traceResolvers({
+  Mutation: {
+    completeAction: async (
+      _,
+      { type }: CompleteActionParams,
+      { con, userId },
+    ): Promise<EmptyResponse> => {
```
rebelchris, line 57: I think you meant GQLEmptyResponse?

src/schema/actions.ts
```diff
@@ -0,0 +1,73 @@
...
+      await con
```
rebelchris, line 58: Why don't we just do `save` here not sure why the manual query builder is needed.

rebelchris (changes requested):
Some minor remarks, overall looks good 👍

src/schema/actions.ts
sshanzel: Oh, that's right. Didn't realize it came from a different thing. Nice catch!

src/schema/actions.ts
sshanzel: Haven't checked if `save` will fail, but let me check the docs.

rebelchris: Ah i see you wan't to ignore duplicates :)
Missed that part initially, in that case this should do and save one query.

rebelchris (approved):

src/entity/Action.ts
```diff
@@ -0,0 +1,19 @@
+import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
+
+export enum ActionType {
+  Notification = 'notification',
+}
+
+@Entity()
+export class Action {
```
idoshamun, line 8: We should consider a more explicit name. `Action` is too generic I believe.
@rebelchris @sshanzel thoughts?

src/entity/Action.ts
```diff
@@ -0,0 +1,19 @@
...
+  @Index()
+  @PrimaryColumn({ type: 'text' })
+  userId: string;
+
+  @Index()
```
idoshamun, line 13: I don't think we need index on this field. We're not gonna search over it I guess

src/schema/actions.ts
```diff
@@ -0,0 +1,73 @@
...
+        .values({ userId, type })
+        .orIgnore()
+        .execute();
+
+      return;
+    },
+  },
+  Query: {
+    actions: (_, __, { con, userId }): Promise<GQLAction[]> =>
+      con.getRepository(Action).findBy({ userId }),
```
idoshamun, line 71: Please use GraphORM for this. It's a safer way to run queries for GraphQL purposes

src/schema/actions.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const typeDefs = /* GraphQL */ `
+  """
+  Action that the user has completed
+  """
+  type Action {
+    """
+    The relevant user that made the action
+    """
+    userId: ID!
```
idoshamun, line 22: This should be a user object, not only the ID. It's a proper way to use GraphQL

src/entity/Action.ts
sshanzel: What do you both think of `CompletedAction`?

src/entity/Action.ts
sshanzel: Hmm. Good point, we may search at very rare cases but not very often. Will remove this one.

src/entity/Action.ts
idoshamun: Perhaps `UserAction` or `OnboardingAction` idk. But something with more context as who and maybe why

rebelchris: This should have been noticed on the DR for future reference.
Let's go with UserAction as it's not particular to onboarding actions it can be used for more things.

idoshamun: @rebelchris agree, my bad :)

src/schema/actions.ts
rebelchris: @idoshamun Just for my understanding what difference does it make for this type?
Just to know for future reference

idoshamun: For this type as it's currently defined not much. But it's future proof in case this entity becomes more complex with subtypes

idoshamun (approved):

PR #1255, opened 2023-05-01
Wt 1281 first post notification

sshanzel:
Introduced a new type of notification `squad_subscribe_to_notification`. Have added the prefix `squad` to the type since the content has a context of the squad and most likely we will have another `subscribe_to_notification` type.

I was planning to use the OneSignal API/SDK to check whether the user has any subscribed devices, but after checking, the said feature is not yet available. So now, it would depend on the FE whether we will render this notification item based on their subscription.

### DIFF
--- a/src/common/links.ts
+++ b/src/common/links.ts
@@ -31,6 +31,7 @@ export const getSourceLink = (
 
 export const scoutArticleLink = `${process.env.COMMENTS_PREFIX}?scout=true`;
 export const squadCreateLink = `${process.env.COMMENTS_PREFIX}?squad=true`;
+export const subscribeNotificationsLink = `${process.env.COMMENTS_PREFIX}/notifications?subscribe=true`;
 
 export const standardizeURL = (url: string): string => {
   const domain = subtractDomain(url);
--- a/src/entity/Notification.ts
+++ b/src/entity/Notification.ts
@@ -32,6 +32,7 @@ export type NotificationType =
   | 'squad_new_comment'
   | 'squad_reply'
   | 'squad_post_viewed'
+  | 'squad_subscribe_to_notification'
   | 'squad_blocked'
   | 'promoted_to_admin'
   | 'demoted_to_member'
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -1,7 +1,11 @@
 import { NotificationType } from '../entity';
 import { NotificationBuilder } from './builder';
 import { NotificationIcon } from './icons';
-import { scoutArticleLink, squadCreateLink } from '../common';
+import {
+  scoutArticleLink,
+  squadCreateLink,
+  subscribeNotificationsLink,
+} from '../common';
 import {
   NotificationBaseContext,
   NotificationCommentContext,
@@ -72,6 +76,8 @@ export const notificationTitleMap: Record<
     `<b>${ctx.doneBy.name}</b> <span class="text-theme-color-cabbage">viewed</span> your post on <b>${ctx.source.name}</b>.`,
   squad_blocked: (ctx: NotificationSourceContext) =>
     `You are no longer part of <b>${ctx.source.name}</b>`,
+  squad_subscribe_to_notification: (ctx: NotificationSourceContext) =>
+    `Congrats on your first post on <b>${ctx.source.name}</b>. Subscribe to get updates about activity in your squad.  Subscribe for updates.`,
   promoted_to_admin: (ctx: NotificationSourceContext) =>
     `Congratulations! You are now an <span class="text-theme-color-cabbage">admin</span> of <b>${ctx.source.name}</b>`,
   demoted_to_member: (ctx: NotificationSourceMemberRoleContext) =>
@@ -210,6 +216,12 @@ export const generateNotificationMap: Record<
       .targetUrl(process.env.COMMENTS_PREFIX)
       .avatarSource(ctx.source)
       .icon(NotificationIcon.Block)
+      .referenceSource(ctx.source),
+  squad_subscribe_to_notification: (builder, ctx: NotificationSourceContext) =>
+    builder
+      .targetUrl(subscribeNotificationsLink)
+      .avatarSource(ctx.source)
+      .icon(NotificationIcon.Bell)
       .referenceSource(ctx.source),
   promoted_to_admin: (builder, ctx: NotificationSourceContext) =>
     builder
--- a/src/workers/notifications/postAdded.ts
+++ b/src/workers/notifications/postAdded.ts
@@ -1,5 +1,5 @@
 import { messageToJson } from '../worker';
-import { Post, SourceMember, SourceType, User } from '../../entity';
+import { Post, SharePost, SourceMember, SourceType, User } from '../../entity';
 import {
   NotificationDoneByContext,
   NotificationPostContext,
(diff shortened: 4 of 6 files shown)


src/common/links.ts
```diff
@@ -31,6 +31,7 @@ export const getSourceLink = (
 
 export const scoutArticleLink = `${process.env.COMMENTS_PREFIX}?scout=true`;
 export const squadCreateLink = `${process.env.COMMENTS_PREFIX}?squad=true`;
+export const subscribeNotificationsLink = `${process.env.COMMENTS_PREFIX}/notifications?subscribe=true`;
```
rebelchris, line 34: This should rather be on the homepage no?

src/entity/Notification.ts
```diff
@@ -32,6 +32,7 @@ export type NotificationType =
   | 'squad_new_comment'
   | 'squad_reply'
   | 'squad_post_viewed'
+  | 'squad_subscribe_to_notification'
```
rebelchris, line 35: What about something like `squad_first_post_notification`?
That seems more explanatory of the notification

src/workers/notifications/postAdded.ts
```diff
@@ -66,6 +66,16 @@ const worker: NotificationWorker = {
             } as NotificationPostContext & Partial<NotificationDoneByContext>,
           }),
         );
+
+        const posts = await con.getRepository(SharePost).findBy({
+          authorId: post.authorId,
+        });
```
rebelchris, line 72: We do need to evaluate whether the use already accepted notifications right?
Else the'll get a pointless notification added.

rebelchris (changes requested):
I think you merged two PRs into one now, right?
Have some small feedback mainly naming wise.

src/common/links.ts
sshanzel: Not really sure where the best possible page is, but if it is going to be the notifications page, there would be at least the banner to see where the content changes once you have accepted.

src/entity/Notification.ts
sshanzel: I was thinking more about the intent. But I'm open to whichever fits. Would love to hear what others think too.

sshanzel:
> I think you merged two PRs into one now, right?

Oh, was not really planning to merge this first than the primary branch. Just so the commit history won't look messy I raised the PR right away 😁

src/entity/Notification.ts
capJavert: `squad_subscribe_to_notification` feels like it matches more with the rest of the values in `NotificationType`.. but we can always flip a coin to decide 😆

rebelchris (changes requested):
Just one more concern around the naming of the actual action

src/workers/notifications/postAdded.ts
```diff
@@ -66,6 +74,23 @@ const worker: NotificationWorker = {
...
           }),
         );
+
+        const posts = await con.getRepository(SharePost).countBy({
+          authorId: post.authorId,
+        });
+        if (posts === 1) {
+          const subscribed = await con.getRepository(UserAction).findOneBy({
+            userId: post.authorId,
+            type: UserActionType.EnableNotification,
```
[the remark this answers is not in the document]
rebelchris: Actually should be `squad_first_post` according to the specs right?
Screenshot 2023-05-02 at 13 08 19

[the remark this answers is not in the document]
rebelchris: Oh wait think you wanting to check if the user is already subscribed.
Got it, but then we also need to validate they don't already have this right? (Case of secondary squad for instance)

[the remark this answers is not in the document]
sshanzel: Yes, this is about whether we should still send a notification for them to subscribe.

> Got it, but then we also need to validate they don't already have this right? (Case of secondary squad for instance)

I think it should not be necessary now since our condition was just `authorId` without checking whichever Squad so this can only happen once. Unless the user deleted their post and then share something again, that is an edge case. Should we cover it or it is highly unlikely?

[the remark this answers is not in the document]
sshanzel: > squad_first_post

They are kinda relative to one another, so it is really a bit confusing. In the task I have for migration and retro checks, it will cover this case.

src/notifications/generate.ts
```diff
@@ -72,6 +76,8 @@ export const notificationTitleMap: Record<
     `<b>${ctx.doneBy.name}</b> <span class="text-theme-color-cabbage">viewed</span> your post on <b>${ctx.source.name}</b>.`,
   squad_blocked: (ctx: NotificationSourceContext) =>
     `You are no longer part of <b>${ctx.source.name}</b>`,
+  squad_subscribe_to_notification: (ctx: NotificationSourceContext) =>
+    `Congrats on your first post on <b>${ctx.source.name}</b>. Subscribe to get updates about activity in your squad.  Subscribe for updates.`,
```
rebelchris, line 80: Asked a [question](https://dailydotdev.slack.com/archives/C0506KU4RH9/p1683088513646799) to sab now, this copy seems a bit off.

Also did you notice the small copy is different, not sure we support it, but worth checking maybe?

Screenshot 2023-05-03 at 06 35 25

src/workers/notifications/postAdded.ts
```diff
@@ -66,6 +74,23 @@ const worker: NotificationWorker = {
             } as NotificationPostContext & Partial<NotificationDoneByContext>,
           }),
         );
+
```
rebelchris, line 77: Should we not also push the first post type here?
It seems like it can be in the same thing, but maybe you covered this in the other PR?

rebelchris (commented):
Looks good, small remark on copy let's wait what Sab says as it looks a bit odd.
Otherwise one question, but no blockers.

src/notifications/generate.ts
sshanzel: Oh, they might have changed it recently (or the design have a different text on in-app and notifications page). It used to have the Squad name in the text, hence, I prefixed the type with `squad` and used the source object. Will check this with them.

src/workers/notifications/postAdded.ts
sshanzel: Yes, that should be in the PR for completed actions. This one handles the new notification type.

src/notifications/generate.ts
sshanzel: Oh, nice. Thanks for asking it to Sab. I think they should be the same as it is the same entity when given back.

rebelchris: Maybe just add that question in the thread whether those can be the same copy :)

src/workers/notifications/postAdded.ts
```diff
@@ -66,6 +74,23 @@ const worker: NotificationWorker = {
...
+        const posts = await con.getRepository(SharePost).countBy({
```
idoshamun, line 78: Idea for optimization: What if we add a migration to add `UserAction` for everyone who already published a post and then the only thing we need to do here is to check the `UserAction` table like below. This will remove a query to count the number of articles every time a user publishes. Wdyt?

sshanzel: Good point! Much easier to check a single row than a potential many! Let's do that 🚢

PR #1256, opened 2023-05-02
feat: we should auto populate the usernames for auth onboarding

TomDowling:
Created a new query `generateUniqueUsername `, that takes the users name on signup and attempts to generate a username for them. It has 3 attempts to do so or it will exit without a suggestion. If the name of the user is already a username in the system, it will attempted 3 times to add a random string of numbers at the end to generate a unique username

WT-804 #done

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -1893,7 +1893,7 @@ describe('query generateUniqueUsername', () => {
     query GenerateUniqueUsername($name: String!) {
       generateUniqueUsername(name: $name)
   }`;
-  
+
   it('should return ', async () => {
     const res = await client.query(QUERY, { variables: { name: 'John Doe' } });
     expect(res.errors).toBeFalsy();
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -1887,3 +1887,16 @@ describe('DELETE /v1/users/me', () => {
     expect(cookies['da3'].value).toBeFalsy();
   });
 });
+
+describe('query generateUniqueUsername', () => {
+  const QUERY = `
+    query GenerateUniqueUsername($name: String!) {
+      generateUniqueUsername(name: $name)
+  }`;
+  
+  it('should return ', async () => {
+    const res = await client.query(QUERY, { variables: { name: 'John Doe' } });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.generateUniqueUsername).toEqual('johndoe');
+  });
+});
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -1894,9 +1894,17 @@ describe('query generateUniqueUsername', () => {
       generateUniqueUsername(name: $name)
   }`;
 
-  it('should return ', async () => {
+  it('should return a unique username', async () => {
     const res = await client.query(QUERY, { variables: { name: 'John Doe' } });
     expect(res.errors).toBeFalsy();
     expect(res.data.generateUniqueUsername).toEqual('johndoe');
+  });
+
+  it('should return a unique username with a random string', async () => {
+    await con.getRepository(User).update({ id: '1' }, { username: 'johndoe' });
+
+    const res = await client.query(QUERY, { variables: { name: 'John Doe' } });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.generateUniqueUsername).not.toEqual('johndoe');
   });
 });
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -33,6 +33,8 @@ import { GraphQLResolveInfo } from 'graphql';
 import { TypeOrmError, NotFoundError } from '../errors';
 import { deleteUser } from '../directive/user';
 import { ActivePost } from '../entity/posts/ActivePost';
+import { randomInt } from 'crypto';
+import { In } from 'typeorm';
 
 export interface GQLUpdateUserInput {
   name: string;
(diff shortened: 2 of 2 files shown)


__tests__/users.ts
```diff
@@ -1887,3 +1887,16 @@ describe('DELETE /v1/users/me', () => {
...
   });
 });
+
+describe('query generateUniqueUsername', () => {
+  const QUERY = `
+    query GenerateUniqueUsername($name: String!) {
+      generateUniqueUsername(name: $name)
+  }`;
+
+  it('should return ', async () => {
```
rebelchris, line 1897: Think the name here is missing a bit?

src/schema/users.ts
```diff
@@ -694,6 +704,36 @@
...
+        const usernameCheck = await ctx
+          .getRepository(User)
+          .findBy({ username });
+
+        if (usernameCheck.length === 0) {
+          break;
+        }
+
+        const random = Math.floor(Math.random() * 100);
+        username = `${username.substring(0, 37)}${random}`;
```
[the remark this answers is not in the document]
capJavert: `crypto.getRandomValues` should do it I think, it does have a bit of a raw interface, I would say this one is the right direction https://stackoverflow.com/a/55598692/9086023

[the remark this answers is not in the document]
TomDowling: I have created a function in the utils to `createSecureRandomNumber` which should solve this issue

idoshamun:
Not long ago @sshanzel added a rate limiting feature to the API. I suggest limiting the new query to make sure users will not abuse us. We can use IP-based (or any other thing like tracking id) limiting because this query cannot be behind authentication

sshanzel:
> We can use IP-based (or any other thing like tracking id) limiting because this query cannot be behind authentication

Currently, it will fallback to tracking id if user id is not found, so it should be safe to be used with our current anonymous users 🥳 

@TomDowling let me know if you need any help with how to use it! 🚀

src/schema/users.ts
```diff
@@ -694,6 +704,36 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const name = args.name.toLowerCase().replace(/[^a-zA-Z0-9]/g, '');
+      let attemptCount = 0;
+
+      if (name.length === 0) {
+        return '';
+      }
+
+      let username = name.substring(0, 39);
+
+      while (attemptCount < 3) {
```
sshanzel, line 721: I think to improve the performance check, we can generate 3 usernames already, then do a single `findBy` query so we won't have to do 3 awaits worst-case scenario. Then decide from there which to send. We can probably return an array of strings to provide multiple suggestions but probably out of scope for now.

idoshamun: That's a great suggestion, I love it! We can even create more than 3 if we do it in one query

TomDowling:
I have added the rateLimit to the query and set the limit to 20, @idoshamun is this an okay amount, or should I reduce it?

idoshamun:
@TomDowling should be fine for now :)

src/schema/users.ts
```diff
@@ -694,6 +706,48 @@ export const resolvers: IResolvers<any, Context> = {
...
+          .findBy({ username: In(generatedUsernames) });
+      })();
+
+      for (const usernameCheck in usernameChecks) {
+        generatedUsernames = generatedUsernames.filter(
+          (item) => item !== usernameChecks[usernameCheck].username,
+        );
+      }
+
+      if (generatedUsernames.length === 0) {
```
rebelchris, line 745: Non blocking request, could you perhaps add a log here (ctx.log.info) so we can keep track of these fails and see if we need to enhance the number of generated usernames.

rebelchris (approved):
One non blocking comment.
Would love to get @sshanzel and @idoshamun review on this one as well.

src/schema/users.ts
```diff
@@ -694,6 +706,48 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const name = args.name.toLowerCase().replace(/[^a-zA-Z0-9]/g, '');
+
+      if (name.length === 0) {
+        return '';
+      }
+
+      const username = name.substring(0, 39);
+      let generatedUsernames = [username];
+
+      const generateRandomUsernames = (count: number) => {
```
idoshamun, line 723: I don't think we need it as a function, we can just run this code. We don't reuse it anyway

src/schema/users.ts
```diff
@@ -694,6 +706,48 @@ export const resolvers: IResolvers<any, Context> = {
...
+        for (let i = 0; i < count; i++) {
+          const random = randomInt(100);
+          const randomUsername = `${username.substring(0, 37)}${random}`;
+          generatedUsernames.push(randomUsername);
+        }
+      };
+
+      generateRandomUsernames(4);
+
+      const usernameChecks = await (async () => {
```
idoshamun, line 733: Same here, it doesn't need to be an anonymous function

src/schema/users.ts
```diff
@@ -694,6 +706,48 @@ export const resolvers: IResolvers<any, Context> = {
...
+        return ctx
+          .getRepository(User)
```
idoshamun, line 735: Let's select only the username here to be more compact

idoshamun (changes requested):
A bit more fine tuning and it's ready!

idoshamun (approved):
Way to go!

sshanzel (approved):
Great work right there! 🥳

rebelchris (approved):
Love it 👍

PR #1257, opened 2023-05-02
feat: trigger completed actions

sshanzel:
Retro checks for completed action:
- Have taken the same path as we did on the markdown content update.
- One, it is much easier to develop, as it is more predictable running functions.
- Also testable as shown in the newly introduced test suite.

Completed actions implemented:
- create_squad
- join_squad
- squad_first_comment
- squad_first_post
- squad_invite
- my_feed

Actions not included in this PR:
- edit_welcome_post: will happen on the edit welcome post mutation.
- browser_extension: will happen onload at FE to send mutation to complete the action
- enabled_notification: there is currently no API on onesignal/node to fetch by user id. Fetching the devices is also limited to 300 rows per request. We will have to send a mutation once detected in FE.


src/schema/feeds.ts
```diff
@@ -1,4 +1,14 @@
-import { FeedAdvancedSettings, AdvancedSettings } from '../entity';
+import {
```
sshanzel, line 1: @rebelchris let me know what you think of the changes here. Or do you want to have a subscription on the alerts myFeed property changes?

src/entity/UserAction.ts
```diff
@@ -2,6 +2,11 @@ import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
 
 export enum UserActionType {
   Notification = 'notification',
```
rebelchris, line 4: This one is wrong in this version right? We described it as `enabled_notification`

src/schema/actions.ts
```diff
@@ -48,6 +49,19 @@ export const typeDefs = /* GraphQL */ `
   }
 `;
 
+export const insertOrIgnoreAction = (
```
rebelchris, line 52: Gotta love it!

src/schema/feeds.ts
rebelchris: This seems like it could trigger many, many times?
Not sure on the implication maybe the less friction for now would be listening on alerts.

Not particular against this solution, but maybe let's ask @idoshamun if he sees any issue with this?

src/workers/notifications/postAdded.ts
```diff
@@ -66,6 +74,18 @@ const worker: NotificationWorker = {
             } as NotificationPostContext & Partial<NotificationDoneByContext>,
           }),
         );
+
+        const posts = await con.getRepository(Post).countBy({
+          authorId: post.authorId,
+        });
```
rebelchris, line 80: Don't think this one is complete.
We would need to check inside this squad scope as well right?

rebelchris: Maybe we should also explicitly check for squad type post as they might be author of old articles?

src/workers/notifications/postAdded.ts
```diff
@@ -66,6 +74,18 @@ const worker: NotificationWorker = {
...
+
+        if (posts === 1) {
```
rebelchris, line 82: This would just be a "Extra"? 
Because technically they could delete a post, then post again (I do know ignore will trigger)

src/workers/notifications/utils.ts
```diff
@@ -64,8 +66,21 @@ export async function articleNewCommentHandler(
 
   const { post, source } = postCtx;
   const excludedUsers = [comment.userId];
-
   const isReply = !!comment.parentId;
+
+  if (source.type === SourceType.Squad) {
+    const comments = await con
+      .getRepository(Comment)
+      .countBy({ userId: comment.userId });
```
rebelchris, line 74: Would it not be simpler to always try the insert or ignore directly?
Seems we don't particular would care about querying first right?

Also the query doesn't take specific squads in concern

rebelchris (changes requested):
Wow love the approach, have some minor suggestion but like the way this is going.

src/schema/feeds.ts
sshanzel: Good point on the frequency which would make it somehow ideal to be listening on alerts.

src/entity/UserAction.ts
sshanzel: Yes, this should be fixed in the PR where we utilized this. Was just a placeholder initially.

src/workers/notifications/postAdded.ts
sshanzel: Though we check the Source Type to be Squad, I agree having to check the post type would be best.

src/workers/notifications/postAdded.ts
sshanzel: Oh, but when they delete a post we just set the property `deleted = true` right?

src/workers/notifications/utils.ts
sshanzel: Yeah, was also thinking that initially but hesitated. Now that you mentioned it, let's do that.

src/workers/notifications/postAdded.ts
rebelchris: Yeah we do, but this would just get any posts count right? 👀
So the count could be different than what you want.

sshanzel: Totally right!

src/workers/notifications/postAdded.ts
sshanzel: Indeed. Should be available on the condition now 🥳

sshanzel:
@rebelchris have now added the retro check and its tests.

rebelchris:
@sshanzel For the enable_notifications one idea I have is that we can actually export a csv of all current used_id in OneSignal and add manual query for those once we go live? (If it allows us to have less FE overhead?) 
Would love to hear what you and @idoshamun think about this.

sshanzel:
> @sshanzel For the enable_notifications one idea I have is that we can actually export a csv of all current used_id in OneSignal and add manual query for those once we go live? (If it allows us to have less FE overhead?)
Would love to hear what you and @idoshamun think about this.

Great idea. I think we can work with their exported CSV.

idoshamun:
> @sshanzel For the enable_notifications one idea I have is that we can actually export a csv of all current used_id in OneSignal and add manual query for those once we go live? (If it allows us to have less FE overhead?) Would love to hear what you and @idoshamun think about this.

Great! It's easy to achieve actually

bin/completedUserActions.ts
```diff
@@ -0,0 +1,118 @@
...
+const runFilterStream = async (con: DataSource) =>
+  con
+    .createQueryBuilder()
+    .select('DISTINCT a."userId"')
+    .addSelect(`'${UserActionType.MyFeed}'`, 'type')
+    .from(Alerts, 'a')
+    .where('a.filter IS FALSE')
+    .stream();
+
+export const retroCheckActions = async (ds?: DataSource): Promise<void> => {
```
idoshamun, line 94: Instead of selecting and then inserting, why don't you use insert from select?
You will not have to deal with streaming or anything like that and it's much more reliable

src/entity/UserAction.ts
```diff
@@ -1,7 +1,13 @@
 import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
 
 export enum UserActionType {
-  Notification = 'notification',
+  Notification = 'enabled_notification',
```
idoshamun, line 4: Let's keep the same convention

```suggestion
  Notification = 'enable_notification',
```

bin/completedUserActions.ts
sshanzel: Oh, cool. Just thought of doing the same process as the past executions, let's do that 🥳

src/schema/feeds.ts
idoshamun: You don't want to add anything unnecessary for the feed call. It may have a tremendous impact.
We already have a mechanism to identify when the user enable my feed, let's use it.

src/workers/notifications/postAdded.ts
idoshamun: I think we discussed in the other PR that we don't need to count the posts if we add retroactively the actions

src/schema/feeds.ts
sshanzel: Yep. I just realized we used the `alerts.filter` for this exact case, when the user updated their filters. We can probably do the inline direct from the updateAlerts mutation.

src/workers/notifications/squadMemberJoined.ts
```diff
@@ -24,6 +31,14 @@ const worker: NotificationWorker = {
         role: SourceMemberRoles.Admin,
       },
     });
+
+    const actionType =
```
idoshamun, line 35: This is a bit more tricky than that, because now admins can promote other admins.
And these new admins didn't really create a squad

idoshamun (changes requested):
Great start but still some work to do

src/workers/notifications/postAdded.ts
sshanzel: Yes, I will rebase this once the PR is merged.

src/workers/notifications/squadMemberJoined.ts
sshanzel: Yes, though it was discussed to provide the checklist for owners even if it came from a promotion.

Sab mentioned:

> We need to understand that this means if a user becomes a Squad admin (owner) then they see the owner checklist and it replaces any older checklists that may or may not have been completed

https://dailydotdev.slack.com/archives/C0506KU4RH9/p1681985308241999?thread_ts=1681978406.223319&cid=C0506KU4RH9

Another relevant discussion:

https://dailydotdev.slack.com/archives/C0506KU4RH9/p1683019582216549

idoshamun: Ok!

src/schema/alerts.ts
```diff
@@ -110,6 +111,10 @@ export const updateAlerts = async (
   const repo = con.getRepository(Alerts);
   const alerts = await repo.findOneBy({ userId });
 
+  if (data.filter === false) {
```
sshanzel, line 114: @rebelchris though we agreed on using `my_feed` but I realized `alert.filter` is more accurate and much easier to use as it is a boolean than a string. Do you think we can move forward with this?

sshanzel:
@idoshamun have updated it now to use `insert select` query 🚢 

@rebelchris have created a ticket for doing the OneSignal subscribed users export -> import user_action table.

https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11/backlog?selectedIssue=WT-1315

src/schema/alerts.ts
rebelchris: What's the difference between the two?
I meant the one we set once someone created my feed (thought it was that one, but open to the other if it's easier and the same)

rebelchris: Perfect then, this was probably the one I meant initially :)

rebelchris (approved):
Really neat work here Lee,
Nothing to add from my side, just make double sure the mutations work as they won't be able to revert.

bin/completedUserActions.ts
```diff
@@ -0,0 +1,97 @@
...
+  User,
+} from '../src/entity';
+import createOrGetConnection from '../src/db';
+import { SourceMemberRoles } from '../src/roles';
+import { DataSource } from 'typeorm';
+
+const insertSelectAction = (
+  con: DataSource,
+  [query, params]: [string, unknown[]],
+) => con.query(`INSERT INTO user_action("userId", type) ${query}`, params);
```
idoshamun, line 18: I think we should ignore conflicts right?

bin/notificationSubscribedUsers.ts
```diff
@@ -0,0 +1,9028 @@
+import { UserAction, UserActionType } from '../src/entity';
+import createOrGetConnection from '../src/db';
+
+const subscribedUsers = [
```
idoshamun, line 4: Please remove the array from this file. It should be provided as a runtime argument.
I don't want to expose user ids in our codebase

sshanzel: Oh, right. How would we receive the values and what would be the type of the argument? Will it be a comma-separated string?

idoshamun: I think a path to a file and then we can read the file in runtime

bin/completedUserActions.ts
sshanzel: Indeed we should 🚀

bin/notificationSubscribedUsers.ts
```diff
@@ -0,0 +1,30 @@
+import { UserAction, UserActionType } from '../src/entity';
+import createOrGetConnection from '../src/db';
+
+const subscribedUsers = [];
```
sshanzel, line 4: WIP 🔨

bin/notificationSubscribedUsers.ts
```diff
@@ -0,0 +1,49 @@
...
+  stream.on('data', function ([userId]) {
+    if (
+      !userId?.trim() ||
+      existing.some(({ userId: existingId }) => existingId === userId)
+    ) {
+      return;
+    }
+
+    console.log('pushing: ', userId);
+    subscribed[userId] = true;
```
sshanzel, line 31: Used object, for easier filtering unique ids as users can have multiple devices from this CSV.

idoshamun (approved):
Looks great just one comment and you can merge without my followup approval after validating locally

bin/notificationSubscribedUsers.ts
```diff
@@ -0,0 +1,49 @@
+import { UserAction, UserActionType } from '../src/entity';
+import createOrGetConnection from '../src/db';
+import * as fs from 'fs';
+import { parse } from 'csv-parse';
+
+(async () => {
+  const con = await createOrGetConnection();
+  const existing = await con
```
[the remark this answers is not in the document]
sshanzel: That makes sense. Just thought of doing it explicitly if possible, but yes, let's just keep it to the ignoring function.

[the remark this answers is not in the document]
idoshamun: Just to give you clear guidance, less queries the better. The more you do on the database side the less you have to think about scale

sshanzel:
Oh, and the other one left would be the `squad_invite`. Though we can move forward if we will add another thing to catch user creation. 

Reference for the actions: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/498401281/Squads+V5#Types-of-action

idoshamun:
@sshanzel what about enable notifications? This will also require frontend side

sshanzel:
> @sshanzel what about enable notifications? This will also require frontend side

Oh, that is right. Though we have a separate runner for the notifications, I agree it would be better to do it in one go. I am just so excited as this is my first time doing this use-case 😂

idoshamun:
Lol yeah. So let's run it once the frontend is done and merged :)

bin/notificationSubscribedUsers.ts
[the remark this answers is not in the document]
sshanzel: Thank you for the insight! 🙏

PR #1260, opened 2023-05-07
feat: add welcome post when creating 

sshanzel:
Add a welcome post when a user has created a Squad.

As we have been trying to move away from snapshots, I have also updated the tests to use an accurate expectation based on the description.

### DIFF
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -655,7 +655,7 @@ describe('query sourceFeed', () => {
     now = new Date(),
     first = 10,
   ): string => `{
-    sourceFeed(source: "${source}", ranking: ${ranking}, now: "${now.toISOString()}", first: ${first}) {
+    sourceFeed(source: "${source}", ranking: ${ranking}, now: "${now.toISOString()}", first: ${first}, supportedTypes: ["article", "share", "welcome"]) {
       ${feedFields()}
     }
   }`;
@@ -669,6 +669,16 @@ describe('query sourceFeed', () => {
     await con.getRepository(Post).update({ id: 'p5' }, { banned: true });
     const res = await client.query(QUERY('b'));
     expect(res.data).toMatchSnapshot();
+  });
+
+  it('should display a welcome post first in source feed', async () => {
+    const res1 = await client.query(QUERY('b', Ranking.TIME));
+    expect(res1.data.sourceFeed.edges[0].node.id).not.toEqual('p5');
+    await con
+      .getRepository(Post)
+      .update({ id: 'p5' }, { type: PostType.Welcome });
+    const res2 = await client.query(QUERY('b'));
+    expect(res2.data.sourceFeed.edges[0].node.id).toEqual('p5');
   });
 
   it('should not display a banned post for community source', async () => {
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -52,8 +52,7 @@ import { DataSource } from 'typeorm';
 import createOrGetConnection from '../src/db';
 import { randomUUID } from 'crypto';
 import { usersFixture } from './fixture/user';
-import { base64, unbase64 } from 'graphql-relay/utils/base64';
-import { addYears } from 'date-fns';
+import { base64 } from 'graphql-relay/utils/base64';
 import { changeYearToNextYear } from '../src/common/date';
 
 let app: FastifyInstance;
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -666,13 +666,17 @@ describe('query sourceFeed', () => {
 
   it('should return a single source feed', async () => {
     const res = await client.query(QUERY('b'));
-    expect(res.data).toMatchSnapshot();
+    res.data.sourceFeed.edges.forEach(({ node }) =>
+      expect(node.source.id).toEqual('b'),
+    );
   });
 
   it('should display a banned post in source feed', async () => {
     await con.getRepository(Post).update({ id: 'p5' }, { banned: true });
     const res = await client.query(QUERY('b'));
-    expect(res.data).toMatchSnapshot();
+    expect(
+      res.data.sourceFeed.edges.some(({ node }) => node.id === 'p5'),
+    ).toBeTruthy();
   });
 
   it('should display a welcome post first in source feed', async () => {
@@ -727,7 +731,9 @@ describe('query sourceFeed', () => {
       .getRepository(Post)
       .update({ id: 'p5' }, { banned: true, sourceId: 'community' });
     const res = await client.query(QUERY('community'));
-    expect(res.data).toMatchSnapshot();
+    expect(
+      res.data.sourceFeed.edges.every(({ node }) => node.id !== 'p5'),
+    ).toBeTruthy();
   });
 
   it('should throw an error when accessing private source', async () => {
@@ -749,7 +755,9 @@ describe('query sourceFeed', () => {
       },
     ]);
     const res = await client.query(QUERY('b'));
-    expect(res.data).toMatchSnapshot();
+    res.data.sourceFeed.edges.forEach(({ node }) =>
+      expect(node.source.id).toEqual('b'),
+    );
   });
 });
 
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -677,7 +677,7 @@ describe('query sourceFeed', () => {
     await con
       .getRepository(Post)
       .update({ id: 'p5' }, { type: PostType.Welcome });
-    const res2 = await client.query(QUERY('b'));
+    const res2 = await client.query(QUERY('b', Ranking.TIME));
     expect(res2.data.sourceFeed.edges[0].node.id).toEqual('p5');
   });
 
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -1,4 +1,4 @@
-import { feedToFilters, Ranking } from '../src/common';
+import { createSquadWelcomePost, feedToFilters, Ranking } from '../src/common';
 import {
   AdvancedSettings,
   ArticlePost,
@@ -16,6 +16,7 @@ import {
   SharePost,
   Source,
   SourceMember,
+  SourceType,
   User,
   View,
 } from '../src/entity';
@@ -51,6 +52,9 @@ import { DataSource } from 'typeorm';
 import createOrGetConnection from '../src/db';
 import { randomUUID } from 'crypto';
 import { usersFixture } from './fixture/user';
+import { base64, unbase64 } from 'graphql-relay/utils/base64';
+import { addYears } from 'date-fns';
+import { changeYearToNextYear } from '../src/common/date';
 
 let app: FastifyInstance;
 let con: DataSource;
@@ -654,8 +658,9 @@ describe('query sourceFeed', () => {
     ranking: Ranking = Ranking.POPULARITY,
     now = new Date(),
     first = 10,
+    after = '',
   ): string => `{
-    sourceFeed(source: "${source}", ranking: ${ranking}, now: "${now.toISOString()}", first: ${first}, supportedTypes: ["article", "share", "welcome"]) {
+    sourceFeed(source: "${source}", ranking: ${ranking}, now: "${now.toISOString()}", first: ${first}, supportedTypes: ["article", "share", "welcome"], after: "${after}") {
       ${feedFields()}
     }
   }`;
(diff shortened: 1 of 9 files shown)


src/schema/feeds.ts
```diff
@@ -829,7 +830,14 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       (ctx, { source }: SourceFeedArgs, builder, alias) =>
         sourceFeedBuilder(ctx, source, builder, alias),
       feedPageGenerator,
-      applyFeedPaging,
+      (ctx, args, page, builder, alias) => {
+        const qb = builder.orderBy(
+          `CASE WHEN ${alias}.type = '${PostType.Welcome}' THEN 1 ELSE 0 END`,
+          'DESC', // this will be replaced once we introduce the column named `pinned` on v6.
+        );
```
sshanzel, line 837: The DR has not mentioned a new property to identify whether a post is pinned, understandably since pinning is introduced only at v6. So this boolean identifier will be replaced by that new property.

Though I would like to get some thoughts if anyone wants to introduced `pinned` property now or wait until v6.

src/common/post.ts
```diff
@@ -80,3 +83,52 @@ export const fetchLinkPreview = async (
...
+  con: DataSource | EntityManager,
+  source: SquadSource,
+  adminId: string,
+) => {
+  const content = getWelcomeContent(source.name);
+  const repo = con.getRepository(WelcomePost);
+  const id = await generateShortId();
+
+  return repo.save(
+    repo.create({
```
idoshamun, line 122: we don't need repo.create really. You can pass the object directly to save

src/common/post.ts
```diff
@@ -80,3 +83,52 @@ export const fetchLinkPreview = async (
...
+      id,
+      shortId: id,
+      title: WELCOME_POST_TITLE,
+      sourceId: source.id,
+      authorId: adminId,
+      content,
+      contentHtml: markdown.render(content),
+      image: defaultImage.welcomePost,
+      banned: true,
```
idoshamun, line 131: Please confirm that it will not yield reputation penalization

src/common/post.ts
```diff
@@ -80,3 +83,52 @@ export const fetchLinkPreview = async (
...
+export const createSquadWelcomePost = async (
+  con: DataSource | EntityManager,
+  source: SquadSource,
+  adminId: string,
+) => {
+  const content = getWelcomeContent(source.name);
+  const repo = con.getRepository(WelcomePost);
+  const id = await generateShortId();
+
+  return repo.save(
```
idoshamun, line 121: I think you need to add the visibility props as well

src/schema/feeds.ts
idoshamun: It's not the right place to put the `orderBy` in. Every feed builder provides a page generator that adds ordering to the query. In this case, we use `feedPageGenerator`. You need to create a new one that will combine pins and creation time

src/schema/sources.ts
```diff
@@ -972,12 +976,15 @@ export const resolvers: IResolvers<any, Context> = {
             memberPostingRank: sourceRoleRank[memberPostingRole],
             memberInviteRank: sourceRoleRank[memberInviteRole],
           });
+          await repo.insert(source);
           // Add the logged-in user as admin
+          await createSquadWelcomePost(entityManager, source, ctx.userId);
```
idoshamun, line 981: I think it's better to put it after adding new source member just to keep it logically right.
Create source -> Add member -> Member publishes welcome post

src/schema/feeds.ts
rebelchris: Just a note bear in mind that pinning vs welcome post pinning might be different ranking still.
So possibly keep it separate until we start v6 for now :)

sshanzel: > Every feed builder provides a page generator that adds ordering to the query.

Yes, that's why I added the `order by` on the paging function of the `sourceFeed` resolver only, so no other feed is affected. But yeah, we can create a new `feedPageGenerator` to make it predictable and reusable.

sshanzel: > So possibly keep it separate until we start v6 for now :)

Awesome, thanks for letting me know 🚢

src/schema/sources.ts
sshanzel: Ah, yes, that makes sense. Will update quickly 🚀

src/schema/feeds.ts
sshanzel: @idoshamun just wanted to mention that `order by` commands can be stacked (like it was used here), the final query would contain comma-separated columns for the said command which it will first sort by the order of how it was declared.

So the order by priority would be then:
- pinned
- create date time

src/common/post.ts
```diff
@@ -80,3 +83,51 @@ export const fetchLinkPreview = async (
 
   throw new Error('Internal server error!');
 };
+
+export const WELCOME_POST_TITLE =
+  process.env.WELCOME_POST_TITLE ??
+  'Hello World! Start your squad journey here';
+
+const getWelcomeContent = (
+  squadName: string,
```
capJavert, line 92: Just to keep it extendable for the future maybe accept `SquadSource`, it can even be `Pick<SquadSource, 'name'>` since if we need any other property function will not change its signature.

capJavert (commented):
Non blocking, don't have anything else to add that others did not mention.

src/common/post.ts
```diff
@@ -1,14 +1,17 @@
...
 import { ValidationError } from 'apollo-server-errors';
 import { isValidHttpUrl } from './links';
+import { markdown } from './markdown';
+import { generateShortId } from '../ids';
 
 export const defaultImage = {
   urls: process.env.DEFAULT_IMAGE_URL?.split?.(',') ?? [],
   ratio: parseFloat(process.env.DEFAULT_IMAGE_RATIO),
   placeholder: process.env.DEFAULT_IMAGE_PLACEHOLDER,
+  welcomePost: process.env.WELCOME_POST_BANNER,
```
idoshamun, line 14: Please make sure that we use proper cloudinary transformations.
Can you drop here the link you use?

src/workers/postBannedRep.ts
```diff
@@ -12,6 +12,8 @@ interface Data {
   post: ChangeObject<Post>;
 }
 
+const punishableTypes = [PostType.Article];
```
rebelchris, line 15: What I see:

punisher

rebelchris (changes requested):
Looks great, just Ido's comment remains I see.

src/common/post.ts
sshanzel: It should be the link below:
https://daily-now-res.cloudinary.com/image/upload/f_auto/public/welcome_post

__tests__/feeds.ts
```diff
@@ -655,7 +655,7 @@ describe('query sourceFeed', () => {
     now = new Date(),
     first = 10,
   ): string => `{
-    sourceFeed(source: "${source}", ranking: ${ranking}, now: "${now.toISOString()}", first: ${first}) {
+    sourceFeed(source: "${source}", ranking: ${ranking}, now: "${now.toISOString()}", first: ${first}, supportedTypes: ["article", "share", "welcome"]) {
```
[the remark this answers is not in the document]
sshanzel: > If we support freeForm won't it support both out of the box already?

Yes, it will be just a matter of adding another parameter for supported types.

src/common/post.ts
idoshamun: Let's use this: https://daily-now-res.cloudinary.com/image/upload/f_auto,q_auto/public/welcome_post

src/common/post.ts
```diff
@@ -80,3 +83,57 @@ export const fetchLinkPreview = async (
...
+    id,
+    shortId: id,
+    title: WELCOME_POST_TITLE,
+    sourceId: source.id,
+    authorId: adminId,
+    content,
+    contentHtml: markdown.render(content),
+    image: defaultImage.welcomePost,
+    banned: true,
+    visible: true,
```
idoshamun, line 136: This is also needed
`visibleAt: new Date()`,

src/schema/feeds.ts
```diff
@@ -686,6 +688,66 @@ const applyFeedPaging = (
...
+
+const applyFeedPagingWithPin = (
+  ctx: Context,
+  page: FeedPage,
+  builder: SelectQueryBuilder<Post>,
+  alias,
+): SelectQueryBuilder<Post> => {
+  const priority = `CASE WHEN ${alias}.type = '${PostType.Welcome}' THEN 1 ELSE 0 END`;
+  let newBuilder = builder
+    .addSelect(`${alias}."createdAt" as "createdAt"`)
```
idoshamun, line 727: Let's add `createdAt` as required field in GraphORM instead. I want to avoid selecting the same field twice in this case.

src/schema/feeds.ts
```diff
@@ -686,6 +688,66 @@ const applyFeedPaging = (
...
+    .limit(page.limit)
+    .orderBy(`priority`, 'DESC')
+    .addOrderBy(`${alias}."createdAt"`, 'DESC');
+
+  if (page.timestamp) {
+    // to ensure pinned post (or welcome post) be pinned on top of the feed, we need to ensure they have greater date than the rest of the result
+    // that is solved by making the year of pinned posts to the next year and since we retained the other parts of the date, sorting should be accurate.
+    // the current condition is about being a welcome post, but on v6 this will be based on the `pinned` property.
+    const nextYearDifference = `(DATE_PART('year', CURRENT_DATE) - DATE_PART('year', ${alias}."createdAt") + 1)`;
+    newBuilder = newBuilder.andWhere(
```
idoshamun, line 738: My only concern here is whether this query still uses the `createdAt` index we have. But I guess we can check it after we deploy. But we need to be mindful to the fact it may ignore indexes

src/schema/feeds.ts
sshanzel: Just checked the graphorm, looks to have the `createdAt` already, I'll try to remove this.

src/common/post.ts
```diff
@@ -1,14 +1,28 @@
...
+import { generateShortId } from '../ids';
+
+const defaultWelcomeBanner =
+  'https://daily-now-res.cloudinary.com/image/upload/f_auto,q_auto/public/welcome_post';
 
 export const defaultImage = {
   urls: process.env.DEFAULT_IMAGE_URL?.split?.(',') ?? [],
   ratio: parseFloat(process.env.DEFAULT_IMAGE_RATIO),
   placeholder: process.env.DEFAULT_IMAGE_PLACEHOLDER,
+  welcomePost: process.env.WELCOME_POST_BANNER ?? defaultWelcomeBanner,
```
idoshamun, line 25: There's no need to get it from env var and have it hard coded. Let's choose one :)

src/schema/feeds.ts
sshanzel: Good point. Let's closely monitor the performance difference vs the previous one once merged.

src/common/post.ts
sshanzel: Cool. Let me update that one.

src/schema/feeds.ts
sshanzel: Also, the current implementation forces ranking by `time`, which is basically our FE implementation. Would you want to open the opportunity to request for `popularity` based for source feed query? That will be another formula to make for adjusting the scoring system. Or we can tackle it when the time comes?

idoshamun: We can tackle when the time comes. Hopefully we will migrate this responsibility to the feed service anyway

idoshamun (approved):

sshanzel:
@rebelchris would love to hear what you think! 🚀

sshanzel:
@rebelchris @capJavert would love to hear what you both think! 🚀

src/schema/feeds.ts
```diff
@@ -686,6 +688,65 @@ const applyFeedPaging = (
...
+  let newBuilder = builder
+    .addSelect(priority, 'priority')
+    .limit(page.limit)
+    .orderBy(`priority`, 'DESC')
+    .addOrderBy(`${alias}."createdAt"`, 'DESC');
+
+  if (page.timestamp) {
+    // to ensure pinned post (or welcome post) be pinned on top of the feed, we need to ensure they have greater date than the rest of the result
+    // that is solved by making the year of pinned posts to the next year and since we retained the other parts of the date, sorting should be accurate.
+    // the current condition is about being a welcome post, but on v6 this will be based on the `pinned` property.
```
capJavert, line 735: So the next year hack is just temporary for v5?

capJavert (approved):
Nice work, one question for my understanding.

src/schema/feeds.ts
sshanzel: Yes, this will use the pinned property for the condition once we introduce it in v6.

capJavert: Cool, tnx for clarification.

sshanzel: Though we will keep the next year mutation until we find a better solution for pagination including two properties.

PR #1261, opened 2023-05-07
refactor: temporarily disable subscribe to notification

sshanzel:
Until we do the retro check for those who have already subscribed, we don't want to show it to those who have existing subscriptions.

This also got me thinking after we do the retro checks for completed actions specifically the first post, should we also send a notification for those unsubscribed for PN yet?


idoshamun:
Unsubscribing is currently an edge case. We can ignore it but well thought :)

idoshamun (approved):

PR #1262, opened 2023-05-08
feat: edit freeform post and welcome post

sshanzel:
The mutation is all about editing a freeform post, but since the welcome post is a type of freeform post, we can utilize the same mutation for updating the welcome post itself. There are just a few differences where title is not possible to be updated.

TODO:
- ~~CDC listener on this table then publish an event to notify mentioned users. Awaiting Sab's response as this would require copy and design references.~~


src/schema/posts.ts
```diff
@@ -859,6 +925,80 @@ export const resolvers: IResolvers<any, Context> = {
...
+            'Content has a maximum length of 80 characters',
+          );
+        }
+
+        const updated: Partial<EditablePost> = {};
+
+        if (post.type === PostType.Freeform) {
+          updated.title = title;
+
+          if (image && process.env.CLOUDINARY_URL) {
```
capJavert, line 975: Not blocking but just checking on how do we validate required `process.env` in general? Should we just throw here or at least log that `CLOUDINARY_URL` is missing. Looks like something that could silently happen and it would require quite some times to figure out in production.

capJavert (commented):
Some general comments here to consider since it still has some TODOs. Overall nice chunk of work. 👍

src/migration/1683523942931-PostMentionEntity.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class PostMentionEntity1683523942931 implements MigrationInterface {
+  name = 'PostMentionEntity1683523942931';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "post_mention" ("postId" character varying(14) NOT NULL, "mentionedByUserId" character varying(36) NOT NULL, "mentionedUserId" character varying(36) NOT NULL, CONSTRAINT "PK_36d21500cf18ea0e76c661ca32f" PRIMARY KEY ("postId", "mentionedByUserId", "mentionedUserId"))`,
+    );
+  }
```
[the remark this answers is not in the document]
sshanzel: Which specific constraint do you mean?

The current one already states all three of the column as below:
```
CONSTRAINT "PK_36d21500cf18ea0e76c661ca32f" PRIMARY KEY ("postId", "mentionedByUserId", "mentionedUserId")
```

src/schema/posts.ts
```diff
@@ -859,6 +925,80 @@ export const resolvers: IResolvers<any, Context> = {
...
+          throw new ValidationError(
+            'Title has a maximum length of 80 characters',
+          );
+        }
+
+        if (content.length > MAX_CONTENT_LENGTH) {
+          throw new ValidationError(
+            'Content has a maximum length of 80 characters',
+          );
+        }
```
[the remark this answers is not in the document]
sshanzel: Not sure, we have a specific rule for spacing between the contents. This was just a general trim so empty space won't work.

src/schema/posts.ts
sshanzel: We don't have a staging environment to test this out so we manually test this on prod. This is an existing implementation for the user update profile, etc., so if one does not work, it won't across the board.

src/schema/posts.ts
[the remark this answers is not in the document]
sshanzel: Not sure, we have a specific rule for spacing between the contents. This was just a general trim so empty space won't work.

src/migration/1683523942931-PostMentionEntity.ts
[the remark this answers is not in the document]
capJavert: Sorry, I think `FOREIGN KEY` constrains, to be more specific:
- `postId` to be actually be valid id from `post` table
- `mentionedUserId` to be valid id from `user` table
- `commentByUserId` to be valid id from `user` table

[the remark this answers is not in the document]
capJavert: Sorry, I think `FOREIGN KEY` constrains, to be more specific:
- `postId` to be actually be valid id from `post` table
- `mentionedUserId` to be valid id from `user` table
- `commentByUserId` to be valid id from `user` table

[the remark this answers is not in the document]
sshanzel: We've decided to create a new table to support the post mention but we wouldn't want to limit anything if something arises in the future that requires similar work.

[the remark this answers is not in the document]
capJavert: Seen the thread on slack 👍 .. I don't see what would be the limiting factor here though.. seems logical to add foreign keys as we do in other places.

src/schema/posts.ts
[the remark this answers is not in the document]
capJavert: We should thing about it, up until this point our users mostly wrote few words or a sentence to accompany the link or write a short comment but with free form I would expect much more different content and hacks like above are quite common in rich text editing, especially in v6.

src/schema/posts.ts
capJavert: Don't mind it for this PR, your implementation is fine for sure.. just something that caught my eye, maybe a backlog candidate.

src/migration/1683523942931-PostMentionEntity.ts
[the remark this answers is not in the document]
rebelchris: For the comment mention table we do use individual constraints to link it to user table though.
Should those not be added?

See CommentMention migration

src/schema/posts.ts
```diff
@@ -508,6 +523,28 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      Avatar image for the squad
+      """
+      image: Upload
+      """
+      Title of the post (max 80 chars)
+      """
+      title: String
+      """
+      Content of the post (max 4000 chars)
```
rebelchris, line 543: Where was this limit set from actually?

src/schema/posts.ts
```diff
@@ -859,6 +925,80 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const { id, image } = args;
+      const { con, userId } = ctx;
+      const title = args.title?.trim() ?? '';
+      const content = args.content?.trim() ?? '';
+
+      await con.transaction(async (manager) => {
+        const repo = manager.getRepository(Post);
+        const post = (await repo.findOneByOrFail({ id })) as
+          | WelcomePost
+          | FreeformPost;
```
rebelchris, line 942: Does this not force cast it to these types?

src/schema/posts.ts
```diff
@@ -859,6 +925,80 @@ export const resolvers: IResolvers<any, Context> = {
...
+              ? SourcePermissions.WelcomePostEdit
+              : SourcePermissions.Post;
+          await ensureSourcePermissions(ctx, post.sourceId, permission);
+        }
+
+        if (!editablePostTypes.includes(post.type)) {
+          throw new ForbiddenError(
+            'Editing post outside of type welcome_post and freeform is not allowed',
+          );
+        }
```
rebelchris, line 956: I would maybe do this one before the permissions to avoid a lookup

src/schema/posts.ts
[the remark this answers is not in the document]
rebelchris: I agree maybe we should use trimAll or regex variant of it.
But can also be introduced in v6 (so not urgent)

src/common/cloudinary.ts
```diff
@@ -53,3 +53,8 @@ const avatarPreset = 'avatar';
 
 export const uploadAvatar = (userId: string, stream: Readable) =>
   uploadFile(`${avatarPreset}_${userId}`, avatarPreset, stream);
+
+const postPreset = 'post';
+
+export const uploadPostImage = (postId: string, stream: Readable) =>
```
rebelchris, line 59: @idoshamun Is there anything here we need to keep in mind regards to cloudinary rules?
Having this cost issue in mind.

rebelchris (changes requested):
Neat! Some minor remarks, but almost good to go from my side.

src/schema/posts.ts
sshanzel: Definitely, also that we don't have any setup for at least mocking this in our test cases.

src/schema/posts.ts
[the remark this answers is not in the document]
sshanzel: There can be numerous ways to tackle this and must be appropriate, like 2 line breaks are max, etc. - we will need to come up with a rule. I think we can apply it on the v6, this update content would only be accessible to admins and moderators, I don't think they would be putting annoying empty spaces for their newcomers.

src/migration/1683523942931-PostMentionEntity.ts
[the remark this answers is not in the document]
sshanzel: Yep, user FKs seem reasonable. Let's keep those.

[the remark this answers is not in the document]
sshanzel: Just the fact that mentioning squad might come into place or something.

src/common/cloudinary.ts
sshanzel: I agree, also that in v6 we will allow everyone to upload images which we must monitor the usage.

src/schema/posts.ts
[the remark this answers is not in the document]
capJavert: Agreed, let's not block due to this but for sure include it in v6 🚀

src/common/cloudinary.ts
idoshamun: @rebelchris this is a great catch! @sshanzel please talk to Denis to understand what transformations and upload preset he applies to make sure we're aligned on this. Please tag me on the relevant lines once you apply that. Until that let's block the PR

src/schema/posts.ts
sshanzel: We fetch through `Post` entity, so it can be either of the 4, this is so we can fetch both types where we allow the editting.

src/common/cloudinary.ts
sshanzel: Code used as inspiration based on the discussion:
https://github.com/dailydotdev/gatekeeper/blob/48842a882167a6e61d82a6e1e74e8c191c73d8b8/src/common/cloudinary.ts#L17

Relevant change:
https://github.com/dailydotdev/daily-api/pull/1262/files#diff-44c49f4dbf41a19d0e49a6f2c185d26132919a5e2b76d00c0a75fa7de72d9370R59-R84

Relevant Slack discussion:
https://dailydotdev.slack.com/archives/C03C3JZPRU4/p1683634554381399?thread_ts=1683633804.026429&cid=C03C3JZPRU4

__tests__/posts.ts
```diff
@@ -2138,3 +2141,207 @@ describe('mutation checkLinkPreview', () => {
...
+    loggedUser = '2';
+
+    return testMutationErrorCode(
+      client,
+      { mutation: MUTATION, variables: { id: 'p1', title: 'Test' } },
+      'FORBIDDEN',
+    );
+  });
+
+  it('should not update title if the post type is "welcome"', async () => {
```
idoshamun, line 2261: Are you sure welcome post title cannot be edited?

__tests__/posts.ts
```diff
@@ -2138,3 +2141,207 @@ describe('mutation checkLinkPreview', () => {
...
+    const title = 'Updated title';
+    const res = await client.mutate(MUTATION, {
+      variables: { id: 'p1', title },
+    });
+    expect(res.errors).toBeFalsy();
+    const actual = await con.getRepository(Post).findOneBy({ id: 'p1' });
+    expect(actual.title).toEqual(title);
+  });
+
+  it('should allow moderator to do update even when user is not the author of the post', async () => {
```
idoshamun, line 2286: Are you sure mods and admins can edit my post?

src/common/cloudinary.ts
```diff
@@ -53,3 +53,35 @@ const avatarPreset = 'avatar';
...
+        public_id: name,
+        upload_preset: postPreset,
+      },
+      (err, callResult) => {
+        if (err) {
+          return reject(err);
+        }
+
+        return resolve(
+          cloudinary.v2.url(callResult.public_id, {
```
idoshamun, line 75: Can you also check if you can provide `sign_url: true` in the params or something like that. I'd like to start using signed urls when possible

src/common/cloudinary.ts
```diff
@@ -53,3 +53,35 @@ const avatarPreset = 'avatar';
...
+            quality: 'auto',
+          }),
+        );
+      },
+    );
+    stream.pipe(outStream);
+  });
+
+export const uploadPostImage = (postId: string, stream: Readable) =>
+  uploadPostFile(`${postPreset}_${postId}`, stream);
```
idoshamun, line 87: `postId` should be enough as a name I believe.
One more question, we should support adding images to the content of the post not only the cover image. Is it covered here or in a different PR?

src/schema/posts.ts
```diff
@@ -621,6 +658,31 @@ export const typeDefs = /* GraphQL */ `
   }
 `;
 
+const saveMentions = (
```
idoshamun, line 661: I think we have a similar function for the comments, can't we generalize with function params perhaps?

src/schema/posts.ts
```diff
@@ -859,6 +925,80 @@ export const resolvers: IResolvers<any, Context> = {
...
+          if (image && process.env.CLOUDINARY_URL) {
+            const upload = await image;
+            updated.image = await uploadPostImage(
+              id,
+              upload.createReadStream(),
+            );
+          }
+        }
+
+        const mentions = await getMentions(
```
idoshamun, line 984: Shouldn't this be inside the if block?

idoshamun (changes requested):
That's a lot of work! 🤯
Still some polishing to do

__tests__/posts.ts
sshanzel: From the moderation spec, it did says it should. Let me confirm with Sab regarding freeform posts.

src/common/cloudinary.ts
sshanzel: It would be covered in the iteration of v6 where we will now allow uploading image content. As part of v5, it should not be the case yet. So in v6 we will introduce a new mutation for that case.

src/schema/posts.ts
sshanzel: Ah, right. I forgot we are doing a transaction, so even if I save the mentions first on the code they will still run at the same time.

__tests__/posts.ts
sshanzel: For the v5 iteration, it should not be the case yet.

capJavert (commented):
I don't see anything popping out aside from comments that Ido already made. 👍

sshanzel:
@rebelchris to keep this PRs scope smaller I think we can create a separate ticket for the new notification type since the sending of mutation is async and will only happen once we deploy the FE, so nothing is going to get missed. Still awaiting Sab's response to rectify some concerns about the supposed notification type. WDYT?

src/schema/posts.ts
```diff
@@ -508,6 +523,28 @@ export const typeDefs = /* GraphQL */ `
...
+      image: Upload
+      """
+      Title of the post (max 80 chars)
+      """
+      title: String
+      """
+      Content of the post (max 4000 chars)
+      """
+      content: String
+    ): EmptyResponse! @auth
```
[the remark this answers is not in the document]
sshanzel: Just wanted to lessen the query since the updates considered in this mutation will be based on all the inputs which the consuming end already has. With that, fetching one post shouldn't hurt the performance, let's return it as you suggested.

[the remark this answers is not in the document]
idoshamun: This is a valid way of thinking and I appreciate it. Bare in mind that the markdown and mentions are not provided as inputs so it will be missing

capJavert:
> @rebelchris to keep this PRs scope smaller I think we can create a separate ticket for the new notification type since the sending of mutation is async and will only happen once we deploy the FE, so nothing is going to get missed. Still awaiting Sab's response to rectify some concerns about the supposed notification type. WDYT?
> 
> cc @capJavert

Agreed 👍

rebelchris (approved):
Love this!
Great test cases as well and agree with taking notification separate, just make sure to add a task for it so we don't forget.

sshanzel:
@idoshamun let me know what you think 🥳

PR #1264, opened 2023-05-10
fix: add welcome posts to existing squads

rebelchris:
Retro command to add welcome post to all squads that don't have one yet.
Should be run once and can be deleted after.

WT-1320 #done


capJavert (commented):
General question, does this run in a transaction by default? I think it was noted in some of the previous version but latest docs specifically call `startTransaction` https://orkhan.gitbook.io/typeorm/docs/transactions#using-queryrunner-to-create-and-control-state-of-single-database-connection

idoshamun (approved):

idoshamun:
@capJavert we don't need transaction for that I believe. It's idempotency process. So if for some reason some fail we can run it again and no side effects will happen

PR #1265, opened 2023-05-10
fix: add dummy user and adjust deletion of shared posts

rebelchris:
Introduces dummy user that gets set for deleted users shared posts

Also perhaps we should consider adding this for welcome/freeform to?

WT-1317

### DIFF
--- a/src/directive/user.ts
+++ b/src/directive/user.ts
@@ -12,6 +12,7 @@ import {
   Post,
   PostReport,
   Settings,
+  SharePost,
   SourceDisplay,
   SourceRequest,
   Upvote,
@@ -41,6 +42,10 @@ export const deleteUser = async (
       await entityManager.getRepository(SourceDisplay).delete({ userId });
       await entityManager.getRepository(SourceRequest).delete({ userId });
       await entityManager.getRepository(Upvote).delete({ userId });
+      // Manually set shared post to 404 dummy user
+      await entityManager
+        .getRepository(SharePost)
+        .update({ authorId: userId }, { authorId: '404' });
       await entityManager
         .getRepository(Post)
         .update({ authorId: userId }, { authorId: null });
--- a/src/directive/user.ts
+++ b/src/directive/user.ts
@@ -13,7 +13,6 @@ import {
   Post,
   PostReport,
   Settings,
-  SharePost,
   SourceDisplay,
   SourceRequest,
   Upvote,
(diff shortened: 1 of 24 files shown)


src/migration/1683721525235-DummyUser.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DummyUser1683721525235 implements MigrationInterface {
+  name = 'DummyUser1683721525235';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `INSERT INTO "public"."user" ("id", "name", "image", "username") VALUES ('404', 'Inactive user', 'https://daily-now-res.cloudinary.com/image/upload/f_auto/placeholders/placeholder3', 'inactive_user')`,
```
idoshamun, line 8:
```suggestion
      `INSERT INTO "public"."user" ("id", "name", "image", "username") VALUES ('404', 'Inactive user', 'https://daily-now-res.cloudinary.com/image/upload/f_auto,q_auto/placeholders/placeholder3', 'inactive_user')`,
```

src/directive/user.ts
```diff
@@ -41,6 +42,10 @@ export const deleteUser = async (
       await entityManager.getRepository(SourceDisplay).delete({ userId });
       await entityManager.getRepository(SourceRequest).delete({ userId });
       await entityManager.getRepository(Upvote).delete({ userId });
+      // Manually set shared post to 404 dummy user
+      await entityManager
+        .getRepository(SharePost)
+        .update({ authorId: userId }, { authorId: '404' });
       await entityManager
```
idoshamun, line 49: This will override the above

src/directive/user.ts
```diff
@@ -41,6 +42,10 @@ export const deleteUser = async (
       await entityManager.getRepository(SourceDisplay).delete({ userId });
       await entityManager.getRepository(SourceRequest).delete({ userId });
       await entityManager.getRepository(Upvote).delete({ userId });
+      // Manually set shared post to 404 dummy user
+      await entityManager
```
idoshamun, line 46: I think we should apply it on every post that is no `ArticlePost`.
Maybe we should apply it also for comments?

rebelchris: Good point, can do that actually :) 
And also for comments seems like a good way to preserve conversations.

PR #1267, opened 2023-05-12
feat: new notification type for mention in content

sshanzel:
A new notification type for when a user is mentioned in the welcome post's content.

~~Still awaiting Sab's response on whether this should be a general notification for both freeform and welcome post content mention or if will there be separate ones.~~

Relevant discussion: https://dailydotdev.slack.com/archives/C0506KU4RH9/p1683521956748589

~~This might change a tiny bit based on what will be the response but wanted to prepare the nitty-gritty parts.~~

Note: this PR should be ready now.

I will raise the `streams` PR once we've concluded the naming here.


rebelchris (approved):
Nice one! I love it

src/notifications/generate.ts
```diff
@@ -145,6 +147,19 @@ export const generateNotificationMap: Record<
...
+  post_mention: (
+    builder,
+    ctx: NotificationPostContext & NotificationDoneByContext,
+  ) =>
+    builder
+      .referencePost(ctx.post)
+      .icon(NotificationIcon.Comment)
+      .description((ctx.post as WelcomePost).content)
+      .targetPost(ctx.post)
+      .setTargetUrlParameter([
```
idoshamun, line 159: This is a generic post mention notification but seems like the parameters here are specific to welcome post

src/workers/newNotificationMail.ts
```diff
@@ -476,6 +477,9 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
   demoted_to_member: async () => {
     return null;
   },
+  post_mention: async () => {
```
idoshamun, line 480: I think we should send an email on mention. Please double check

src/workers/notifications/postMention.ts
```diff
@@ -0,0 +1,39 @@
...
+import { NotificationWorker } from './worker';
+import { ChangeObject } from '../../types';
+import { buildPostContext } from './utils';
+
+interface Data {
+  postMention: ChangeObject<PostMention>;
+}
+
+const worker: NotificationWorker = {
+  subscription: 'api.comment-mention-notification',
```
idoshamun, line 16: Seems like the wrong name. Please also make sure to add this new worker to the index file and pulumi

src/workers/notifications/postMention.ts
```diff
@@ -0,0 +1,39 @@
...
+  handler: async (message, con) => {
+    const { postMention }: Data = messageToJson(message);
+    const { postId, mentionedByUserId, mentionedUserId } = postMention;
+    const postCtx = await buildPostContext(con, postId);
+
+    if (!postCtx) {
+      return;
+    }
+
+    const doneBy = await con
```
idoshamun, line 26: Need to make sure the user exists, otherwise skip

src/workers/newNotificationMail.ts
sshanzel: Yes, I believe there should be one. Let me check again with Sab if it is now ready.

src/notifications/generate.ts
```diff
@@ -145,6 +147,26 @@ export const generateNotificationMap: Record<
...
       .targetPost(ctx.post, ctx.comment)
       .avatarManyUsers([ctx.commenter]),
+  post_mention: (
+    builder,
+    ctx: NotificationPostContext & NotificationDoneByContext,
+  ) =>
+    builder
+      .referencePost(ctx.post)
+      .icon(NotificationIcon.Comment)
+      .description((ctx.post as WelcomePost).content)
```
idoshamun, line 157: content is too long here, shouldn't we truncate it or something?
Also to make it more generic please change the casting to FreeformPost

idoshamun (approved):
One last change, let's not block it because of the email and I'll let Sab know

src/notifications/generate.ts
sshanzel: Good point. I was thinking that this could already be truncated based on how long comments currently work. But will double-check that one.

idoshamun: Yes you can apply the same mechanism for sure :)

idoshamun:
Also, still waiting for the streams PR. We can't merge before that

idoshamun (approved):

PR #1268, opened 2023-05-12
fix: do not perform update if no post properties changed

capJavert:
Without this mutation throws unexpected error when you just press save in the UI with the same post content since updated object is empty on the API side and typeorm does not like that.


rebelchris (approved):
Ah first time seeing this, nice spot.

sshanzel (commented):
Great catch! Though I think we should add this to the `errorFormatter` in the `index` file to make it easier to understand in the FE as I see this can occur in the future again.

idoshamun (commented):
Should we consider a test case to make sure we don't throw an exception?

idoshamun (approved):

rebelchris (approved):

PR #1274, opened 2023-05-16
refactor: stop sending notification on welcome post to members

sshanzel:
If the post is a welcome post, we should not send a notification to users by returning early on the postAdded worker.


rebelchris (approved):
Looks good :)

idoshamun (approved):

idoshamun:
@sshanzel can you just add a test case please?

sshanzel:
> @sshanzel can you just add a test case please?

Right. We should definitely have one. Coming right up!

PR #1275, opened 2023-05-16
feat: pinned at property on post 

sshanzel:
- Creates a new column on the Post entity.
- Updates existing welcome posts to be pinned
- Updated the function to make new welcome posts automatically pinned.

WT-1343 #done


src/entity/posts/Post.ts
```diff
@@ -149,6 +149,10 @@ export class Post {
   @Column({ default: null })
   visibleAt: Date;
 
+  @Column({ default: null })
+  @Index()
```
idoshamun, line 153: I think the right index will be a composite index that involves sourceId, pinnedAt, createdAt.
Because this is gonna be they way we will filter on this property. We will check for all posts of a given source id and then sort by pinned and created at

idoshamun:
@sshanzel I think we should update `pinnedAt` for welcome posts in the migration right?

src/entity/posts/Post.ts
sshanzel: Good point. Do we have any pre-existing samples for the composite index? I tried checking the docs I didn't find any relevant steps. Shouldn't having them all with `Index` work, out of the box?

rebelchris (approved):
Should we set existing welcome to be pinnedAt now?
And introduce default for welcome post to be `pinnedAt` as well?

(Can also be a seperate PR if you want)

sshanzel:
> Should we set existing welcome to be pinnedAt now?
> And introduce default for welcome post to be `pinnedAt` as well?

That's already part of this PR I believe. The migration updates existing welcome posts. Then the new ones are set through the update at `/common/post` file.

src/entity/posts/Post.ts
idoshamun: Adding `Index` to every column will create separate indexes.
You can add something like this: `@Index('IDX_notification_user_id_created_at', ['userId', 'public', 'createdAt'])`
above the class definition

sshanzel: Wow. That's the first time I got to know this. Thanks. I have applied it now.

Do we keep the standalone index of the `pinnedAt` like the others? Or we should remove it?

idoshamun (approved):

PR #1277, opened 2023-05-16
refactor: source feed to utilize pinned at property

sshanzel:
We don't have to use the case when statement anymore since we can now just use the `pinnedAt` property.

Have refactored the previous hack we had for pinned post where we update the date to be next year. Instead, we passed the relevant parameters in order to generate the right query in a more stable way.

WT-1344 #done


bin/completedUserActions.ts
```diff
@@ -108,6 +108,4 @@ export const retroCheckActions = async (ds?: DataSource): Promise<void> => {
   console.log('finished retro checking');
 };
 
-if (process.env.NODE_ENV !== 'test') {
-  retroCheckActions();
-}
+// retroCheckActions();
```
rebelchris, line 111: This shouldn't be here?

src/schema/feeds.ts
```diff
@@ -695,21 +695,39 @@ const feedPageGeneratorWithPin: PageGenerator<
   unknown
 > = {
   connArgsToPage: ({ first, after }: FeedArgs) => {
-    const cursor = getCursorFromAfter(after);
+    const unbased = unbase64(after);
```
rebelchris, line 698: Why do you need all of this?
Not really sure I get this, a simple datetime sort is enough right?

sshanzel: We need the reference to the pinned date as the sorting is also used in this query, and it is required to be part of the condition for the next pages.

bin/completedUserActions.ts
sshanzel: Right. Let me revert this one.

__tests__/__snapshots__/feeds.ts.snap
```diff
@@ -2496,11 +2496,33 @@ Object {
 }
 `;
 
+exports[`query sourceFeed should display return the right posts after the first page being a mix of pinned and unpinned posts 1`] = `
```
sshanzel, line 2499: The mix ones are the first page, not necessarily the result here which is the second page, but we wanted to see that it will return the correct second page which should be unpinned since the last item of the mixed first page is unpinned.

idoshamun (changes requested):
I think the approach here should be slightly different.
When you use both the pin time and the created at time to generate a cursor it may create an edge case.
Let's take an example:
Post A: Pinned at 17/5/23 but created at 10/5/23
Post B: Created at 16/5/23
In the following implementation if you provide to the query Post A's cursor, you will not get Post B in the response as it will ask for posts created before 10/5/23.
If the post is pinned we should ignore its timestamp in the pagination. I suggest also to have proper test cases to check it.
Other test cases may involve that Post B created at 18/5/23 (after the pinned time) and we want to make sure it returns on the next page.

sshanzel:
> If the post is pinned we should ignore its timestamp in the pagination. I suggest also to have proper test cases to check it.

Yes, this is what we are already doing by returning early and not going through the `timestamp` condition.

> Other test cases may involve that Post B created at 18/5/23 (after the pinned time)

But it should not be possible to be created after pinned time since created time will always come first, right?

sshanzel:
Have fixed the test case now. Also refactored for cleaner and reusable function but the result should still be the same 🚀

src/common/pagination.ts
```diff
@@ -33,3 +33,25 @@ export const getCursorFromAfter = (after?: string): string => {
   }
   return unbase64(after).split(':')[1];
 };
+
+export const getArgsFromAfter = <
```
idoshamun, line 37: Love it!

idoshamun (approved):
Awesome! Looks great

PR #1278, opened 2023-05-16
chore: stop running unnecessary commands

sshanzel:
I noticed our test runners are executing the `completedUserActions` file.

Screenshot 2023-05-16 at 11 31 49 PM

### DIFF
--- a/__tests__/workers/completedUserActions.ts
+++ b/__tests__/workers/completedUserActions.ts
@@ -263,3 +263,7 @@
 //     expect(completed).toBeTruthy();
 //   });
 // });
+
+it('should contain no test', () => {
+  expect(true).toBeTruthy();
+});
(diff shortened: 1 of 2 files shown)


idoshamun (commented):
Is this the right solution?
Basically it will make the script not work in case we want to run it.
Pretty sure there's another reason for that. Wdyt?

rebelchris:
@idoshamun should we actually not simply remove them?
They should only have had this one use and it can only cause confusion down the line maybe?

sshanzel:
> Pretty sure there's another reason for that. Wdyt?

I realized we have a dedicated test for it. Maybe we should actually delete those files as Chris suggested?

idoshamun:
I usually tend to keep utility files but it's your call you can either remove them or disable their tests. I don't have strong opinion here

sshanzel:
> I usually tend to keep utility files but it's your call you can either remove them or disable their tests. I don't have strong opinion here

Let's disable the test, when I created a new bin file, I referenced old implementations which can probably useful to new comers in the future.

idoshamun (approved):

rebelchris (approved):

PR #1281, opened 2023-05-18
refactor: edit mutation to update title of welcome post

sshanzel:
This is to allow the welcome post's title to be updated as part of v6.


src/schema/posts.ts
```diff
@@ -1028,23 +1028,23 @@ export const resolvers: IResolvers<any, Context> = {
 
         if (content.length > MAX_CONTENT_LENGTH) {
           throw new ValidationError(
-            'Content has a maximum length of 80 characters',
+            'Content has a maximum length of 4000 characters',
```
rebelchris, line 1031: Title also has a max length perhaps we should also check that here?

rebelchris (changes requested):
Just one suggestion otherwise looks amazing

src/schema/posts.ts
sshanzel: There is, we are already checking it from 1024. I just fixed the typo here.

rebelchris (approved):
Looks great

idoshamun (approved):
Awesome!

PR #1282, opened 2023-05-19
refactor: stop using active post view

idoshamun:
Whenever we introduce a schema change to Post it required to replace the view which causes deadlocks. We introduce a new feature to GraphORM that simulates the view functionality.


idoshamun:
@rebelchris we should. I should have mentioned that it's the next step. Otherwise we will introduce a downtime or maybe even won't be able to pass the migration step due to the deadlocks

PR #1285, opened 2023-05-19
fix: added a generic content-published creation subscriber

rebelchris:
In favor of individual publishing methods we want to streamline content via one channel `content-published` this can mean either creating/updating content or rejecting content (community picks)

Please take some time to go over the PR as it introduces a lot of changes, and would love to hear if you think we can make it a bit more abstracted in places.
(My vision was to make as much re-usable - independant of update or create)

WT-1377

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -101,13 +101,13 @@ const wsMemory = 3072;
 const wsLimits: pulumi.Input<{
   [key: string]: pulumi.Input<string>;
 }> = {
-  cpu: '1',
+  cpu: '500m',
   memory: `${wsMemory}Mi`,
 };
 
 const bgLimits: pulumi.Input<{
   [key: string]: pulumi.Input<string>;
-}> = { cpu: '500m', memory: '256Mi' };
+}> = { cpu: '250m', memory: '256Mi' };
 
 const readinessProbe: k8s.types.input.core.v1.Probe = {
   httpGet: { path: '/health', port: 'http' },
@@ -168,7 +168,7 @@ if (isAdhocEnv) {
       limits,
       readinessProbe,
       livenessProbe,
-      metric: { type: 'memory_cpu', cpu: 80 },
+      metric: { type: 'memory_cpu', cpu: 85 },
       createService: true,
       enableCdn: true,
       disableLifecycle: true,
@@ -185,7 +185,7 @@ if (isAdhocEnv) {
       limits: wsLimits,
       readinessProbe,
       livenessProbe,
-      metric: { type: 'memory_cpu', cpu: 80 },
+      metric: { type: 'memory_cpu', cpu: 85 },
       disableLifecycle: true,
     },
     {
@@ -212,7 +212,7 @@ if (isAdhocEnv) {
       limits,
       readinessProbe,
       livenessProbe,
-      metric: { type: 'memory_cpu', cpu: 80 },
+      metric: { type: 'memory_cpu', cpu: 85 },
       createService: true,
       serviceType: 'ClusterIP',
       disableLifecycle: true,
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -2267,27 +2267,27 @@ describe('mutation editPost', () => {
     );
   });
 
-  it('should not update title if the post type is "welcome"', async () => {
-    loggedUser = '1';
-    const title = 'Updated title';
-    const res = await client.mutate(MUTATION, {
-      variables: { id: 'p1', title },
-    });
-    expect(res.errors).toBeFalsy();
-    expect(res.data.editPost.title).not.toEqual(title);
-  });
-
-  it('should update title if the post type is "freeform"', async () => {
+  it('should update title of the post if it is either freeform or welcome post', async () => {
     loggedUser = '1';
     await con
       .getRepository(Post)
       .update({ id: 'p1' }, { type: PostType.Freeform });
     const title = 'Updated title';
-    const res = await client.mutate(MUTATION, {
+    const res1 = await client.mutate(MUTATION, {
       variables: { id: 'p1', title },
     });
-    expect(res.errors).toBeFalsy();
-    expect(res.data.editPost.title).toEqual(title);
+    expect(res1.errors).toBeFalsy();
+    expect(res1.data.editPost.title).toEqual(title);
+
+    await con
+      .getRepository(Post)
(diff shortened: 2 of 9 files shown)


src/workers/postUpdated.ts
```diff
@@ -44,26 +51,29 @@ const worker: Worker = {
...
+      const { reject_reason, post_id, updated_at, submission_id } = data;
       await con.transaction(async (entityManager) => {
-        // For now, we only allow Squad posts to be updated through this flow
-        const databasePost = await entityManager
-          .getRepository(ArticlePost)
-          .findOneBy({ id: post_id, origin: PostOrigin.Squad });
+        if (reject_reason) {
+          // Check if we have a submission id, we need to notify
+          if (!submission_id) {
+            logger.info({ data }, 'received rejection without submission id');
```
idoshamun, line 60: I think we should return at this point, right?

src/workers/postUpdated.ts
```diff
@@ -44,26 +51,29 @@ const worker: Worker = {
...
-        if (
-          !databasePost ||
-          databasePost.metadataChangedAt.toISOString() >=
-            updatedDate.toISOString()
-        ) {
+          const submissionRepo = con.getRepository(Submission);
+          const submission = await submissionRepo.findOneBy({
+            id: submission_id,
+          });
+          if (submission.status === SubmissionStatus.Started) {
```
idoshamun, line 67: Just in case the submission doesn't exist

```suggestion
          if (submission?.status === SubmissionStatus.Started) {
```

src/workers/postUpdated.ts
```diff
@@ -82,8 +92,10 @@ const worker: Worker = {
             : data?.extra?.creator_twitter;
 
         const authorId = await findAuthor(entityManager, creatorTwitter);
-        const title = data?.title || databasePost.title;
-        const becomesVisible = !!title?.length;
+
+        const { private: privacy } = await entityManager
```
idoshamun, line 96: Let's select only the relevant field then. It's a minor optimization but I want us to get used to selecting relevant fields when possible :)

idoshamun (changes requested):
I added some comments but overall I think we need to refactor `postUpdated` worker and add more functions. Right now it's difficult to read as it's one big functions.
Let's separate it to different cases and utility functions that will make it more readable.
The handler function should be a high level execution of a workflow. Primarily include function calls and some conditions. This way it will be easier to understand the flow and then deep dive to specifics

src/workers/postUpdated.ts
```diff
@@ -75,76 +292,49 @@ const worker: Worker = {
...
+          // Handle creation of new post
+          try {
+            combinedPost = await createPost({
+              logger,
+              entityManager,
+              data: fixedData,
+              submission_id: data?.submission_id,
+            });
+          } catch (e) {
+            return;
```
idoshamun, line 313: Why do we ignore all errors?

src/workers/postUpdated.ts
```diff
@@ -38,33 +47,241 @@ interface Data {
...
-        }
-
-        if (
-          !databasePost ||
-          databasePost.metadataChangedAt.toISOString() >=
-            updatedDate.toISOString()
-        ) {
-          return;
+        if (reject_reason) {
+          return await handleRejection({ logger, data, entityManager });
```
idoshamun, line 284: await is not needed here

src/workers/postUpdated.ts
```diff
@@ -75,76 +292,49 @@ const worker: Worker = {
...
+          // Handle update of existing post
+          combinedPost = await updatePost({
+            entityManager,
+            data: fixedData,
+            id: post_id,
+          });
+        }
+
+        // After add or update:
+        if (combinedPost.visible) {
```
idoshamun, line 325: Should we always do this check? I think we should do it only on creation of if `visible` was changed right?

src/workers/postUpdated.ts
```diff
@@ -75,76 +292,49 @@ const worker: Worker = {
...
-          await entityManager
-            .getRepository(SharePost)
-            .update(
-              { sharedPostId: databasePost.id },
-              { visible: true, visibleAt: updatedDate, private: privacy },
-            );
-        }
-
-        await addKeywords(entityManager, mergedKeywords, data.post_id);
+        await addKeywords(entityManager, mergedKeywords, combinedPost.id);
```
idoshamun, line 337: This should also happen on creation and if the tags were changed since last time.
Also bare in mind that we probably have to delete the old keywords and apply the new ones

src/workers/postUpdated.ts
```diff
@@ -38,33 +47,241 @@ interface Data {
...
+  const postCreatedAt = new Date();
+  data.id = postId;
+  data.shortId = postId;
+  data.createdAt = postCreatedAt;
+  data.score = Math.floor(postCreatedAt.getTime() / (1000 * 60));
+  data.origin = data?.scoutId
+    ? PostOrigin.CommunityPicks
+    : data.origin ?? PostOrigin.Crawler;
+
+  // TODO: Not sure if these are still needed?
```
idoshamun, line 148: Not needed, that's fine :)

src/workers/postUpdated.ts
rebelchris: Basically they are handled inside the function, I just needed a quick way to return if it failed.
We can do it based on result if you prefer?

idoshamun (changes requested):
Refactor looks awesome now! Much more readable.
Added some comments about the internals :)

src/workers/postUpdated.ts
idoshamun: Yes please. For me seeing an empty `catch` usually signals a bad practice

src/workers/postUpdated.ts
rebelchris: Hmm, yeah originally `newPost` would check for title atleast, in this new flow it's generic so there could be a case where it's not yet.
Not sure if we could ever create new post that have share post attached actually.

src/workers/postUpdated.ts
rebelchris: This happens for both flows, I didn't change the internals of this.

src/workers/postUpdated.ts
rebelchris: Oh wait you are right, let me adjust it back

idoshamun (approved):
Woohoo!!

PR #1288, opened 2023-05-24
feat: create post mutation

sshanzel:
Introduce a new mutation to create a new freeform post.

Title and content should now be required.


src/common/post.ts
```diff
@@ -145,3 +147,55 @@ export const createSquadWelcomePost = async (
...
+  | 'sourceId'
+  | 'id'
+  | 'shortId'
+>;
+
+export const createFreeFormPost = async (
+  con: DataSource | EntityManager,
+  args: CreatePost,
+) =>
+  con.getRepository(FreeformPost).save({
```
idoshamun, line 172: We can set `shortId` as `id` in here instead of getting externally

src/schema/posts.ts
```diff
@@ -538,6 +534,31 @@ export const typeDefs = /* GraphQL */ `
...
     ): EmptyResponse @auth
 
+    """
+    To allow user to create freeform posts
+    """
+    createPost(
+      """
+      ID of the squad to post to
+      """
+      id: ID!
```
idoshamun, line 544: sourceId is better. just ID may confuse with post id

src/schema/posts.ts
```diff
@@ -983,6 +1004,64 @@ export const resolvers: IResolvers<any, Context> = {
...
+    createPost: async (
+      source,
+      args: EditPostArgs,
+      ctx: Context,
+      info,
+    ): Promise<GQLPost> => {
+      const { id: sourceId, image } = args;
+      const { con, userId } = ctx;
+      const title = args.title?.trim() ?? '';
+      const content = args.content?.trim() ?? '';
```
idoshamun, line 1016: In `validatePost` we also trim them. Let's unify somehow

__tests__/posts.ts
```diff
@@ -2169,6 +2169,156 @@ describe('mutation checkLinkPreview', () => {
   });
 });
 
+describe('mutation createPost', () => {
```
idoshamun, line 2172: Let's add a test to check that we don't allow creating a post in a squad I don't belong to

idoshamun: And another test for mentioning people outside the squad. This should not be allowed

src/schema/posts.ts
```diff
@@ -983,6 +1004,64 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const id = await generateShortId();
+
+      if (!title) {
+        throw new ValidationError('Title can not be an empty string!');
+      }
+
+      if (!content) {
+        throw new ValidationError('Content can not be an empty string!');
+      }
```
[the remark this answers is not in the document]
sshanzel: Oh, my bad. I was planning to use the validate post in the `editPost` too, but these properties are not required on that end. WDYT?

src/schema/posts.ts
```diff
@@ -538,6 +534,31 @@ export const typeDefs = /* GraphQL */ `
       comment: String
     ): EmptyResponse @auth
 
+    """
+    To allow user to create freeform posts
+    """
+    createPost(
```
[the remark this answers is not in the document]
sshanzel: Agree with that 🚢

src/common/post.ts
```diff
@@ -145,3 +147,58 @@ export const createSquadWelcomePost = async (
...
+  FreeformPost,
+  'title' | 'content' | 'image' | 'contentHtml'
+>;
+
+export type CreatePost = Pick<
+  FreeformPost,
+  'title' | 'content' | 'image' | 'contentHtml' | 'authorId' | 'sourceId' | 'id'
+>;
+
+export const createFreeFormPost = async (
```
idoshamun, line 161: Another thought, I see that this function doesn't really create the freeform as it just saves it to the db. It doesn't handle the markdown rendering, mentions, and so on. Isn't it confusing?

idoshamun (approved):
Basically I approve this PR anyway but see my comment if you'd like to rename or change the functionality. Up to you

src/common/post.ts
sshanzel: Good point. Let's use the most accurate label. Will refactor 🚢

PR #1289, opened 2023-05-26
fix: await post content upload id

sshanzel:
Names are always the same since the promise was not awaited.


idoshamun:
We can close this PR thanks to https://github.com/dailydotdev/daily-api/pull/1290

PR #1290, opened 2023-05-26
feat: delete zombie images

idoshamun:
Freeform post introduces the ability to upload images to our cloudinary storage. It can easily get bloated by unused images if not handled carefully. This PR monitors all the uploaded images and deleted 30 days old images that never got submitted to any post.

WT-1368 #done


idoshamun:
@sshanzel do we need to support comments at this stage? If so I need to add another worker probably

sshanzel (approved):
The edit process flow was neat! Amazing! 🚀

sshanzel:
> @sshanzel do we need to support comments at this stage? If so I need to add another worker probably

We can probably do it on the v7 since that is only when we will allow uploading in the comments section.

idoshamun:
> We can probably do it on the v7 since that is only when we will allow uploading in the comments section so we can get this going as we ship v6.

Let's add a task for that then

PR #1296, opened 2023-05-30
fix: delete welcome and freeform post

sshanzel:
Deleting of posts for Welcome and Freeform post should be allowed.

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -722,7 +722,6 @@ export const typeDefs = /* GraphQL */ `
 `;
 
 const nullableImageType = [PostType.Freeform, PostType.Welcome];
-const squadPostTypes = [PostType.Share, PostType.Freeform, PostType.Welcome];
 
 const saveHiddenPost = async (
   con: DataSource,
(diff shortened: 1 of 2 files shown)


capJavert (changes requested):
👌 Nice

We probably need to update tests? 

Also remove mention of shared post and createSharedPost helper (since it no longer applies to only that one).

capJavert (changes requested):
👌 Nice

We probably need to update tests? 

Also remove mention of shared post and createSharedPost helper (since it no longer applies to only that one).

sshanzel:
> We probably need to update tests?
>
> Also remove mention of shared post and createSharedPost helper (since it no longer applies to only that one).

Right, we need to add tests for this. Though for the helper, I think we can keep it as it is used by most of the tests related to deleting which would work the same for other types.

capJavert:
> > We probably need to update tests?
> > Also remove mention of shared post and createSharedPost helper (since it no longer applies to only that one).
> 
> Right, we need to add tests for this. Though for the helper, I think we can keep it as it is used by most of the tests related to deleting which would work the same for other types.

Agreed, we can just rename it maybe 👍

sshanzel:
I think we don't even have to touch it, we won't just use it on the new tests that we will be creating since it won't be useful.

src/schema/posts.ts
```diff
@@ -980,7 +981,7 @@ export const resolvers: IResolvers<any, Context> = {
       await ctx.con.transaction(async (manager) => {
         const repo = manager.getRepository(Post);
         const post = await repo.findOneBy({ id });
-        if (post.type === PostType.Share) {
+        if (squadPostTypes.includes(post.type)) {
```
idoshamun, line 984: I think we should not check the type here even.
If user has access to delete a post within squad, why does the type matter?

sshanzel: Makes sense. I think this was just an additional check for `article` types.

idoshamun (approved):

sshanzel:
@capJavert let me know what you think 🥳

capJavert (approved):

PR #1298, opened 2023-05-31
feat: add referralOrigin column and referral campaign query

capJavert:
- Add referralOrigin column
- Add referralCampaign query to get user progress in campaign
- Add getShortUrl util

TODO
- [x] Sort out Heimdall modifications with platform (not blocker necessarily for this PR)


src/entity/User.ts
```diff
@@ -94,6 +94,10 @@ export class User {
   @Index('IDX_user_referral')
   referralId?: string | null;
 
+  @Column({ length: 36, nullable: true })
```
capJavert, line 97: @idoshamun I copied length from referral, I think it is plenty but we might wanna double to be save if we have lengthier origins?

capJavert: Also, do we have referrals outside of squads, asking because we might be able to prefill `squad` to existing users with `referralId`?

src/common/links.ts
```diff
@@ -51,3 +52,30 @@ export function isValidHttpUrl(link: string): boolean {
     return false;
   }
 }
+
+export const getShortUrl = async (url: string | URL): Promise<string> => {
+  const urlSHortenerSecret = process.env.URL_SHORTENER_SECRET;
+
+  if (!urlSHortenerSecret) {
+    return url.toString();
+  }
```
capJavert, line 61: It would be great if we could add some kind of runtime check if all env variables are present before starting the app in production. It could run in CI/CD. Because if we were to run the app in production we could be serving not shortened URLs if we forget to add `URL_SHORTENER_SECRET`. Similar thing already exists with `CLOUDINARY_URL`.

Just wanted to see if we looked at this before or have any ideas?

.env
```diff
@@ -37,3 +37,4 @@ HEIMDALL_ORIGIN=http://localhost:7000
 KRATOS_ORIGIN=http://localhost:7000
 SUBMIT_ARTICLE_THRESHOLD=250
 ANALYTICS_URL=http://localhost:5000
+URL_SHORTENER_BASE_URL=https://dly.to
```
idoshamun, line 40: I think it's not needed for local env if we don't have the key right?
Same goes for adhoc.yaml

src/common/links.ts
idoshamun: Valid point. Let's log it and think how to approach later

src/entity/User.ts
idoshamun: Let's use `type: 'text'` instead of setting the length. We can set default value to `squad` as you suggested to prefill evertyhing

idoshamun (changes requested):
A bit more tweaking but overall looks great

src/entity/User.ts
capJavert: @idoshamun would we use custom `bin` to pre-fill current values because I don't know if we can conditionally set values of `referralOrigin` depending if `referralId` has value?

idoshamun: @capJavert you can add it as statement to the migration to deal with that. Way easier.
Let's also add a check when creating a new user that if referralId is set we will set referralOrigin to squad to have backwards compatibility for users with old version

src/common/links.ts
capJavert: Logged here https://dailydotdev.atlassian.net/browse/WT-1428

src/entity/User.ts
capJavert: Added in a68d66d6adfacc2516057140f5836f047c469027 and 2932aa49de456303ecf2c075c099379bea8e85fb

sshanzel (approved):
Nice work here! Just had one non-blocking comment, but everything looks good great 🚀

idoshamun (approved):

src/common/links.ts
```diff
@@ -51,3 +52,31 @@ export function isValidHttpUrl(link: string): boolean {
     return false;
   }
 }
+
+export const getShortUrl = async (url: string | URL): Promise<string> => {
```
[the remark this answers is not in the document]
capJavert: Yeah, I actually prefer URL constructor (as seen in this PR), but let's make it `string` for now since I think it looks clunky to accept 2 types and return single type.

capJavert:
Heimdall adjustments made here https://github.com/dailydotdev/heimdall/pull/71

PR #1300, opened 2023-06-02
feat: Making commentary optional for sharePost

TomDowling:
Making the commentary optional on the sharePost as well as trimming the whitespace leading and trailing on the commentary before creating

### DIFF
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -237,10 +237,13 @@ export const createSharePost = async (
 ): Promise<SharePost> => {
   let strippedCommentary = commentary;
 
-  if (commentary) {
+  if (commentary?.length) {
     strippedCommentary = commentary.trim();
     await validateCommentary(strippedCommentary);
+  } else {
+    strippedCommentary = null;
   }
+
   const id = await generateShortId();
   try {
     const { private: privacy } = await con
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -232,10 +232,15 @@ export const createSharePost = async (
   sourceId: string,
   userId: string,
   postId: string,
-  commentary: string,
+  commentary: string | null,
   visible = true,
 ): Promise<SharePost> => {
-  await validateCommentary(commentary);
+  let strippedCommentary = commentary;
+
+  if (commentary) {
+    strippedCommentary = commentary.trim();
+    await validateCommentary(strippedCommentary);
+  }
   const id = await generateShortId();
   try {
     const { private: privacy } = await con
@@ -248,7 +253,7 @@ export const createSharePost = async (
       sourceId,
       authorId: userId,
       sharedPostId: postId,
-      title: commentary,
+      title: strippedCommentary,
       sentAnalyticsReport: true,
       private: privacy,
       origin: PostOrigin.UserGenerated,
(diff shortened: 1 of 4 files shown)


src/schema/posts.ts
```diff
@@ -1286,19 +1286,20 @@ export const resolvers: IResolvers<any, Context> = {
...
         sourceId,
-      }: { id: string; commentary: string; sourceId: string },
+      }: { id: string; commentary?: string; sourceId: string },
       ctx,
       info,
     ): Promise<GQLPost> => {
       await ctx.con.getRepository(Post).findOneByOrFail({ id });
       await ensureSourcePermissions(ctx, sourceId, SourcePermissions.Post);
 
+      const strippedCommentary = commentary.trim();
```
idoshamun, line 1296: If commentary is null this will break 😭

idoshamun (changes requested):
Let's add a test where we provide empty commentary to see that everything works

sshanzel (commented):
Kindly do the same thing on the `submitExternalLink` too as it is also a mutation for sharing a post 🥳

src/schema/posts.ts
```diff
@@ -1250,6 +1250,7 @@ export const resolvers: IResolvers<any, Context> = {
           throw new ValidationError('URL is not valid');
         }
 
+        const strippedCommentary = commentary ? commentary.trim() : '';
```
sshanzel, line 1253: Should we probably keep the `null` value if it is empty? WDYT? @idoshamun

sshanzel (approved):
Looks good to me overall! 🥳 

Just had one non-blocking comment 🚢

src/schema/posts.ts
idoshamun: Yes, let's set it to null

idoshamun: Maybe it makes more sense to even put this `commentary ? commentary.trim() : null`` inside the `createSharePost` function. This way we have it only in one place

TomDowling: Good idea, reduce the duplicate code. This change has been done now

sshanzel: I noticed the change, but we are now missing the `trim` part of the commentary when there is an actual value. Ido's suggestion would probably be the best route to go about it.

`commentary ? commentary.trim() : null`

idoshamun (approved):

src/entity/posts/utils.ts
```diff
@@ -232,10 +232,15 @@ export const createSharePost = async (
...
+  commentary: string | null,
   visible = true,
 ): Promise<SharePost> => {
-  await validateCommentary(commentary);
+  let strippedCommentary = commentary;
+
+  if (commentary) {
+    strippedCommentary = commentary.trim();
+    await validateCommentary(strippedCommentary);
+  }
```
sshanzel, line 243:
```suggestion
  let strippedCommentary = commentary?.trim();

  if (commentary.length) {
    await validateCommentary(strippedCommentary);
  } else {
    strippedCommentary = null;
  }
```

sshanzel (approved):
Nice work! Just one last thing before merging. We still don't set it to null but have suggested a commit.

sshanzel (approved):
Thanks for fixing the tests!

PR #1303, opened 2023-06-05
refactor: post title html property

sshanzel:
Note: The mutation for making the title commentary as HTML will be worked in a different PR.


sshanzel:
> @sshanzel how are we going to retroactively fill the column with data?

We don't have to, we will only use it in FE if it is available like how it was used below:

[packages/shared/src/components/post/share/SharePostTitle.tsx](https://github.com/dailydotdev/apps/blob/82c9b2638c105342ac00ee877da7c241bcbc2e63/packages/shared/src/components/post/share/SharePostTitle.tsx#L9-L17)

idoshamun:
Let's keep it this way for now I guess @sshanzel

PR #1304, opened 2023-06-05
feat: share commentary to allow mentions

sshanzel:
Allow share commentary to mention users.

As shown in the test, we were able to mention and also have shown to not work on non-existent users.

This utility function is used in all 3 avenues:
- Create a Squad and share a post right with it.
- Share a post on a Squad.
- Share a post with an external link to a Squad.

### DIFF
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -246,7 +246,9 @@ export const createSharePost = async (
   const id = await generateShortId();
   try {
     const mentions = await getMentions(con, commentary, userId, sourceId);
-    const titleHtml = generateTitleHtml(commentary, mentions);
+    const titleHtml = commentary?.length
+      ? generateTitleHtml(commentary, mentions)
+      : null;
     const { private: privacy } = await con
       .getRepository(Source)
       .findOneBy({ id: sourceId });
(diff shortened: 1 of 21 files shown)


src/entity/posts/utils.ts
```diff
@@ -227,6 +230,11 @@ export const createExternalLink = async (
   });
 };
 
+export const generateTitleHtml = (
+  title: string,
+  mentions: MentionedUser[],
+): string => `<p>${renderMentions(title, mentions)}</p>`;
```
idoshamun, line 236: We have to be careful here because this is gonna be in a `innerHtml`.
We have to make sure it's a safe string with no html tags or javascript snippets, or in other words, we have to escape the html. You can have a look at the markdown renderer to see how we escape it.
You need to apply it on the title. Please also add some tests to check that

idoshamun (changes requested):
It looks good but I'd like you to add one test for testing html injection so let's submit a commentary like:
`</p><style>html { color: red !important; }</style><p>` and let's make sure it's safe.

sshanzel:
> It looks good but I'd like you to add one test for testing html injection so let's submit a commentary like:
> </p><style>html { color: red !important; }</style><p> and let's make sure it's safe.

Oh, right. I completely missed this. Let me write one up quickly.

sshanzel:
@idoshamun escaped! 🥳

https://github.com/dailydotdev/daily-api/pull/1304/commits/c649241aa23b165cfae1e07badc5dd1397ea6282

idoshamun (approved):

PR #1305, opened 2023-06-05
refactor: post mention preview using title or content

sshanzel:
Changes: 
- Improved generics on the notification context so we will only have to do it once.
- Check whether the mention was from the title or from the content, and display the appropriate preview.
- `doneTo` property whenever applicable as an optional property.

### DIFF
--- a/src/common/markdown.ts
+++ b/src/common/markdown.ts
@@ -134,11 +134,12 @@ export const saveMentions = (
   return query.orIgnore().execute();
 };
 
-export const checkHasMention = (content: string, username: string) =>
-  content.split(' ').some((word) => {
-    if (word.charAt(0) !== '@') return false;
+export const checkHasMention = (content: string, username: string) => {
+  const lines = content.split('\n');
 
-    const mention = word.substring(1);
+  return lines.some((line) => {
+    const words = line.split(' ');
 
-    return mention === username;
+    return words.some((word) => word === `@${username}`);
   });
+};
--- a/src/common/markdown.ts
+++ b/src/common/markdown.ts
@@ -133,3 +133,12 @@ export const saveMentions = (
 
   return query.orIgnore().execute();
 };
+
+export const checkHasMention = (content: string, username: string) =>
+  content.split(' ').some((word) => {
+    if (word.charAt(0) !== '@') return false;
+
+    const mention = word.substring(1);
+
+    return mention === username;
+  });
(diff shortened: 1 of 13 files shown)


src/common/markdown.ts
```diff
@@ -133,3 +133,12 @@ export const saveMentions = (
...
   return query.orIgnore().execute();
 };
+
+export const checkHasMention = (content: string, username: string) =>
+  content.split(' ').some((word) => {
+    if (word.charAt(0) !== '@') return false;
+
+    const mention = word.substring(1);
+
+    return mention === username;
```
idoshamun, line 143: Wouldn't it be simpler to do
```suggestion
    return word === `@${username}`
```

src/notifications/types.ts
```diff
@@ -55,6 +56,7 @@ export type NotificationSourceRequestContext = NotificationBaseContext & {
 
 export type NotificationDoneByContext = NotificationBaseContext & {
   doneBy: Reference<User>;
+  doneTo?: Reference<User>;
```
idoshamun, line 59: What's the purpose of this field? Isn't that always the notification user?

src/notifications/generate.ts
```diff
@@ -149,12 +150,17 @@ export const generateNotificationMap: Record<
...
     builder,
-    ctx: NotificationPostContext & NotificationDoneByContext,
+    ctx: NotificationPostContext<FreeformPost> & NotificationDoneByContext,
   ) =>
     builder
       .referencePost(ctx.post)
       .icon(NotificationIcon.Comment)
-      .description((ctx.post as WelcomePost).content, true)
+      .description(
+        checkHasMention(ctx.post.title, ctx.doneTo.username)
```
idoshamun, line 159: Let's add test to cover that, usually we don't check the description but here we have some logic to test

src/notifications/types.ts
sshanzel: Yes, but we only have the `userId` of the user to notify in this context but we need the username. Or do we have a pre-existing property for the relevant User object?

src/common/markdown.ts
sshanzel: Indeed, that is much simpler! 🤯

src/notifications/generate.ts
sshanzel: I agree. I was looking for some examples and have found it now 🚢

idoshamun (approved):

PR #1312, opened 2023-06-14
fix: hide posts with `showOnFeed` false by default

idoshamun:
`showOnFeed` is a new property added few weeks ago to support complex flows of Yggdrasil. It allows us to add posts to the database without showing them on the feed


rebelchris (approved):
PR is good, just for my knowledge in which cases do we add posts that are not supposed to be visible? (What's the reason for those posts)

PR #1313, opened 2023-06-14
feat: add downvote table

capJavert:
Add downvote table. mutations will be done in separate PR.

Also added `downvotes` column to `post` table.


idoshamun (commented):
Can you please add `replicate identity full` to the migration? (you can look at existing migrations)
This will make sure whenever we want to track cdc events using debezium we will have the before and after data

idoshamun (approved):

rebelchris (approved):
Looks great

PR #1314, opened 2023-06-14
feat: add db triggers to handle upvotes count

capJavert:
As per comments in the [DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/544768013/Downvoting+and+reporting+posts#Other-notes) replacing inline increments with db triggers to avoid race conditions while upvoting and in the future downvoting.

Also added downvote triggers.

Also updated two tests that were right but expected wrong values due to test setup, see 4962c92f3eac11f3af9ab6e4ff06fe9f3ec5be33 commit message.

As additional note, in general we could apply this for comments too.

### DIFF
--- a/src/migration/1686749310179-UpvoteTrigger.ts
+++ b/src/migration/1686749310179-UpvoteTrigger.ts
@@ -1,0 +1,42 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class UpvoteTrigger1686749310179 implements MigrationInterface {
+    name = 'UpvoteTrigger1686749310179'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      queryRunner.query(`
+        CREATE OR REPLACE FUNCTION increment_upvote()
+          RETURNS TRIGGER
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          UPDATE post SET upvotes = upvotes + 1 WHERE id = NEW."postId";
+          RETURN NEW;
+        END;
+        $$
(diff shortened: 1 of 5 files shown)


src/migration/1686749310179-UpvoteTrigger.ts
```diff
@@ -0,0 +1,42 @@
...
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          UPDATE post SET upvotes = upvotes + 1 WHERE id = NEW."postId";
+          RETURN NEW;
+        END;
+        $$
+      `)
+      queryRunner.query('CREATE TRIGGER upvote_create_trigger AFTER INSERT ON "upvote" FOR EACH ROW EXECUTE PROCEDURE increment_upvote()')
```
idoshamun, line 19: CREATE OR REPLACE may be better just in case. Also for the other trigger

capJavert: @idoshamun yeah I tried it but `OR REPLACE` for `CREATE TRIGGER` is only available from version [14](https://www.postgresql.org/docs/14/sql-createtrigger.html), we are on version [11](https://www.postgresql.org/docs/11/sql-createtrigger.html) from what I can see.

idoshamun: Oh right! Thanks for letting me know

idoshamun (approved):

rebelchris (approved):
Nice, yeah these functions are pretty cool actually.

PR #1315, opened 2023-06-15
feat: add downvoted property to gql post

capJavert:
Add `downvoted` property to gql post and boot to be used in companion.

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -394,6 +394,7 @@ const COMPANION_QUERY = parse(`query Post($url: String) {
           author {
             id
           }
+          downvoted
         }
       }`);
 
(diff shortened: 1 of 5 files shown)


src/routes/boot.ts
```diff
@@ -394,6 +394,7 @@ const COMPANION_QUERY = parse(`query Post($url: String) {
           author {
             id
           }
+          downvoted
```
rebelchris, line 397: Not sure, but I thought we also had a specific companion boot test that should be affected.

rebelchris (approved):
One small comment, but looks great, nice job on the test.

src/routes/boot.ts
idoshamun: Please validate it before merging :)

capJavert: @rebelchris this one 92972601b44b0444789e6d45d00d305be11b1256?

rebelchris: Ah yes, sorry thought that was the regular boot (sometimes hard to see in the PR screen 😅)

PR #1316, opened 2023-06-16
feat: add downvote mutations

capJavert:
Downvote mutations. 

Adjust upvote mutations to revert downvotes.

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -66,6 +66,7 @@ import { FileUpload } from 'graphql-upload/GraphQLUpload';
 import { insertOrIgnoreAction } from './actions';
 import { generateShortId, generateUUID } from '../ids';
 import { ContentImage } from '../entity/ContentImage';
+import { Downvote } from '../entity/Downvote';
 
 export interface GQLPost {
   id: string;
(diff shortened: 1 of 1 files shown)


src/schema/posts.ts
```diff
@@ -1327,6 +1366,76 @@ export const resolvers: IResolvers<any, Context> = {
...
+            postId: id,
+            userId: ctx.userId,
+          });
+
+          if (upvote) {
+            await entityManager.getRepository(Upvote).delete({
+              postId: id,
+              userId: ctx.userId,
+            });
+          }
```
idoshamun, line 1398: We can just delete the object without checking if it exists to reduce number of queries

src/schema/posts.ts
```diff
@@ -1327,6 +1366,76 @@ export const resolvers: IResolvers<any, Context> = {
...
+        if (downvote) {
+          await entityManager.getRepository(Downvote).delete({
+            postId: id,
+            userId: ctx.userId,
+          });
+
+          await entityManager
+            .getRepository(HiddenPost)
+            .delete({ postId: id, userId: ctx.userId });
+
```
idoshamun, line 1432: Let's turn it into a function. We have this logic twice here already

idoshamun (changes requested):
See my comments and also add relevant tests for upvote mutation. We should check that it cancels the downvote and removes the hidden post

src/schema/posts.ts
capJavert: @idoshamun there is a simular check if `cancelUpvote`, we can then remove it from there also?

idoshamun: Basically yes

src/schema/posts.ts
capJavert: moved to `revertPostDownvote` 👍

idoshamun (approved):

rebelchris (approved):

PR #1317, opened 2023-06-17
feat: clear out zombie images on comment html

sshanzel:
- Refactored the usage of `enum` to follow a PascalCase setup.
- New pubsub topics for:
  - Comment edited.
  - Comment deleted.
- New subscription for post commented and comment commented to update uploaded images.
- Created a generator function for creating workers related to clearing out content images.


src/workers/cdc.ts
```diff
@@ -193,6 +195,14 @@ const onCommentChange = async (
         data.payload.after.id,
       );
     }
+  } else if (data.payload.op === 'u') {
+    await notifyCommentEdited(logger, data.payload.after);
```
idoshamun, line 199: Let's check here that it was the content that got edited

src/workers/cdc.ts
```diff
@@ -193,6 +195,14 @@ const onCommentChange = async (
...
+  } else if (data.payload.op === 'd') {
+    await notifyCommentDeleted(logger, {
+      ...data.payload.before,
+      content: undefined,
```
idoshamun, line 203: Why do you want to change these props? I think we should keep it as is. We can't be 100% sure who's gonna subscribe to this topic, and this is not an expected behavior

__tests__/workers/commentMarkdownImages.ts
```diff
@@ -0,0 +1,69 @@
...
+beforeEach(async () => {
+  jest.resetAllMocks();
+  await con.getRepository(ContentImage).save([
+    { serviceId: '1', url: 'https://daily.dev/1.jpg' },
+    { serviceId: '2', url: 'https://daily.dev/2.jpg' },
+    { serviceId: '3', url: 'https://daily.dev/3.jpg' },
+  ]);
+});
+
+it('should do nothing for post without a content html', async () => {
```
idoshamun, line 34:
```suggestion
it('should do nothing for comment without a content html', async () => {
```

idoshamun (changes requested):
This looks great! Love the generators approach.
Please see my comments and you also have a pulumi error due to duplicate subscription names

src/workers/cdc.ts
sshanzel: I agree, but we would need a reference on which comment was deleted. The `after` is `null` once deleted, so I had to take the `before` for the IDs and applied the empty content as part of being a `null` after.

src/workers/cdc.ts
sshanzel: The only thing that can change from here is the `content` (and `contentHtml`) but I agree on making it explicit just to be sure 🚢

src/workers/index.ts
```diff
@@ -40,6 +42,8 @@ export const workers: Worker[] = [
   deleteUserFromMailingList,
   commentUpvotedRep,
   commentUpvoteCanceledRep,
+  commentMarkdownImages,
```
idoshamun, line 45: I think you miss some workers here. I see more new subscriptions then workers here

sshanzel: Yes, I thought we can put 2 workers (edit and delete) in a single subscription. I am currently adding a dedicated one for deleting purposes.

src/workers/cdc.ts
idoshamun: The topic itself means this comment is deleted. It's not trivial to mess with the properties of the data. Again, someone in the future may subscribe to event and will not figure out what happens here. I suggest we find another way to indicate deletion, e.g. providing a flag to the generator

sshanzel: The value of `data.payload.after` is `null`, the subscriber won't have any reference to which comment was deleted if we pass just the `after` object. We have two options:
- We will pass the `id` only in here to still have the reference on the subscription, or
- We will send a flag to the generator to clear **all** `ContentImage` with the `type` we sent with all the `referenceId` that is not present on the relevant table anymore. Might become a heavy process in the future when as we will fetch all `Comment` type (for example)  then we will query all reference IDs that do not exist on the `Comments` table anymore.

idoshamun: I think you misunderstood me. My suggestion is to send `before` as you did already but without manipulating `content` and `contentHtml`. Just pass the object as is

sshanzel: Oh, right. Yep. That makes sense. Let me do that 🚀

sshanzel:
I have now added an optional parameter to allow the generator to make some decisions based on some values and to keep the openness of the function to changes 🚢

idoshamun (approved):

rebelchris (approved):
Looks amazing!

PR #1320, opened 2023-06-21
fix: add disallow handles for username and squads

rebelchris:
Added a disallow handle table to verify usernames and source handles against.

@idoshamun Let me know if this is inline with what you had in mind.
It was hard to make it super uniform to not have to evaluate seperate.

WT-1467 #done


src/entity/DisallowHandle.ts
```diff
@@ -0,0 +1,23 @@
+import { Entity, EntityManager, Index, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class DisallowHandle {
+  @PrimaryColumn({ unique: true })
+  @Index({ unique: true })
+  value: string;
+}
+
+export const checkDissalowHandle = async (
```
capJavert, line 10: checkDisallowHandle* typo?

src/schema/users.ts
```diff
@@ -756,10 +757,23 @@ export const resolvers: IResolvers<any, Context> = {
...
+          }
+          return handle.value;
+        },
+      );
 
-      for (const usernameCheck in usernameChecks) {
+      for (const usernameCheck in disallowedUsernames) {
         generatedUsernames = generatedUsernames.filter(
-          (item) => item !== usernameChecks[usernameCheck].username,
+          (item) => item !== disallowedUsernames[usernameCheck],
```
capJavert, line 776: This should be better?
```js
generatedUsernames
    .filter(item => !disallowedUsernames.includes(item))
```

src/entity/DisallowHandle.ts
```diff
@@ -0,0 +1,23 @@
...
+): Promise<boolean> => {
+  const handle = await entityManager
+    .getRepository(DisallowHandle)
+    .createQueryBuilder()
+    .select('value')
+    .where('value = :value', {
+      value,
+    })
+    .getRawOne();
+  return !!handle;
```
capJavert, line 22: Don't know if we can cache this since value will almost never change as table would be static.

src/entity/DisallowHandle.ts
```diff
@@ -0,0 +1,23 @@
+import { Entity, EntityManager, Index, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class DisallowHandle {
+  @PrimaryColumn({ unique: true })
+  @Index({ unique: true })
```
idoshamun, line 6: We don't need to index primary key. It is indexed by default

src/entity/DisallowHandle.ts
```diff
@@ -0,0 +1,23 @@
+import { Entity, EntityManager, Index, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class DisallowHandle {
+  @PrimaryColumn({ unique: true })
```
idoshamun, line 5: A single primary column is unique by default. But I don't mind being explicit about it if you wish

src/entity/DisallowHandle.ts
idoshamun: Should be relatively fast. So I wouldn't worry about cache at this point but great thinking

src/entity/User.ts
```diff
@@ -249,7 +250,11 @@ export const addNewUser = async (
 
   return con.transaction(async (entityManager) => {
     const isUniqueUser = await checkUsernameAndEmail(entityManager, data);
-    if (!isUniqueUser) {
+    const isDissalowHandle = await checkDissalowHandle(
```
idoshamun, line 253: We can merge both calls using `Promise.all`. Not major impact but could be nicer

src/migration/1687349045215-DisallowHandle.ts
```diff
@@ -0,0 +1,5518 @@
...
+  name = 'DisallowHandle1687349045215';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "disallow_handle" ("value" character varying NOT NULL, CONSTRAINT "PK_8c84e48365905231826700dd8b4" PRIMARY KEY ("value"))`,
+    );
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX "IDX_8c84e48365905231826700dd8b" ON "disallow_handle" ("value") `,
+    );
+    await queryRunner.query(
```
idoshamun, line 13: Can we turn it into a script perhaps? This will make the migration relatively heavy and we constantly run it for tests and so on

src/schema/sources.ts
```diff
@@ -805,7 +806,11 @@ export const resolvers: IResolvers<any, Context> = {
         .getRepository(Source)
         .findOneBy({ handle: handle.toLowerCase() });
 
-      return !!source;
+      const disallowHandle = await ctx
```
idoshamun, line 809: Can ber merged we the above using `Promise.all`

idoshamun: Also why don't you use `checkDissalowHandle`?
We can add the `toLowerCase` to the function as well to keep it consistent

src/schema/sources.ts
```diff
@@ -967,6 +972,15 @@ export const resolvers: IResolvers<any, Context> = {
       });
       try {
         const sourceId = await ctx.con.transaction(async (entityManager) => {
+          const disallowHandle = await entityManager
```
idoshamun, line 975: Why don't you use `checkDissalowHandle`?

src/schema/sources.ts
```diff
@@ -1046,6 +1060,15 @@ export const resolvers: IResolvers<any, Context> = {
       try {
         const editedSourceId = await ctx.con.transaction(
           async (entityManager) => {
+            const disallowHandle = await entityManager
```
idoshamun, line 1063: Why don't you use `checkDissalowHandle`?

src/schema/users.ts
```diff
@@ -756,10 +757,23 @@ export const resolvers: IResolvers<any, Context> = {
         where: { username: In(generatedUsernames) },
         select: ['username'],
       });
+      const disallowHandles = await ctx.getRepository(DisallowHandle).find({
```
idoshamun, line 760: Can be merged with the above

idoshamun (changes requested):
Overall implementation is great. I added some tips for optimization and unification.
Also, I think there are two test cases missing:
* When creating a new user with disallowed handle
* When an existing user wants to change the username to a disallowed handle

src/schema/users.ts
rebelchris: Ah super yes, I kind of worked on the old code, but good point 👍

src/entity/DisallowHandle.ts
rebelchris: Agreed with Ido, great point but for now it's ok, not super heavy.
Also we might be adding new values in the first period.

src/migration/1687349045215-DisallowHandle.ts
rebelchris: Sure was my first thought, but then thought we might want them locally everything as we do with keywords now.
Will convert to script for now 🙏

rebelchris:
@idoshamun Refactored a bit, also missed the username change so thanks for that.
For the newUser we already have the test, but it sits in private test suite.

Does it suffice?

src/migration/1687349045215-DisallowHandle.ts
```diff
@@ -0,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DisallowHandle1687349045215 implements MigrationInterface {
+  name = 'DisallowHandle1687349045215';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "disallow_handle" ("value" character varying NOT NULL, CONSTRAINT "PK_8c84e48365905231826700dd8b4" PRIMARY KEY ("value"))`,
+    );
+    await queryRunner.query(
```
idoshamun, line 10: Probably this index was created due to the `{unique: true}`. We don't need it so let's drop it from the entity and regenerate the migration. Sorry for the trouble! I'll not block for this

src/routes/private.ts
```diff
@@ -62,7 +63,8 @@ export default async function (fastify: FastifyInstance): Promise<void> {
       const user = await con
         .getRepository(User)
         .findOneBy({ username: search });
-      return res.status(200).send({ isTaken: !!user });
+      const disallowHandle = await checkDisallowHandle(con, search);
```
idoshamun, line 66: Can be merged with the above as well

idoshamun (approved):
Looks awesome! Had some non blocking comments so once you apply them just merge :)

src/entity/DisallowHandle.ts
capJavert: Works 👍

PR #1321, opened 2023-06-27
fix: add tags to slack report post

rebelchris:
Added the newly added tags to the slack port report worker.

WT-1450 #done


capJavert (approved):
Looks good, one question for my understanding, how do we test this out locally?

rebelchris:
@capJavert Generally don't test it locally, just tested test cases for now.
You can run locally by providing right slack key, but it's ok for now.

idoshamun (approved):

PR #1322, opened 2023-06-28
feat: added comment report table and migration

capJavert:
Add comment_report table and migration.

### DIFF
--- a/src/entity/CommentReport.ts
+++ b/src/entity/CommentReport.ts
@@ -1,0 +1,28 @@
+import { Column, Entity, Index, ManyToOne, PrimaryColumn } from 'typeorm';
+import { User } from './User';
+
+@Entity()
+export class CommentReport {
+  @PrimaryColumn({ type: 'text' })
+  @Index('IDX_comment_report_comment_id')
+  commentId: string;
+
+  @PrimaryColumn({ length: 36 })
+  @Index('IDX_comment_report_user_id')
+  userId: string;
+
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ length: 12 })
+  reason: string;
+
+  @Column({ type: 'text', nullable: true })
+  note: string;
+
+  @ManyToOne(() => User, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  user: Promise<User>;
+}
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Looks alright to me!

src/entity/CommentReport.ts
```diff
@@ -0,0 +1,28 @@
...
+  commentId: string;
+
+  @PrimaryColumn({ length: 36 })
+  @Index('IDX_comment_report_user_id')
+  userId: string;
+
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ length: 12 })
```
sshanzel, line 17: We have a report reason that pertains to misinformation and has thought of using `MISINFORMATION` as its key. The limit is 12, but the said key is 14 character. We can maybe extend the length to 16? WDYT?
```suggestion
  @Column({ length: 16 })
```

src/migration/1687956058691-CommentReport.ts
```diff
@@ -0,0 +1,20 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class CommentReport1687956058691 implements MigrationInterface {
+    name = 'CommentReport1687956058691'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE TABLE "comment_report" ("commentId" text NOT NULL, "userId" character varying(36) NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "reason" character varying(12) NOT NULL, "note" text, CONSTRAINT "PK_cef0f7f862ea6ffd5a99dd640eb" PRIMARY KEY ("commentId", "userId"))`);
```
sshanzel, line 7: In relation to the initial comment about extending the length.
```suggestion
        await queryRunner.query(`CREATE TABLE "comment_report" ("commentId" text NOT NULL, "userId" character varying(36) NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "reason" character varying(16) NOT NULL, "note" text, CONSTRAINT "PK_cef0f7f862ea6ffd5a99dd640eb" PRIMARY KEY ("commentId", "userId"))`);
```

src/entity/CommentReport.ts
rebelchris: Oh good catch, could even be safe and make it 36 doens't add much overhead I personally think.

src/migration/1687956058691-CommentReport.ts
capJavert: Added `36` as Chris suggested.

PR #1323, opened 2023-06-28
feat: added flags column to post table

capJavert:
Added `flags` column to `post` table.

Added indexes for jsonb column type. Choose `HASH` due to our flags being values and [HASH index type provides best performance for `=` operator](https://scalegrid.io/blog/using-jsonb-in-postgresql-how-to-effectively-store-index-json-data-in-postgresql/#6) (more general alternative would be BTREE).

Migrations for initial values and mutations to keep `flags` column in sync will be added in separate PR. 

Debezium filter is added to filter out the messages during migration. This filter should be removed after migration.


src/entity/posts/Post.ts
```diff
@@ -173,4 +183,11 @@ export class Post {
   @Column({ type: 'integer', default: 0 })
   @Index('IDX_post_downvotes')
   downvotes: number;
+
+  @Column({ type: 'jsonb', default: {} })
+  @Index('IDX_post_flags_sentAnalyticsReport', { synchronize: false })
+  @Index('IDX_post_flags_banned', { synchronize: false })
+  @Index('IDX_post_flags_deleted', { synchronize: false })
+  @Index('IDX_post_flags_promoteToPublic', { synchronize: false })
```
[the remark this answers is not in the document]
rebelchris: I think the idea was more the keep the existing indexes available.
So currently we'll have `IDX_post_deleted` which should now map.

At least this was my understanding of what we wanted to achieve.

rebelchris (changes requested):
I'll need to check this one a bit more carefully locally, but have one discussion point.

src/entity/posts/Post.ts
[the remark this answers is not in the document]
capJavert: They are available still, but added new ones for these new `jsonb` fields. Or do you mean update existing ones to new column?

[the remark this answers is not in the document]
capJavert: Yeah sure, not blocking us, I can continue work on migrations and mutations adjustment until we get a confirmation.

src/migration/1687960180394-PostFlags.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class PostFlags1687960180394 implements MigrationInterface {
+    name = 'PostFlags1687960180394'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
```
idoshamun, line 6: @capJavert shouldn't we set here the flags based on existing values or is it part of another migration?

src/entity/posts/Post.ts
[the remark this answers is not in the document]
idoshamun: LGTM

src/migration/1687960180394-PostFlags.ts
rebelchris: That was my main concern above, you mentioned they should be the same as existing ones right?
Would suggest to do it in the move-over PR though in that case, while migrating the values to flags

Just wanted to make sure I understood correctly.

capJavert: @idoshamun yeah, that one is done in a separate PR, that one also keeps them up to date. https://github.com/dailydotdev/daily-api/pull/1333

idoshamun: We should consider including in this PR the implementation that updates the flags instead of separating the creation of the column and updating the column value.
It's not the actually move over but just making sure they are always up to date. Wdyt?

capJavert: The PR I mentioned actually targets merge into this one so feel free to review https://github.com/dailydotdev/daily-api/pull/1333 and then we can merge all in one go to `main`

rebelchris (approved):

idoshamun (approved):

src/migration/1687960180394-PostFlags.ts
```diff
@@ -0,0 +1,22 @@
...
+        await queryRunner.query(`ALTER TABLE "post" ADD "flags" jsonb NOT NULL DEFAULT '{}'`);
```
capJavert, line 7: @idoshamun should we set DEFAULT as:
```
ALTER TABLE "post" ADD "flags" jsonb NOT NULL DEFAULT '{"sentAnalyticsReport":true,"banned":false,"deleted":false,"private":false,"visible":true,"showOnFeed":true,"promoteToPublic":false}''
```

To have default values available for new created posts? It makes sense to set those since not all are `false` by default and we can not do "false if missing".

idoshamun: Let's do it only for the true by default values

capJavert: #savingdbstorage works!

capJavert:
@idoshamun this is ready for review, full migration plan:
1. We deploy this PR that contains Debezium message filter and migration to add flags column (ALTER TABLE)
2. After deployment we run the migration script from https://github.com/dailydotdev/daily-api/pull/1333 in chunks
  - this will sync all the flags from current columns to `flags` column
  - only thing left is to decide how we want to run it (eg. what will the chunks be)
3. After migration we merge https://github.com/dailydotdev/daily-api/pull/1333 which contains adjustments for all the mutations to keep `flags` in sync with column values.

We will now have `flags` synced with existing column values. Next step then will be to make API and feed actually use values from `flags` which can be done in a separate PR.

Also please take a look at https://github.com/dailydotdev/daily-api/pull/1333 again for sanity check once more

.infra/application.properties
```diff
@@ -11,8 +11,11 @@ debezium.source.table.include.list=public.comment,public.comment_upvote,public.c
...
 debezium.source.plugin.name=pgoutput
 debezium.source.heartbeat.interval.ms=60000
-debezium.transforms=Reroute
+debezium.transforms=Reroute,filter
 debezium.transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
 debezium.transforms.Reroute.topic.regex=(.*)
 debezium.transforms.Reroute.topic.replacement=%topic%
+debezium.transforms.filter.type=io.debezium.transforms.Filter
+debezium.transforms.filter.language=jsr223.groovy
+debezium.transforms.filter.condition=!(valueSchema.field('op') && value.op == 'u' && value.before && value.before.flags == '{}' && value.after.flags != '{}')
```
rebelchris, line 20: I'm not familiar with this so can't really comment on whether this is correct or not.
Assume you tested it and it does what it needs to.

rebelchris (approved):
Nothing weird to me.
Did you not rather want to introduce the script in this PR so you can run it right away?
Keep in mind deployments on API take a long time.

PR #1325, opened 2023-06-29
fix: reporting hidden post not correctly writes report to database

capJavert:
This surfaced now that we added downvote option.

If you already downvoted since downvote also calls `saveHiddenPost` the report after that would always silently fail since `saveHiddenPost` would return `false` as the `hidden_post` row existed.

FIx here returns `true` if post is already hidden and throws error in all other cases as it should be to notify frontend of any API errors.


rebelchris (approved):
Nice find and great work on adding test case for it

Dont have to wait on ido for this one 👍

PR #1326, opened 2023-06-30
fix: invalidating cdn cache on cloudinary

sshanzel:
After further investigation, we will have to manually invalidate the cache for replacing an asset with the same name/public id.

Relevant sources:
https://support.cloudinary.com/hc/en-us/articles/202520852-I-ve-replaced-an-existing-asset-with-a-new-file-but-my-site-or-application-still-shows-the-old-asset-Why-is-that-

By default, invalidation is disabled (for good reasons), but we should invalidate in this case for replacing a thumbnail.

https://cloudinary.com/documentation/managing_assets#invalidating_cached_media_assets_on_the_cdn

Relevant API:
https://cloudinary.com/documentation/image_upload_api_reference#upload_optional_parameters

_Also just fixed some types while at it_

### DIFF
--- a/src/common/cloudinary.ts
+++ b/src/common/cloudinary.ts
@@ -38,7 +38,11 @@ interface UploadResult {
   id: string;
 }
 
-type UploadFn = (name: string, stream: Readable) => Promise<UploadResult>;
+type UploadFn = (
+  name: string,
+  stream: Readable,
+  options?: OptionalProps,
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Interesting, good find.
Looks good to me, let's proceed no need to wait for Ido on this.

capJavert (approved):
Nice 👍

PR #1327, opened 2023-06-30
feat: report comment reason map

sshanzel:
This was inspired by the existing mapping of our post report reasons. They already work as needed, so I thought of why not have the same process.


rebelchris (approved):
Neat!
Good to go.

PR #1329, opened 2023-06-30
fix: cloud errors

rebelchris:
Had a quick look at our google cloud errors.
Mostly some edge cases of functions that call on null values.

(Main introductions where due to freeform post not having a sharedPost or commentary at first)

See separate commits per instance.

### DIFF
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -212,11 +212,13 @@ const updatePost = async ({
   }
 
   if (databasePost.tagsStr !== data.tagsStr) {
-    await removeKeywords(
-      entityManager,
-      databasePost.tagsStr.split(','),
-      data.id,
-    );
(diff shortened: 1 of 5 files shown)


capJavert (approved):
Nice 👍

src/workers/postUpdated.ts
```diff
@@ -212,11 +212,13 @@ const updatePost = async ({
   }
 
   if (databasePost.tagsStr !== data.tagsStr) {
-    await removeKeywords(
-      entityManager,
-      databasePost.tagsStr.split(','),
-      data.id,
-    );
+    if (databasePost.tagsStr.length) {
```
idoshamun, line 215:
```suggestion
    if (databasePost.tagsStr?.length) {
```

idoshamun (approved):

PR #1330, opened 2023-07-03
feat: open squads query

sshanzel:
The flow displayed here came from an existing `sources` query. Link below.

https://github.com/dailydotdev/daily-api/blob/c2b0ac4d435295e51e431bde855f849f21a11fa8/src/schema/sources.ts#L769-L787

Now it got me thinking, why don't we just add a parameter to which filters we want to apply to the existing query? Thoughts?


rebelchris (approved):
Neat approach, I like it! 🙏
Let's wait for one other approval then we can merge if test resolve.

sshanzel:
I have updated the tests to avoid using snapshots to make it all easier as we also favor it.

idoshamun (approved):

PR #1332, opened 2023-07-03
feat: report comment mutation

sshanzel:
Similar to posts, comments should be allowed to be reported. The process is similar but much simpler as there is no capability of hiding the reported comment, yet (since raised, [Sab is inclined to have this](https://dailydotdev.slack.com/archives/C0592NL7YCU/p1688381192909049?thread_ts=1688378301.974239&cid=C0592NL7YCU) in the future).


src/schema/comments.ts
```diff
@@ -785,6 +810,32 @@ export const resolvers: IResolvers<any, Context> = {
...
+    reportComment: async (
+      source,
+      { commentId: id, reason, note }: ReportCommentArgs,
+      ctx: Context,
+    ): Promise<GQLEmptyResponse> => {
+      if (!reportCommentReasons.has(reason)) {
+        throw new ValidationError('Reason is invalid');
+      }
+
+      await ctx.getRepository(Comment).findOneByOrFail({ id });
```
capJavert, line 822: Do we need this, would `insert` not just fail due to referential integrity?

sshanzel: While that can work, the `insert` is inside the try-catch and it ignores any thrown error.

rebelchris: It would throw a TypeOrmError of foreign key constraint no?
I think you can indeed bypass this above line  (see for example saveHiddenPost function)

rebelchris (approved):
Just one additional remark, further looks good to me :)

src/schema/comments.ts
sshanzel: I tried but the CommentReport schema seems to be missing the FK relating to the `commentId`. Do you want me to update the schema itself?

capJavert: I see post report also missing FK, it might be to avoid reports being deleted when post/comments are deleted?

Let's keep the query in my opinion then, don't wan to add additional work when minor thing.

capJavert (approved):

src/schema/comments.ts
sshanzel: Ah, that makes sense, yes. To not cascade deleting the reports. Thank you both for the reviews! 🚀

rebelchris: Yeah agreed, wasn't aware of the nuance of the missing FK, but let's leave it.
Not a big thing.

idoshamun: @sshanzel can we please select the relevant fields only? In this case it's enough to ask for id. Comment can be big due to the content column

sshanzel: Ah, that is true. I will keep that in check for future fetches.

sshanzel:
@idoshamun I have now refactored the fetch to only include a single column for efficiency 🥳

src/schema/comments.ts
```diff
@@ -785,6 +810,34 @@ export const resolvers: IResolvers<any, Context> = {
...
+      try {
+        await ctx.getRepository(CommentReport).insert({
+          commentId: id,
+          userId: ctx.userId,
+          reason,
+          note,
+        });
+      } catch (err) {
+        if (err?.code !== TypeOrmError.DUPLICATE_ENTRY) {
+          ctx.log.error({ err }, 'failed to save report to database');
```
idoshamun, line 835: Shouldn't we throw the error here?

idoshamun (approved):
Non blocking comment :)

src/schema/comments.ts
sshanzel: We wanted to ignore the error and continue forward. I think was some sort of insert or ignore.

idoshamun: It doesn't really makes sense though. Because if the db is not available for example it will not get saved and the user will see that it succeeded. Is it the same for post report as well?
It's ok to not throw it for duplicates but not for all errors

sshanzel: Yes, currently works the same. We might need to have a discussion on how we're going to tackle these scenarios moving forward. Would raise it in our next web team meeting.

idoshamun: Let's fix them both for now. Should be straightforward. We should throw error when it's not duplicate

sshanzel: Awesome. I'll raise a PR quickly.

PR #1333, opened 2023-07-03
feat: flags migration and update during mutations

capJavert:
Also implements [WT-1484](https://dailydotdev.atlassian.net/browse/WT-1484?atlOrigin=eyJpIjoiYjM5NGRhODcxZGYxNGZhMGFjNjM3OThmZjAwM2MzMjciLCJwIjoiaiJ9)

Changes:
- script to set all the flags in `flags` column per their original values in `post` columns
  - has the option to do it by `createdAt` to avoid deadlocks on db during mass update (because we will have to update all posts)
- added `updateFlagsStatement` (automatically creates update statement that merges current flags value with those that need to be updated f59656c4426953b2cd8be3723d54126ada7c10ca)
  - can be used in `repo.update` easily as we would update other columns for entity
 - for each flag updated all the places where their original columns are set to also update `flags` column
 - exposed public flags through graphql to be accessed on frontend
 - since we are doing stepped migration this does not actually change that `flags` are used during read in different places on the API, that would be done in the separate PR

[WT-1484]: https://dailydotdev.atlassian.net/browse/WT-1484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/common/post.ts
```diff
@@ -158,7 +164,7 @@ export type CreatePost = Pick<
   'title' | 'content' | 'image' | 'contentHtml' | 'authorId' | 'sourceId' | 'id'
 >;
 
-export const saveFreeformPost = async (
+export const createFreeformPost = async (
```
capJavert, line 167: Change to better show for what this function is used for since if it was also for updating it would need separate logic for updating flags (to avoid overwrite).

rebelchris (approved):
Looks solid to me 🙏

src/migration/1688029905946-PostFlagsSync.ts
```diff
@@ -0,0 +1,30 @@
...
+        await queryRunner.query('UPDATE post SET flags = jsonb_set(flags, \'{deleted}\', \'false\'::jsonb) WHERE "deleted" = FALSE')
+
+        await queryRunner.query('UPDATE post SET flags = jsonb_set(flags, \'{private}\', \'true\'::jsonb) WHERE "private" = TRUE')
+        await queryRunner.query('UPDATE post SET flags = jsonb_set(flags, \'{private}\', \'false\'::jsonb) WHERE "private" = FALSE')
+
+        await queryRunner.query('UPDATE post SET flags = jsonb_set(flags, \'{visible}\', \'true\'::jsonb) WHERE "visible" = TRUE')
+        await queryRunner.query('UPDATE post SET flags = jsonb_set(flags, \'{visible}\', \'false\'::jsonb) WHERE "visible" = FALSE')
+
+        await queryRunner.query('UPDATE post SET flags = jsonb_set(flags, \'{showOnFeed}\', \'true\'::jsonb) WHERE "showOnFeed" = TRUE')
+        await queryRunner.query('UPDATE post SET flags = jsonb_set(flags, \'{showOnFeed}\', \'false\'::jsonb) WHERE "showOnFeed" = FALSE')
```
idoshamun, line 23: We should turn it into one query that updates all the relevant posts where flags are not set to the default value. We also have to make sure it will not trigger a CDC update. I think there's an SQL statement that we can run before the update query to disable it for this transaction. But I don't remember precisely the command.

idoshamun (commented):
Let's add a test case for the new `flags` property in the Post type for graphql.
Also it's highly important to monitor the deployment of this PR due to potential high impact of the migration

src/migration/1688029905946-PostFlagsSync.ts
capJavert: > We should turn it into one query that updates all the relevant posts where flags are not set to the default value.

@idoshamun not sure about what you mean here, so you want all the updates in single `UPDATE` statement?

idoshamun: Exactly! But we also have to take into account the CDC side effect to it. If you update the post table it will generate thousands of CDC messages with no actual value. We can run an SQL statement to disable it for the transaction. Anyway happy to go on a call if needed.

src/migration/1687960180394-PostFlags.ts
```diff
@@ -5,17 +5,36 @@ export class PostFlags1687960180394 implements MigrationInterface {
...
+            LANGUAGE PLPGSQL
+            AS
+          $$
+          BEGIN
+            NEW."flags" = '{"sentAnalyticsReport":true,"banned":false,"deleted":false,"private":false,"visible":true,"showOnFeed":true,"promoteToPublic":false}' || NEW."flags";
+            RETURN NEW;
+          END;
+          $$
+        `)
+        await queryRunner.query('CREATE TRIGGER post_default_flags_trigger BEFORE INSERT ON "post" FOR EACH ROW EXECUTE PROCEDURE post_default_flags()')
```
capJavert, line 26: Added so that before insert we merge all the flags provided by insert statement with defaults.

This also takes care of the case where in graphql if we query for example `flags.private` in our `Post` query (as seen in `should return default values for public flags` test) it would return:
```json
{
 "flags": {
    "private": null
  }
}
```

Which I think it is not really good from client side. Granted we could take care of this with graphorm transform but then querying from db directly would again have null values. 

Since currently we had all the fields populated by default for each of the columns I think this should follow case.

This continues on points from https://github.com/dailydotdev/daily-api/pull/1323#discussion_r1253482759

bin/syncProductFlags.ts
```diff
@@ -0,0 +1,28 @@
+import '../src/config';
+import createOrGetConnection from '../src/db';
+
+(async (): Promise<void> => {
+  const createdAfterArgument = process.argv[2];
```
capJavert, line 5: Idea is to run this script in batches to avoid deadlocks on the DB because we will have to update all the rows.

bin/syncProductFlags.ts
```diff
@@ -0,0 +1,43 @@
+import '../src/config';
+import createOrGetConnection from '../src/db';
+
+(async (): Promise<void> => {
+  const createdFromArgument = process.argv[2];
+  const createdToArgument = process.argv[3];
```
capJavert, line 6: Idea is to run this script in batches by createdAt to avoid deadlocks on the DB because we will have to update all the rows.

.infra/application.properties
```diff
@@ -11,11 +11,8 @@ debezium.source.table.include.list=public.comment,public.comment_upvote,public.c
 debezium.source.column.exclude.list=public.post.tsv,public.post.placeholder,public.post.views,public.post.score,public.post.discussionScore,public.notification."readAt"
 debezium.source.plugin.name=pgoutput
 debezium.source.heartbeat.interval.ms=60000
-debezium.transforms=Reroute,filter
+debezium.transforms=Reroute
 debezium.transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
 debezium.transforms.Reroute.topic.regex=(.*)
 debezium.transforms.Reroute.topic.replacement=%topic%
-debezium.transforms.filter.type=io.debezium.transforms.Filter
```
rebelchris, line 18: What is correct here?
I see in the previous PR you introduce it and here it's gone again?

bin/syncProductFlags.ts
rebelchris: I think normally we run it in a stream preferred.
That can handle most things.

src/migration/1689068187186-PostFlagsDefaultValueTrigger.ts
```diff
@@ -0,0 +1,27 @@
...
+            LANGUAGE PLPGSQL
+            AS
+          $$
+          BEGIN
+            NEW."flags" = '{"sentAnalyticsReport":true,"banned":false,"deleted":false,"private":false,"visible":true,"showOnFeed":true,"promoteToPublic":false}' || NEW."flags";
+            RETURN NEW;
+          END;
+          $$
+        `)
+        await queryRunner.query('CREATE TRIGGER post_default_flags_trigger BEFORE INSERT ON "post" FOR EACH ROW EXECUTE PROCEDURE post_default_flags()')
```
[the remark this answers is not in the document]
rebelchris: Just wondering what happens when we decide to introduce a new flag?

rebelchris (commented):
More generic questions.
One suggestion of what we normally run for large entities.
Although @idoshamun probably knows best if there are implications with that.

src/migration/1689068187186-PostFlagsDefaultValueTrigger.ts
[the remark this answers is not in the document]
capJavert: We would update the function in another migration and add the new flag. Because of `CREATE OR REPLACE` that would replace the current function.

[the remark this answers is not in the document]
capJavert: I agree it is clunky but think this is the best way to achieve it since JSON field is not really supported in a best way through typeorm.

.infra/application.properties
capJavert: This removes it since we don't need filtering anymore after migration. It is planned as a one time thing currently to disable CDC and avoid many messages when we do the mass update of `flags` column.

PR #1334, opened 2023-07-04
fix: equality check on tinybird

sshanzel:
Note: do not merge until the other side have been updated to manage the additional column.


src/common/post.ts
```diff
@@ -202,3 +204,24 @@ export const validatePost = (args: ValidatePostArgs): ValidatePostArgs => {
 
   return { title, content };
 };
+
+export const getPostsTinybirdExport = (con: DataSource, latest: Date) =>
```
rebelchris, line 208: Why did you decide to move this.
Maybe keeping it isolated in tinybird cron is more explicit as to what needs review from Viktor moving forward?

Just to keep the constraint of changes in that file need his approval.

idoshamun: Agree

idoshamun (changes requested):
Great initiative to add the tests. Let's just accommodate Chris's feedback before

rebelchris (approved):

vpol:
Could we please keep the order of fields? So in this case can we make `source_type` field go after `post_private`? Thanks

rebelchris:
@vpol Added the latest content_curation change and changed the order.
Please let me know if this is ok for merging

idoshamun (approved):
Looks good to me. @vpol review and let's get it merged

PR #1336, opened 2023-07-04
fix: added promote to public flag mutation

rebelchris:
Decided to make it one uniform toggle for promoting/demoting.
Based of WT-1481 open PR so can only merge once that's in main.

WT-1482 #done 
WT-1483 #done


idoshamun (approved):

capJavert (commented):
Looks good 👍 , I still think that specific mutations (`postPromoteToPublic` and `postDemoteFromPublic` as in DR) would be easier to follow and find from frontend side of things. For example if this mutation adds additional logic like downvote has with blocking it would be much more code paths to follow since it would need to support `true` and `false` vs two different mutations would be written more straightforward (as we have on downvote and cancelDownvote).

rebelchris:
@capJavert I think often with these cases let's tackle them IF they come up.
It's an easy fix then, and makes the code more lean now.

We do follow this approach for other examples to have one uniform flag updating system.

capJavert (approved):
Looks good, one comment and I think just update `should return all the public flags` test's mutation to include `promoteToPublic`.

idoshamun:
@rebelchris please check this: https://dailydotdev.slack.com/archives/C0592NL7YCU/p1689596826331729
before merging

src/migration/1688029905946-PostFlagsSync.ts
```diff
@@ -0,0 +1,30 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class PostFlagsSync1688029905946 implements MigrationInterface {
```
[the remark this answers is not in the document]
rebelchris: Removed :)

rebelchris:
@capJavert / @idoshamun Decided to move to seperate mutations anyway.
Using same function for now, but better extendable.

Also had to modify the key for database abstraction, do note PG doens't have "date" so "text" is the actual representation in jsonb

src/migration/1689851339253-PromoteDateTime.ts
```diff
@@ -0,0 +1,19 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "IDX_post_flags_promoteToPublic"`);
+    await queryRunner.query(
+      `CREATE INDEX "IDX_post_flags_promoteToPublic" ON post USING HASH (((flags->'promoteToPublic')::text))`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "IDX_post_flags_promoteToPublic"`);
+    await queryRunner.query(
```
idoshamun, line 15: This should be a drop statement

src/migration/1689851339253-PromoteDateTime.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class PromoteDateTime1689851339253 implements MigrationInterface {
+  name = 'PromoteDateTime1689851339253';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "IDX_post_flags_promoteToPublic"`);
+    await queryRunner.query(
```
idoshamun, line 8: @capJavert mentioned something about using HASH for text. Should we consider a different index?

src/migration/1689851339253-PromoteDateTime.ts
idoshamun: Oh don't mind me. Sorry!

sshanzel (approved):
Looks good to me! 🚀 

Don't have anything to add on top of the current comments.

src/migration/1689851339253-PromoteDateTime.ts
rebelchris: @capJavert Any suggestion for this from your side?
You probably have more knowledge from your recent research.

capJavert: It comes down to do we want to do `>=` and `<=` queries on that flags column but since we have to use `text` because jsonb does not support date we can leave the `HASH` because we would not be able to for example say `flags->promoteToPublic > '2023-07-07'`.

Not sure if we would have the need to query all the promoted posts to specific date on API side but if we do in that case we might be better off to changing the type to `number` and store timestamp. 

Then we could also move to `BTREE` to support `>=` and `<=` on index level.

rebelchris: We can store isostrings as numbers as well, I read that was the alternative.

rebelchris: @capJavert // @idoshamun We could choose UNIX time, but not sure it implied further complications down the line?
https://stackoverflow.com/a/29269257 (it's a number basically)

capJavert: Yeah I twas thinking of UNIX time yes and then use BTREE. If we needed it for frontend in any case I think we can always transform on graphorm level.

idoshamun: BTREE it is then

src/migration/1689851339253-PromoteDateTime.ts
```diff
@@ -0,0 +1,21 @@
...
+      `CREATE INDEX "IDX_post_flags_promoteToPublic" ON post USING BTREE (((flags->>'promoteToPublic')::integer))`,
```
rebelchris, line 9: Note we have to cast with `-->` operator. You can read more on it [here](https://www.postgresql.org/docs/9.5/functions-json.html#FUNCTIONS-JSON-OP-TABLE) and [here](https://stackoverflow.com/a/39173083).

Also subnote, the type can only be `integer` which doesn't allow `bigint` so we can't store milliseconds, hence I store it as seconds value.
For this specific case it should be ok, but good to note.

rebelchris:
@idoshamun and @capJavert Adjusted, please see my remarks, would like to hear if you see any objections else we can go.

Ido: I decided to automatically add 7 days on API level as you are the only user.

src/migration/1689851339253-PromoteDateTime.ts
idoshamun: Thanks for mentioning. We have to make sure platform team reads it properly

src/schema/posts.ts
```diff
@@ -203,6 +203,11 @@ export const typeDefs = /* GraphQL */ `
     Whether the post's source is private or not
     """
     private: Boolean
+
+    """
+    The unix timestamp (seconds) the post will be promoted to public to
+    """
+    promoteToPublic: Int
```
idoshamun, line 210: Do we want to transform it to Date before sending back to the client? Also not sure we even want to provide it to public. So maybe we should remove it even

idoshamun (changes requested):
Just realized we also need to change this condition to include flags changes. Otherwise it will not reflect in the feed.
https://github.com/dailydotdev/daily-api/blob/main/src/workers/cdc.ts#L322

src/schema/posts.ts
rebelchris: We only want it to conditionally render the action to promote or demote not actually use it as it's value

idoshamun: Can we make sure then it gets back only the system mods? I think our auth directive supports it

rebelchris: @idoshamun I can only find that option for the whole query, not particularly one field, but maybe I'm missing some knowledge on that.
I also know we can do the nullIfNotLoggedIn kind of thing, but not sure that supports this set.

idoshamun: Yes, we can also do that as part of the graphorm if you wish. Because right now the system mods are in-memory it should be easy to do. You can check this function signature, I think you get there the context as well as the query response. Let me know if you need more info

rebelchris:
@idoshamun All concerns should be resolved now

idoshamun (approved):

PR #1338, opened 2023-07-05
fix: public property now correctly calculated from private

capJavert:
While working on new squad create and edit pages I noticed `public` was always `true`.


rebelchris (approved):
Interesting we didn't see this before 🤯
Learning curve to always do test both cases I guess.

idoshamun (approved):
Amazing catch!

PR #1340, opened 2023-07-06
feat: first visit in cache

sshanzel:
We now generate the `firstVisit` and turned it into a number to make it easier to consume at the frontend.

Note: we might need another value in the cache in the future relating to the user, should we probably store an object instead (with `firstVisit` as its property) so that we will only have to fetch once whenever we introduce new properties instead of touching the cache twice? Thoughts?

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -11,3 +11,13 @@ export const fallbackImages = {
 };
 
 export const REDIS_CHANGELOG_KEY = 'boot:latest_changelog';
+
+export enum StorageTopic {
+  Boot = 'boot',
+}
+
+export const generateStorageKey = (
+  topic: StorageTopic,
+  key: string,
+  identifier = 'global',
+): string => `${topic}:${key}:${identifier}`;
--- a/src/config.ts
+++ b/src/config.ts
@@ -19,5 +19,5 @@ export enum StorageTopic {
 export const generateStorageKey = (
   topic: StorageTopic,
   key: string,
-  identifier = 'global',
+  identifier: string, // mostly used for user id - "global" for global keys
 ): string => `${topic}:${key}:${identifier}`;
--- a/src/redis.ts
+++ b/src/redis.ts
@@ -44,8 +44,16 @@ export function deleteKeysByPattern(pattern: string): Promise<void> {
   );
 }
 
-export const setRedisObject = (key, value) =>
+export const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
+
+export const setRedisObject = (key: string, value: string | Buffer | number) =>
   ioRedisPool.execute((client) => client.set(key, value));
+
+export const setTimedRedisObject = (
+  key: string,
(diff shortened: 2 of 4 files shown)


src/config.ts
```diff
@@ -11,3 +11,13 @@ export const fallbackImages = {
...
+
+export enum StorageTopic {
+  Boot = 'boot',
+}
+
+export const generateStorageKey = (
+  topic: StorageTopic,
+  key: string,
+  identifier = 'global',
+): string => `${topic}:${key}:${identifier}`;
```
sshanzel, line 23: We have a function to generate storage keys in the `apps` repo, so I've thought of doing the same here.

src/redis.ts
```diff
@@ -44,8 +44,16 @@ export function deleteKeysByPattern(pattern: string): Promise<void> {
   );
 }
 
-export const setRedisObject = (key, value) =>
+export const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
+
+export const setRedisObject = (key: string, value: string | Buffer | number) =>
   ioRedisPool.execute((client) => client.set(key, value));
 
+export const setTimedRedisObject = (
```
idoshamun, line 52: The name is a bit confusing. It's about setting expiration to an object. `Timed` is confusing

src/routes/boot.ts
```diff
@@ -57,8 +66,13 @@ export type BootUserReferral = Partial<{
   referralOrigin?: string;
 }>;
 
+interface AnonymousUser extends BootUserReferral {
+  id: string;
+  firstVisit: number;
```
idoshamun, line 71: Why don't we want to return a string here?
We always return a string for dates

src/config.ts
```diff
@@ -11,3 +11,13 @@ export const fallbackImages = {
...
 export const REDIS_CHANGELOG_KEY = 'boot:latest_changelog';
+
+export enum StorageTopic {
+  Boot = 'boot',
+}
+
+export const generateStorageKey = (
+  topic: StorageTopic,
+  key: string,
+  identifier = 'global',
```
idoshamun, line 22: Why do we set a default value here?

idoshamun (changes requested):
Overall looks great but I'm missing some tests.
Your sidenote is very relevant. Let's keep it in mind in case we have more properties we need to store

src/config.ts
sshanzel: Ah, in FE, it is common to have such values to be consistent across users. In BE, I realized the changelog was the same case, so I kept it. Though we can make it required here in BE since it is not as frequent as in FE.

src/routes/boot.ts
sshanzel: Right. Let's stick to using strings. Can't remember exactly which string type we were using (so I used the number as it is easy to remember and works cross-platform), found out it was ISO string.

idoshamun (approved):
Amazing!

PR #1341, opened 2023-07-07
feat: add referralUrl field to source

capJavert:
Adding new `referralUrl` field to source to be used on frontend for invite links. Supports public and private squads.

Added in request caching with [dataloader](https://github.com/graphql/dataloader). 

Idea was to also create a generic service `DataLoaderService` that we can extend with additional loaders to provide in request caching capabilities. 

What does this enable: we can batch a lot of same database calls, 3party services calls and any promise to be batched and fetched only once in request context.

Example here is that we cache calls to our URL shortening service to avoid repeated calls while resolving multiple sources.

### DIFF
--- a/src/dataLoaderService.ts
+++ b/src/dataLoaderService.ts
@@ -4,7 +4,7 @@ import { getShortUrl } from './common';
 import { SourceMember } from './entity';
 import { GQLSource } from './schema/sources';
 
-const defaultCacheKeyFn = <K>(key: K) => {
+export const defaultCacheKeyFn = <K>(key: K) => {
   if (typeof key === 'object') {
     return JSON.stringify(key);
   }
--- a/src/dataLoaderService.ts
+++ b/src/dataLoaderService.ts
@@ -1,6 +1,16 @@
 import DataLoader, { BatchLoadFn } from 'dataloader';
 import { Context } from './Context';
 import { getShortUrl } from './common';
+import { SourceMember } from './entity';
+import { GQLSource } from './schema/sources';
+
+const defaultCacheKeyFn = <K>(key: K) => {
+  if (typeof key === 'object') {
+    return JSON.stringify(key);
+  }
+
+  return key.toString();
+};
 
 export class DataLoaderService {
   protected loaders: Record<string, DataLoader<unknown, unknown>>;
@@ -14,9 +24,11 @@ export class DataLoaderService {
   protected getLoader<K, V>({
     type,
     loadFn,
+    cacheKeyFn,
   }: {
     type: string;
     loadFn: (params: K) => Promise<V> | V;
+    cacheKeyFn: (key: K) => string;
   }): DataLoader<K, V> {
     if (!this.loaders[type]) {
       const batchLoadFn: BatchLoadFn<K, V> = async (keys) => {
@@ -32,13 +44,7 @@ export class DataLoaderService {
       };
 
       this.loaders[type] = new DataLoader(batchLoadFn, {
-        cacheKeyFn: (key: K) => {
-          if (typeof key === 'object') {
-            return JSON.stringify(key);
-          }
-
-          return key.toString();
-        },
+        cacheKeyFn,
         maxBatchSize: 30,
         name: `${DataLoaderService.name}.${type}`,
       });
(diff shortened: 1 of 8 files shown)


src/schema/sources.ts
```diff
@@ -814,7 +821,26 @@ export const resolvers: IResolvers<any, Context> = {
...
+          process.env.COMMENTS_PREFIX,
+        );
+
+        if (source.public) {
+          referralUrl.pathname = `/squads/${source.handle}`;
+          referralUrl.searchParams.append('cid', 'squad');
+          referralUrl.searchParams.append('userid', ctx.userId);
+        }
+
+        source.referralUrl = await getShortUrl(referralUrl.toString(), ctx.log);
```
idoshamun, line 840: This ideally should be part of a graphql field resolver for `referralUrl` or as part of graphorm.
Implementing it only here can result in inconsistencies throughout our api

capJavert: @idoshamun Yeah, understand the concern. 

We would have to optimize `getShortUrl` though to avoid it being run multiple times during single request in that case. I did try in graphql fields and it would run multiple times for each entity inside feed. I used https://vlio20.github.io/utils-decorators/#memoizeAsync for that kind of cases before. Thoughts?

idoshamun: We can do an in memory cache in the context or use dataloaders. I think it should solve our problem

capJavert: Yeah, memoizeAsync does the same thing. Never used dataloaders in graphql if you mean this? https://github.com/graphql/dataloader I'll give it a go.

src/schema/sources.ts
```diff
@@ -1256,5 +1262,48 @@ export const resolvers: IResolvers<any, Context> = {
...
+              select: ['referralToken'],
+              where: { sourceId: source.id, userId: ctx.userId },
+            });
+
+          referralToken = sourceMember?.referralToken;
+        }
+
+        if (!referralToken) {
+          return null;
+        }
```
capJavert, line 1297: Wandering if there is a better way to request this in terms of our graphql resolver. 

My logic is that of course `currentMember` might not be available in query context so I have to refetch but it feels clunky.

Also should I just use new `DataLoaderService` to have the whole `referralUrl` cached in request.

src/schema/sources.ts
capJavert: Implemented in 8b036510b771a53773d17f6b75d34b4b4ec981da

src/dataLoaderService.ts
```diff
@@ -0,0 +1,56 @@
...
+export class DataLoaderService {
+  protected loaders: Record<string, DataLoader<unknown, unknown>>;
+  private ctx: Context;
+
+  constructor({ ctx }: { ctx: Context }) {
+    this.loaders = {};
+    this.ctx = ctx;
+  }
+
+  protected getLoader<K, V>({
```
idoshamun, line 14: This is a very specific implementation but that's ok for now. I don't expect to have too many dataloaders. Thankfully graphorm solves a lot of issues

src/schema/sources.ts
idoshamun: If `currentMember` is available in the `source` parameter we should use it. It's currently missing in your implementation. Another option is to move `referralUrl` to be inside the `SourceMember` object instead of `Source`. It will make the implementation a bit nicer I believe, but up to you to decide.

As for the second question I'm not sure I understand. Can you elaborate please?

Also, have you checked that the dataloader works in terms of reducing the number of requests?

idoshamun (commented):
Overall implementation looks great!

src/schema/sources.ts
idoshamun: > Other way around it would need to source.handle which is again not guaranteed.

You're right. So let's keep it in `source` but just add a check to see if it was already requested. Please make sure that it's indeed doesn't generate another query.

> Also should I just use new DataLoaderService to have the whole referralUrl cached in request.
Yes. It's better this way I believe. Most probably it will rarely be helpful but better have it there

src/dataLoaderService.ts
capJavert: Yeah, I think graphorm works really well for anything db related but when you have 3party service calls (like shortener) it becomes inefficient.

For my knowledge, I would be Interested in what terms you think this is specific implementation? We can also catch up on a short call outside of this to discuss. I would appreciate it.

rebelchris (approved):

src/dataLoaderService.ts
```diff
@@ -0,0 +1,114 @@
...
+    type,
+    loadFn,
+    cacheKeyFn,
+  }: {
+    type: string;
+    loadFn: (params: K) => Promise<V> | V;
+    cacheKeyFn: (key: K) => string;
```
capJavert, line 31: I made `cacheKeyFn` required when creating a new loader so that we can customize the keys per loader.

For example `referralUrl` data loader accepts `source` but if we used the default cache key fn the key would actually contain many properties depending on the `source` object you provided which would then create cache misses because `source` which has `currentMember` and `source` without it would be different cache keys.

By providing custom cache key below we can say that for `referralUrl` we wanna cache per user per source and everything else is irrelevant.

idoshamun (approved):
Amazing! Love your work on the dataloader

PR #1342, opened 2023-07-07
refactor: throwing of error when unable to save to db

sshanzel:
When the error is not about duplicated entries, we should throw an error.

### DIFF
--- a/src/schema/comments.ts
+++ b/src/schema/comments.ts
@@ -832,7 +832,7 @@ export const resolvers: IResolvers<any, Context> = {
         });
       } catch (err) {
         if (err?.code !== TypeOrmError.DUPLICATE_ENTRY) {
-          ctx.log.error({ err }, 'failed to save report to database');
+          throw new Error('Failed to save report to database');
         }
       }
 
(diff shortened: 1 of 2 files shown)


TomDowling (approved):
Looks good to me

src/schema/comments.ts
```diff
@@ -832,7 +832,7 @@ export const resolvers: IResolvers<any, Context> = {
         });
       } catch (err) {
         if (err?.code !== TypeOrmError.DUPLICATE_ENTRY) {
-          ctx.log.error({ err }, 'failed to save report to database');
```
rebelchris, line 835: I think we still need the logs, don't think the error does that but might be wrong

capJavert: It should be caught by `fastify.setErrorHandler` and respond with `500` https://github.com/dailydotdev/daily-api/blob/7c6d96af88937563dfadc28aaae33c21dd103651/src/index.ts#L43-L46

idoshamun: Yap, Ante's right

idoshamun (approved):

rebelchris (approved):

PR #1344, opened 2023-07-11
feat: identify pre onboarding v2 users

sshanzel:
To make it easier to identify if the user is valid for v2 or not, we calculate the dates on the API instead of FE to avoid any discrepancy.

### DIFF
--- a/src/featureFlags.ts
+++ b/src/featureFlags.ts
@@ -87,8 +87,7 @@ const getSubmitArticleState = (flags: IFlags, user: User) => {
 
 interface UpdateParam {
   feature: string;
-  replacement: Flag['value'];
-  onCheckValidity(value: Flag['value']): boolean;
+  checkIsApplicable(value: Flag['value']): boolean;
 }
 
 export const adjustAnonymousFlags = (
@@ -97,9 +96,11 @@ export const adjustAnonymousFlags = (
 ): IFlags => {
   const updatedFlags = { ...flags };
 
-  updates.forEach(({ feature, onCheckValidity, replacement }) => {
-    if (flags?.[feature]?.enabled && onCheckValidity(flags[feature].value)) {
-      updatedFlags[feature] = { enabled: true, value: replacement };
+  updates.forEach(({ feature, checkIsApplicable }) => {
+    if (!flags?.[feature]?.enabled) return;
+
+    if (!checkIsApplicable(flags[feature].value)) {
+      updatedFlags[feature] = { enabled: false, value: '' };
     }
   });
 
--- a/src/featureFlags.ts
+++ b/src/featureFlags.ts
@@ -3,6 +3,7 @@ import pTimeout from 'p-timeout';
 import flagsmith, { IFlags } from './flagsmith';
 import { DataSource } from 'typeorm';
 import { Feature, User } from './entity';
+import { Flag } from 'flagsmith-nodejs/sdk/models';
 
 const FLAGSMITH_TIMEOUT = 1000;
 export const DEFAULT_FLAGS = {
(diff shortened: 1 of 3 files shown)


src/routes/boot.ts
```diff
@@ -359,9 +363,11 @@ const anonymousBoot = async (
     middleware ? middleware(con, req, res) : {},
     getAnonymousFirstVisit(req.trackingId),
   ]);
+  const isPreOnboardingV2 = new Date(firstVisit) < onboardingV2Requirement;
   return {
     user: {
       firstVisit,
+      isPreOnboardingV2,
```
rebelchris, line 370: Just a suggestion why dont you immediately change the flag value in api?
We have done this before so from the frontend you'll only have to rely on flags

sshanzel: Oh, never knew we've done that in the past, but we can definitely do that to make it easier.

src/routes/boot.ts
```diff
@@ -359,14 +366,26 @@ const anonymousBoot = async (
...
   ]);
+  const isPreOnboardingV2 = firstVisit
+    ? new Date(firstVisit) < onboardingV2Requirement
+    : false;
+
+  const anonymousFlags = adjustAnonymousFlags(flags, [
+    {
+      replacement: 'control',
+      feature: 'onboarding_v2',
+      onCheckValidity: (value) => value !== 'control' && isPreOnboardingV2,
```
rebelchris, line 377: @idoshamun just one mention for this experiment that it won't show to x% of selected numbers as we further down it here.
Just wanted to mention it for alignment.

src/routes/boot.ts
```diff
@@ -346,6 +350,9 @@ const getAnonymousFirstVisit = async (trackingId: string) => {
   return finalValue;
 };
 
+// We released the firstVisit at July 10, 2023 - so to have a buffer time, the requirement is the day after.
+export const onboardingV2Requirement = new Date(2023, 6, 11);
```
idoshamun, line 354: We will have to update it once we want to start the experiment. Just noting it here to remember

src/featureFlags.ts
```diff
@@ -84,6 +85,27 @@ const getSubmitArticleState = (flags: IFlags, user: User) => {
   return flags?.submit_article;
 };
 
+interface UpdateParam {
+  feature: string;
+  replacement: Flag['value'];
+  onCheckValidity(value: Flag['value']): boolean;
```
idoshamun, line 91: The name here is a bit confusing. I'd suggest changing it to `isUserApplicable`. It should return `true` if the we should keep the flag as is, and `false` if not. So in the case of the onboarding, if the user visited before the date they shouldn't be applicable and we return false.

src/featureFlags.ts
```diff
@@ -84,6 +85,27 @@ const getSubmitArticleState = (flags: IFlags, user: User) => {
...
+
+export const adjustAnonymousFlags = (
+  flags: IFlags,
+  updates: UpdateParam[],
+): IFlags => {
+  const updatedFlags = { ...flags };
+
+  updates.forEach(({ feature, onCheckValidity, replacement }) => {
+    if (flags?.[feature]?.enabled && onCheckValidity(flags[feature].value)) {
+      updatedFlags[feature] = { enabled: true, value: replacement };
```
idoshamun, line 102: If the user is not applicable for the flag, I suggest setting it to `enabled: false` at this point without value. This way we know in the data that this user is not part of an experiment

src/featureFlags.ts
sshanzel: Agreed, turning the value to false will be much better.

src/featureFlags.ts
```diff
@@ -84,6 +85,28 @@ const getSubmitArticleState = (flags: IFlags, user: User) => {
...
+  flags: IFlags,
+  updates: UpdateParam[],
+): IFlags => {
+  const updatedFlags = { ...flags };
+
+  updates.forEach(({ feature, checkIsApplicable }) => {
+    if (!flags?.[feature]?.enabled) return;
+
+    if (!checkIsApplicable(flags[feature].value)) {
+      updatedFlags[feature] = { enabled: false, value: '' };
```
idoshamun, line 103:
```suggestion
      updatedFlags[feature] = { enabled: false, value: null };
```

src/routes/boot.ts
capJavert: Could we not created flagsmith flag for these kind of things? Feels clunky to leave it like this.

rebelchris (approved):
Nothing from my side, just reminder to change the date when you decide to launch 🙏

sshanzel:
Looking nice!

image

PR #1347, opened 2023-07-12
fix: move to graphORM query for support

rebelchris:
Moved the sources query (unused for now) to graphORM paginated query to support the custom types we have there.
The one we needed was `membersCount` it also is more inline with our existing queries.


idoshamun:
@rebelchris it looks great! But I'm holding my approval until the incident resolves

idoshamun (approved):

PR #1348, opened 2023-07-12
feat: filter read messages in cdc

capJavert:
During debezium upgrade to 2.0 we encountered a lot of snapshot messages with `op=r` property. 

This ensures those messages are ignored and not sent to pubsub and cdc worker.

Before merging this we need to test out locally that debezium 2.0 starts-up ok and that this fix really handles all the excess messages.


capJavert:
@idoshamun this filter handles read operation filtering. When we confirm this works we can return the flags migration related filter as the separate transform.

idoshamun (approved):

PR #1351, opened 2023-07-17
feat: notification preference entity

sshanzel:
- As mentioned in the ticket, convert the type `NotificationType` to an `enum`.
- Introduced `NotificationPreference` entity to accommodate user's preference towards notification for specific types.
- The DR only mentions the `userId` and `notificationType` to be primary keys, but we need to add more criteria to find the uniqueness to avoid errors as users can have the same notification type registered in the table. For example, you want to mute two different comment threads that you've been a part of. Both rows will have the same `userId` and `notificationType`, so in this case, I added the unique ID that the preference is referring to, with regards to the example, it means the `commentId`.
- I have also created a `/notifications` directory in the entities folder.

WT-1522 #done


src/migration/1689590210380-NotificationPreference.ts
```diff
@@ -0,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
sshanzel, line 1: Currently looking at why the migration didn't include the FKs to cascade on delete.

src/entity/notifications/NotificationPreferenceComment.ts
```diff
@@ -0,0 +1,13 @@
+import { ChildEntity, ManyToOne, PrimaryColumn } from 'typeorm';
+import { NotificationPreference } from './NotificationPreference';
+import { NotificationPreferenceType } from '../../notifications/common';
+import { Comment } from '../Comment';
+
+@ChildEntity(NotificationPreferenceType.Comment)
+export class NotificationPreferenceComment extends NotificationPreference {
+  @PrimaryColumn({ type: 'text', default: null })
+  commentId?: string;
```
idoshamun, line 9: This should not be optional in case it's a `NotificationPreferenceComment`. We must have a commentId for that

src/entity/notifications/NotificationPreferencePost.ts
```diff
@@ -0,0 +1,13 @@
+import { ChildEntity, ManyToOne, PrimaryColumn } from 'typeorm';
+import { Post } from '../posts';
+import { NotificationPreferenceType } from '../../notifications/common';
+import { NotificationPreference } from './NotificationPreference';
+
+@ChildEntity(NotificationPreferenceType.Post)
+export class NotificationPreferencePost extends NotificationPreference {
+  @PrimaryColumn({ type: 'text', default: null })
+  postId?: string;
```
idoshamun, line 9: Same comment here

src/entity/notifications/NotificationPreferenceSource.ts
```diff
@@ -0,0 +1,13 @@
+import { ChildEntity, ManyToOne, PrimaryColumn } from 'typeorm';
+import { Source } from '../Source';
+import { NotificationPreferenceType } from '../../notifications/common';
+import { NotificationPreference } from './NotificationPreference';
+
+@ChildEntity(NotificationPreferenceType.Source)
+export class NotificationPreferenceSource extends NotificationPreference {
+  @PrimaryColumn({ type: 'text', default: null })
```
idoshamun, line 8: And here

idoshamun (changes requested):
The primary key is a great catch @sshanzel! Something we didn't think through during the DR.
The problem with the current implementation is that every new subtype should expand the primary key. If we introduce new subtype in the future we will have to recalculate the primary key which is a bad practice.
My suggestion is to introduce a new column in the base type `uniqueKey`.
We will create a primary key using `userId, notificationType, uniqueKey`. This will ensure no dups. We will set `uniqueKey` to be the id of the relevant object.
Tagging @capJavert as he wrote the DR

capJavert:
> The primary key is a great catch @sshanzel! Something we didn't think through during the DR. The problem with the current implementation is that every new subtype should expand the primary key. If we introduce new subtype in the future we will have to recalculate the primary key which is a bad practice. My suggestion is to introduce a new column in the base type `uniqueKey`. We will create a primary key using `userId, notificationType, uniqueKey`. This will ensure no dups. We will set `uniqueKey` to be the id of the relevant object. Tagging @capJavert as he wrote the DR

Ah this kinda stayed open from initial idea to use entity with multiple tables. Good catch indeed!

@idoshamun but can we have different constraints for this `uniqueKey` column for different entities, not sure that is possible and without that we lose cascade on delete for example?

idoshamun:
> @idoshamun but can we have different constraints for this uniqueKey column for different entities, not sure that is possible and without that we lose cascade on delete for example?

@capJavert I'm not suggesting to change the current structure suggested by Lee. So we will keep the foreign keys as they are. But we will duplicate them into the `uniqueKey` field. So in the case of Source preference `uniqueKey` and `sourceId` will be the same, in the case of Post then `postId` and `uniqueKey` will be the same

capJavert:
> > @idoshamun but can we have different constraints for this uniqueKey column for different entities, not sure that is possible and without that we lose cascade on delete for example?
> 
> @capJavert I'm not suggesting to change the current structure suggested by Lee. So we will keep the foreign keys as they are. But we will duplicate them into the `uniqueKey` field. So in the case of Source preference `uniqueKey` and `sourceId` will be the same, in the case of Post then `postId` and `uniqueKey` will be the same

Gotcha, didn't understand the usage of `uniqueKey`. Tnx for clarifying 🤝

sshanzel:
I also missed it during the review. Agreed. Let's do the `uniqueKey` and keep the FKs which is beneficial for cascading delete.

idoshamun (approved):
Awesome!!
@capJavert please note in the DR that's better to query this table using `uniqueKey` and not the specific `sourceId`, `postId` and so on due to the primary key. We want to utilize the indexes as much as possible

src/entity/notifications/NotificationPreference.ts
```diff
@@ -0,0 +1,28 @@
...
+  NotificationPreferenceStatus,
+  NotificationPreferenceType,
+  NotificationType,
+} from '../../notifications/common';
+
+@Entity()
+@TableInheritance({ column: { type: 'varchar', name: 'type' } })
+export class NotificationPreference {
+  @PrimaryColumn({ type: 'text' })
+  uniqueKey: string;
```
rebelchris, line 12: Would this not work as simply using existing naming for `id`?
Or what do you plan to merge as the uniqueKey?

rebelchris (approved):
Just one non-blocking comment

src/entity/notifications/NotificationPreference.ts
sshanzel: Yes, since we will have the constraint on the dedicated child entities. This is so we will have the uniqueness of the row as adding new primary keys (new child entities) over time will have an additional overhead resource-wise. 

> I assumed at first we just have a generated ID as it adds no value correct?

Yes, it holds no "value" in terms of definition, either generated id or using the reference id should work.

sshanzel: @rebelchris merging this in a few moments. Let me know if you have further concerns 🥳

rebelchris: Fine if everyone else agrees :)
Not really blocking, just a stickler for naming that get's reused 🙏

capJavert: I am fine with `referenceId`

idoshamun: Fine as well

PR #1352, opened 2023-07-18
feat: notification preferences query

sshanzel:
As we introduced a notification preferences entity instead of just the muted ones, we should now support querying the said entity.

One concern though, the preferences here can grow in time, pulling all the preferences on every explicit subscription or muted can become a lot over time. Though on the initial rollout, there shouldn't be much to worry about (since we also filter by type) but just wanted to have it mentioned.

Thoughts are, we keep this or make it paginated though we will introduce another query to fetch a single preference based on the referenced id and the `notificationType` to identify the status and use it however the consuming end wants (verifying subscribed or muted already and reversing the action).

WT-1525 #done


idoshamun:
@sshanzel what's the use case of querying them all?

sshanzel:
> @sshanzel what's the use case of querying them all?

None yet at the initial release, but if we will ever present a list of notifications preferences, to provide options to users like removing muted selections, etc.

sshanzel:
> @sshanzel if it's not needed I wouldn't proceed with this PR as it can potentially be a huge object as you mentioned. Is it part of the DR?

Not it was not, let's enforce the `type` requirement.

idoshamun:
If it's not needed then I suggest to close the PR at this point.
There's no need to add additional functionality that is not required at the moment.
Let's be more aware next time 🙏

sshanzel:
> If it's not needed then I suggest to close the PR at this point.
There's no need to add additional functionality that is not required at the moment.
Let's be more aware next time 🙏

Oh, this PR was required from the DR, just the requirement of having an optional `type` on the query.

sshanzel:
Would love to hear what you guys think. 

cc @idoshamun @rebelchris @capJavert

rebelchris (commented):
Nothing much on the code itself, but also concerned this might become very big.
At what point would we decide to query this?

As a once-off and then map to items on the frontend?
Alternative would be to only query the specific ID and get 1 result? (As it's only invoked on user action (options menu))

Sorry should have been tackled in the DR, but having some concerns with load of this massive query (potentially)

sshanzel:
Should now be accurate with the updates from the DR.

__tests__/notifications.ts
```diff
@@ -291,6 +302,118 @@ describe('query notifications', () => {
...
+  `;
+
+  it('should not authorize when not logged-in', () =>
+    testQueryErrorCode(
+      client,
+      { query: QUERY, variables: { data: [] } },
+      'UNAUTHENTICATED',
+    ));
+
+  const prepareNotificationPreferences = async () => {
```
rebelchris, line 325: I would suggest making this more extended for future proofing.

For instance:
- add multiple of the same type for different posts but same user
- also have on test that only queries one of the types and other should not auto return


Its just a failsafe for when code gets accidentally changed
Non blocking so will approve already

rebelchris (approved):
Nothing further from my side, just minor test suggestion

__tests__/notifications.ts
sshanzel: > add multiple of the same type for different posts but same user

Sure. Let me add another one.

> also have on test that only queries one of the types and other should not auto return

We should have this covered as we already have other data for user `1` but with `source` type. First first test's return length was just one and verified to have the correct type (post).

sshanzel: Added! Thanks for the review! 🚀

src/schema/notifications.ts
```diff
@@ -250,6 +316,17 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+    notificationPreferences: (
+      _,
+      { data }: NotificationPreferenceInput,
+      { con, userId },
+    ): Promise<GQLNotificationPreference[]> => {
+      const params = data.reduce((args, value) => {
+        return [...args, { ...value, userId }];
+      }, []);
+
+      return con.getRepository(NotificationPreference).find({ where: params });
```
idoshamun, line 328: Please use graphorm. It's much safer.

sshanzel: Agreed. Should be updated now.

idoshamun (approved):

PR #1353, opened 2023-07-18
feat: muting notification

sshanzel:
We created a map that defines which type of `NotificationPreference` a `NotificationType` belongs to. This is to understand whether we should put the `uniqueKey` on the designated properties unique for each entity.

It was mentioned to include a mutation to `unmute` a preference. I thought of making it a more general kind of unmuting since its sole purpose would be to remove the actual row, I have instead renamed it a bit to be applicable to removing `subscribed` in the future. Though I am cool if we would want this to be just removing `muted` rows only, just let me know.


src/notifications/common.ts
```diff
@@ -35,3 +45,66 @@ export enum NotificationPreferenceType {
...
+  NotificationPreferenceSource;
+
+export const saveNotificationPreference = async (
+  con: DataSource | EntityManager,
+  userId: string,
+  referenceId: string,
+  notificationType: NotificationType,
+  status: NotificationPreferenceStatus,
+) => {
+  const type = notificationPreferenceMap[notificationType];
```
rebelchris, line 69: Can you test this by passing null or something.
Think it might be breaking here already without coming to the validationError below.

src/notifications/common.ts
```diff
@@ -35,3 +45,66 @@ export enum NotificationPreferenceType {
...
+
+  const params: Partial<NotificationPreferenceUnion> = {
+    type,
+    userId,
+    status,
+    notificationType,
+    uniqueKey: referenceId,
+  };
+
+  switch (type) {
```
rebelchris, line 83: Why don't we use a map here, I think they look slightly cleaner and easier to add to 😂 (but maybe i'm lazy)

rebelchris (approved):
Looks good, nothing breaking.
Minor suggestion

src/notifications/common.ts
sshanzel: Shouldn't break as it is a valid key, then from the map it will return `undefined`.

Also, from GraphQL standpoint it should break as it is a required parameter.

src/notifications/common.ts
sshanzel: Not sure how we can do that as we are trying to set some values. Can you give an example?

.infra/Pulumi.adhoc.yaml
```diff
@@ -28,5 +28,6 @@ config:
     typeormUser: postgres
     tz: UTC
     urlPrefix: http://api.local.com
+  api:image: api-image:tilt-3eb951d47824a3d1
```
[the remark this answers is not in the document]
sshanzel: Oops, that's right 😅

src/notifications/common.ts
```diff
@@ -35,3 +45,66 @@ export enum NotificationPreferenceType {
...
+      .insert()
+      .values(params)
+      .orIgnore()
+      .execute();
+  } catch (err) {
+    if (err.code === TypeOrmError.FOREIGN_KEY) {
+      throw new NotFoundError('Invalid reference id');
+    }
+
+    throw new Error('Something went wrong');
```
idoshamun, line 108: Please throw the original error so we keep info of what happened
```suggestion
    throw err;
```

idoshamun (approved):
Please fix my suggestion before merging but looks great

src/notifications/common.ts
rebelchris: Very quick and dirty, but something like this, let me know what you think.

```js
const type = 'a';
const params = {};
const referenceId = '123';

const typeToParams = {
  'a': 'commentId'
};

params[typeToParams[type]] = referenceId;

console.log(params);
```

sshanzel: Let's see how we can turn that to typescript.

PR #1354, opened 2023-07-18
revert: debezium flags filter

capJavert:
Reverted after migration was executed on production through script.


idoshamun (approved):
Thanks!

PR #1356, opened 2023-07-19
feat: use growthbook as an experimentation platform

idoshamun:
Add support for growthbook.io as experimentation and feature flags platform. This will replace flagsmith and will allow us much more flexible experiments.

Boot now includes a new object `exp` that contains the recent active experiments of the user (as `e`) and encrypted features settings (as `f`).

The frontend uses it to decide what values to value to assign the feature.

Still WIP as I'm not refreshing the flags when necessary. There should be real-time support but I'm still checking it

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -113,12 +113,14 @@ const readinessProbe: k8s.types.input.core.v1.Probe = {
   httpGet: { path: '/health', port: 'http' },
   failureThreshold: 2,
   periodSeconds: 2,
+  initialDelaySeconds: 20,
 };
 
 const livenessProbe: k8s.types.input.core.v1.Probe = {
   httpGet: { path: '/liveness', port: 'http' },
   failureThreshold: 3,
   periodSeconds: 5,
+  initialDelaySeconds: 20,
 };
 
 let appsArgs: ApplicationArgs[];


src/workers/experimentAllocated.ts
```diff
@@ -0,0 +1,57 @@
...
+    const data: Data = messageToJson(message);
+    await ioRedisPool.execute(async (client) => {
+      const key = `exp:${data.user_id}`;
+      const exps = parseHash(await client.hgetall(key));
+      const current = {
+        variation: data.variation_id,
+        timestamp: new Date(data.server_timestamp),
+      };
+      exps[data.experiment_id] = current;
+      const drop = keysToDrop(exps);
```
capJavert, line 45: This is to only ever have last X experiments in redis?

src/routes/index.ts
```diff
@@ -32,7 +32,12 @@ export default async function (fastify: FastifyInstance): Promise<void> {
 
   // Debugging endpoint
   fastify.post('/e', (req, res) => {
-    req.log.info({ body: req.body }, 'events received');
+    req.log.debug({ body: req.body }, 'events received');
+    return res.status(204).send();
+  });
+
+  fastify.post('/e/x', (req, res) => {
+    req.log.debug({ body: req.body }, 'allocation received');
```
capJavert, line 40: This should post `api.experiment-allocated` topic in the future?

capJavert (commented):
Looks simple enough to me. Question, when should `api.experiment-allocated` topic be posted?

src/routes/boot.ts
```diff
@@ -361,11 +386,12 @@ const anonymousBoot = async (
   middleware?: BootMiddleware,
   shouldLogout = false,
 ): Promise<AnonymousBoot> => {
-  const [visit, flags, extra, firstVisit] = await Promise.all([
+  const [visit, flags, extra, firstVisit, exp] = await Promise.all([
     visitSection(req, res),
     getUserFeatureFlags(req, con),
     middleware ? middleware(con, req, res) : {},
     getAnonymousFirstVisit(req.trackingId),
+    getExperimentation(req.trackingId),
```
rebelchris, line 394: Just wondering does growthbook allow the conversion from trackingId to userId eventually?

rebelchris (approved):
Nothing crazy from my side.
Keen to have a more advanced system in play

src/routes/index.ts
idoshamun: Nope, it's just a debugging endpoint for local development. The real endpoint is in analytics-api

src/routes/boot.ts
idoshamun: Ideally they should be the same. We use `trackingId` as `userId` once the user registers

PR #1357, opened 2023-07-19
feat: skipping muted notifications

sshanzel:
Skip notifications based on notification types as below:
- `comment_reply`
- `squad_reply`
- `squad_member_joined`
- `squad_post_added`
- `article_new_comment`

~~WIP: still waiting on a response from the threads below but already went ahead and fixed the assumptions but this PR should be complete already.~~

Threads:
- ~~https://dailydotdev.slack.com/archives/C02E2C3C13R/p1689776730496559~~
- ~~https://dailydotdev.slack.com/archives/C0592NL7YCU/p1689778547622899~~ (answered)

Note: will update this PRs description based on the response.

### DIFF
--- a/src/notifications/common.ts
+++ b/src/notifications/common.ts
@@ -56,7 +56,7 @@ export const notificationPreferenceMap: Partial<
   [NotificationType.SquadMemberJoined]: NotificationPreferenceType.Source,
 };
 
-export const commentNotificationTypes = [
+export const commentReplyNotificationTypes = [
   NotificationType.CommentReply,
   NotificationType.SquadReply,
 ];
--- a/src/notifications/common.ts
+++ b/src/notifications/common.ts
@@ -51,9 +51,15 @@ export const notificationPreferenceMap: Partial<
 > = {
   [NotificationType.ArticleNewComment]: NotificationPreferenceType.Post,
   [NotificationType.CommentReply]: NotificationPreferenceType.Comment,
+  [NotificationType.SquadReply]: NotificationPreferenceType.Comment,
   [NotificationType.SquadPostAdded]: NotificationPreferenceType.Source,
   [NotificationType.SquadMemberJoined]: NotificationPreferenceType.Source,
 };
+
+export const commentNotificationTypes = [
+  NotificationType.CommentReply,
+  NotificationType.SquadReply,
+];
 
 type NotificationPreferenceUnion = NotificationPreferenceComment &
   NotificationPreferencePost &
--- a/src/workers/notifications/postAdded.ts
+++ b/src/workers/notifications/postAdded.ts
@@ -1,10 +1,7 @@
 import { messageToJson } from '../worker';
 import {
-  NotificationPreference,
-  NotificationPreferenceSource,
   Post,
   PostType,
-  SourceMember,
   SourceType,
   User,
   UserAction,
--- a/src/workers/notifications/postAdded.ts
+++ b/src/workers/notifications/postAdded.ts
@@ -1,5 +1,6 @@
 import { messageToJson } from '../worker';
 import {
+  NotificationPreference,
   NotificationPreferenceSource,
   Post,
   PostType,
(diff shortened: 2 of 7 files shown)


rebelchris (approved):
Looks great, minor suggestion non blocking from my side atleast.

src/notifications/common.ts
```diff
@@ -51,10 +51,16 @@ export const notificationPreferenceMap: Partial<
 > = {
   [NotificationType.ArticleNewComment]: NotificationPreferenceType.Post,
   [NotificationType.CommentReply]: NotificationPreferenceType.Comment,
+  [NotificationType.SquadReply]: NotificationPreferenceType.Comment,
   [NotificationType.SquadPostAdded]: NotificationPreferenceType.Source,
   [NotificationType.SquadMemberJoined]: NotificationPreferenceType.Source,
 };
 
+export const commentNotificationTypes = [
```
idoshamun, line 59: We can use `notificationPreferenceMap` to generate a reverse map. From `NotificationPreferenceType` to `NotificationType[]`

src/workers/notifications/postAdded.ts
```diff
@@ -61,13 +65,20 @@ const worker: NotificationWorker = {
...
+            notificationType: NotificationType.SquadPostAdded,
+          }),
+        ]);
+        const ignored = mutes.map(({ userId }) => userId);
+        ignored.push(post.authorId);
         const members = await con.getRepository(SourceMember).find({
           where: {
             sourceId: source.id,
-            userId: Not(In([post.authorId])),
+            userId: Not(In(ignored)),
```
idoshamun, line 81: I suggest using a subquery here instead of fetching the users and then providing them back to the database.
It's much more scalable and fast

src/workers/notifications/squadMemberJoined.ts
```diff
@@ -26,10 +26,17 @@ const worker: NotificationWorker = {
...
+      notificationType: NotificationType.SquadMemberJoined,
+      status: NotificationPreferenceStatus.Muted,
+    });
+    const ignored = mutes.map(({ userId }) => userId);
+    ignored.push(member.userId);
+    const admins = await con.getRepository(SourceMember).find({
       where: {
         sourceId: member.sourceId,
-        userId: Not(In([member.userId])),
+        userId: Not(In(ignored)),
```
idoshamun, line 39: Same comment about subquery here

src/workers/notifications/utils.ts
```diff
@@ -127,10 +131,15 @@ export async function articleNewCommentHandler(
     }));
   }
 
-  return users.map((userId) => ({
-    type,
-    ctx: { ...ctx, userId },
-  }));
+  const muted = await con.getRepository(NotificationPreferencePost).findBy({
```
idoshamun, line 134: Shouldn't we filter the squad members above as well?

idoshamun (changes requested):
Please see my comments

src/notifications/common.ts
sshanzel: There could be of `NotificationPreferenceType` but not applicable to the case. Let's say we have another `NotificationType` relevant to comments, it should not be included in the use case here. Probably I should rename this to be more explicit that this is about replies.

src/workers/notifications/utils.ts
```diff
@@ -25,6 +31,33 @@ export const uniquePostOwners = (
...
+  referenceId: string,
+  where: ObjectLiteral,
+) =>
+  con
+    .getRepository(SourceMember)
+    .createQueryBuilder('sm')
+    .select('"userId"')
+    .where(where)
+    .andWhere(
+      `
```
idoshamun, line 46: Please use typeorm query builder for this purpose. It's not recommended to use raw queries. We should use it as last resort

idoshamun (changes requested):
Only one request. Overall love the new refactor!

src/workers/notifications/utils.ts
sshanzel: Finally found how to work it out 🚀

idoshamun (approved):
Amazing!

PR #1358, opened 2023-07-19
feat: restrict open squad content for blocked members

capJavert:
Restricting access to post and squad pages for public sources (squads).

Blocked user of open squad should not be able to:
- view squad page
- view squad posts
- comment on squad posts

Discussion [here](https://dailydotdev.slack.com/archives/C0592NL7YCU/p1689770365788359)

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -9,6 +9,7 @@ import {
   ensureSourcePermissions,
   GQLSource,
   SourcePermissions,
+  sourceTypesWithMembers,
 } from './sources';
 import { Context } from '../Context';
 import { traceResolverObject } from './trace';
@@ -44,6 +45,7 @@ import {
   PostMention,
   PostReport,
   PostType,
+  SourceType,
   Toc,
   Upvote,
   UserActionType,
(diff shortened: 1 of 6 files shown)


src/schema/posts.ts
```diff
@@ -881,12 +882,18 @@ export const resolvers: IResolvers<any, Context> = {
...
-      const post = await ctx.con.getRepository(Post).findOneOrFail({
-        select: ['sourceId', 'private'],
+      const partialPost = await ctx.con.getRepository(Post).findOneOrFail({
+        select: ['id', 'sourceId', 'private'],
+        relations: ['source'],
         where: { id },
       });
-      if (post.private) {
-        await ensureSourcePermissions(ctx, post.sourceId);
+      const postSource = await partialPost.source;
```
capJavert, line 890: `source` is already preloaded with `relations` above, this is just to unwrap the `Promise`. 

Alternatively I could switch this to custom query to select just `source.type` if needed.

src/schema/sources.ts
```diff
@@ -573,6 +575,13 @@ export const canAccessSource = async (
   validateRankAgainstId?: string,
 ): Promise<boolean> => {
   if (permission === SourcePermissions.View && !source.private) {
+    if (sourceTypesWithMembers.includes(source.type)) {
+      const isMemberBlocked = member?.role === SourceMemberRoles.Blocked;
+      const canAccess = !isMemberBlocked;
+
+      return canAccess;
+    }
```
capJavert, line 583: More detailed alternative to this would be to create a set of permissions for anonymous users and then use those when calling `ensureSourcePermission` on feed, source, post queries. 

All of those currently just check `View` permission on source so this makes sure that even though source is public blocked member is still checked.

rebelchris: I think for the use case at hand we simply want to say if you are blocked from public you don't even have view rights, so would cover as is.

When we get more requests we can evaluate separate rules for members and anonymous maybe?

src/schema/posts.ts
idoshamun: That's fine

rebelchris (approved):
Looks ok to me, Ido will maybe have some alternative for loading, but from high level it looks good.

I think until we get more detailed requests this will do.
Also thanks a lot for adding the detailed test cases 🙏

src/schema/sources.ts
idoshamun: What would be the tradeoffs between the two?

idoshamun (approved):
Please note that the PR title is not following our conventions 🥲

src/schema/sources.ts
capJavert: We missed it in planning so  main trade-off is time we would need to properly define how it would work and propagate that through all the API mutations.

idoshamun: So let's keep it in mind for next time :)
We are already a bit behind schedule here

PR #1360, opened 2023-07-20
feat: setting welcomePost as showOnFeed as false by default

TomDowling:
Setting the showOnFeed flag as false by default for welcomePosts


rebelchris (changes requested):
this is great, part of this ticket was also to retro fix existing squads (which will need a script probably)
Would you want to take that one on as well? (else let me know I can support on it)

idoshamun:
@rebelchris I think a simple migration can also be useful as a retro fix. Should be straightforward

src/common/post.ts
```diff
@@ -145,6 +145,7 @@ export const createSquadWelcomePost = async (
       banned: true,
       private: true,
       visible: true,
+      showOnFeed: false,
```
idoshamun, line 148: Currently `flags` is not used as a primary source so we should add `showOnFeed` as a main property of the object

idoshamun (changes requested):
Please check my comment. Also we should update the tests I guess right?

rebelchris:
@idoshamun Added all comments and a new test case as we didn't explicitly test for it

src/migration/1689854830946-WelcomeShowOnFeed.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class WelcomeShowOnFeed1689854830946 implements MigrationInterface {
+  name = 'WelcomeShowOnFeed1689854830946';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `UPDATE post SET flags = flags || '{"showOnFeed": false}' WHERE type = 'welcome'`,
```
idoshamun, line 8: We need to set `showOnFeed` itself as well here

idoshamun: I see it only in `down` the `up` wasn't changed

rebelchris:
@idoshamun rushing was a bad idea, fixed both now

idoshamun (approved):
Perfect!

sshanzel (approved):
Looks great!

PR #1364, opened 2023-07-24
feat: add show posts on feed setting for SourceMember

capJavert:
- added new `flags.hideFeedPosts` to `SourceMember` entity
  - this flag is used to hide source (squad) posts that they are member of from their feed
- updated `feedToFilters` in feed generator to exclude sources depending on the setting
- added mutations so source members can set that option from the client
- expose flags on `SourceMember` entity in gql


src/common/feedGenerator.ts
```diff
@@ -113,10 +113,19 @@ export const feedToFilters = async (
...
+      ? con
+          .getRepository(SourceMember)
+          .createQueryBuilder('sm')
+          .select('sm."sourceId"')
+          .where('sm."userId" = :userId', { userId })
+          .andWhere(
+            new Brackets((qb) => {
+              qb.where(
+                `(sm."flags"->'showPostsOnFeed')::boolean = TRUE`,
+              ).orWhere(`sm."flags"->'showPostsOnFeed' IS NULL`);
```
capJavert, line 125: Had to do this kind of query because if `flags.showPostsOnFeed` on feed is not set the query `WHERE (flags->'showPostsOnFeed')::boolean != false` would not return correctly, it just does not work with nullish values for json paths. That is why I have `OR` condition to also check for `NULL`.

> Why did we not just set `flags.showPostsOnFeed` to `true` by default?

During initial post flags PR https://github.com/dailydotdev/daily-api/pull/1333#discussion_r1259503040 we figured that if any of our flags have value (like true by default) we would need this kind of trigger to actually preserve those default during insert.

capJavert:
Test failing due to other PR but this should be ready for review.

src/common/feedGenerator.ts
idoshamun: A better approach would be to reverse the flag. So as we discussed in the post table flags, it's better to have flags that are false/null by default. So maybe we should call it `hideFromFeed`, and only if it's true we will not show it on the feed. May reduce the edge cases here, right?
Some lessons learned from previous migrations

src/entity/SourceMember.ts
```diff
@@ -38,6 +42,10 @@ export class SourceMember {
   @Column({ type: 'text' })
   @Index('IDX_source_member_referralToken', { unique: true })
   referralToken: string;
+
+  @Column({ type: 'jsonb', default: {} })
+  @Index('IDX_source_member_flags_showPostsOnFeed', { synchronize: false })
```
idoshamun, line 47: There's no need to index this field by itself. It's better to index it with user id as they go together

idoshamun (changes requested):
Please see my suggestion. Overall love the implementation and how you highlighted the complexity of the query.
Don't we miss returning the value of the flag to the user?

src/common/feedGenerator.ts
capJavert: @idoshamun I thought about that but it would not make a difference in regards to the query, we would still need `NOT NULL` check since our `flags->hideFromFeed` would still not have value by default and query `WHERE (flags->'hideFromFeed')::boolean != true` would still not return matches for all those source members.

src/entity/SourceMember.ts
```diff
@@ -7,7 +7,17 @@ import { SourceMemberRoles } from '../roles';
...
+  showPostsOnFeed: boolean;
+}>;
+
+export type SourceMemberFlagsPublic = Pick<
+  SourceMemberFlags,
+  'showPostsOnFeed'
+>;
+
 @Entity()
+@Index('IDX_source_member_userId_flags_showPostsOnFeed', { synchronize: false })
```
capJavert, line 20: Had to still use this since `@Index` does not play well with JSON fields.

src/schema/sources.ts
```diff
@@ -1278,6 +1313,30 @@ export const resolvers: IResolvers<any, Context> = {
 
       return getSourceById(ctx, info, sourceId);
     },
+    hideSourceFeedPosts: async (_, { sourceId }: { sourceId: string }, ctx) => {
+      await ensureSourcePermissions(ctx, sourceId, SourcePermissions.View);
```
rebelchris, line 1317: Optional, but you could abstract this into a top level function as only the value is changing, example is [this](https://github.com/dailydotdev/daily-api/pull/1336/files#diff-e3ad1f5d3fe340ded80e09a15a95f7d80be472d110fe2bdbbce21dba440b925aR883).

rebelchris (commented):
Nothing smart to add, if there is a way we can omit the null check that be great.
But unsure at this point, happy to debug together if need be though.

src/common/feedGenerator.ts
idoshamun: @capJavert have you tried using the `COALESCE` function? `COALESCE(flags->'hideFromFeed', false)`

capJavert: @idoshamun nice one, looks like it works, gonna refactor to `hideFromFeed` first thing in the morning so we can close this.

rebelchris (approved):
Nice work!
The coalesce really makes it cleaner 🙏

idoshamun (approved):
Love it!

PR #1365, opened 2023-07-24
fix: return null is auth on scalar value

rebelchris:
Added a manual evaluation if field type is a `GraphQLScalarType` in which case we should simply return a null value.
Otherwise we can always throw forbidden error as it will be a query or mutation.

WT-1551 #done

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -3188,7 +3188,7 @@ describe('flags field', () => {
     }
   }`;
 
-  it('should return all the public flags', async () => {
+  it('should return all the public flags for anonymous user', async () => {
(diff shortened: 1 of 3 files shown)


idoshamun (changes requested):
Can you please add tests to check the new change? Try to get a field with the right permissions (maybe this one already exists) and without.

rebelchris:
@idoshamun I already added tasks for this in the previous flow.
Not 100% sure why it didn't throw the error in the wrong permission flow though..

idoshamun (approved):

idoshamun (approved):

PR #1366, opened 2023-07-24
fix: notification preference query with empty array

sshanzel:
Realized a bug while working on the FE. We formulate the parameters on the where clause using `.reduce`. This means when an array has zero length, then it would be an empty array for the condition which will return all data.


__tests__/notifications.ts
```diff
@@ -354,43 +354,15 @@ describe('query notificationPreferences', () => {
       'UNAUTHENTICATED',
     ));
 
-  const prepareNotificationPreferences = async () => {
```
sshanzel, line 357: Just a duplicated code that happened through some merging. There should be another one a bit above this code block.

rebelchris (approved):

PR #1369, opened 2023-07-26
fix: missing property on notification type

sshanzel:
In order for us to show the current preference of the user towards the said notification type, we need to fetch it first and the reference id is one of the requirements to do so.

Screenshot 2023-07-26 at 3 07 58 PM

### DIFF
--- a/src/schema/notifications.ts
+++ b/src/schema/notifications.ts
@@ -100,6 +100,10 @@ export const typeDefs = /* GraphQL */ `
     """
     type: String!
     """
+    Referenced entity's id of the notification
+    """
+    referenceId: String!
+    """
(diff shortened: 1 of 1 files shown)


idoshamun (approved):

sshanzel:
@rebelchris @idoshamun after further development, I noticed that even though we strictly set the proper `type` of the `NotificationPreference`, it is getting overridden by what was defined in the class. Using the base class results in always setting it to `NotificationPreference` string as shown below. I have pushed a commit to fix this and ended up with a switch case, I tried doing a map of the class but wasn't working.

Screenshot 2023-07-26 at 7 24 51 PM

PR #1370, opened 2023-07-26
fix: correct notification preference type

sshanzel:
after further development, I noticed that even though we strictly set the proper `type` of the `NotificationPreference`, it is getting overridden by what was defined in the class. Using the base class results in always setting it to `NotificationPreference` string as shown below. I have pushed a commit to fix this and ended up with a switch case, I tried doing a map of the class but wasn't working.

Screenshot 2023-07-26 at 7 24 51 PM

### DIFF
--- a/__tests__/notifications.ts
+++ b/__tests__/notifications.ts
@@ -29,6 +29,7 @@ import { subDays } from 'date-fns';
 import request from 'supertest';
 import { FastifyInstance } from 'fastify';
 import {
+  notificationPreferenceMap,
   NotificationPreferenceStatus,
   NotificationPreferenceType,
   NotificationType,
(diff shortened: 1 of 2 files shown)


capJavert (approved):
Looks like it's a running issue from even before in our repo and typeorm:
- https://github.com/dailydotdev/daily-api/pull/1154
- https://github.com/typeorm/typeorm/issues/9033

Looks ok from my side 👍

idoshamun (commented):
We should add tests to check the right type then. To make sure we don't introduce it again

__tests__/notifications.ts
```diff
@@ -551,6 +552,9 @@ describe('mutation muteNotificationPreference', () => {
       .findOneBy(params);
 
     expect(muted).toBeTruthy();
+    expect(muted.type).toEqual(
+      notificationPreferenceMap[params.notificationType],
+    );
```
capJavert, line 557: This test I think covers the type check during insert 👍

sshanzel: Yes, this is the one I also used to make the test fail, and applied the fix to make this go green.

sshanzel: cc @idoshamun

idoshamun: Missed that!

idoshamun (approved):

PR #1371, opened 2023-07-26
feat: split cron job into testable components (PT-327)

mykytanikitenko:


src/common/tinybird.ts
```diff
@@ -0,0 +1,139 @@
...
+export type fetchfn = (
+  url: RequestInfo,
+  init?: RequestInit,
+) => Promise<Response>;
+
+export enum TinybirdDatasourceMode {
+  APPEND = 'append',
+}
+
+export interface ITinybirdClient {
```
idoshamun, line 14: @mykytanikitenko we don't follow the `I` prefix for interfaces on this project

src/common/tinybird.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    if (!response.ok) {
+      const text = await response.text();
+      throw new Error(`tinybird response ${response.status}: ${text}}`);
+    }
+
+    return (await response.json()) as PostDatasourceResult;
+  }
+
+  public static async Json2Csv<T>(records: T[]): Promise<string> {
```
idoshamun, line 105: Functions should start with lowercase

```suggestion
  public static async json2Csv<T>(records: T[]): Promise<string> {
```

src/common/tinybird.ts
```diff
@@ -0,0 +1,139 @@
...
+    return Object.keys(params)
+      .map(
+        (key: string): string =>
+          encodeURIComponent(key) + '=' + encodeURIComponent(params[key]),
+      )
+      .join('&');
+  }
+}
+
+const json2csv = promisify(jsonexport);
```
idoshamun, line 139: This needs to go above the first usage. So in this case, above the class definition

src/cron/exportToTinybird.ts
```diff
@@ -24,64 +53,144 @@ export const getPostsTinybirdExport = (con: DataSource, latest: Date) =>
...
+  }
+
+  public async exportAndLog(): Promise<void> {
+    let result: TinybirdExportResult;
+
+    try {
+      result = await this.export();
+    } catch (err) {
+      this.logger.error(
+        { error: err.message, stack: err.stack },
```
idoshamun, line 158: We usually log errors like this

```suggestion
        { err },
```

idoshamun (approved):
Thanks for the hard work in breaking the tinybird exporter down to testable components.
Actually on this project we use less interfaces and classes, this is more of platform team way. But that's perfectly fine as it helps us to test this exporter.
Would love to hear @rebelchris thoughts

__tests__/common/tibybird.ts
```diff
@@ -0,0 +1,206 @@
...
+  it('Json2Csv date', async () => {
+    const records = [
+      {
+        column1: 'test',
+        column2: new Date('2023-07-27T16:47:33+0000'),
+      },
+    ];
+
+    const got = await TinybirdClient.json2Csv(records);
+    expect(got).toMatchSnapshot();
```
rebelchris, line 157: Why did you decide on snapshot here?
It doesn't really matter but since it's a small value maybe inline would be best, to avoid snapshot issues down the line?

(Same goes for the other snapshots below here)

__tests__/cron/exportToTinybird.ts
```diff
@@ -0,0 +1,345 @@
...
+    const now = new Date();
+    const latest = new Date(now.getTime() - 10000);
+    const repo = new PostsRepository(con);
+
+    const posts = await repo.getForTinybirdExport(latest);
+    posts.forEach((post) => {
+      post.created_at = '';
+      post.metadata_changed_at = '';
+    });
+    expect(posts).toMatchSnapshot();
```
rebelchris, line 58: We had some issues in the past where the order of the snapshot can change, thus the inline object test was a better solution.
Would probably be best to revert to that to avoid flaky order issues.

__tests__/cron/exportToTinybird.ts
```diff
@@ -0,0 +1,345 @@
...
+        return {} as PostDatasourceResult;
+      },
+    } as ITinybirdClient;
+
+    const postsMetadataRepository = new PostsMetadataRepository(
+      tinybirdMock,
+      dataSource,
+    );
+
+    const posts: TinybirdPost[] = [
```
rebelchris, line 85: I think we can also use ` await saveFixtures(con, ArticlePost, postsFixture);` something like this right?

__tests__/cron/exportToTinybird.ts
```diff
@@ -0,0 +1,345 @@
...
+      logger,
+      postsRepositoryMock,
+      postsMetadataRepositoryMock,
+    );
+
+    await service.exportAndLog();
+  });
+
+  test.skip('test real tinybird', async () => {
+    //it('test real tinybird', async () => {
```
rebelchris, line 321: Is this one supposed to be here?

rebelchris (commented):
Looks great, thanks for the split up.
Nothing major, just some generic comments.

__tests__/cron/exportToTinybird.ts
mykytanikitenko: Done

__tests__/cron/exportToTinybird.ts
mykytanikitenko: By accident :)

__tests__/cron/exportToTinybird.ts
mykytanikitenko: This exact test is not intended to use a real database

__tests__/cron/exportToTinybird.ts
```diff
@@ -0,0 +1,444 @@
...
+    const latest = new Date(now.getTime() - 10000);
+    const repo = new PostsRepository(con);
+
+    const posts = await repo.getForTinybirdExport(latest);
+    posts.forEach((post) => {
+      post.created_at = '';
+      post.metadata_changed_at = '';
+    });
+    expect(posts).toHaveLength(7);
+    expect(posts).toContainEqual({
```
rebelchris: I think you can actually use this one as a one-stop-solution: `toStrictEqual`
See for example: https://github.com/dailydotdev/daily-api/commit/163d70ac1cb6dc0697ea307bba056fb2e22066ad

rebelchris (approved):
One non-blocking comment.
Also keep in mind @vpol should give a go before merging as it includes changes in tinybird file.

vpol (approved):

PR #1372, opened 2023-07-26
feat: set NODE_ENV to development for local env

capJavert:
`NODE_ENV` was undefined in our local environment for daily-api
Since most developers would assume it should be defined we inject it as `development` for our local setup.

This fixes some edge cases in development like send email infinite loop https://github.com/dailydotdev/daily-api/pull/1302 which was fixed but the fix did not work anymore since our `NODE_ENV` was `undefined`.


idoshamun (approved):

rebelchris (approved):

sshanzel (approved):
Nice catch! 🚀

PR #1374, opened 2023-07-27
fix: article new comment muting

sshanzel:
When we send a notification for `article_new_comment`, the entity's reference id is based on the comment's id, not the post. But when we want to mute `article_new_comment` from the notifications page, what we need is the post id though as mentioned what was tied to the notification was the comment's id.

In the FE, we just render the list, we have no way of identifying whether we need to fetch the post id for certain notification types just to mute a notification. To make it all simpler, we are doing it in the BE to do the verifications.

### DIFF
--- a/__tests__/notifications.ts
+++ b/__tests__/notifications.ts
@@ -17,6 +17,7 @@ import {
   NotificationPreferencePost,
   Post,
   User,
+  Comment,
   Source,
   NotificationPreferenceSource,
   NotificationPreference,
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Seems like the right fix 🙏
Let me know when I can re-test it please

sshanzel:
@idoshamun let us know what you think.

idoshamun (approved):
Love this approach! Way to go :)

PR #1375, opened 2023-07-30
fix: flaky anonymous test

rebelchris:
Converted flaky anonymous feed test to use inline object equality vs snapshot 👀


__tests__/feeds.ts
```diff
@@ -325,7 +325,101 @@ describe('query anonymousFeed', () => {
...
+          },
+          tags: ['javascript', 'webdev'],
+          title: 'Included Post',
+          url: 'http://ip1.com',
+        },
+      },
+    ];
+    res.data.anonymousFeed.edges.forEach((post) =>
+      expect(snapshot).toContainEqual(post),
+    );
```
idoshamun, line 422: The only problem with this test is that if the snapshot contains something extra, you will not know it. You currently check that whatever returns is within the snapshot but not the other way around.
You also don't check the order of the returned array. I'm not sure that's an ideal solution. Any alternatives?

rebelchris: Well the order is what flaked and not sure it matters at this point correct?
We can also check for equality on a set object, but the order seems to flake (snapshot this is a snapshot issue, so should be care about that)

idoshamun: In the feed's case, the order matters. What may be the root cause of the flakiness in your opinion?

rebelchris: From what ive read snapshot evaluations may change order, as even with hard passing a certain order it would get overwritten on certain environments.

Can also have another look to see if anything messed up the order before or to stable it better

idoshamun: Can't we then have equality check on the whole object without snapshots?

rebelchris: @idoshamun Let me know what you think of my changes 👍

idoshamun (approved):

PR #1377, opened 2023-08-01
fix: notification preference query on post related types

sshanzel:
Similar to the PR for muting notification types relating to `article_new_comment` (https://github.com/dailydotdev/daily-api/pull/1374) - we should also update the query counterpart and the clearing of notifications.

Note: I will also raise a PR to make the FE work with the new requirements.

### DIFF
--- a/__tests__/notifications.ts
+++ b/__tests__/notifications.ts
@@ -331,7 +331,7 @@ const prepareNotificationPreferences = async () => {
       userId: '1',
       sourceId: sourcesFixture[0].id,
       referenceId: sourcesFixture[0].id,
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
Seems good for now, bit of a edge-case.
We'll need to log a alternative way for future proofing I think.

sshanzel:
I agree. Not the most ideal in the long term if we get more similar scenarios in the future. Quick random thought is probably adding another column in the notification entity to hold the right reference for muting/subscribing.

capJavert (approved):

idoshamun (approved):

PR #1378, opened 2023-08-02
fix: add yggdrasil id to content flow

rebelchris:
We decided to add the Yggdrasil id to the post API.
For now it can simply be set, no need to expose it anywhere.

### DIFF
--- a/src/migration/1690963996461-PostYggdrasilId.ts
+++ b/src/migration/1690963996461-PostYggdrasilId.ts
@@ -1,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class PostYggdrasilId1690963996461 implements MigrationInterface {
+  name = 'PostYggdrasilId1690963996461';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "post" ADD "yggdrasilId" uuid`);
+    await queryRunner.query(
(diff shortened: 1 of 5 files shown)


denisb0 (approved):

src/migration/1690963996461-PostYggdrasilId.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "post" ADD "yggdrasilId" uuid`);
+    await queryRunner.query(
+      `CREATE INDEX "IDX_yggdrasil_id" ON "post" ("yggdrasilId") `,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "public"."IDX_yggdrasil_id"`);
```
capJavert, line 14: Will we query by this? From what I see on discussion it will be used to get back content from platform. Just thinking we might not need an index.

rebelchris: Eventually I think we will need the index perhaps to search for (at this point no)
But need to see what @idoshamun thinks if it will become needed or not. (good thinking though)

idoshamun: I think we will need this index once we introduce incremental updates

PR #1379, opened 2023-08-03
chore: disable temporarily cdc for changes in yggdrasil id

idoshamun:
I'd like to backfill yggdrasil ids which is something like 50k records. I don't want to trigger CDC and flood the queue. Please make sure that my condition is correct in your review.


idoshamun:
Closing as I applied everything locally and there's no need to keep it

PR #1381, opened 2023-08-07
fix: remove posts from private sources in most upvoted feed

omBratteng:
Creates a new feed filter to filter out squad posts from `Most upvoted` and `Best discussed` feeds

WT-1550


idoshamun (approved):
Looks great to me! Let's wait for @rebelchris

__tests__/feeds.ts
```diff
@@ -1183,6 +1183,35 @@ describe('query mostUpvotedFeed', () => {
...
+    const sourceRepo = con.getRepository(Source);
+    await sourceRepo.update({ id: 'squad' }, { type: SourceType.Squad });
+    const repo = con.getRepository(Post);
+    await repo.update({ id: 'p1' }, { upvotes: 20 });
+    await repo.update({ id: 'p3' }, { upvotes: 15 });
+    await repo.update({ id: 'p2' }, { upvotes: 30, sourceId: 'squad' });
+
+    const res = await client.query(QUERY(30));
+    expect(res.errors).toBeFalsy();
+    expect(res.data).toMatchSnapshot();
```
rebelchris, line 1197: Nothing blocking, but in general snapshots can give us some flakiness on array comparison (as they are not guaranteed order apparently)
For now no action needed, but maybe for future good note to adjust to other evaluations.

src/common/feedGenerator.ts
```diff
@@ -222,6 +222,7 @@ export const applyFeedWhere = (
   removeHiddenPosts = true,
   removeBannedPosts = true,
   allowPrivateSources = true,
+  allowSquadPosts = true,
```
rebelchris, line 225: More of a question, is there any existing flow that uses this as default currently?

rebelchris (approved):
Nothing major, looks good to me.

__tests__/feeds.ts
omBratteng: Yeah, I noticed that the array order was different between when I ran the all the test and the specific test file

rebelchris: Yep, array order of snapshot is not guaranteed unfurtionally (you'll find many github issues in Jest repo for this)
https://dailydotdev.slack.com/archives/C02E2C3C13R/p1690297213537409?thread_ts=1690296315.741339&cid=C02E2C3C13R

src/common/feedGenerator.ts
idoshamun: feed and bookmarks

__tests__/feeds.ts
omBratteng: That's quite unfortunate, since snapshot is a great tool.

rebelchris: Agreed! 
Hope they'll either support it soon or we can find a good workaround or something.

As mentioned for now don't worry, let's see if they get caught on flakiness at all.

PR #1382, opened 2023-08-08
fix: set private flag to match squad source on free form posts

omBratteng:
Sets the private attribute on freeform posts to match the private attribute of the squad source

WT-1628 #done


idoshamun (approved):

sshanzel (approved):
Nice catch! Amazing fix! 🚀

PR #1383, opened 2023-08-09
chore: cleanup and dependency updates

rebelchris:
Cleanup some false test cases, update dependencies


1683868176744-PostMentionTracking.ts
```diff
@@ -1,17 +0,0 @@
-import { MigrationInterface, QueryRunner } from 'typeorm';
```
idoshamun, line 1: Why did you delete that migration?

__tests__/cron/exportToTinybird.ts
```diff
@@ -67,7 +72,7 @@ describe('PostsRepository', () => {
       post_private: 0,
       post_type: 'article',
       source_id: 'a',
-      source_type: 'Source',
+      source_type: 'machine',
```
sshanzel, line 75: Good find with all these.

1683868176744-PostMentionTracking.ts
rebelchris: Was situated in the root of the project, not migration folder.
AFAIK they don't get triggered anyway. (and looked to be outdated)

sshanzel: Oh, in that case, I think we might have to move it to the correct migrations folder if the replica identity is not accurate with `PostMention`.

__tests__/comments.ts
```diff
@@ -169,9 +174,10 @@ const saveSquadFixture = async (sourceId: string) => {
       image: 'sample2/image',
     },
   ]);
-  await con
-    .getRepository(Source)
-    .save(createSource(sourceId, 'A', 'http://a.com', SourceType.Squad, true));
+  await con.getRepository(Source).update({ id: sourceId }, { private: true });
+  // await con
+  //   .getRepository(Source)
+  //   .save(createSource(sourceId, 'A', 'http://a.com', SourceType.Squad, true));
```
sshanzel, line 180: I think we can now remove this comment.

sshanzel (commented):
Always nice to keep things updated. Just had a few comments but nothing blocking :shipit:

sshanzel (approved):
Again, thanks for doing the dependency updates! 🚀

PR #1384, opened 2023-08-09
feat: search recommendation entity

sshanzel:
Search recommendation entity to display under the Search bar.

WT-1646-search-recommendation #done


idoshamun (approved):

rebelchris (commented):
I think your missing the replica identity 🥲

sshanzel:
> I think your missing the replica identity 🥲

Oh, there was no mention of any workers watching this entity. Are we supposed to send an event when a new row is created?

rebelchris:
I think more for the table to become available in metabase

sshanzel:
> I think more for the table to become available in metabase

Oh, I never knew that. I thought it was more about working with background workers. Now I understand, thank you for letting me know, I learned something today. Would you prefer a separate one or would it be okay to include it in the recommended search PR?

rebelchris:
This is atleast my understanding of when we need it.
@idoshamun can you confirm before we action anything 🤘

idoshamun:
Actually the replica identity is required for debezium and cdc purposes only. So Lee was right, we don't need it in this case

PR #1385, opened 2023-08-09
feat: graphql query for recommended search questions

sshanzel:
A new query to fetch recommended search questions based on the user's 5 recent upvotes then randomized the result to fetch only 3 items.

WT-1647 #done


src/schema/posts.ts
```diff
@@ -1002,6 +1013,35 @@ export const resolvers: IResolvers<any, Context> = {
         },
       );
     },
+    searchQuestionRecommendations: async (
+      source,
+      _,
+      ctx: Context,
+    ): Promise<GQLPostQuestion[]> => {
+      const res = await ctx.con
```
rebelchris, line 1021: Would graphorm not be preferred here?

sshanzel: Makes sense. Let me see how it would be able to easily work with the complex we have here.

sshanzel: Done! 🥳

rebelchris (approved):
Looks good to me 🙏

src/schema/posts.ts
```diff
@@ -469,6 +472,12 @@ export const typeDefs = /* GraphQL */ `
     image: String!
   }
 
+  type PostQuestion {
+    id: String!
+    postId: String!
```
idoshamun, line 477: This needs to be a post object

src/schema/posts.ts
```diff
@@ -1002,6 +1013,31 @@ export const resolvers: IResolvers<any, Context> = {
...
+                .from(Upvote, 'u')
+                .where({ userId: ctx.userId })
+                .orderBy('u."createdAt"', 'DESC')
+                .limit(5),
+            'upvoted',
+            `"${builder.alias}"."postId" = upvoted."postId"`,
+          )
+          .orderBy('rand', 'DESC')
+          .limit(3)
+          .getRawMany(),
```
idoshamun, line 1038: This is not needed for graphorm

src/schema/posts.ts
```diff
@@ -1002,6 +1013,31 @@ export const resolvers: IResolvers<any, Context> = {
...
     },
+    searchQuestionRecommendations: async (
+      source,
+      _,
+      ctx: Context,
+      info,
+    ): Promise<GQLPostQuestion[]> =>
+      graphorm.query(ctx, info, (builder) => ({
+        queryBuilder: builder.queryBuilder
+          .addSelect('random()', 'rand')
```
idoshamun, line 1024: We don't need to add it as a select. We can put it directly in the order by

sshanzel: Oh, cool. Nice to know we can do that.

idoshamun (approved):
Amazing!

PR #1386, opened 2023-08-10
feat: send feedback to magni

sshanzel:
The service for Magni is not up yet but I wanted to set the baseline for the integration with the said service so the FE will have an easier understanding of what to expect with the API proxy.

WT-1649 #done


src/integrations/magni.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const magniOrigin = process.env.MAGNI_ORIGIN;
+
+export const postFeedback = async (req: FastifyRequest): Promise<Response> =>
+  fetch(`${magniOrigin}/feedback`, {
+    ...fetchOptions,
+    method: 'post',
+    body: JSON.stringify(req.params),
+    headers: {
+      cookie: req.headers.cookie,
```
idoshamun, line 13: This is not required, instead you have to provide `X-User-Id` header as mentioned in the DR.
image

src/integrations/magni.ts
```diff
@@ -0,0 +1,16 @@
+import fetch, { Response } from 'node-fetch';
+import { fetchOptions } from '../http';
+import { FastifyRequest } from 'fastify';
+
+const magniOrigin = process.env.MAGNI_ORIGIN;
+
+export const postFeedback = async (req: FastifyRequest): Promise<Response> =>
```
idoshamun, line 7: It's better to be more type-safe than this. I suggest that this function will accept the exact parameters expected by magni instead of relying on the req object.

src/routes/search.ts
```diff
@@ -0,0 +1,28 @@
...
+import { postFeedback } from '../integrations';
+import { isNullOrUndefined } from '../common/object';
+
+export interface FeedbackArgs {
+  value: number;
+  chunkId: string;
+}
+
+export default async function (fastify: FastifyInstance): Promise<void> {
+  fastify.post<{ Body: FeedbackArgs }>('/feedback', async (req, res) => {
```
idoshamun, line 11: This should be a graphql mutation isn't it? @rebelchris

rebelchris: Not sure actually, since these are only proxying request I suggested we keep it the same as we do for heimdall.
What would be pro/con of using mutation over this?

idoshamun: It's needed for multiple reasons:
* Keep consistent with the rest of our codebase which is graphql
* Provide type-safety and clear expectations on input/output
* I want us to look at it as something other than a proxy per se. But more like a remote database. Imagine Magni is our database.

We somehow missed this point in the web team's DR, I guess. Good to note for next time

Sidenote: we don't proxy heimdall. We just call it as part of our boot endpoint.

rebelchris: Alright would agree with that, just something that is "vague" and not standardized at the moment.
Maybe some day we can spend some time to setup some guidelines.

As in 
- prefer graphQL
- prefer graphORM
- etc

For heimdall we do have exclusions, like the route to check email etc, they are not via graphORM from top of my head.

idoshamun: I agree that we need guidelines. For now anything API project related should use graphql and graphorm as much as possible to keep our conventions. Heimdall is available on sso.daily.dev so it's not proxy by api

sshanzel: Amazing. Thank you both for the inputs. I shall make the changes.

src/integrations/magni.ts
sshanzel: Right. Let me fix this quickly.

src/integrations/magni.ts
```diff
@@ -0,0 +1,23 @@
...
+import { fetchOptions } from '../http';
+
+const magniOrigin = process.env.MAGNI_ORIGIN;
+
+export interface SearchResultFeedback {
+  chunkId: string;
+  feedback: number;
+}
+
+export const postFeedback = async (
```
idoshamun, line 11: @vpol we need your eyes here to double check

idoshamun (approved):
Thanks for addressing the feedback swiftly! 🙏

rebelchris (approved):
Nothing breaking from my side. (but await viktor please)

src/integrations/magni.ts
vpol: Works for me.

PR #1387, opened 2023-08-10
fix: extend redis cache on each visit

rebelchris:
Decided to simply set the value, which in turn will reset the TTL.
Also added dedicated test case to ensure TTL get's reset.

Tested locally as well by comparing TTL on redis pool.

WT-1655 #done


sshanzel (approved):
Thanks for the fix! 🚀

idoshamun (approved):
Thanks for the hotfix

PR #1388, opened 2023-08-10
chore: massive cleanup

rebelchris:
After some investigation found some unused code and routes.
I checked traces and no trace found.


sshanzel (approved):
Wow. Thank you so much for the effort. Can't believe we have finally removed the compat routes.

idoshamun (approved):
What can go wrong... 😬

PR #1389, opened 2023-08-10
feat: add user post table and mutations

capJavert:
- [x] add `user_post` table
- [x] update mutations to set `UserPost` entity in addition to old entities
- [x] update current tests
- [x] add new `votePost` mutation (will be used by frontend after data migration is done in other PR)
- [x] add new `userState` field
- [x] added `flags.feedbackDismiss` field and and `dismissPostFeedback` mutation (for engagement loops feature)
- [x] tests for new mutation and fields

### DIFF
--- a/src/entity/UserPost.ts
+++ b/src/entity/UserPost.ts
@@ -43,6 +43,9 @@ export class UserPost {
   @UpdateDateColumn()
   updatedAt: Date;
 
+  @Column({ default: null, nullable: true })
+  votedAt: Date;
+
   @Column({ type: 'smallint', default: UserPostVote.None })
   vote: UserPostVote;
 
--- a/src/entity/UserPost.ts
+++ b/src/entity/UserPost.ts
@@ -22,12 +22,6 @@ export enum UserPostVote {
   Down = -1,
 }
 
-export const userPostDefaultData = Object.freeze({
-  vote: UserPostVote.None,
-  hidden: false,
-  flags: {},
-});
-
 @Entity()
 @Index(['postId', 'userId'], { unique: true })
 export class UserPost {
@@ -47,13 +41,13 @@ export class UserPost {
   votedAt: Date;
 
   @Column({ type: 'smallint', default: UserPostVote.None })
-  vote: UserPostVote;
+  vote: UserPostVote = UserPostVote.None;
 
-  @Column({ default: false })
-  hidden: boolean;
+  @Column({ type: 'boolean', default: false })
+  hidden = false;
 
   @Column({ type: 'jsonb', default: {} })
-  flags: UserPostFlags;
+  flags: UserPostFlags = {};
 
   @ManyToOne(() => Post, {
     lazy: true,
--- a/src/entity/UserPost.ts
+++ b/src/entity/UserPost.ts
@@ -11,10 +11,10 @@ import { Post } from './posts';
 import { User } from './User';
 
 export type UserPostFlags = Partial<{
-  feedbackDimiss: boolean;
+  feedbackDismiss: boolean;
 }>;
 
-export type UserPostFlagsPublic = Pick<UserPostFlags, 'feedbackDimiss'>;
+export type UserPostFlagsPublic = Pick<UserPostFlags, 'feedbackDismiss'>;
 
 export enum UserPostVote {
   Up = 1,
--- a/src/entity/UserPost.ts
+++ b/src/entity/UserPost.ts
@@ -22,6 +22,12 @@ export enum UserPostVote {
   Down = -1,
 }
 
+export const userPostDefaultData = Object.freeze({
+  vote: UserPostVote.None,
+  hidden: false,
+  flags: {},
+});
+
 @Entity()
 @Index(['postId', 'userId'], { unique: true })
 export class UserPost {
--- a/src/entity/UserPost.ts
+++ b/src/entity/UserPost.ts
@@ -37,7 +37,7 @@ export class UserPost {
   @UpdateDateColumn()
   updatedAt: Date;
 
-  @Column({ type: 'smallint' })
+  @Column({ type: 'smallint', default: UserPostVote.None })
   vote: UserPostVote;
 
   @Column({ default: false })
--- a/src/entity/UserPost.ts
+++ b/src/entity/UserPost.ts
@@ -1,0 +1,60 @@
+import {
+  Column,
+  CreateDateColumn,
+  Entity,
+  Index,
+  ManyToOne,
+  PrimaryColumn,
+  UpdateDateColumn,
+} from 'typeorm';
+import { Post } from './posts';
+import { User } from './User';
+
+export type UserPostFlags = Partial<{
+  feedbackDimiss: boolean;
+}>;
+
+export type UserPostFlagsPublic = Pick<UserPostFlags, 'feedbackDimiss'>;
+
+export enum UserPostVote {
+  Up = 1,
+  None = 0,
+  Down = -1,
+}
+
+@Entity()
+@Index(['postId', 'userId'], { unique: true })
+export class UserPost {
+  @PrimaryColumn({ type: 'text' })
+  postId: string;
+
+  @PrimaryColumn({ type: 'text' })
+  userId: string;
+
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @UpdateDateColumn()
+  updatedAt: Date;
+
+  @Column({ type: 'smallint' })
+  vote: UserPostVote;
+
+  @Column({ default: false })
+  hidden: boolean;
+
+  @Column({ type: 'jsonb', default: {} })
+  flags: UserPostFlags;
+
+  @ManyToOne(() => Post, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  post: Promise<Post>;
+
+  @ManyToOne(() => User, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  user: Promise<User>;
+}
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -167,6 +167,16 @@ const obj = new GraphORM({
       flags: {
         jsonType: true,
       },
+      userState: {
+        relation: {
+          isMany: false,
+          customRelation: (ctx, parentAlias, childAlias, qb): QueryBuilder => {
+            return qb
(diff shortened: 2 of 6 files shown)


src/schema/posts.ts
```diff
@@ -799,6 +840,31 @@ export const typeDefs = /* GraphQL */ `
...
+
+    """
+    Dismiss user post feedback
+    """
+    dismissPostFeedback(
+      """
+      Id of the post
+      """
+      id: ID!
+    ): EmptyResponse @auth
```
capJavert, line 867: Decided to only expose what we currently support, which is dismiss only.

src/graphorm/index.ts
```diff
@@ -167,6 +169,27 @@ const obj = new GraphORM({
...
+          },
+        },
+        transform: (value: GQLUserPost, ctx: Context) => {
+          if (!ctx.userId) {
+            return null;
+          }
+
+          if (!value) {
+            return userPostDefaultData;
+          }
```
capJavert, line 188: Wanted to achieve that clients can fetch and get valid `userState` object even when `user_post` entry is empty for some user <-> post combination. Think it will be better DX instead of always looking for null types. Interested to see what you think.

src/schema/posts.ts
```diff
@@ -1516,6 +1635,112 @@ export const resolvers: IResolvers<any, Context> = {
...
+            );
+          } else {
+            await entityManager.getRepository(UserPost).insert({
+              postId: id,
+              userId: ctx.userId,
+              flags: {
+                feedbackDismiss: true,
+              },
+            });
+          }
```
capJavert, line 1731: Had to do this since `save` does not support custom update queries and we would overwrite other flags in the future if just doing `save` since `typeorm` does not support nested merging of json fields. That's why we needed to have `updateFlagsStatement` helper.

src/entity/UserPost.ts
```diff
@@ -0,0 +1,66 @@
...
+  @PrimaryColumn({ type: 'text' })
+  userId: string;
+
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @UpdateDateColumn()
+  updatedAt: Date;
+
+  @Column({ type: 'smallint', default: UserPostVote.None })
```
idoshamun, line 46: Can we use tinyint perhaps? Much smaller

```suggestion
  @Column({ type: 'tinyint', default: UserPostVote.None })
```

src/graphorm/index.ts
idoshamun: We can go around it with optional chaining but happy with this implementation as well. We just have to make sure we maintain this default object

src/schema/posts.ts
```diff
@@ -1550,6 +1667,112 @@ export const resolvers: IResolvers<any, Context> = {
...
+              hidden: false,
+            });
+
+            break;
+          case UserPostVote.Down:
+            await userPostRepo.save({
+              postId: id,
+              userId: ctx.userId,
+              vote: UserPostVote.Down,
+              hidden: true,
```
idoshamun, line 1699: Do we hide the post on downvote? I don't remember by now 😓

src/schema/posts.ts
```diff
@@ -1550,6 +1667,112 @@ export const resolvers: IResolvers<any, Context> = {
...
+
+        await ctx.con.transaction(async (entityManager) => {
+          const userPost = await entityManager
+            .getRepository(UserPost)
+            .findOneBy({
+              postId: id,
+              userId: ctx.userId,
+            });
+
+          if (userPost) {
```
idoshamun, line 1743: If you want we can do it in one query, not highly important for this endpoint but more for learning purposes:
https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/
insert on conflict ftw!

src/entity/UserPost.ts
capJavert: `tinyint` is [not supported in psql ](https://www.postgresql.org/docs/8.1/datatype.html) from what I know. `smallint` should be closest alternative.

idoshamun (commented):
One thing that I missed in the DR noticed now is that we will not have the vote date. Can we add a column to indicate the last vote timestamp?

src/entity/UserPost.ts
idoshamun: Ok!

src/schema/posts.ts
capJavert: Yes 😆 you can see it in `downvote` mutation as well.

src/schema/posts.ts
```diff
@@ -421,6 +455,11 @@ export const typeDefs = /* GraphQL */ `
     All the flags for the post
     """
     flags: PostFlagsPublic
+
+    """
+    User state for the post
+    """
+    userState: UserPost
```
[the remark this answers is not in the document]
capJavert: Oh, good catch 👌

src/schema/posts.ts
capJavert: Yeah I tried `repo.upsert` from `typeorm` but would still not work because it does not support passing `UPDATE` statement for column like `repo.update`. I would have to do this in raw SQL which I thought is worse so I went with above. Is there any way that you know it can be done through typeorm?

capJavert: Mentioned it here https://github.com/dailydotdev/daily-api/pull/1389#discussion_r1291021136

idoshamun: You can do it with typeorm query builder, not the actual orm methods. You can look for `onConflict` in our codebase, I believe there are few examples

capJavert: @idoshamun thanks 👌 implemented in 1afbd774c2790040af8b37c7d7b7cd4c514634ee

One note, `onConflict` seems to be deprecated in `typeorm` https://github.com/typeorm/typeorm/issues/8168 and proposed `orUpdate` replacement does not support our case since it only allows to specify which columns will be overwritten in `UPDATE` query.

So I am thinking we might just leave it with my previous implementation without using `onConflict` which will be removed in the future versions. WDYT?

src/schema/posts.ts
```diff
@@ -1550,6 +1667,96 @@ export const resolvers: IResolvers<any, Context> = {
       });
       return { _: true };
     },
+    votePost: async (
```
rebelchris, line 1670: Just checking should we not also support the "old" data here for the time being? (not sure, just raising it as a question for now)

src/schema/posts.ts
```diff
@@ -1550,6 +1667,96 @@ export const resolvers: IResolvers<any, Context> = {
...
+        if (err?.code === TypeOrmError.FOREIGN_KEY) {
+          throw new NotFoundError('Post or user not found');
+        }
+
+        throw err;
+      }
+
+      return { _: true };
+    },
+    dismissPostFeedback: async (
```
rebelchris, line 1726: Happy with this for now, down the line we probably would like a generic function inside as we do for post flags.

rebelchris (approved):
Nothing really blocking from my side.
One small question on retro support.

src/schema/posts.ts
capJavert: After migration PR there is no point in writing to old tables since all data will also be copied into `user_post`. Old migrations are there only to support BC for extension. So idea is to keep these new ones clean and when we start using them their data will already be filled.

src/graphorm/index.ts
capJavert: @rebelchris interested to hear what you think its better here 🙏

src/migration/1691668733329-UserPost.ts
```diff
@@ -0,0 +1,35 @@
...
+            AS
+          $$
+          BEGIN
+            UPDATE user_post SET "votedAt" = now();
+            RETURN NEW;
+          END;
+          $$
+        `)
+        await queryRunner.query('CREATE TRIGGER voted_insert_trigger AFTER INSERT ON "user_post" FOR EACH ROW WHEN (NEW.vote != 0) EXECUTE PROCEDURE voted_at_time()')
+        await queryRunner.query('CREATE TRIGGER voted_update_trigger AFTER UPDATE ON "user_post" FOR EACH ROW WHEN (OLD.vote IS DISTINCT FROM NEW.vote) EXECUTE PROCEDURE voted_at_time()')
```
capJavert, line 26: `typeorm` does not support auto filling date based on specific column change and since their subscriptions don't work with all the typeorm methods and custom queries I decided to update `votedAt` with triggers. This should also keep that column in sync for any future change outside of API.

src/schema/posts.ts
idoshamun: Sounds good to me. Your call :)
Thanks for checking

src/migration/1691668733329-UserPost.ts
idoshamun: @capJavert I think it's better to have before insert and before update and manipulate the payload to reduce the number of CDC messages. Usually it's better to reduce db manipulations when possible

src/migration/1691668733329-UserPost.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "user_post" DROP CONSTRAINT "FK_61c64496bf096b321869175021a"`);
```
idoshamun, line 30: Remember to drop the triggers and procedure here

src/graphorm/index.ts
rebelchris: This is fine with me, as Ido mentioned just important to keep the default in place and up to date.
Perhaps we can create a test case that hardcodes these so we atleast get notified when they don't match?

src/migration/1691668733329-UserPost.ts
capJavert: Implemented here fec4ae2a7957eee8a9b6a40faa430e4f1e072c69

idoshamun (approved):

src/graphorm/index.ts
capJavert: @rebelchris per your comment I looked how this could be more streamlined to avoid double updating, utilised `repo.create` in dee9abe93f4ecb1848fdcf0e86aa25f25684da79, seems cleaner. WDYT?

note if obscure: `repo.create` just creates an object it does not do any db calls

rebelchris: Thats great indeed!

rebelchris (approved):
Looks great

PR #1390, opened 2023-08-11
feat: search sessions history

sshanzel:
A query that accepts the limit and last id for pagination (both optional for the initial request). The query here should return the user's search history.

The limit's default value is 30. Do we maybe set a max value?

WT-1651 #done


src/integrations/magni.ts
```diff
@@ -21,3 +28,36 @@ export const postFeedback = async (
       'Content-Type': 'application/json',
     },
   });
+
+  if (!res.ok) throw new ValidationError(await res.text());
```
sshanzel, line 32: Refactored the `postFeedback` to check for an `ok` response.

src/schema/search.ts
```diff
@@ -1,11 +1,33 @@
...
 import { GQLEmptyResponse } from './common';
 
+type GQLSearchSession = Pick<SearchSession, 'id' | 'prompt' | 'createdAt'>;
+type GQLSearchSessionParams = Pick<SearchSessionParams, 'limit' | 'lastId'>;
+
 export const typeDefs = /* GraphQL */ `
+  type SearchSession {
+    id: String!
+    prompt: String!
+    createdAt: String!
```
sshanzel, line 21: Since the value is coming from an external service, we are unaware of the format for the returned date. Hence, made it a String for safety.

rebelchris (approved):

src/integrations/magni.ts
```diff
@@ -21,3 +28,36 @@ export const postFeedback = async (
...
+  { limit = 30, lastId }: SearchSessionParams = {},
+): Promise<SearchSession[]> => {
+  const params = new URLSearchParams({ limit: limit.toString() });
+
+  if (lastId) params.append('lastId', lastId);
+
+  const url = `${magniOrigin}/sessions?${params.toString()}`;
+  const res = await fetch(url, {
+    ...fetchOptions,
+    headers: { 'X-User-Id': userId },
```
idoshamun, line 55: Don't you miss here the content type header?

src/schema/search.ts
idoshamun: It's better to set everything as Date. It should be ok, if not we will parse it.
External service should be treated just like a database when we approach designing the API

src/schema/search.ts
```diff
@@ -1,11 +1,33 @@
...
+  }
+
+  extend type Query {
+    """
+    Get user's search history
+    """
+    searchSessionHistory(limit: Int, lastId: String): [SearchSession]! @auth
```
idoshamun, line 28: Usually we return a pagination object for such cases. This is the same as reading history and so on

src/schema/search.ts
```diff
@@ -15,6 +37,14 @@ export const typeDefs = /* GraphQL */ `
 `;
 
 export const resolvers: IResolvers<unknown, Context> = traceResolvers({
+  Query: {
+    searchSessionHistory: async (
```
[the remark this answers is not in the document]
idoshamun: It makes sense though, to be explicit because the client isn't aware of the different schemas

src/integrations/magni.ts
sshanzel: We should not need it here as we are not passing anything to `body` to define its type.

src/schema/search.ts
sshanzel: Cool. Let me build a paginator for something outside the graphorm query.

idoshamun: You can utilize the same components and make them generic :)

sshanzel: Yes, I mean, I will use our existing building blocks but make it applicable to our new requirement.

src/integrations/magni.ts
sshanzel: Good point. Let me check with the platform team what will be the supposed expectation gere

sshanzel:
@idoshamun I have re-used the pagination in the graphorm but on a new function. Should be found in this commit https://github.com/dailydotdev/daily-api/pull/1390/commits/c51d225ebcd93491e81ccfa01d7de4ef2b0ead86.

I am not entirely sure with the current setup, but I will be glad to do a coding session when possible.

src/schema/search.ts
```diff
@@ -47,6 +70,12 @@ export const typeDefs = /* GraphQL */ `
   }
 
   extend type Query {
+    """
+    Get user's search history
+    """
+    searchSessionHistory(limit: Int, lastId: String): SearchSessionConnection!
```
idoshamun, line 76: Please note that we usually have different signature here. If you check other paginated queries you will see. We accept a cursor as `after` and it comes before the limit if I'm not mistaken

src/schema/search.ts
```diff
@@ -63,6 +92,18 @@ export const typeDefs = /* GraphQL */ `
...
   Query: {
+    searchSessionHistory: async (
+      _,
+      { limit, lastId }: GQLSearchSessionParams,
+      ctx,
+    ): Promise<ConnectionRelay<GQLSearchSession>> => {
+      return graphorm.queryPaginatedIntegration(
+        () => !!lastId,
+        (nodeSize) => nodeSize === limit,
+        (node) => base64(`timestamp:${node.createdAt}`),
```
idoshamun, line 103: In our case the cursor should be created by the id as it's the property that we pass for the pagination

idoshamun: Let's have a session when you come back from your holiday and I'll explain :)

sshanzel: Amazing. Looking forward to it! 🚀

idoshamun (approved):

PR #1391, opened 2023-08-11
refactor: add retries to kratos calls

idoshamun:


src/kratos.ts
```diff
@@ -15,21 +16,41 @@ const addKratosHeaderCookies = (req: FastifyRequest): RequestInit => ({
...
+      const res = await fetch(endpoint, {
+        ...fetchOptions,
+        ...addKratosHeaderCookies(req),
+        ...opts,
+      });
+      if (res.status < 300) {
+        return { res: await res.json(), headers: res.headers };
+      }
+      const err = new KratosError(res.status, await res.text());
+      if (res.status >= 500) {
```
rebelchris, line 47: Should we maybe add some logs here, just in case to maybe debug down the line?

rebelchris (approved):
Just considering maybe some logs are also helpful?

sshanzel (approved):
Looks great. Hopefully, we will see an immediate reduction of "logged out" users (if not completely fixed).

PR #1395, opened 2023-08-14
feat: set log level through LOG_LEVEL env

capJavert:
Add env to control log level of the app. Useful in development when debugging events for `/e` endpoint.

Falls back to `info` if not defined per [docs](https://github.com/fastify/fastify/blob/main/docs/Reference/Logging.md#enable-logging).

### DIFF
--- a/.infra/Pulumi.adhoc.yaml
+++ b/.infra/Pulumi.adhoc.yaml
@@ -16,6 +16,7 @@ config:
     jwtIssuer: Daily API Staging
     jwtSecret: '|r+.2!!!.Qf_-|63*%.D'
     kratosOrigin: http://heimdall-kratos-public
+    logLevel: info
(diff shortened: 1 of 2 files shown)


capJavert:
I suppose we don't have to add `logLevel` it to prod yaml for now?

.infra/Pulumi.adhoc.yaml
```diff
@@ -16,6 +16,7 @@ config:
     jwtIssuer: Daily API Staging
     jwtSecret: '|r+.2!!!.Qf_-|63*%.D'
     kratosOrigin: http://heimdall-kratos-public
+    logLevel: info
```
capJavert, line 19: I suppose we don't have to add `logLevel` it to prod yaml for now?

PR #1396, opened 2023-08-14
feat: copy user post data from old to new table

capJavert:
- [x] select and insert vote data from `upvote` and `downvote` tables to `user_post`
  - we chunk inserts per createdAt
- [x] insert hidden post data from `hidden_post` table to `user_post`
  - hidden post chunking is done with limit/offset because it does not have `createdAt`
- [x] `votedAt` is updated per old table `createdAt` to preserve vote timestamp

Migration is split into two scripts due to different chunking options as noted above. Did not want to complicate further.

### DIFF
--- a/bin/copyUserPostVoteData.ts
+++ b/bin/copyUserPostVoteData.ts
@@ -1,6 +1,6 @@
 import '../src/config';
 import createOrGetConnection from '../src/db';
-import { Downvote, HiddenPost, Upvote } from '../src/entity';
+import { Downvote, Upvote } from '../src/entity';
 
 (async (): Promise<void> => {
   const entityArgument = process.argv[2];
@@ -10,7 +10,6 @@ import { Downvote, HiddenPost, Upvote } from '../src/entity';
   const argumentToEntityMap = {
     Upvote: Upvote,
     Downvote: Downvote,
-    HiddenPost: HiddenPost,
   };
   const entity = argumentToEntityMap[entityArgument];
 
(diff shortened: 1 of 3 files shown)


bin/copyUserPostVoteData.ts
```diff
@@ -0,0 +1,92 @@
...
+          WHERE "createdAt" > '${createdFromDate.toISOString()}' AND "createdAt" < '${createdToDate.toISOString()}'
+          ON CONFLICT ("postId", "userId") DO UPDATE SET vote = 1;
+        `);
+        await manager.query(`
+          INSERT INTO user_post ("postId", "userId", "votedAt")
+          SELECT "postId", "userId", "createdAt" as "votedAt"
+          FROM upvote
+          WHERE "createdAt" > '${createdFromDate.toISOString()}' AND "createdAt" < '${createdToDate.toISOString()}'
+          ON CONFLICT ("postId", "userId") DO UPDATE SET "votedAt" = EXCLUDED."votedAt";
+        `);
```
capJavert, line 68: Additional query here is to set `votedAt` per old table `createdAt`. We can not set it in one query because `voted_at_time` function in triggers will always overwrite to `now()`. I think it should be fine as long as we don't do many updates in a chunk. For info, we did 100k per chunk easily in posts flags migration.

rebelchris: We can also temporary disable triggers for this script and re-enable at the end?
https://stackoverflow.com/questions/3942258/how-do-i-temporarily-disable-triggers-in-postgresql#:~:text=Alternatively%2C%20if%20you%20are%20wanting,triggers%20for%20the%20current%20session.

rebelchris (approved):
Nothing really urgent from my side.
Just a alternative suggestion.

bin/copyUserPostVoteData.ts
idoshamun: Since we don't have a lot of data I believe we don't even need to split it to different batches. Happy with the current implementation even if it requires an extra query. It's just a one off

idoshamun (approved):

PR #1397, opened 2023-08-15
fix: add feed config names

rebelchris:
Changed the final feed version to be mapped to feed_config_name.
Supporting only the two names at the moment.

I decided to keep the internal based on version, not sure that was intended behaviour?

WT-1661 #done


idoshamun (approved):
Looks ok to me. @pnvasanth can you have a look please?

pnvasanth:
> Looks ok to me. @pnvasanth can you have a look please?

LGTM :rocket:

PR #1398, opened 2023-08-17
feat: integrate user_post throughout the application

capJavert:
Integrating user_post table https://github.com/dailydotdev/daily-api/pull/1389 so it is a source of truth across the application instead of the old tables. 

- [x] replace triggers and add tests for upvotes/downvotes count
- [x] add user_post table to cdc and adjust handlers to read from it
- [x] update postUpvotes query to read from user_post
- [x] update upvoted and downvoted field resolvers to read from user_post
- [x] update workers to read from user_post
- [x] update feedGenerator to use hidden property from user_post

Only thing that writes to old tables now are the mutations which we can remove together with entities after extension adoption.

I went over all the subscriptions [mentioned in the DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/591462420/WT-1620+User+post+data+migration) and they do fire when doing actions through web app too.


__tests__/posts.ts
```diff
@@ -889,7 +890,7 @@ describe('query postUpvotes', () => {
     postUpvotes(id: $id) {
       edges {
         node {
-          createdAt
+          createdAt: votedAt
```
capJavert, line 893: This change will need to be done on `POST_UPVOTES_BY_ID_QUERY` on the clients to preserve interfaces. This is because in `UserPost` the `votedAt` field is the actual time of voting vs `createdAt` that is just insertion date for the row.

capJavert: Tested it, we just need to adjust `Upvote` interface for `votedAt` instead of `createdAt`.

src/migration/1692265909018-VoteTriggers.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export class VoteTriggers1692265909018 implements MigrationInterface {
+    name = 'VoteTriggers1692265909018'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query('DROP TRIGGER IF EXISTS upvote_create_trigger ON upvote')
+      await queryRunner.query('DROP TRIGGER IF EXISTS upvote_delete_trigger ON upvote')
+
+      await queryRunner.query('DROP TRIGGER IF EXISTS downvote_create_trigger ON downvote')
+      await queryRunner.query('DROP TRIGGER IF EXISTS downvote_delete_trigger ON downvote')
```
capJavert, line 11: We remove old triggers to not have `post` be double updated. We don't remove procedures for now to be able to revert migration if needed.

src/schema/posts.ts
```diff
@@ -488,7 +498,7 @@ export const typeDefs = /* GraphQL */ `
   }
 
   type UpvoteEdge {
-    node: Upvote!
+    node: UserPost!
```
idoshamun, line 501: This is a breaking change, to make it backwards compatible we need to introduce `createdAt` field which will be identical to `votedAt`

idoshamun (approved):
That's massive! I'm so happy with the new design 💪

src/graphorm/index.ts
```diff
@@ -422,6 +431,7 @@ const obj = new GraphORM({
     },
   },
   UserPost: {
+    requiredColumns: ['votedAt'],
```
rebelchris, line 434: Why is it required here as seen it's an optional field in the entity

rebelchris (approved):
Wow amazing work here, besides Ido's remark nothing blocking from me

src/graphorm/index.ts
capJavert: From what I understand its not a matter of optional but just so that this column is always added to select db query no matter what user requests from gql query. `Upvote.createdAt` was like that so this is why I also added `votedAt`. cc: @idoshamun is my reasoning ok or do we use `requiredColumns` for something else too?

rebelchris: Ah yes sorry think I'm thinking of the wrong place.
Here it indeed says it will always be added.

capJavert: Yeah, but valid comments because I was wondering if it is needed in the first place.

src/schema/posts.ts
capJavert: @idoshamun from what I see we don't use said `createdAt` anywhere in webapp or extension. 

Also the said `UpvoteEdge` is only used for `postUpvotes` query for sorting, test is confirmed and updated [here](https://github.com/dailydotdev/daily-api/pull/1398/files#diff-6e46397f1e2cc9f81f6d251217c5f625accc5596fbd5e11dcdb644922166ddb9R919-R958). 

So even though its breaking change it should not affect any current client (aside from updating types to prevent using wrong field in the future [as mentioned here](https://github.com/dailydotdev/daily-api/pull/1398#discussion_r1297372829)).

To support this breaking change we would need to introduce the:
- GraphORMType
- this type would use `from` to pull from `user_post` table
- we would have a `createdAt` fields override with `select: 'votedAt'` (since `createdAt` also exists for `user_post`)
  - this will also be confusing I think
- then we would use that type just for this one case
- it would require maintenance down the line even though we completely move to new vote design

So if there is not easier way to do the above I think we can go with this breaking change.

cc: @rebelchris

idoshamun (approved):

src/graphorm/index.ts
capJavert: Cool, since we order by `votedAt` then it should be here ✅

__tests__/posts.ts
capJavert: Adjusted here https://github.com/dailydotdev/apps/pull/2063

PR #1399, opened 2023-08-17
feat: freeform content workflow

omBratteng:
WT-1630


Tiltfile
```diff
@@ -5,7 +5,7 @@ docker_build(
   'api-image',
   context='.',
   dockerfile='./Dockerfile.dev',
-  ignore=['./node_modules', './.infra'],
+  ignore=['./node_modules', './.infra', '__tests__'],
```
omBratteng, line 8: Not sure if it is needed, it just kept on rebuilding whenever I did a change in tests (often)

src/workers/postUpdated.ts
```diff
@@ -182,22 +197,42 @@ const updatePost = async ({
...
+    ? databasePost.visibleAt ?? data.metadataChangedAt
     : null;
 
+  if (content_type in allowedFieldsMapping) {
+    const allowedFields = [
+      'id',
+      'title',
+      'visible',
+      'visibleAt',
+      'flags',
```
omBratteng, line 225: Unsure if `title` should be allowed or not

__tests__/workers/postUpdated.ts
```diff
@@ -237,6 +251,34 @@ it('should update post and modify keywords', async () => {
   expect(postKeywords.length).toEqual(2);
 });
 
+it('should update freeform post and only modify allowed fields', async () => {
+  // await createDefaultKeywords();
```
rebelchris, line 255: Probably missed removal :)

src/workers/cdc.ts
```diff
@@ -292,6 +295,19 @@ const onPostChange = async (
     if (data.payload.after.visible) {
       await notifyPostVisible(logger, data.payload.after);
     }
+    if (data.payload.after.type === PostType.Freeform) {
+      const freeform = data as unknown as ChangeMessage<FreeformPost>;
```
rebelchris, line 299: Why is the `as unknown` needed here?

src/workers/postUpdated.ts
```diff
@@ -182,22 +197,42 @@ const updatePost = async ({
...
-    databasePost.metadataChangedAt.toISOString() >= updatedDate.toISOString()
+    databasePost.metadataChangedAt.toISOString() >=
+      data.metadataChangedAt.toISOString()
   ) {
     return null;
   }
 
   const title = data?.title || databasePost.title;
-  const updateBecameVisible = !databasePost.visible && !!title?.length;
+  const updateBecameVisible =
```
rebelchris, line 207: Don't think this is particularly right, would mean we always set freeform to true, which might be the case now. but in some instances it might actually be set to false for whatever reason.
I think we can keep title.length, but take the DB value for freeform ones?

src/workers/postUpdated.ts
rebelchris: Title should not be allowed to be updated in this flow.

rebelchris: Also technically the ID should never change here.

src/workers/postUpdated.ts
```diff
@@ -291,6 +327,7 @@ const fixData = async ({
   // Try and fix generic data here
   return {
     mergedKeywords,
+    content_type: data?.content_type as PostType,
```
rebelchris, line 330: In this flow do you need `content_type` at all?

__tests__/workers/postUpdated.ts
```diff
@@ -237,6 +251,34 @@ it('should update post and modify keywords', async () => {
...
+  });
+  const post = await con.getRepository(ArticlePost).findOneBy({ id: 'p2' });
+  expect(post.metadataChangedAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
+  expect(post.visible).toEqual(true);
+  expect(post.flags.visible).toEqual(true);
+  expect(post.visibleAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
+  expect(post.contentCuration).toEqual(['news', 'story', 'release']);
+  expect(post.siteTwitter).toEqual('text');
+
+  expect(post.title).toEqual('test');
```
rebelchris, line 277: This should be `freeform post`

rebelchris (changes requested):
Looks great, some minor comments and question.
Overall very neat implementation 👏

__tests__/workers/postUpdated.ts
omBratteng: Yeah, removed it now

__tests__/workers/postUpdated.ts
omBratteng: Removed title from allowed fields, and set this to now be equal to `freeform post`.

src/workers/postUpdated.ts
omBratteng: Right now only the updatePost uses `content_type`, but I added it to fixData in case createPost would need it in the future.
I can remove it from fixData if that is preferred.

omBratteng:
For some reason I can comment on this comment https://github.com/dailydotdev/daily-api/pull/1399#discussion_r1297179041
> Also technically the ID should never change here.

Yeah. I think I actually could removed the always allowed fields, and just move the entire if section a bit up to line 210-212 someting.

src/workers/postUpdated.ts
omBratteng: Removed the title.

omBratteng: Yeah, the ID should never change, but to not remove it from the data object, where it is defined a bit further up, I added it to the "always allowed" fields. Same with the rest.
https://github.com/dailydotdev/daily-api/blob/c00f3713688ca1bca848d15bcdbaa2bbc74bfef7/src/workers/postUpdated.ts#L212

src/workers/cdc.ts
```diff
@@ -292,20 +295,50 @@ const onPostChange = async (
...
         if (
           isChanged(freeform.payload.before, freeform.payload.after, 'content')
         ) {
           await notifyPostContentEdited(logger, data.payload.after);
         }
       }
     }
+
+    if (data.payload.after.type === PostType.Freeform) {
+      const freeform = data as unknown as ChangeMessage<FreeformPost>;
```
rebelchris, line 327: you probably also don't need the `as unknown` here.
Really can't recall why it's introduced 😅

src/workers/postUpdated.ts
rebelchris: Ah i see, no that's fine with me.
Just didn't see any use so wasn't sure.

Let's keep it for now :)

src/workers/postUpdated.ts
omBratteng: Changed it to use the database value.

rebelchris (approved):
Good from my side.
@idoshamun would love to get your view as well.

denisb0 (approved):
LGTM
hopefully didn't miss anything important because of limited understanding of the project and language.

omBratteng:
@denisb0 think the most important might be what's getting sent to Yggdrasil
https://github.com/dailydotdev/daily-api/blob/c00f3713688ca1bca848d15bcdbaa2bbc74bfef7/src/workers/cdc.ts#L304-L308

denisb0:
Yes, matches what I am expecting

Tiltfile
idoshamun: Perfect!

src/workers/cdc.ts
```diff
@@ -292,20 +295,50 @@ const onPostChange = async (
...
+
+    if (data.payload.after.type === PostType.Freeform) {
+      const freeform = data as ChangeMessage<FreeformPost>;
+      if (
+        Math.abs(
+          freeform.payload.before.content.length -
+            freeform.payload.after.content.length,
+        ) >= FREEFORM_POST_MINIMUM_CHANGE_LENGTH
+      ) {
+        await notifyContentRequested(logger, {
```
idoshamun, line 334: Let's abstract this to a function that accepts freeform and triggers the message just in case

src/workers/postUpdated.ts
```diff
@@ -182,22 +197,41 @@ const updatePost = async ({
...
 
   const title = data?.title || databasePost.title;
-  const updateBecameVisible = !databasePost.visible && !!title?.length;
+  const updateBecameVisible =
+    content_type === PostType.Freeform
+      ? databasePost.visible
+      : !databasePost.visible && !!title?.length;
 
   data.id = databasePost.id;
-  data.metadataChangedAt = updatedDate;
```
idoshamun, line 194: Why did you remove this line?

omBratteng: I moved it into the fixData function

omBratteng: https://github.com/dailydotdev/daily-api/blob/d614e55bc016f47c98744117e877d52435f7283d/src/workers/postUpdated.ts#L341-L342

src/workers/cdc.ts
omBratteng: Which part?

idoshamun: The call to notifyContentRequested

omBratteng: Extracted the call to a separate function and the if statements

idoshamun (approved):

PR #1402, opened 2023-08-20
fix: new banner optional flow

rebelchris:
Moving the banner call to a boot Redis cached flow.
this way we don't have to call it from frontend when not needed.

I need to set it to "false" on deletion as there is no fallback, and otherwise it would default to "true" on the evaluation.

WT-1671 #done

Related PRs:
[Streams](https://github.com/dailydotdev/streams/pull/14)
[Apps](https://github.com/dailydotdev/apps/pull/2061)


.infra/workers.ts
```diff
@@ -191,6 +191,10 @@ export const workers: Worker[] = [
     topic: 'api.v1.post-visible',
     subscription: 'api.post-changelog-added-v2',
   },
+  {
```
idoshamun, line 194: We need to add the second worker as well for the deleted banner

rebelchris: Well spotted accidentally removed at testing a other flow 😅

idoshamun (approved):

idoshamun (approved):

PR #1405, opened 2023-08-21
refactor: introduce flexible design to support new feed configs

idoshamun:
This PR is quite big and it breaks our feed service integration down to multiple components:
* Client - To dispatch the request to the feed service, also allows to cache the response
* Config - Different feeds have different config, hence I introduced a class to generate them according to different rules. (e.g. config based on the feed preferences)
* Generator - Takes the two above and generates a feed

I added comments wherever necessary, if something's not clear do let me know.

Note:  I remove the change in configuration based on the feed version for the popular feed. Hopefully very soon we'll have a dedicated feed config for popular. Other than one test for the case above, existing tests were green right away meaning it should be good to go.


src/integrations/feed/index.ts
```diff
@@ -0,0 +1,12 @@
...
+  FeedPreferencesConfigGenerator,
+  SimpleFeedConfigGenerator,
+} from './configs';
+export { FeedClient, CachedFeedClient } from './clients';
+export {
+  FeedGenerator,
+  feedGenerators,
+  versionToFeedGenerator,
+  cachedFeedClient,
+} from './generators';
```
capJavert, line 12: We use index export in webapp lately, we should continue doing it for api too, this is a good start.

src/integrations/feed/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    feedId: string | undefined,
+    pageSize: number,
+    offset: number,
+  ): Promise<FeedConfig>;
+}
+
+/**
+ * An interface for a feed service client
+ */
+export interface IFeedClient {
```
capJavert, line 54: I like `I` prefix for interfaces, `E` for enum eg. `EActionType` and etc. It makes it easier to know if something is interface, class or enum just by looking at name/import. Maybe we should standardize in that direction from now on 👌

src/integrations/feed/generators.ts
```diff
@@ -0,0 +1,69 @@
...
+const client = new FeedClient();
+export const cachedFeedClient = new CachedFeedClient(client, ioRedisPool);
+const opts = {
+  includeBlockedTags: true,
+  includeAllowedTags: true,
+  includeBlockedSources: true,
+  includeSourceMemberships: true,
+};
+
+export const feedGenerators: Record<string, FeedGenerator> = Object.freeze({
```
capJavert, line 49: I think it would be better for this to have some type other then `string`, it breaks other typescript features down the line with iterators like reduce and such. So something like `Record<FeedVersion, FeedGenerator>` ..or just `type` with predefined values would work better.

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
...
+        return cached;
+      }
+    } catch (e) {
+      ctx?.log.error({ err: e }, 'failed to get feed from cache');
+    }
+    const cloneConfig = { ...config };
+    delete cloneConfig.offset;
+    const feedRes = await this.client.fetchFeed(ctx, feedId, cloneConfig);
+    if (feedRes.length) {
+      setTimeout(() => this.updateCache(feedId, config, feedRes));
```
capJavert, line 163: Why `setTimeout`?

__tests__/integrations/feed.ts
```diff
@@ -0,0 +1,258 @@
...
+});
+
+describe('CachedFeedClient', () => {
+  it('should fetch subsequent pages from cache', async () => {
+    const mockClient = mock<IFeedClient>();
+    mockClient.fetchFeed.mockResolvedValueOnce(feedResponse);
+    const feedClient = new CachedFeedClient(mockClient, ioRedisPool);
+    const page0 = await feedClient.fetchFeed(ctx, 'id', config);
+    expect(page0).toEqual(feedResponse.slice(0, 2));
+    await new Promise((resolve) => setTimeout(resolve, 50));
```
capJavert, line 98: `process.nextTick` should work right?

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
...
+        `${key}:time`,
+        new Date().toISOString(),
+        'EX',
+        ONE_DAY_SECONDS,
+      );
+      pipeline.set(
+        `${key}:posts`,
+        JSON.stringify(items),
+        'EX',
+        ONE_DAY_SECONDS,
```
capJavert, line 143: I seen this before all around, maybe not applicable for this but while we are refactoring maybe create a key generator to abstract this syntax to something easier to repeat all around

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  private async internalFetchFeed(config: FeedConfig): Promise<{
+    data: { post_id: string; metadata: Record<string, string> }[];
+  }> {
+    const res = await fetch(this.url, {
+      ...this.fetchOptions,
+      method: 'POST',
+      body: JSON.stringify(config),
+    });
+    if (res.status >= 200 && res.status < 300) {
```
capJavert, line 33: `res.ok` should do, its `200-299`

capJavert (commented):
Nice abstraction 👌, pretty readable. Few general comments and ideas.

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
+import { FeedConfig, FeedResponse, IFeedClient } from './types';
+import fetch, { RequestInit } from 'node-fetch';
+import { fetchOptions as globalFetchOptions } from '../../http';
+import pRetry, { AbortError } from 'p-retry';
+import { IORedisPool } from '@dailydotdev/ts-ioredis-pool';
+import { Context } from '../../Context';
+
+const ONE_DAY_SECONDS = 24 * 60 * 60;
```
rebelchris, line 8: We already have this in `src/redis` maybe make it a common util export instead?

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
...
+    url = process.env.INTERNAL_FEED,
+    fetchOptions: RequestInit = globalFetchOptions,
+  ) {
+    this.url = url;
+    this.fetchOptions = fetchOptions;
+  }
+
+  private async internalFetchFeed(config: FeedConfig): Promise<{
+    data: { post_id: string; metadata: Record<string, string> }[];
+  }> {
```
rebelchris, line 27: Not really blocking, but maybe make this a interface return type? (Just more readable to me)

rebelchris: Actually would `FeedResponse` not do?

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
...
+      throw new AbortError(`feed service request is invalid: ${res.status}`);
+    }
+    throw new Error(`unexpecetd response from feed service: ${res.status}`);
+  }
+
+  async fetchFeed(ctx, feedId, config): Promise<FeedResponse> {
+    const res = await pRetry(() => this.internalFetchFeed(config), {
+      retries: 5,
+    });
+    if (!res.data.length) {
```
rebelchris, line 47: Could it technically not even return data?
Not sure we need to optionally chain it maybe?

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (offset) {
+      return true;
+    }
+    const updateKey = `${this.getCacheKeyPrefix(feedId)}:update`;
+    const [lastGenerated, lastUpdated] = await this.redisPool.execute(
+      async (client) => {
+        return client.mget(`${key}:time`, updateKey);
+      },
+    );
+    const ttl = 3 * 60 * 1000;
```
rebelchris, line 92: This should also be a global defined named variable I think

src/integrations/feed/clients.ts
```diff
@@ -0,0 +1,167 @@
...
+import pRetry, { AbortError } from 'p-retry';
+import { IORedisPool } from '@dailydotdev/ts-ioredis-pool';
+import { Context } from '../../Context';
+
+const ONE_DAY_SECONDS = 24 * 60 * 60;
+
+/**
+ * Naive implementation of a feed client that fetches the feed from the feed service
+ */
+export class FeedClient implements IFeedClient {
```
rebelchris, line 13: In which case would you use FeedClient directly and not the Cached version?
Just thinking we might want to restrict which one you can call?

src/integrations/feed/configs.ts
```diff
@@ -0,0 +1,80 @@
+import { FeedConfig, FeedConfigGenerator } from './types';
+import { feedToFilters } from '../../common';
+
+type Options = {
```
rebelchris, line 4: Should we not also have something for old config: `removeHiddenPosts`
And `allowSquadPosts` (the public ones)

src/integrations/feed/generators.ts
```diff
@@ -0,0 +1,69 @@
...
+  '11': new FeedGenerator(
+    cachedFeedClient,
+    new FeedPreferencesConfigGenerator(
+      { feed_config_name: 'personalise' },
```
rebelchris, line 53: We have a `FeedConfigName` enum which I think would be best to use to avoid string crashing down the line.

src/integrations/feed/generators.ts
```diff
@@ -0,0 +1,69 @@
...
+    ),
+  ),
+  '14': new FeedGenerator(
+    cachedFeedClient,
+    new FeedPreferencesConfigGenerator({ feed_config_name: 'vector' }, opts),
+  ),
+  popular: new FeedGenerator(
+    cachedFeedClient,
+    new SimpleFeedConfigGenerator({ feed_config_name: 'personalise' }),
+  ),
```
rebelchris, line 64: This is just a demo for now I assume?
Will the end result be something as `feed_config_name: Popular`?

src/integrations/feed/types.ts
```diff
@@ -0,0 +1,66 @@
+import { Context } from '../../Context';
+
+export type FeedResponse = [postId: string, metadata: string | undefined][];
+
+export type FeedProvider = {
```
rebelchris, line 5: With these options, do we actually want to make these known to API?
Is it not something feed service handles.

Mainly raising it to have the API be as dumb as possible and simple request a feed with some basic props.

src/schema/feeds.ts
```diff
@@ -857,15 +858,16 @@ const feedResolverV2: IFieldResolver<
...
       page,
     ) =>
-      generatePersonalizedFeed({
-        con: ctx.con,
-        pageSize: page.limit,
-        offset: page.offset,
-        feedVersion: args.version,
-        userId: ctx.userId || ctx.trackingId,
-        feedId: args.feedId,
+      (args.feedId
```
rebelchris, line 861: Not super important, but can't we clean out some logic of this feedResolver as well?
Seems very heavy still.

rebelchris (commented):
Wow amazing work!
Overall looks great, have some minor comments

src/integrations/feed/clients.ts
idoshamun: I don't want to block returning the feed for storing it in the cache. I think it's higher priority to return asap. Wdyt?
Sidenote: it was part of the previous implementation

src/integrations/feed/generators.ts
idoshamun: Agree! Will refactor

src/integrations/feed/types.ts
idoshamun: I don't like it tbh 🙈 I just had no other alternative because I wanted to name the actual implementation `FeedService`. :lolsob:

__tests__/integrations/feed.ts
idoshamun: I'll try!

src/integrations/feed/clients.ts
idoshamun: Cache is gonna move to feed service side, so I don't want to introduce any changes here. I copied it as it was before. Wdyt?

src/integrations/feed/clients.ts
idoshamun: Will do!

src/integrations/feed/clients.ts
idoshamun: Good point!

src/integrations/feed/clients.ts
idoshamun: `FeedResponse` is a bit different. I'll create a dedicated type, no worries

src/integrations/feed/clients.ts
idoshamun: Good catch!

src/integrations/feed/clients.ts
idoshamun: Yap!

src/integrations/feed/clients.ts
idoshamun: There can be temporary feeds which you will not want to cache and anyway caching is gonna move to feed service side. So I think it's better to keep it flexible and have both options

src/integrations/feed/configs.ts
idoshamun: It's not supported by feed service at the moment

src/integrations/feed/generators.ts
idoshamun: I deleted it in this PR but I can recover it :)

src/integrations/feed/generators.ts
idoshamun: It's not a demo, it's what currently exists. In the next PR I'll change it, I just don't want to introduce so many changes.
First refactor then significant changes

src/integrations/feed/clients.ts
rebelchris: Got it!
Just means keeping a good eye on review including anything feeds to see which ones are used and should be used.

src/integrations/feed/generators.ts
rebelchris: Would prefer the enum, just so it's more "strict"

src/integrations/feed/generators.ts
rebelchris: Got it, that's fine.
So how does feed service make a popular feed based on personalise? 
It's then basically the same as my feed?

src/integrations/feed/clients.ts
capJavert: Yes, that works of course

src/integrations/feed/types.ts
capJavert: Thats also one of the reasons for it, you can have interface and based implementation named similarly. But ok, I jumped the gun 😆

src/integrations/feed/clients.ts
capJavert: Get that but if `this.updateCache` is a promise and we don't await it will resolve even without `setTimeout` right?

src/integrations/feed/configs.ts
```diff
@@ -0,0 +1,80 @@
...
+  includeAllowedTags?: boolean;
+  includeBlockedTags?: boolean;
+  includeBlockedSources?: boolean;
+  includeSourceMemberships?: boolean;
+};
+
+function getDefaultConfig(
+  baseConfig: Partial<FeedConfig>,
+  userId: string | undefined,
+  feedId: string | undefined,
```
sshanzel, line 14: The `baseConfig` has a property `user_id`, why did we end up having the second parameter as the `userId`? The same goes for the `feedId`, though understandable as we want to set a default value for it, if so, we should maybe move it as the last parameter and set the default value from there.

If some of the properties are not really expected from the `constructor` when setting the `FeedConfig`, we should maybe use `Omit` to make it straightforward that those values would only come from the generate function parameters.

src/integrations/feed/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  blocked_tags?: string[];
+  blocked_sources?: string[];
+  squad_ids?: string[];
+  providers?: Record<string, FeedProvider>;
+};
+
+export interface FeedConfigGenerator {
+  generate(
+    ctx: Context,
+    userId: string | undefined,
```
sshanzel, line 44: From the `Context`, we should be able to fetch the user's id right?

sshanzel (commented):
Looks amazing! It is much easier to navigate how we generate the feed now. Even the tests show how quick you can setup the tests and consistently get the expected results. Most have been covered already from Chris and Ante's reviews, only had some comments about optional parameters.

src/schema/feeds.ts
idoshamun: Will do!

src/integrations/feed/types.ts
idoshamun: Actually, I'm gonna use it in the next PR partially. We actually want to give the API the flexibility to play around and experiment with some stuff based on user's state. But we can add it as necessary if you prefer

rebelchris: Hmm might just make things messy from API side.
From that perspective I'd hoped most logic stayed on feed service, but open to discussion

idoshamun: Ok, let's talk about it when we get the chance. And we can have stricter rules in case someone changes it

src/integrations/feed/types.ts
idoshamun: So some feeds may ask for a different userId or not provide a userId at all even if you are logged in. So I wanted to be explicit about it

src/integrations/feed/index.ts
idoshamun: I love index exports and try to use them where I can. It creates a design of small libraries within the project. Make a lot of sense to me

src/integrations/feed/clients.ts
capJavert: Got it, just saying it is not like `setTimeout` moves it to to the "background" it moves it further down event loop stack.

This should do the same thing in sending a response early and promise will still execute async without blocking the response:
```js
if (feedRes.length) {
  this.updateCache(feedId, config, feedRes);
}

return feedRes.slice(offset, pageSize + offset);
```

So everything is the same I just think `setTimeout` is not needed.

idoshamun: Looking into the implementation of `updateCache` I see there's a top level `await` there but if there wasn't the situation would have been different. In this case you're right it's almost the same.
But let's imagine the `await` or the async operation was down the line in this function then using `setTimeout` would make a different than not having await.
The event callback doesn't continue to the next event unless the current event finished processing or is now idle because it waits for an I/O operation.
Meaning, `updateCache` would start its execution immediately until the first encounter of an I/O operation. So if before the first encounter there's heavy processing it will delay sending back the response. It's mostly nuances but sometimes it matters. This is at least according to my understanding

capJavert: When talking in scope of this function having or not having `setTimeout` the `return` line will always execute before anything related to `updateCache`. 

When talking in scope of the whole request processing here there is no way of knowing when `setTimeout` callback will get executed or in what state the queues are since `fetchFeed` is also async and there is more async stuff that executed until actual `req.send` is called internally by fastify.. at least not without performance testing. It can also change while we update the code of course.

If it was like this before let's not change it because it will not hurt response time. It's just a potential head scratcher for anyone looking at the code.

src/integrations/feed/generators.ts
idoshamun: Right now it's very similar that's the problem

__tests__/__snapshots__/feeds.ts.snap
```diff
@@ -156,6 +156,8 @@ Object {
 }
 `;
 
+exports[`mutation addFiltersToFeed should not authorize when not logged-in 1`] = `null`;
```
rebelchris, line 159: Just wondering if this was intended?

src/integrations/feed/generators.ts
rebelchris: Should be `FeedVersion` type now

rebelchris (approved):
Looks alright to me.
Non breaking comments

__tests__/__snapshots__/feeds.ts.snap
idoshamun: Nope

capJavert (approved):
Looking great 👌

PR #1411, opened 2023-08-23
fix: extra invalidation on image response

rebelchris:
We already invalidate the cloudinary flow, but cloudinary makes it so that the URL will remain the same on a update.
To ensure a browser reload we added a optional version number which is current date timestamp.

WT-1670 #done 

Further discussion: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1692365364463109


src/schema/posts.ts
```diff
@@ -1410,7 +1410,7 @@ export const resolvers: IResolvers<any, Context> = {
             UploadPreset.PostBannerImage,
             { invalidate: true },
           );
-          updated.image = coverImageUrl;
+          updated.image = `${coverImageUrl}v=${Date.now()}`;
```
sshanzel, line 1413: I think we are missing the question mark for the query parameters.

sshanzel (approved):
Looks great! 🚀

idoshamun:
@rebelchris feels a bit hacky, doesn't cloudinary provide a way to do it via their sdk? Probably there's an argument that returns a new url

rebelchris:
@idoshamun Can look into that :)
As it was already pretty much edge-casey didn't put too much time into it.

rebelchris:
@idoshamun / @sshanzel Went with a force_version if we want to invalidate as well.
It forces the cloudinary datestamp version string instead

idoshamun (approved):

PR #1413, opened 2023-08-23
refactor: update popular feed config

idoshamun:
Set popular feed to use different configs then my feed

### DIFF
--- a/src/integrations/feed/generators.ts
+++ b/src/integrations/feed/generators.ts
@@ -70,7 +70,19 @@ export const feedGenerators: Record<FeedVersion, FeedGenerator> = Object.freeze(
     popular: new FeedGenerator(
       cachedFeedClient,
       new SimpleFeedConfigGenerator({
-        feed_config_name: FeedConfigName.Personalise,
+        providers: {
+          fresh: {
+            enable: true,
+            remove_engaged_posts: true,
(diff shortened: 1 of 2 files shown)


src/integrations/feed/generators.ts
```diff
@@ -70,7 +70,19 @@ export const feedGenerators: Record<FeedVersion, FeedGenerator> = Object.freeze(
...
+            remove_engaged_posts: true,
+            page_size_fraction: 0.1,
+          },
+          engaged: {
+            enable: true,
+            remove_engaged_posts: true,
+            page_size_fraction: 1,
+            fallback_provider: 'fresh',
+          },
+        },
```
capJavert, line 85: Just for understanding, does these options represent what feed_config_name: 'personalise' used to mean or did we do some actual change on how this feed works to achieve something different?

capJavert (approved):

src/integrations/feed/generators.ts
idoshamun: It's a totally different setting to that should be aligned with the meaning of popular feed. You can see this definition as an adhoc feed rather than a fixed configuration (when using the config name). Feed service provides both

capJavert: Got it, tnx for explanation 🤝

PR #1415, opened 2023-08-25
fix: magni session url

rebelchris:
Single session URL was defined wrong and not according to spec.


idoshamun (approved):

sshanzel (approved):
Nice catch! Thanks for the fix 🚀

sshanzel:
You might have to update the test where it mocks the request towards Magni.

PR #1417, opened 2023-08-26
fix: always return search recs when possible

idoshamun:
Previous query took 5 recent upvotes without checking whether there are any questions related to these posts. I added a check to make sure we always return search recs when possible.

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -55,7 +55,6 @@ import {
   UserPost,
   UserPostFlagsPublic,
   UserPostVote,
-  NotificationPreference,
 } from '../entity';
 import { GQLEmptyResponse } from './common';
 import {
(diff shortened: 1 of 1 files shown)


src/schema/posts.ts
```diff
@@ -1122,13 +1122,20 @@ export const resolvers: IResolvers<any, Context> = {
...
         queryBuilder: builder.queryBuilder
           .innerJoin(
-            (query) =>
-              query
+            (query) => {
+              const sub = query
+                .subQuery()
+                .select('id')
+                .from(PostQuestion, 'pq')
+                .where(`"pq"."postId" = "up"."postId"`);
```
rebelchris, line 1130: Maybe we should add a order by upvotes for the subs, but fully optional, this will do fine

rebelchris (approved):

src/schema/posts.ts
idoshamun: There's already an order by votedAt :)

PR #1418, opened 2023-08-31
fix: add workers

rebelchris:
Adding the relevant workers that handle the banner states


sshanzel (approved):
Good catch!

Is this the banner on top of our feed pages?

rebelchris:
@sshanzel It's the banner on top of the header for the monthly dev for instance.

idoshamun (approved):

PR #1419, opened 2023-08-31
feat: utility route to get user's id

idoshamun:
I added a route that returns the user's id. I need it for support cases mainly


sshanzel (approved):
Looks good. Would we need this for anon users that uses `req.trackingId`?

idoshamun:
@sshanzel good catch but only registered users

PR #1420, opened 2023-09-01
feat: update squad image placeholder

capJavert:
Frontend PR https://github.com/dailydotdev/apps/pull/2151

### DIFF
--- a/src/migration/1672583245776-SourceImageDefault.ts
+++ b/src/migration/1672583245776-SourceImageDefault.ts
@@ -7,7 +7,7 @@ export class SourceImageDefault1672583245776 implements MigrationInterface {
     await queryRunner.query(`ALTER TABLE "source"
       ALTER COLUMN "image" SET NOT NULL`);
     await queryRunner.query(`ALTER TABLE "source"
-      ALTER COLUMN "image" SET DEFAULT 'https://daily-now-res.cloudinary.com/image/upload/s--LrHsyt2T--/f_auto/v1692632054/squad_placeholder_sfwkmj'`);
+      ALTER COLUMN "image" SET DEFAULT 'https://daily-now-res.cloudinary.com/image/upload/v1672041320/squads/squad_placeholder.jpg'`);
   }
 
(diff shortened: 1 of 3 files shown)


src/migration/1672583245776-SourceImageDefault.ts
```diff
@@ -7,7 +7,7 @@ export class SourceImageDefault1672583245776 implements MigrationInterface {
     await queryRunner.query(`ALTER TABLE "source"
       ALTER COLUMN "image" SET NOT NULL`);
     await queryRunner.query(`ALTER TABLE "source"
-      ALTER COLUMN "image" SET DEFAULT 'https://daily-now-res.cloudinary.com/image/upload/v1672041320/squads/squad_placeholder.jpg'`);
+      ALTER COLUMN "image" SET DEFAULT 'https://daily-now-res.cloudinary.com/image/upload/s--LrHsyt2T--/f_auto/v1692632054/squad_placeholder_sfwkmj'`);
```
omBratteng, line 10: Shouldn't it be a new migration, instead of altering an already existing migration?

rebelchris: Yeah it should be a new migration, this won't trigger

rebelchris (changes requested):
We need to add a new migration or manually set it in DB

src/migration/1672583245776-SourceImageDefault.ts
capJavert: I know it wont trigger, I just thought there is no point in updating defaults with current squads. 

I'll add the new one if needed 👍

rebelchris: Ah yes, but we want the new default to be something correct?
So if squads get created without image it needs to default to this new one.

capJavert: Yes, mine oversight, will take care od it.

rebelchris (approved):
Good to go

omBratteng (approved):

idoshamun (approved):

PR #1421, opened 2023-09-04
feat: private endpoints to fetch tags, sources, content curations (to be used from snotra)

mykytanikitenko:


src/routes/privateSnotra.ts
```diff
@@ -0,0 +1,54 @@
...
+      return res.status(400).send();
+    }
+
+    const con = await createOrGetConnection();
+    const result = await con.query(`
+       SELECT value AS name
+       FROM keyword
+       WHERE status = 'allow'
+       ORDER BY name ASC
+    `);
```
rebelchris, line 16: We don't really use the query command on API side, prefer to use the build in query builder, examples is the `getSquads` function in boot for instance.

src/routes/privateSnotra.ts
```diff
@@ -0,0 +1,54 @@
...
+    return res.status(200).send(result);
+  });
+
+  fastify.get('/all_sources', async (req, res) => {
+    if (!req.service) {
+      return res.status(400).send();
+    }
+
+    const con = await createOrGetConnection();
+    const result = await con.query(`
```
rebelchris, line 27: Same remark as above

src/routes/privateSnotra.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  fastify.get('/all_content_curations', async (req, res) => {
+    if (!req.service) {
+      return res.status(400).send();
+    }
+
+    const con = await createOrGetConnection();
+    // this query might be hard on a database, we can consider extracting content curations
+    // to a separate table
+    const result = await con.query(`
```
rebelchris, line 46: Same

rebelchris (changes requested):
The queries look fine, we just prefer them as querybuilder objects.
I'm also not sure do we need the new rout instance or can we put them under `/p` in general?

(No real preference from my side there)

idoshamun (changes requested):
I agree with Chris regarding the query building. For all I care we can keep it under `/p/snotra` for now I don't have any preference.
We should also introduce some tests as a safety net. More like this endpoint exists and returns something if we don't want proper tests

__tests__/privateSnotra.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it('should return all tags', async () => {
+    const { body } = await request(app.server)
+      .get('/p/snotra/all_tags')
+      .set('authorization', `Service ${process.env.ACCESS_SECRET}`)
+      .send()
+      .expect(200);
+
+    expect(body).toMatchSnapshot();
```
rebelchris, line 42: Prefer inline tests to snapshot but it's non blocking at this point.

src/routes/privateSnotra.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  fastify.get('/all_content_curations', async (req, res) => {
+    if (!req.service) {
+      return res.status(400).send();
+    }
+
+    const con = await createOrGetConnection();
+    // this query might be hard on a database, we can consider extracting content curations
+    // to a separate table
+    const result = await con
```
rebelchris, line 50: @idoshamun You happen to know any optimisations here?
Not sure how often this endpoint will be called/implications.,

rebelchris (approved):
Pretty much good from my side.
Think for future good to note we prefer to not use snapshots as they get very messy and not correct.

src/routes/privateSnotra.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const con = await createOrGetConnection();
+    const result = await con
+      .createQueryBuilder()
+      .select('id', 'name')
+      .from(Source, 'source')
+      .where('type = :type', { type: 'machine' })
+      .andWhere('private = :private', { private: false })
+      .andWhere('active = :active', { active: true })
+      .orderBy('name', 'ASC')
```
idoshamun, line 36: Do we order by name or id in BQ?

src/routes/privateSnotra.ts
idoshamun: @mykytanikitenko not ideal, as you rightfully pointed, let's add a filter to all posts of the last 30 days or so. Should give us enough. Another option is to use static options for now. Anyway our NLP method returns fixed options. It's very structured

__tests__/privateSnotra.ts
mykytanikitenko: Anyway changed this :)

src/routes/privateSnotra.ts
mykytanikitenko: I added the last 30 days. Honestly, I would prefer query to static options. Once a day during the night, it shouldn't hit the db very hard :)

src/routes/privateSnotra.ts
mykytanikitenko:
```
SELECT * FROM EXTERNAL_QUERY("us.api","SELECT CONCAT('source_', LOWER(id)) AS name FROM source WHERE type = 'machine' AND private = false AND active")
UNION ALL
SELECT * FROM EXTERNAL_QUERY('us.api',"SELECT CONCAT('tag_', LOWER(value)) AS name from keyword where status = 'allow'")
ORDER BY 1
```
This is an original query, it uses a different approach: takes all tags and sources and orders them together later. I added ordering to receive consistent responses, but it is not strongly necessary.

idoshamun: Don't get me wrong, I think the ordering is needed but I'd suggest using id. Name can be changed, id can't

idoshamun (approved):
You have green light from me. Just check my latest comment and decide what you think :)

src/routes/privateSnotra.ts
mykytanikitenko: @idoshamun aah, ok, I got what you mean now. Sorry, I got confused xD. In the sources table, we have columns "id" and "name", the name can be changed, id can not. I am querying ID as NAME, so it is should be fine :)

idoshamun: Ok then we can merge!

PR #1422, opened 2023-09-04
feat: squad leader to mailing list

capJavert:
Add squad leaders to new mailing list.

Topic added https://github.com/dailydotdev/streams/pull/16

We will backfill current owners in a one time import to sendgrid from CSV. I got the query ready.


capJavert:
@rebelchris I am guessing we also need to pre-fill this list once, how do we do that usually? With a script?

.infra/workers.ts
```diff
@@ -215,4 +215,8 @@ export const workers: Worker[] = [
     topic: 'analytics-api.v1.experiment-allocated',
     subscription: 'api.experiment-allocated',
   },
+  {
+    topic: 'api.v1.member-joined-source',
+    subscription: 'api.member-joined-source-mailing',
+  },
```
capJavert, line 221: Added new subscription here, better then cram up this logic into existing that is dedicated to notifications?

__tests__/workers/squadMemberJoinedMailing.ts
```diff
@@ -0,0 +1,94 @@
...
+    await expectSuccessfulBackground(worker, {
+      sourceMember,
+    });
+    expect(getContactIdByEmail).toBeCalledTimes(1);
+    expect(getContactIdByEmail).toBeCalledWith('tsahi@daily.dev');
+    expect(addUserToContacts).toBeCalledTimes(1);
+    expect(addUserToContacts).toBeCalledWith(
+      user,
+      [LIST_DRIP_CAMPAIGN],
+      undefined,
```
capJavert, line 73: I've put `undefined` here, for some reason when mocking `getContactIdByEmail` it was always returning `undefined`. Not sure if due to our jest setup but could not find anything wrong.

src/common/mailing.ts
```diff
@@ -173,7 +175,10 @@ export const updateUserContactLists = async (
       lists.push(LIST_MARKETING_EMAILS);
     } else if (contactId) {
       // If they no longer subscribe to marketing emails remove them from the contact list
-      await removeUserFromList(LIST_MARKETING_EMAILS, contactId);
+      await Promise.all([
+        removeUserFromList(LIST_MARKETING_EMAILS, contactId),
+        removeUserFromList(LIST_DRIP_CAMPAIGN, contactId),
```
capJavert, line 180: Remove when user did not accept marketing campaign [per this comment](https://dailydotdev.slack.com/archives/C03QZBF46JC/p1693838974151479?thread_ts=1692689849.334919&cid=C03QZBF46JC)

src/workers/squadMemberJoinedMailing.ts
```diff
@@ -0,0 +1,49 @@
...
+    try {
+      const { sourceMember } = data;
+
+      const [user, source] = await Promise.all([
+        con.getRepository(User).findOneBy({ id: sourceMember.userId }),
+        con.getRepository(SquadSource).findOneBy({ id: sourceMember.sourceId }),
+      ]);
+
+      // currently we treat admins as owners
+      const isOwner = sourceMember.role === SourceMemberRoles.Admin;
```
capJavert, line 29: It [was agreed](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1693562363131039) we treat admins as owners for now

capJavert: Alternative implementation would be to listen for specific user action `create_squad` on cdc.

rebelchris: Ah i thought for retro filling we could use admins, but for other cases i would use the new squad callback yes.
Sorry bit outdated on what this ticket is for again, ill check in more details tomorrow

rebelchris: Yeah after thinking about it, since we introduce a new subscription let's make it a new topic.
Something like squad created, where we can internally even set a user action for squad_created (might come in handy)
and trigger this new subscription for now.

Think it's a bit more bulletproof.
As for the once-off we probably can simply export and import directly in sendgrid.

idoshamun:
@capJavert is it ready for review?

src/workers/squadMemberJoinedMailing.ts
capJavert: @rebelchris I see all our topics are source related, so maybe `api.v1.source-created` and then we do specific logic for `type=squad`? I am little conflicted on implications of having more different topics so would appreciate feedback on how we do it usually.

capJavert:
@idoshamun it's not ready for review yet. We are solving topics/subscriptions for this https://github.com/dailydotdev/daily-api/pull/1422#discussion_r1315134194

src/workers/squadMemberJoinedMailing.ts
rebelchris: Yeah well could be source created or squad created depends where we put the logic.
Think a cdc that evalutates sources and matches type is lowest on resource.

@idoshamun would you agree?
Also happy to sync on this after my lunch ante?

capJavert: Yeah we can sync and then I can implement in one go.

src/workers/notifications/squadMemberJoined.ts
```diff
@@ -34,12 +34,9 @@ const worker: NotificationWorker = {
       },
     );
 
-    const actionType =
-      member.role === SourceMemberRoles.Admin
-        ? UserActionType.CreateSquad
-        : UserActionType.JoinSquad;
-
-    await insertOrIgnoreAction(con, member.userId, actionType);
```
capJavert, line 42: `UserActionType.CreateSquad` is now handled in `sourceSquadCreated` worker.

src/workers/sourceSquadCreated.ts
```diff
@@ -0,0 +1,65 @@
...
+        .limit(1)
+        .execute();
+
+      if (!owner) {
+        return;
+      }
+
+      const user = await con
+        .getRepository(User)
+        .findOneBy({ id: owner.userId });
```
rebelchris, line 39: Just wondering is this not already retriever on the owner you got above here?

rebelchris (approved):
Nice, looks great to me

src/workers/sourceSquadCreated.ts
```diff
@@ -0,0 +1,65 @@
...
+      const { source } = data;
+      const [owner] = await con
+        .getRepository(SourceMember)
+        .createQueryBuilder('sm')
+        .select('sm."userId" as "userId"')
+        .where('sm."sourceId" = :sourceId', { sourceId: source.id })
+        .andWhere('sm."role" = :role', { role: SourceMemberRoles.Admin })
+        .orderBy('sm."createdAt"', 'ASC')
+        .limit(1)
+        .execute();
```
sshanzel, line 31: Is it not possible to query this through `findOne`?

sshanzel (approved):
Looks great! Just have one non-blocking comment 🚀

src/workers/sourceSquadCreated.ts
sshanzel: Can also do a `const user = await owner.user`.

.infra/workers.ts
```diff
@@ -215,4 +215,8 @@ export const workers: Worker[] = [
     topic: 'analytics-api.v1.experiment-allocated',
     subscription: 'api.experiment-allocated',
   },
+  {
+    topic: 'api.v1.source-squad-created',
+    subscription: 'api.source-squad-created',
```
idoshamun, line 220: My suggestion to make it a bit more explicit

```suggestion
    subscription: 'api.source-squad-created-mailing',
```

src/common/mailing.ts
```diff
@@ -156,6 +156,8 @@ export const getContactIdByEmail = async (email: string) => {
 const LIST_REGISTERED_USERS = '85a1951f-5f0c-459f-bf5e-e5c742986a50';
 // Id of "Weekly recap" contact list on SendGrid
 const LIST_MARKETING_EMAILS = '53d09271-fd3f-4e38-ac21-095bf4f52de6';
+// Id of "Drop campaign" contact list on SendGrid
+export const LIST_DRIP_CAMPAIGN = '919b5624-ccc9-46df-960d-f6a60eb8241b';
```
idoshamun, line 160: A bit more explicit, wdyt?

```suggestion
export const LIST_SQUAD_DRIP_CAMPAIGN = '919b5624-ccc9-46df-960d-f6a60eb8241b';
```

src/workers/cdc.ts
```diff
@@ -488,6 +490,12 @@ const onSourceChange = async (
   logger: FastifyBaseLogger,
   data: ChangeMessage<Source>,
 ) => {
+  if (data.payload.op === 'c') {
+    if (data.payload.after.type === SourceType.Squad) {
+      await notifySourceSquadCreated(logger, data.payload.after);
```
idoshamun, line 495: Shouldn't we just notify on every source created and then filter on the read side?

src/workers/sourceSquadCreated.ts
```diff
@@ -0,0 +1,65 @@
...
+import { insertOrIgnoreAction } from '../schema/actions';
+import { ChangeObject } from '../types';
+import { messageToJson, Worker } from './worker';
+
+interface Data {
+  source: ChangeObject<Source>;
+}
+
+const worker: Worker = {
+  subscription: 'api.source-squad-created',
```
idoshamun, line 17: Since sendgrid api is not the most stable one I'd suggest to separate this to two workers

src/workers/sourceSquadCreated.ts
capJavert: `findOne` does not have order by and we want first admin that joined the squad which is owner.

src/workers/cdc.ts
capJavert: Thought about it, was not sure on the cost of publishing a bunch of messages that we do not use currently. In that case I would just create `api.source-created` topic and then separate subscription that handles only squad stuff. Would that be better approach? Never worked with our topics/subscriptions before so generally interested on what would we prefer.

src/workers/sourceSquadCreated.ts
capJavert: Sure, kinda same thinking from this comment https://github.com/dailydotdev/daily-api/pull/1422#discussion_r1316902768 do we prefer more different subscriptions/different workers?

src/workers/cdc.ts
idoshamun: Since creation of source doesn't happen so often compared to squads I think we should be ok. Your line of thought is fine, cost should be taken into account but it's minor in this case. But I also think about not having to introduce new topics for other use cases in the future. Just like with posts we have a single object and not topic per post type. This is how I see it.

src/workers/sourceSquadCreated.ts
capJavert: I could use `find` though

src/workers/sourceSquadCreated.ts
idoshamun: I usually follow the "do only one thing" approach for workers. We have to think about failures and retries which often happen

src/workers/sourceSquadCreated.ts
capJavert: tnx, added in 4715830bc6968139da3d41afd9bafafea0b614fc

src/workers/sourceSquadCreated.ts
capJavert: good point, added 2ba74f8a27743bb5e654a09fc0bf7350a5fce6ae

src/workers/sourceSquadCreated.ts
capJavert: > do only one thing

👌

src/workers/cdc.ts
rebelchris: ah yes this is my suggestion as we only want to act on squads for now I thought it would be more efficient to only listen to that and act on those.
To save bandwidth.

Might be good to have a chat about this and what we prefer (or document it)

.infra/workers.ts
```diff
@@ -215,4 +215,12 @@ export const workers: Worker[] = [
     topic: 'analytics-api.v1.experiment-allocated',
     subscription: 'api.experiment-allocated',
   },
+  {
+    topic: 'api.v1.source-created',
+    subscription: 'api.source-created-squad',
```
rebelchris, line 220: This should probably be more specific as to what the subscription actually does.
As in assigning the owner data

.infra/workers.ts
```diff
@@ -215,4 +215,12 @@ export const workers: Worker[] = [
...
+  },
+  {
+    topic: 'api.v1.source-created',
+    subscription: 'api.source-created-squad-mailing',
```
rebelchris, line 224: Perhaps also squad-owner-mailing to be more explicit as to what this subscription handles?

src/workers/sourceSquadCreated.ts
```diff
@@ -0,0 +1,37 @@
...
+import { messageToJson, Worker } from './worker';
+
+interface Data {
+  source: ChangeObject<Source>;
+}
+
+const worker: Worker = {
+  subscription: 'api.source-created-squad',
+  handler: async (message, con) => {
+    const data: Data = messageToJson(message);
```
rebelchris, line 14: We need a exit here if it's not a squad source

src/workers/sourceSquadCreatedMailing.ts
```diff
@@ -0,0 +1,51 @@
...
+import { messageToJson, Worker } from './worker';
+
+interface Data {
+  source: ChangeObject<Source>;
+}
+
+const worker: Worker = {
+  subscription: 'api.source-created-squad-mailing',
+  handler: async (message, con) => {
+    const data: Data = messageToJson(message);
```
rebelchris, line 18: Same here as for early exit

rebelchris (changes requested):
Nice work on splitting it up, some minor naming comments

.infra/workers.ts
capJavert: Agreed

src/workers/cdc.ts
```diff
@@ -488,6 +489,10 @@ const onSourceChange = async (
   logger: FastifyBaseLogger,
   data: ChangeMessage<Source>,
 ) => {
+  if (data.payload.op === 'c') {
+    await notifySourceCreated(logger, data.payload.after);
+  }
+
   if (data.payload.op === 'u') {
```
idoshamun, line 496: else if I guess. Not that it really matters

idoshamun (approved):
Awesome!

rebelchris (approved):
Nice one

PR #1423, opened 2023-09-05
fix: remove sub query to improve performance

capJavert:
Remove sub query to improve performance. Can introduce a case where some users will not get questions.


idoshamun (approved):

rebelchris (commented):
Curious to see if we have quicker alternatives for future implementations

capJavert:
> Curious to see if we have quicker alternatives for future implementations

Join instead of sub query might be better I think, but we wanted to remove it to see what happens. Also just caching this in redis as @idoshamun mentioned would be alternative for optimization.

PR #1425, opened 2023-09-05
fix: cache search rec

rebelchris:
Added a redis 3 minute cache for search rec.
Note: It could introduce edge-cases as we json stringify the graphORM data.

Basically means if we ever change the format of selection in the query we should invalidate/renew a key version.

WT-1816 #done


idoshamun (approved):
Awesome!

capJavert (approved):
Nice, we should also see when to reintroduce sub query (or optimize) from this https://github.com/dailydotdev/daily-api/pull/1423

PR #1430, opened 2023-09-09
refactor: cleanup flagsmith usage and set default values

sshanzel:
Cleanup of Flagsmith for the API side.

- Hard-coded values for the following flags:
  - Onboarding v2: always at `v1` then the existing logic to update it as necessary is still in place.
  - Feed version: value is now always `11`.
  - Submit article: we have an existing mechanism that updates the flags to include it when the user is eligible.
    - We will keep the same process to support outdated extension users.
    - We will introduce an additional property to tag whether the user is eligible (though we already have a query that returns the same value, the `query submissionAvailability` - maybe we should remove it from there).
- We should probably run a local script to clear off our cache from flagsmith values. Thoughts?


.infra/workers.ts
```diff
@@ -199,10 +199,6 @@ export const workers: Worker[] = [
     topic: 'api.v1.banner-deleted',
     subscription: 'api.banner-deleted',
   },
-  {
-    topic: 'features-reset',
```
idoshamun, line 203: We can delete this topic from `streams` once it's merged

idoshamun (approved):
Looks great!! We can either let the cache expire or delete manually. Not sure we need a script for that

sshanzel:
Good point! Forgot they actually expire on their own. In that case, we won't have to run anything.

rebelchris (approved):
Cleanup! 🧼 

Looks pretty complete, think you also have the overwrites ok to work backwords compatible 🙏

sshanzel:
> Cleanup! 🧼
> 
> Looks pretty complete, think you also have the overwrites ok to work backwords compatible 🙏

Yep, I will merge them right after one another. We might have some fluctuations with the distribution since extension users will always be on default.

PR #1433, opened 2023-09-11
fix: get the source privacy even when source_id is not sent

omBratteng:


omBratteng:
Is this line still relevant? Shouldn't the sourceID be the ID of the squad source?
https://github.com/dailydotdev/daily-api/blob/72fa8c4b74f433497b7a277776eb702125bd3261/src/workers/postUpdated.ts#L194-L196

src/workers/postUpdated.ts
```diff
@@ -274,6 +275,41 @@ const updatePost = async ({
...
+const getSourcePrivacy = async ({
+  logger,
+  entityManager,
+  data,
+}: GetSourcePrivacyProps): Promise<boolean> => {
+  try {
+    let query = entityManager
+      .getRepository(Source)
+      .createQueryBuilder('source')
+      .select(['source.private']);
```
rebelchris, line 292: Just `private` would do no?

rebelchris (approved):
Just also wait for @idoshamun feedback as well on this

rebelchris:
@omBratteng Let me explain the use-case for `unknown` sources.
Users can share items in squads, when we decided to allow for "any" url to be shared we needed a uniform way of storing these URLs but ensuring they could be accessed by multiple squads over time.

This is what you see as `unknown` sourced content, it's basically URLs we didn't have in the platform yet, but can belong to multiple squads at once.
They will all be used as `sharedPostId` in some other freeFormContent type post.

Hope this makes sense?
Let me know if you like me to show you instead, it's good for you to know how it works and links together.

src/workers/postUpdated.ts
omBratteng: Without `source.` the tests would fail

omBratteng:
That makes senes

idoshamun (approved):
Looks good to me!

src/workers/postUpdated.ts
rebelchris: Oh I think it's because you define `source` in `createQueryBuilder` you can omit that as you already defined the repo.

omBratteng: I think it would potentially crash with the join as well. I'll double check

omBratteng: Couldn't get it to work without it. I'll just leave it there

PR #1434, opened 2023-09-12
feat: provide user attributes as part of boot

idoshamun:
To reduce growthbook payload size we have to use user attributes. This will allow us to move from defining hard coded user ids on GB to having them on our database. Hence, it will reduce significantly the payload.

NOTE: Requires apps changes but not blocking this PR

### DIFF
--- a/__tests__/boot.ts
+++ b/__tests__/boot.ts
@@ -14,6 +14,8 @@ import {
   ALERTS_DEFAULT,
   ArticlePost,
   Banner,
+  Feature,
+  FeatureType,
   MachineSource,
   Notification,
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
So it will become a mix of the two?
Some cases we use GB directly and somecases we internally set to feature value?

I'm a bit concerned about scalability/clarity down the line, but we can discuss it seperatly.

idoshamun:
Everything goes through GB. But these attributes will be used to identify beta testers. By themselves, the attributes will not do a thing unless you set it properly on GB

PR #1435, opened 2023-09-12
fix: test case and update type based on content type

rebelchris:
We needed to distinguish between content_type => post_type mapping.
Basically the old flow was only working for ArticlePost.

Also modified test cases to ensure non existing fields don't throw errors.

Note:
Currently only works for ArticlePost and FreeformPost (expected, but good to note)

### DIFF
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -193,7 +193,7 @@ const updatePost = async ({
   data,
   id,
   mergedKeywords,
-  content_type,
+  content_type = PostType.Article,
 }: UpdatePostProps) => {
   const postType = contentTypeFromPostType[content_type];
   const databasePost = await entityManager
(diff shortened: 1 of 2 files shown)


idoshamun (approved):

capJavert (approved):
Nice, catch.

src/workers/postUpdated.ts
```diff
@@ -166,13 +168,19 @@ const allowedFieldsMapping = {
...
   ],
 };
 
+const contentTypeFromPostType: Record<PostType, typeof Post> = {
+  [PostType.Article]: ArticlePost,
+  [PostType.Freeform]: FreeformPost,
+  [PostType.Share]: SharePost,
+  [PostType.Welcome]: WelcomePost,
+};
+
```
capJavert, line 183: @rebelchris we might wanna extract it to some common file immediately for reuse?

rebelchris: Probably a good one for when we need it, not sure where else we evaluate based on types. 
It's something Yggdrasil uses, and not particularly the API at this point.

Probably another thing Protobuf would fix

capJavert: Yeah, or any kind of cross service standardisation 👍

PR #1443, opened 2023-09-22
fix: only rerun failed tests

rebelchris:
Trying the circleCI rerun option


idoshamun (approved):
@rebelchris just make sure check that it runs all tests

capJavert (commented):
We should probably force a test fail to see how partial running works?

PR #1447, opened 2023-09-26
feat: personalized digest

capJavert:
Implementing personalized digest, [DR](https://dailydotdev.atlassian.net/l/cp/VR2oa90g)

Frontend PR: https://github.com/dailydotdev/apps/pull/2250

Deployment plan:
- merge this PR
- enroll all team members manually through SQL query (members are added as `feature=team` in `feature` table
- test out sending, processing, analyse worker load etc.
- enroll 10k initial users per query below (save them to CSV or temp table https://github.com/dailydotdev/daily-api/pull/1447#discussion_r1344132165)
- analyse and enroll the rest

Issues:
- unsubscribes from sendgrid will currently not persist in our db, see the section I added in DR https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/595918850/WT-1622+Weekly+Digest#Tracking-unsubscribes
- we might have to adjust our logic to avoid errors when we introduce user selectable preference, see: https://github.com/dailydotdev/daily-api/pull/1447#discussion_r1341477599

TODO:
- [x] Implement script to enroll users, will be needed once we want to import more (SQL query to pick initial users below)
```sql
INSERT INTO user_personalized_digest ("userId", "preferredTimezone")
SELECT id AS "userId", COALESCE(timezone, 'Etc/UTC') AS "preferredTimezone" FROM public.user 
WHERE "infoConfirmed" IS TRUE AND "acceptedMarketing" IS TRUE -- IS FALSE for other group
ORDER BY RANDOM() limit 5000
ON CONFLICT DO NOTHING;
```
- [x] Add category to sendgrid email call to allow specific unsubscribe option (minor)
- [x] Integrate real posts fetching through feed service (currently mocked)


.infra/crons.ts
```diff
@@ -43,5 +43,9 @@ export const crons: Cron[] = [
   {
     name: 'refresh-user-subscriptions',
     schedule: '0 */3 * * *',
+  },
+  {
+    name: 'personalized-digest',
+    schedule: '0 0 * * 0'
```
capJavert, line 49: Currently running once a week on Sunday to generate Monday digest. Additional logic shall be added once we integrate user preference for which day they want the digest to send.

src/cron/personalizedDigest.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    await new Promise((resolve, reject) => {
+      personalizedDigestStream.on('error', (error) => {
+        logger.error(
+          { err: error },
+          'streaming personalized digest subscriptions failed',
+        );
+
+        reject(error);
+      });
```
capJavert, line 29: If stream fails for any reason would cron retry? Just wanna make sure that we don't get the flow:
- we start to schedule digest generation
- it fails mid generation (but some digest are already scheduled)
- it retries and schedules more

src/workers/generatePersonalizedDigestForUser.ts
```diff
@@ -0,0 +1,114 @@
...
+  subscription: 'api.generate-personalized-digest-for-user',
+  handler: async (message, con) => {
+    const data = messageToJson<Data>(message);
+    const { personalizedDigest } = data;
+
+    const user = await con.getRepository(User).findOneBy({
+      id: personalizedDigest.userId,
+    });
+
+    if (!user || !user.infoConfirmed) {
```
capJavert, line 68: We delete users who don't confirm in an hour so this is to make sure we don't send emails to those until they confirm.

src/workers/generatePersonalizedDigestForUser.ts
```diff
@@ -0,0 +1,114 @@
...
+      return;
+    }
+
+    const posts = await getMockedPosts({ con });
+
+    const [dayName] = Object.entries(DayOfWeek).find(
+      ([, value]) => value === personalizedDigest.preferredDay,
+    );
+    const userName = user.name?.trim().split(' ')[0] || user.username;
```
capJavert, line 77: Seen this is how firstname is extracted on other places. Not too happy with it becauase some people do not have single name/lastname etc... Should we just use `username`?

capJavert:
Not finished yet but opening for review since most of the stuff is in place.

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (!user || !user.infoConfirmed) {
+      return;
+    }
+
+    const posts = await getMockedPosts({ con });
+
+    const [dayName] = Object.entries(DayOfWeek).find(
+      ([, value]) => value === personalizedDigest.preferredDay,
+    );
+    const userName = user.name?.trim().split(' ')[0] || user.username;
```
capJavert, line 77: Seen this is how firstname is extracted on other places. Not too happy with it becauase some people do not have single name/lastname etc... Should we just use username?

rebelchris: Lets leave it as is, it will impact how sendgrid stores them as well.
Its not ideal but seems to work for now 🙏
Thanks for raising it though

src/migration/1695735320395-UserPersonalizedDigest.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class UserPersonalizedDigest1695735320395 implements MigrationInterface {
+    name = 'UserPersonalizedDigest1695735320395'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
```
rebelchris, line 6: We might want to consider a index here, future proofing if we wants users on a day perhaps?

rebelchris: I also think you'll need `await queryRunner.query(`ALTER TABLE "user_personalized_digest" REPLICA IDENTITY FULL`);` on this one, as it will be used in CDC correct?

src/schema/users.ts
```diff
@@ -967,6 +1025,55 @@ export const resolvers: IResolvers<any, Context> = {
...
+      args,
+      ctx: Context,
+    ): Promise<unknown> => {
+      const repo = ctx.con.getRepository(UserPersonalizedDigest);
+
+      if (ctx.userId) {
+        await repo.delete({
+          userId: ctx.userId,
+        });
+      }
```
rebelchris, line 1073: Not really comment about your code.
But let's say we roll out now to these 10K users.

And 100 will unsubscribe -> deleting this entry.
Then we decide to enroll everyone by running the script.
These will would be added again correct?

How can we ensure that doesn't happen?

src/schema/users.ts
```diff
@@ -967,6 +1025,55 @@ export const resolvers: IResolvers<any, Context> = {
         )
         .andWhere('"userId" = :userId', { userId: ctx.userId })
         .execute(),
+    subscribePersonalizedDigest: async (
```
rebelchris, line 1028: Who calls this?
Just wondering if the hour/day should be defaulted and optional

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,114 @@
...
+    const emailSendDate = getEmailSendDate({ personalizedDigest });
+    const emailPayload: MailDataRequired = {
+      to: {
+        email: user.email,
+        name: user.name,
+      },
+      from: {
+        email: 'informer@daily.dev',
+        name: 'Weekly digest',
+      },
```
rebelchris, line 87: Maybe we should use `baseNotificationEmailData` and overwrite the from alone?

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,114 @@
...
+      sendAt: Math.floor(emailSendDate.getTime() / 1000),
```
rebelchris, line 88: You tested this is the expected UTC from their side?
(and it arrives in your timezone as expected?)

rebelchris (commented):
Pretty clean implementation so far 👏 

Some minor questions, mostly validations really.

src/migration/1695735320395-UserPersonalizedDigest.ts
capJavert: Yeah, users per day we already need, just did not add an index. Will do. 👌

src/schema/users.ts
capJavert: Yes, good catch. Since experiment was not meant or planned we can't in current db schema. What we could do is if we decide to enroll everyone check the db before we do it and do not re-add any users who unsubscribed deleted their entry.

Alternative would be to add status `subscribed | unsubscribed` column.

src/migration/1695735320395-UserPersonalizedDigest.ts
capJavert: REPLICA is already set in the line below.

capJavert: Added in 3a6c0f0594e73d55fc42d4a96b7d25f518ccb91f

src/schema/users.ts
capJavert: Frontend from settings, I'll update once I open that PR.

src/schema/users.ts
rebelchris: Yeah let's opt for the "simplist" one.
Maybe it's enough to simply keep track of these 10K people seperate? (as in CSV) and manually exclude on next step?

capJavert: 🤝

src/cron/personalizedDigest.ts
capJavert: Use `redis` or api db storage to store data on which digests were sent and when. As a mean for deduplication.

sshanzel: That makes sense, though not sure if our cron jobs retry when failing. As as I know they don't.

src/schema/users.ts
capJavert: All are optional and filled up with default values in db if not sent. f212961d1e4a2466bc3df937812741646a985c0d

sshanzel (commented):
Looks great for the initial review! 🚀

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,165 @@
...
+  );
+
+  return templateData;
+};
+
+const worker: Worker = {
+  subscription: 'api.personalized-digest-email',
+  handler: async (message, con, logger) => {
+    if (process.env.NODE_ENV === 'development') {
+      return;
```
capJavert, line 88: Not sure how to mock this locally, `http://localhost:6000/feed.json` used for feeds is not available. Any recommendations?

src/schema/users.ts
capJavert: So my plan is:
- enroll first 10k at random with query in description
- pull them out into CSV file
- when we decide we want to import more users we are gonna manually see which of them unsubscribe to not enroll again

__tests__/users.ts
```diff
@@ -1973,3 +1975,270 @@ describe('query referralCampaign', () => {
...
+
+  it('should require authentication', async () => {
+    await testQueryErrorCode(
+      client,
+      { query: QUERY, variables: {} },
+      'UNAUTHENTICATED',
+    );
+  });
+
+  it('should throw not found exception when user is not subscribed', async () => {
```
idoshamun, line 1997: I think it's a bit weird to throw an exception in this case, isn't it? I'd expect a response that hints that the user is not subscribed. But it's very minor

bin/subscribeUsersToPersonalizedDigest.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (createdFromDate > createdToDate) {
+    throw new Error(
+      'createdFrom argument should be less than createdTo argument',
+    );
+  }
+
+  const con = await createOrGetConnection();
+
+  await con.transaction(async (manager) => {
+    await manager.query(`
```
idoshamun, line 37: Since there's no CDC on this table we can just run the sql query without batching, can't we?

src/cron/personalizedDigest.ts
idoshamun: Cron job do have a retry mechanism. They actually also have a timeout mechanism so we need to make sure it doesn't take too long

src/workers/personalizedDigestEmail.ts
idoshamun: You disabled it for the local env so it will not run?

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    const data = messageToJson<Data>(message);
+    const { personalizedDigest } = data;
+
+    const user = await con.getRepository(User).findOneBy({
+      id: personalizedDigest.userId,
+    });
+
+    if (!user || !user.infoConfirmed) {
```
idoshamun, line 98: Should be enough :P

```suggestion
    if (!user?.infoConfirmed) {
```

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,165 @@
...
+        total_posts: personalizedDigestPostsCount,
+        date_from: format(previousSendDate, personalizedDigestDateFormat),
+        date_to: format(currentDate, personalizedDigestDateFormat),
+        allowed_tags: feedConfig.includeTags,
+        blocked_tags: feedConfig.blockedTags,
+        blocked_sources: feedConfig.excludeSources,
+      },
+    );
+
+    const posts = await con.getRepository(ArticlePost).find({
```
idoshamun, line 126: We need to order it by the same order we receive from the feed

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,165 @@
...
+
+const getPostsTemplateData = async ({ posts }: { posts: ArticlePost[] }) => {
+  const templateData = await Promise.all(
+    posts.map(async (post) => {
+      const source = await post.source;
+
+      const postUrl = new URL(getDiscussionLink(post.id));
+      postUrl.searchParams.append('utm_source', 'notification');
+      postUrl.searchParams.append('utm_medium', 'email');
+      postUrl.searchParams.append('utm_campaign', 'digest');
```
idoshamun, line 69: Don't we have a function already for that in the notifications area?

src/workers/userCreatedAddPersonalizedDigest.ts
```diff
@@ -0,0 +1,22 @@
+import { User } from '../common';
+import { UserPersonalizedDigest } from '../entity/UserPersonalizedDigest';
+import { messageToJson, Worker } from './worker';
+
+interface Data {
+  user: User;
+}
+
+const worker: Worker = {
```
idoshamun, line 9: Can be substituted with a trigger, can't it?

capJavert: Conditions on which we enroll might change based on multiple discussion so I think having a worker gives a bigger flexibility. If logic in triggers can get ugly.

src/workers/personalizedDigestEmail.ts
capJavert: Good point.

src/integrations/feed/clients.ts
```diff
@@ -158,3 +158,40 @@ export class CachedFeedClient implements IFeedClient {
     return feedRes.slice(offset, pageSize + offset);
   }
 }
+
+/**
+ * Client that fetches posts for personalized digest from feed personalise service
+ */
+export class PersonalizedDigestFeedClient implements IFeedClient {
```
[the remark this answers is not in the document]
capJavert: No, the response is different so the map before return needs different handling.

bin/subscribeUsersToPersonalizedDigest.ts
capJavert: Yeah, waited to see what you will say, ~300k rows in one query should be ok in one go. Although down the line script gives us more flexibility I think.

__tests__/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,169 @@
+import { expectSuccessfulBackground, saveFixtures } from '../helpers';
```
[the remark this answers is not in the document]
capJavert: I know, but I would have to hardcode bunch of values and computations in there which would again be sensitive with all the data changes. So snapshot looked like the easiest solution.

.infra/crons.ts
```diff
@@ -43,5 +43,9 @@ export const crons: Cron[] = [
   {
     name: 'refresh-user-subscriptions',
     schedule: '0 */3 * * *',
+  },
+  {
+    name: 'personalized-digest',
+    schedule: '0 0 * * *'
```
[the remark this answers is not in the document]
capJavert: Good point!

__tests__/workers/personalizedDigestEmail.ts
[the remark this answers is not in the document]
capJavert: Because I wanted to make sure all the properties are there and correct when sending an email, if some of them are wrong we can potentially have thousands of invalid emails.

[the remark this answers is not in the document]
capJavert: If we wanna avoid snapshots I would just copy their current values in raw in code and use `toMatchObject`

__tests__/users.ts
capJavert: Came to my mind, I'll try to see if there is a better option. 👍

src/cron/personalizedDigest.ts
capJavert: What is the timeout?

idoshamun: Right now it's 5min but I think you can adjust it in pulumi

__tests__/workers/personalizedDigestEmail.ts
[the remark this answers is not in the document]
idoshamun: Let's leave it as it is for now then

src/cron/personalizedDigest.ts
capJavert: I think 5 minutes should be plenty, but will monitor with initial deployment.

__tests__/users.ts
capJavert: Frontend implementation is here https://github.com/dailydotdev/apps/pull/2250, since it is a separate entity I think 404 makes sense. What I did on the frontend is to handle that error case in our dedicated hook. Let me know what you think.

rebelchris (approved):
Nothing smart from my side to add. Looks solid.

src/cron/personalizedDigest.ts
sshanzel: Oh, nice. Good to know it was brought up, now I know it retries.

src/common/feedGenerator.ts
```diff
@@ -502,7 +502,7 @@ export const tagFeedBuilder = (
   builder.andWhere((subBuilder) => whereTags([tag], subBuilder, alias));
 
 export const fixedIdsFeedBuilder = (
-  ctx: Context,
+  ctx: unknown,
```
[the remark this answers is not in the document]
capJavert: I needed to use it inside worker which does not have the `ctx` available. The `ctx` is not really used in that specific builder but since all builders have that kind of arguments footprint I did not wish to remove it and opted to make parameter `unknown`.

This allows me to just pass `{}` inside the worker call https://github.com/dailydotdev/daily-api/blob/6d963b9a7e6a95d38bef1f25c1514013c3f18542/src/workers/personalizedDigestEmail.ts#L130-L131 and if for ever reason `ctx` becomes used inside `fixedIdsFeedBuilder` typescript will detect worker usage and we will be able to supply or `Pick` specific part of `Context`.

src/workers/personalizedDigestEmail.ts
capJavert: Confirmed it works.

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,178 @@
...
+        first_name: userName,
+        posts: await getPostsTemplateData({ posts }),
+      },
+      asm: {
+        groupId: 23809,
+      },
+      category: 'Digests',
+    };
+
+    await sendEmail(emailPayload);
```
capJavert, line 174: @idoshamun should we wrap this in try catch, I am worried if sendgrid fails this would retry, in development it retries forever, not sure how many retries it does in production?

idoshamun: I don't think we have a limit on number of retries currently, would you like to check?

idoshamun: If it's not possible, I suggest to try and catch. It's better not to ddos sendgrid much more than we will 😂

idoshamun (approved):
Happy with the implementation! Please sort out the retry for sendgrid

src/workers/personalizedDigestEmail.ts
capJavert: Haha ok, for sure we don't want infinite loops/retries again this week 😆 I'll take a look

capJavert: @idoshamun I think for solving retries we need to setup dead lettering https://cloud.google.com/pubsub/docs/handling-failures#dead_letter_topic
In that topic we can for now just log the message payload and exit. Dead letter topic can be enabled per subscription from what I can see so it does not need to be system wide change. Could we setup something like this?

As for solving stuff for this PR I added try catch around the whole resolver so we make sure if anything goes wrong we don't have thousands of messages retrying forever.

idoshamun: You can find here an example of a dead letter provisioning with pulumi:
https://github.com/dailydotdev/analytics-processor/blob/main/.infra/index.ts#L45

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,190 @@
...
+
+const emailTemplateId = 'd-328d1104d2e04fa1ab91e410e02751cb';
+
+const personalizedDigestDateFormat = 'yyyy-MM-dd HH:mm:ss';
+
+const getEmailSendDate = ({
+  personalizedDigest,
+}: {
+  personalizedDigest: UserPersonalizedDigest;
+}): Date => {
```
sshanzel, line 41: I think we can use the same interface or at least base from it.

```suggestion
}: Data): Date => {
```

src/workers/personalizedDigestEmail.ts
```diff
@@ -0,0 +1,190 @@
...
+  );
+
+  return sendDateInPreferredTimezone;
+};
+
+const getPreviousSendDate = ({
+  personalizedDigest,
+}: {
+  personalizedDigest: UserPersonalizedDigest;
+}): Date => {
```
sshanzel, line 56: Same goes here.

```suggestion
}: Data): Date => {
```

sshanzel (approved):
Looks amazing! Very well implemented 🚀

src/workers/personalizedDigestEmail.ts
capJavert: Merging this now and will look into dead letter topic after.

PR #1448, opened 2023-09-27
fix: added feature value option

rebelchris:
Added option to allow values for features.
This way we can negative value people who shouldn't have access.

@idoshamun Let me know if this will work.
According to my plan we'll have to introduce a new evaluation saying if search = 1 show v1, if search = -1 show control.
But keep the `true` evaluation as well.

### DIFF
--- a/src/entity/Feature.ts
+++ b/src/entity/Feature.ts
@@ -7,8 +7,8 @@ export enum FeatureType {
 }
 
 export enum FeatureValue {
-  Up = 1,
-  Down = -1,
+  Allow = 1,
+  Block = -1,
 }
 
 @Entity()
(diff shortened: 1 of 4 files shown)


src/entity/Feature.ts
```diff
@@ -6,6 +6,11 @@ export enum FeatureType {
   Search = 'search',
 }
 
+export enum FeatureValue {
+  Up = 1,
+  Down = -1,
```
idoshamun, line 11: Better names I guess
```suggestion
  Allow = 1,
  Block = -1,
```

idoshamun (approved):
Looks good just a suggestion for renaming

PR #1449, opened 2023-09-27
fix: add notification deletion

rebelchris:
Added workers on existing topics to handle deletion of posts and comment to remove notifications.
Decided to split them in case we eventually wanted to add more logic.

WT-1805 #done

### DIFF
--- a/src/workers/notifications/commentDeleted.ts
+++ b/src/workers/notifications/commentDeleted.ts
@@ -1,0 +1,45 @@
+import { messageToJson, Worker } from '../worker';
+import { Notification, Comment } from '../../entity';
+import { ChangeObject } from '../../types';
+
+interface Data {
+  comment: ChangeObject<Comment>;
+}
+
+const worker: Worker = {
+  subscription: 'api.comment-deleted-notification-cleanup',
+  handler: async (message, con, logger) => {
+    const data: Data = messageToJson(message);
+    const { comment } = data;
+
+    try {
+      await con
+        .getRepository(Notification)
+        .createQueryBuilder()
+        .delete()
+        .where({
+          referenceType: 'comment',
+          referenceId: comment?.id,
+        })
+        .execute();
+      logger.info(
+        {
+          data,
+          messageId: message.messageId,
+        },
+        'deleted notifications due to comment deletion',
+      );
+    } catch (err) {
+      logger.error(
+        {
+          data,
+          messageId: message.messageId,
+          err,
+        },
+        'failed to delete notifications due to comment deletion',
+      );
+    }
+  },
+};
+
+export default worker;
--- a/src/workers/notifications/postDeleted.ts
+++ b/src/workers/notifications/postDeleted.ts
@@ -1,0 +1,45 @@
+import { messageToJson, Worker } from '../worker';
+import { Notification, Post } from '../../entity';
+import { ChangeObject } from '../../types';
+
+interface Data {
+  post: ChangeObject<Post>;
+}
+
+const worker: Worker = {
+  subscription: 'api.post-deleted-notification-cleanup',
+  handler: async (message, con, logger) => {
+    const data: Data = messageToJson(message);
+    const { post } = data;
+
+    try {
+      await con
+        .getRepository(Notification)
+        .createQueryBuilder()
+        .delete()
+        .where({
+          referenceType: 'post',
+          referenceId: post?.id,
+        })
+        .execute();
+      logger.info(
+        {
+          data,
+          messageId: message.messageId,
+        },
+        'deleted notifications due to post deletion',
+      );
+    } catch (err) {
+      logger.error(
+        {
+          data,
+          messageId: message.messageId,
+          err,
+        },
+        'failed to delete notifications due to post deletion',
+      );
+    }
+  },
+};
+
+export default worker;
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -124,6 +124,14 @@ export const workers: Worker[] = [
   },
   // Notifications
   {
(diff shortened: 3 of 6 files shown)


src/workers/notifications/commentDeleted.ts
```diff
@@ -0,0 +1,45 @@
...
+  comment: ChangeObject<Comment>;
+}
+
+const worker: Worker = {
+  subscription: 'api.comment-deleted-notification-cleanup',
+  handler: async (message, con, logger) => {
+    const data: Data = messageToJson(message);
+    const { comment } = data;
+
+    try {
```
idoshamun, line 15: We can remove the try and catch in this case to make sure it gets retried on failure

src/workers/notifications/commentDeleted.ts
```diff
@@ -0,0 +1,45 @@
...
+      await con
+        .getRepository(Notification)
+        .createQueryBuilder()
+        .delete()
+        .where({
+          referenceType: 'comment',
+          referenceId: comment?.id,
```
idoshamun, line 22: This is risky. I suggest validating that id exists before, otherwise we may risk deleting all comment notifications

src/workers/notifications/postDeleted.ts
```diff
@@ -0,0 +1,45 @@
...
+  post: ChangeObject<Post>;
+}
+
+const worker: Worker = {
+  subscription: 'api.post-deleted-notification-cleanup',
+  handler: async (message, con, logger) => {
+    const data: Data = messageToJson(message);
+    const { post } = data;
+
+    try {
```
idoshamun, line 15: Same here

src/workers/notifications/postDeleted.ts
```diff
@@ -0,0 +1,45 @@
...
+      await con
+        .getRepository(Notification)
+        .createQueryBuilder()
+        .delete()
+        .where({
+          referenceType: 'post',
+          referenceId: post?.id,
```
idoshamun, line 22: And here

src/workers/notifications/postDeleted.ts
```diff
@@ -0,0 +1,45 @@
+import { messageToJson, Worker } from '../worker';
+import { Notification, Post } from '../../entity';
+import { ChangeObject } from '../../types';
+
+interface Data {
+  post: ChangeObject<Post>;
+}
+
+const worker: Worker = {
+  subscription: 'api.post-deleted-notification-cleanup',
```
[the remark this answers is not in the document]
rebelchris: Can i suggest to simply evaluate inside the subscriber? 
Feels like overkill to add a separate topic.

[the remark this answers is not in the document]
idoshamun: Ah yes of course. Just didn't know where to put this comment 🙈

rebelchris:
@idoshamun Refactored to check if items exist and for post if it's deleted

src/workers/notifications/commentDeleted.ts
```diff
@@ -0,0 +1,47 @@
...
+      .findOneBy({ id: comment?.id });
+    if (!databaseComment) {
+      return logger.error(
+        {
+          data,
+          messageId: message.messageId,
+        },
+        'failed to delete notifications due to comment deletion',
+      );
+    }
```
idoshamun, line 26: This is not what I meant because at this point the comment is already not in the database as we're dealing with deleted comments. We only need an if statement to check if comment?.id is defined

src/workers/notifications/postDeleted.ts
```diff
@@ -0,0 +1,47 @@
...
+  handler: async (message, con, logger) => {
+    const data: Data = messageToJson(message);
+    const { post } = data;
+
+    const databasePost = await con
+      .getRepository(Post)
+      .findOneBy({ id: post?.id });
+    if (!databasePost || !databasePost?.deleted) {
+      return logger.error(
```
idoshamun, line 19: Let's change the logger level to something else, otherwise we will get false alerts

idoshamun (approved):

src/workers/notifications/postDeleted.ts
```diff
@@ -0,0 +1,47 @@
...
+        'failed to delete notifications due to post deletion',
+      );
+    }
+
+    await con
+      .getRepository(Notification)
+      .createQueryBuilder()
+      .delete()
+      .where({
+        referenceType: 'post',
```
sshanzel, line 33: What about comments from the post such as replies and mentions. Should we delete those as well when the post gets deleted?

rebelchris: I think that's already handled by the system afaik.

sshanzel: My bad, I was not clear. I mean the notifications from those comments.

sshanzel: Oh, wait. You are right. There will be a full cycle. Post deleted -> cascades to delete comments -> worker triggered -> delete notification comments. Great work!

rebelchris: Yes should cascade the comment deletion right? 🙏
Thanks for double checking though

idoshamun: Deleting posts is just a logical deletion, no cascade will happen in this case. We need to delete the relevant comment notifications manually. That's a great catch Lee! We need a follow-up PR I guess

sshanzel: Oh, right. I forgot we don't actually hard-delete post data 😅

idoshamun: Last time we talked to product they asked to keep the comments on deletion but we can double check

PR #1450, opened 2023-09-28
fix: set feed version 15 as default

rebelchris:
Set default from 11 to 15


omBratteng (approved):

idoshamun (approved):
We also need to change on the apps repo I think :)

rebelchris:
@idoshamun Apps is actually set default to 1 😅
I'll adjust just to be aligned, although I think we do this on purpose for the local test (that run on feed v1)

idoshamun:
Ok nvm then

PR #1452, opened 2023-10-03
fix: add comments deletion

omBratteng:
Create new worker to delete comments on post deletion.

WT-1874


idoshamun (approved):

rebelchris (approved):
One request, can you add a notification for comment or add it to the notification specific test?
We need to ensure those get cleaned as well.

omBratteng:
Added notification and test if it is removed

PR #1454, opened 2023-10-04
infra: ignore changes to views and discussion score

idoshamun:
We have cron jobs that update frequently the views and discussion score of posts. These changes generate quite a lot of cdc messages but we ignore them on our cdc worker. This change ignores them on debezium level to reduce the overall load on the system.

Please extra check my conditions!


capJavert (commented):
Edge case that might happen is that if any other post field is updated in same transaction/query as discussionStore or views then those messeges would also be filtered out. So we need to watch out that both fields included in "or" statement are always updated separately of other fields.

idoshamun:
@capJavert that's a solid point!
Validated that it happens here:
* https://github.com/dailydotdev/daily-api/blob/main/src/cron/updateViews.ts
* https://github.com/dailydotdev/daily-api/blob/main/src/cron/updateDiscussionScore.ts

capJavert (approved):

PR #1456, opened 2023-10-04
refactor: replace p-retry with custom implementation based on retry

idoshamun:
Since we're using an old p-retry version and it's not possible to upgrade at this point we decided to roll our own based on the retry package (same as p-retry)


src/integrations/retry.ts
```diff
@@ -0,0 +1,84 @@
+import retry, { OperationOptions } from 'retry';
```
rebelchris, line 1: I see you based it on a earlier version, did we not want to stick to the latest one?

src/integrations/retry.ts
```diff
@@ -0,0 +1,84 @@
...
+
+          if (!operation.retry(err)) {
+            throw operation.mainError();
+          }
+        } catch (finalError) {
+          reject(finalError);
+        }
+      }
+    });
+  });
```
capJavert, line 66: Should we do some tests for this?

src/integrations/retry.ts
idoshamun: This is the latest one actually: https://github.com/tim-kos/node-retry/blob/master/package.json#L7

rebelchris: Ah I compared to this? https://github.com/sindresorhus/p-retry/blob/main/index.js

src/integrations/retry.ts
idoshamun: Good point! Added

src/integrations/retry.ts
idoshamun: Yes that's what I thought :)
So we're all clear I guess

rebelchris (approved):

PR #1457, opened 2023-10-04
fix: add index optimization

rebelchris:
We had another missed index on the source images.

Screenshot 2023-10-04 at 13 48 11


capJavert (commented):
We should add it to typeorm entity and then from that auto generate migration?

rebelchris:
@capJavert good point, added :)

idoshamun:
@rebelchris where do we use this query? It looks pretty weird to me

rebelchris:
@idoshamun It's the devcard query for sources
Screenshot 2023-10-04 at 14 48 40

idoshamun (approved):
Thanks a lot for showing me the query!
We'll need to optimize devcard in the future

PR #1458, opened 2023-10-04
refactor: send whoami calls only when jwt is about to expire or not available

idoshamun:
To reduce the load on heimdall side, we'll call the whoami only when we actually must. This should also optimize subsequent calls to boot in the same session.

### DIFF
--- a/.env
+++ b/.env
@@ -39,3 +39,6 @@ ANALYTICS_URL=http://localhost:5000
 
 EXPERIMENTATION_KEY=6FA4C8D3DD2400B9D614BB7A229FD69B
 #GROWTHBOOK_CLIENT_KEY=override_me_in_env.development
+
+JWT_PUBLIC_KEY_PATH=../apps/.cert/public.pem
+JWT_PRIVATE_KEY_PATH=../apps/.cert/key.pem
--- a/__tests__/boot.ts
+++ b/__tests__/boot.ts
@@ -51,6 +51,8 @@ import {
 import { SourcePermissions } from '../src/schema/sources';
 import { getEncryptedFeatures } from '../src/growthbook';
 import { base64 } from 'graphql-relay/utils/base64';
+import { cookies } from '../src/cookies';
+import { signJwt } from '../src/auth';
 
 let app: FastifyInstance;
 let con: DataSource;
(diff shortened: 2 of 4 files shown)


src/auth.ts
```diff
@@ -93,13 +96,13 @@ const plugin = async (
     publicKey = Buffer.from('test');
     privateKey = Buffer.from('test');
   } else {
-    publicKey = fs.readFileSync(process.env.JWT_PUBLIC_KEY_PATH);
-    privateKey = fs.readFileSync(process.env.JWT_PRIVATE_KEY_PATH);
+    loadAuthKeys();
```
rebelchris, line 99: This won't make any diff on prod?

.env
```diff
@@ -39,3 +39,6 @@ ANALYTICS_URL=http://localhost:5000
 
 EXPERIMENTATION_KEY=6FA4C8D3DD2400B9D614BB7A229FD69B
 #GROWTHBOOK_CLIENT_KEY=override_me_in_env.development
+
+JWT_PUBLIC_KEY_PATH=../apps/.cert/public.pem
+JWT_PRIVATE_KEY_PATH=../apps/.cert/key.pem
```
rebelchris, line 44: Feels a bit weird to link to other project like this 🤣

__tests__/boot.ts
```diff
@@ -361,6 +363,32 @@ describe('logged in boot', () => {
...
   });
+
+  it('should not dispatch whoami when jwt is available', async () => {
+    const accessToken = await signJwt(
+      {
+        userId: '1',
+        roles: [],
+      },
+      15 * 60 * 1000,
+    );
```
omBratteng, line 374: Should there be a test if the jwt has expired?

idoshamun: There's a test with no jwt, do you think we need an expired one as well?

rebelchris (approved):
Minor remark, more on technicality not implementation.

__tests__/boot.ts
rebelchris: Probably a good case actually yes :)

.env
idoshamun: Hehe yes quick and dirty, maybe we should think how to approach it down the line

src/auth.ts
idoshamun: Nope, I made it compatible with adhoc scripts that I had to run to check jwt validity

__tests__/boot.ts
omBratteng: Doesn't hurt to have one 🙃

.env
rebelchris: Yeah leave it for now.
We can actually simply move the certs to API I think.
Just means adjusting the scripts here  😅

__tests__/boot.ts
idoshamun: I'll create one quickly

idoshamun: Done! You can check again

omBratteng: blueblob_jump

omBratteng (approved):

PR #1459, opened 2023-10-05
fix: allow cors from vercel previews

rebelchris:
Proposal to allow cors from the vercel previews.
It's still narrowed down to only our domains there.


omBratteng (approved):

capJavert (approved):
Nice, let's try!

PR #1461, opened 2023-10-05
feat:  add `updateSharePost` mutation support

DragosIorgulescu:
- allows both creating and updating an existing post record

WT-1801


src/entity/posts/utils.ts
```diff
@@ -248,13 +248,14 @@ export const generateTitleHtml = (
 ): string =>
   `<p>${renderMentions(markdown.utils.escapeHtml(title), mentions)}</p>`;
 
-export const createSharePost = async (
+export const upsertSharePost = async (
```
DragosIorgulescu, line 251: I went ahead with re-using the existing mutation rather than creating a new one. Let me know your thoughts

src/schema/posts.ts
```diff
@@ -1622,19 +1626,22 @@ export const resolvers: IResolvers<any, Context> = {
...
 
-      const newPost = await createSharePost(
+      const newPost = await upsertSharePost(
         ctx.con,
         sourceId,
         ctx.userId,
         id,
         commentary,
+        true,
+        postId,
```
DragosIorgulescu, line 1644: lack of named arguments usage results in this, I can refactor this further if we want to clean it up

rebelchris (commented):
Ideally we should make a new call here.
It could potentially become different evaluation and I think makes it a bit safer.

src/entity/posts/utils.ts
capJavert: I agree with Chris it should be separate thing.

src/schema/posts.ts
capJavert: In API project we don't use named arguments most of the times. I am up for it to start but maybe see what @idoshamun thinks.

idoshamun: It's for legacy reasons, I'm up to using them. It's not a guideline :)

rebelchris: I prefer named as well, I think for new/refactored stuff we can start implementing it (if scope stays small)

src/schema/posts.ts
```diff
@@ -1643,6 +1662,28 @@ export const resolvers: IResolvers<any, Context> = {
...
+      {
+        id,
+        sourceId,
+        commentary,
+      }: { id: string; sourceId: string; commentary: string },
+      ctx,
+      info,
+    ): Promise<GQLPost> => {
+      await ctx.con.getRepository(Post).findOneByOrFail({ id });
+      await ensureSourcePermissions(ctx, sourceId, SourcePermissions.Post);
```
rebelchris, line 1676: I think we need some form of test + check if you are eliged to edit this post.
You can see an example in the editPost basically.

src/schema/posts.ts
```diff
@@ -1643,6 +1662,28 @@ export const resolvers: IResolvers<any, Context> = {
       );
       return getPostById(ctx, info, newPost.id);
     },
+    updateSharePost: async (
```
rebelchris, line 1665: probably stick with editSharedPost to keep it uniform with the other call

src/schema/posts.ts
DragosIorgulescu: good point, I was meaning to ask about this -- this should now take that into account as well

src/entity/posts/utils.ts
```diff
@@ -340,6 +341,56 @@ export const createSharePost = async (
...
+      },
+      {
+        title: strippedCommentary,
+        titleHtml,
+      },
+    );
+
+    if (mentions.length) {
+      await saveMentions(con, postId, userId, mentions, PostMention);
+    }
```
capJavert, line 379: What happens when we remove mentions on post edit, we should remove them right?

src/entity/posts/utils.ts
```diff
@@ -340,6 +341,56 @@ export const createSharePost = async (
...
+  commentary: string | null,
+) => {
+  let strippedCommentary = commentary;
+
+  if (commentary?.length) {
+    strippedCommentary = commentary.trim();
+    await validateCommentary(strippedCommentary);
+  } else {
+    strippedCommentary = null;
+  }
```
capJavert, line 358: We should add a test to validate commentary properly removes on edit.

src/entity/posts/utils.ts
```diff
@@ -340,6 +341,56 @@ export const createSharePost = async (
...
+  try {
+    const mentions = await getMentions(con, commentary, userId, sourceId);
+    const titleHtml = commentary?.length
+      ? generateTitleHtml(commentary, mentions)
+      : null;
+
+    await con.getRepository(SharePost).update(
+      {
+        id: postId,
+        authorId: userId,
```
capJavert, line 369: Do we need `authorId` here, `psotId` should be unique?

capJavert (commented):
Almost there, some missing tests and questions.

src/entity/posts/utils.ts
DragosIorgulescu: agreed, it's just another layer of security, to avoid the possibility of loading someone else's post, we also scope it by the author and current user ID, we can remove it if you feel it's overkill, since we also have a validation for this

src/entity/posts/utils.ts
DragosIorgulescu: I added [a test for that](https://github.com/dailydotdev/daily-api/blob/c1d9b8d5de8acb2353e1a671ec40aabf895afdc3/__tests__/posts.ts#L2090-L2100) already 🙏  unless you are referring to something else?

capJavert: My bad, missed it due to different name, tnx for checking

src/entity/posts/utils.ts
capJavert: From perspective of a function it is called `updateSharePost` so if it was reused to update shared post outside of concrete usage case of editing mutation it would not exactly do that, it is now basically `updateSharePostFromAuthor`. 

So since we have checks on mutation level I would remove `authorId` for that case and make it more reusable. If it makes sense?

src/entity/posts/utils.ts
DragosIorgulescu: that's a good point, and as far as I can see, this use case is not being covered on the `editPost` mutation either -- maybe we should fix this there as well?

src/entity/posts/utils.ts
DragosIorgulescu: that's a fair point 👍 updated

src/entity/posts/utils.ts
DragosIorgulescu: after [discussing in slack](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1697103348449099), it seems not removing mentions is intentional, for now

capJavert (approved):

__tests__/posts.ts
```diff
@@ -2011,6 +2011,126 @@ describe('mutation sharePost', () => {
...
+    });
+    await con.getRepository(SharePost).save({
+      id: 'sharePost',
+      shortId: 'sharePost',
+      sourceId: 's1',
+      type: PostType.Share,
+      title: 'Foo Bar',
+      authorId: '1',
+    });
+  });
```
rebelchris, line 2050: I think this already happens on top level, we should aim to avoid re-running extra queries to avoid timings

__tests__/posts.ts
```diff
@@ -2011,6 +2011,126 @@ describe('mutation sharePost', () => {
...
+    loggedUser = '2';
+
+    return testMutationErrorCode(
+      client,
+      { mutation: MUTATION, variables },
+      'FORBIDDEN',
+    );
+  });
+
+  it('should update the post w/ a trimmed commentary', async () => {
```
rebelchris, line 2078: Typo I think?

__tests__/posts.ts
```diff
@@ -2011,6 +2011,126 @@ describe('mutation sharePost', () => {
...
+    loggedUser = '1';
+    await con.getRepository(User).update({ id: '2' }, { username: 'lee' });
+    const params = { ...variables };
+    params.commentary = `<style>html { color: red !important; }</style>`;
+    const res = await client.mutate(MUTATION, { variables: params });
+    expect(res.errors).toBeFalsy();
+    const post = await con
+      .getRepository(SharePost)
+      .findOneBy({ id: variables.id });
+    expect(post.titleHtml).toMatchSnapshot();
```
rebelchris, line 2130: I know we probably do it for the other one, but can you perhaps change this to inline evaluation?
Slowly want to get rid of snapshots as they always turn to be pests.

src/entity/posts/utils.ts
```diff
@@ -340,6 +341,49 @@ export const createSharePost = async (
...
+  commentary: string | null,
+) => {
+  let strippedCommentary = commentary;
+
+  if (commentary?.length) {
+    strippedCommentary = commentary.trim();
+    await validateCommentary(strippedCommentary);
+  } else {
+    strippedCommentary = null;
+  }
```
rebelchris, line 358: I would say maybe uniform this with the one in `validatePost`.
Ideally we have two extra functions there: `validateTitle` and `validateCommentary` the validatePost can then re-use this.

It's just to ensure all logic is the same accross posts.

src/schema/posts.ts
```diff
@@ -1643,6 +1662,35 @@ export const resolvers: IResolvers<any, Context> = {
...
+        commentary,
+      }: { id: string; sourceId: string; commentary: string },
+      ctx,
+      info,
+    ): Promise<GQLPost> => {
+      const post = await ctx.con.getRepository(Post).findOneByOrFail({ id });
+
+      if (post.authorId !== ctx.userId) {
+        throw new ForbiddenError(
+          `Editing other people's posts is not allowed!`,
```
rebelchris, line 1679: Since this is now re-used logic, can we maybe either abstract it to a function or use a enum for the error message?
(in case of changing the message for instance, but option 1 would be my preference)

src/entity/posts/utils.ts
rebelchris: Ah I see this was existing, and evaluates a little different as it's technically a title, but different name.
Maybe we can add the trim stuff inside this `validateCommentary`?

rebelchris (changes requested):
Nice almost there 🤘

__tests__/posts.ts
DragosIorgulescu: It was not happening at the top level actually (this file could use some cleanup overall IMO) -- but I did replace it with the existing function for generating fixtures in this file 🙏

src/schema/posts.ts
DragosIorgulescu: I started extracting this into a function, but realized in the process that this is not entirely re-used logic, the way it is used in `editPost` is based on slightly different conditions. 

The exception management could use some structure here, but it feels like a more involved refactor that could be done separately. What are your thoughts?

__tests__/posts.ts
rebelchris: The w/ in description

src/schema/posts.ts
rebelchris: Ah ok then never mind, it's more a nice to have for now.

__tests__/posts.ts
DragosIorgulescu: updated the test descriptions 🙏

__tests__/__snapshots__/posts.ts.snap
```diff
@@ -77,6 +77,10 @@ exports[`mutation editPost should allow mention as part of the content 1`] = `
 "
 `;
 
+exports[`mutation editSharePost should share to squad with mentioned users 1`] = `"<p>Test <a href=\\"http://localhost:5002/lee\\" data-mention-id=\\"2\\" data-mention-username=\\"lee\\">@lee</a> @non-existent</p>"`;
```
rebelchris, line 80: Think this got reintroduced?

rebelchris (approved):
Looks good to me, just not sure if you re-introduced a snapshot?

__tests__/__snapshots__/posts.ts.snap
DragosIorgulescu: good catch, yeah, removed 🙏

PR #1462, opened 2023-10-05
feat: send sendgrid events to analytics

idoshamun:


src/routes/redirects.ts
```diff
@@ -21,32 +16,22 @@ const sendAnalyticsEvent = async (
...
-    const events = [
+    await sendAnalyticsEvent([
       {
         event_timestamp: now,
-        event_id: generateEventId(now),
         event_name: 'page view',
         event_page: '/get',
         app_platform: 'redirector',
         query_params: queryStr.length > 2 ? queryStr : undefined,
-        session_id: boot.visit.sessionId,
```
rebelchris, line 32: There shouldnt be any linking of boot ids to tracking?
These related elements would all get new ids correct?

rebelchris (approved):
Lol you might as well tackle the unsubscribe rule 👀

src/routes/redirects.ts
idoshamun: Nope, I just unified the logic here and moved it internally

PR #1464, opened 2023-10-06
feat: process & store post questions on post worker processing

DragosIorgulescu:
- always create questions for new posts (if present)
- only update questions for existing posts if there aren't any already present

ref [WT-1866](https://dailydotdev.atlassian.net/browse/WT-1866)

[WT-1866]: https://dailydotdev.atlassian.net/browse/WT-1866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -8,6 +8,7 @@ import {
   Post,
   PostKeyword,
   PostOrigin,
+  PostQuestion,
   PostType,
   SharePost,
   Source,
@@ -98,6 +99,23 @@ const createDefaultKeywords = async () => {
     status: 'synonym',
     synonym: 'ab-testing',
   });
+};
+
+const DEFAULT_QUESTIONS = [
+  'What is the one thing you need to truly understand neural networks?',
+  'How do neural networks learn?',
+  'What is the process of training a neural network?',
+];
+
+const createDefaultQuestions = async (postId: string) => {
+  const repo = con.getRepository(PostQuestion);
+
+  for (const question of DEFAULT_QUESTIONS) {
+    await repo.insert({
+      postId,
+      question,
+    });
+  }
 };
 
 const createSharedPost = async (id = 'sp1') => {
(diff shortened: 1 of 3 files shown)


__tests__/workers/postUpdated.ts
```diff
@@ -100,6 +101,23 @@ const createDefaultKeywords = async () => {
...
+
+const createDefaultQuestions = async (postId: string) => {
+  const repo = con.getRepository(PostQuestion);
+
+  for (const question of DEFAULT_QUESTIONS) {
+    await repo.insert({
+      postId,
+      question,
+    });
+  }
```
rebelchris, line 118: No prio, but think this would also work:
```
await repo.save(DEFAULT_QUESTIONS.map((question) => ({ postId, question })));
```

__tests__/workers/postUpdated.ts
```diff
@@ -593,3 +611,142 @@ it('should not update already approved post', async () => {
   expect(submission.id).toEqual(uuid);
   expect(submission.status).toEqual(SubmissionStatus.Accepted);
 });
+
+describe('on post create', () => {
+  beforeEach(async () => {
+    await saveFixtures(con, Source, [
```
rebelchris, line 617: The main before each already has all sources do you need this specific one?

__tests__/workers/postUpdated.ts
```diff
@@ -593,3 +611,142 @@ it('should not update already approved post', async () => {
...
+      expect(questionsAfter.length).toEqual(3);
+      expect(questionsAfter[0].question).toEqual(
+        'What is the one thing you need to truly understand neural networks?',
+      );
+      expect(questionsAfter[1].question).toEqual(
+        'How do neural networks learn?',
+      );
+      expect(questionsAfter[2].question).toEqual(
+        'What is the process of training a neural network?',
+      );
```
rebelchris, line 663: I think maybe do a array check, safer in case the order ever changes

__tests__/workers/postUpdated.ts
```diff
@@ -593,3 +611,142 @@ it('should not update already approved post', async () => {
...
+        source_id: COMMUNITY_PICKS_SOURCE,
+        submission_id: uuid,
+        extra: {
+          questions: DEFAULT_QUESTIONS,
+        },
+      });
+
+      const post = await con
+        .getRepository(Post)
+        .findOneBy({ title: 'With questions' });
```
rebelchris, line 650: I don't think you need the actual post here, should be able to match if you pass a already mocked post basically.
It would avoid time on database read-writes

__tests__/workers/postUpdated.ts
```diff
@@ -593,3 +611,142 @@ it('should not update already approved post', async () => {
...
+      expect(questionsAfter[1].question).toEqual(
+        'How do neural networks learn?',
+      );
+      expect(questionsAfter[2].question).toEqual(
+        'What is the process of training a neural network?',
+      );
+    });
+  });
+
+  describe('when data does not include questions', () => {
```
rebelchris, line 667: This one would already be covered in general as the other test don't include questions.

rebelchris (changes requested):
Nice work!
Only have some ideas for optimising test cases, otherwise really well done on this, super clean

__tests__/workers/postUpdated.ts
DragosIorgulescu: great call, thank you 🙏

__tests__/workers/postUpdated.ts
DragosIorgulescu: ah, missed that, thanks for pointing it out 👍

__tests__/workers/postUpdated.ts
DragosIorgulescu: agreed, but since the other tests don't explicitly test for this, I'd like to still keep this one as a sanity check -- in case the other ones get updated in the future to include questions for some reason

rebelchris (approved):
Good from my side.

With API projects we always need to wait for @idoshamun approval as well, so don't merge it yet please.

Another note, please see our existing [guidelines](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/375652412/Git#Branch-names) on PR title naming.

idoshamun (approved):
Perfect!

capJavert (approved):
Smooth 🚀

PR #1466, opened 2023-10-06
feat: import digest users from csv file

capJavert:
Import provided list from Dave to DB for digest experiment. Reused the same script from now.


idoshamun:
@capJavert my SQL client allows to easily import a csv if you want, may be easier

capJavert:
> @capJavert my SQL client allows to easily import a csv if you want, may be easier

I just don't know if we can also pull the user's timezone and assign through import?

idoshamun:
Missed that part! I'll review later

bin/subscribeUsersToPersonalizedDigest.ts
```diff
@@ -1,45 +1,52 @@
 import '../src/config';
 import createOrGetConnection from '../src/db';
+import fs from 'fs/promises';
 
-(async (): Promise<void> => {
-  const createdFromArgument = process.argv[2];
-  const createdToArgument = process.argv[3];
+const parseCSV = <T>(csv: string, splitBy: string): T[] => {
```
rebelchris, line 5: We also have import { parse } from 'csv-parse'; installed.
see for instance notificationSubscribedUsers

rebelchris (approved):
Wondering if it wouldn't be easier if we go with the manual CSV import in mysql?
I think it's fairly easy to either get dave to include timezone or we could once-off query.

But it's fully optional, no strong opinion.
I also noticed you do quite a lot of CSV manipulation manually, we do have a package already for it.

bin/subscribeUsersToPersonalizedDigest.ts
capJavert: I had that one from some other scripts of mine 😆 , did not wish to install dependencies, but also did not look if we have some already, yes I'll use the package. 

Manual CSV import don't think supports the timezone assignment, see https://github.com/dailydotdev/daily-api/pull/1466#issuecomment-1751183586

We could bottom line do CSV import and then run a query to set timezone based on the value of `user.timezone` column. 

We could also see if Dave can pull it and provide timezones in CSV. Again the above seemed fastest.

rebelchris: Ah yeah the timezone was more around the initial CSV could potentially have it.
But also happy with your propsal tbh, don't want to spend to long on the task.

bin/subscribeUsersToPersonalizedDigest.ts
```diff
@@ -1,45 +1,52 @@
...
   await con.transaction(async (manager) => {
     await manager.query(`
-      INSERT INTO user_personalized_digest ("userId", "preferredTimezone")
-      SELECT id AS "userId", COALESCE(timezone, 'Etc/UTC') AS "preferredTimezone" FROM public.user WHERE "infoConfirmed" IS TRUE
-      AND "createdAt" > '${createdFromDate.toISOString()}' AND "createdAt" < '${createdToDate.toISOString()}'
-      ON CONFLICT DO NOTHING;
-    `);
+        INSERT INTO user_personalized_digest ("userId", "preferredTimezone")
+        SELECT id AS "userId", COALESCE(timezone, 'Etc/UTC') AS "preferredTimezone" FROM public.user WHERE id IN (${users
+          .map((item) => (item.user_id ? `'${item.user_id}'` : null))
```
idoshamun, line 46: More like an FYI because it's a script file not really something will run in production, typeorm supports query params of type array so you can easily use that next time. We have plenty of examples throughout the codebase

idoshamun (approved):

PR #1467, opened 2023-10-08
refactor: use views for search recs

idoshamun:
Earlier today we checked how many users see the search recommendations and it was relatively low. We guess that it's because upvote is less frequent than views.

We'll try to change that and see if it increases adoption.


rebelchris (approved):
Lets give it a try,
Did notice lots of unrelated changes maybe set linter more explicit from my side

idoshamun:
@rebelchris it's my stupid local settings that were messed up. I'll fix that

PR #1468, opened 2023-10-08
refactor: streamline retryFetch and convert most usage to retry

idoshamun:
I refactor our `retryFetch` function to support more generic use cases so we'll be able to apply system-wide. It's best practice to use retries and now that we know our retry library don't introduce issues we can actually apply it :P

### DIFF
--- a/src/integrations/retry.ts
+++ b/src/integrations/retry.ts
@@ -1,6 +1,6 @@
 import retry, { OperationOptions } from 'retry';
 import isNetworkError from './networkError';
-import fetch, { RequestInit } from 'node-fetch';
+import fetch, { RequestInfo, RequestInit, Response } from 'node-fetch';
 
 export class AbortError extends Error {
   public originalError: Error;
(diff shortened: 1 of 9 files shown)


src/integrations/retry.ts
```diff
@@ -66,23 +81,29 @@ export async function asyncRetry<T>(
...
     if (res.status < 500) {
-      throw new AbortError(`request is invalid: ${res.status}`);
+      throw new AbortError(err);
     }
-    throw new Error(`unexpecetd response from feed service: ${res.status}`);
+    throw err;
   }, retryOpts);
 }
+
+export async function retryFetchParse<T>(
```
rebelchris, line 102: I'd only argue this placement inside `feed` dir is a bit off.

src/kratos.ts
```diff
@@ -33,28 +33,26 @@ const fetchKratos = async (
...
-      });
-      if (res.status < 300) {
-        return { res: await res.json(), headers: res.headers };
-      }
-      const err = new KratosError(res.status, await res.text());
-      if (res.status >= 500) {
-        req.log.warn({ err }, 'unexpected error from kratos');
-        throw err;
+  try {
+    const res = await retryFetch(endpoint, {
```
rebelchris, line 37: Why don't we use `retryFetchParse` here as well?

rebelchris (approved):
Nice, thanks for that, have some minor non blocking suggestions

src/kratos.ts
idoshamun: Because we need the response headers. Parse returns only the body parsed as json

src/integrations/retry.ts
idoshamun: It's not in the `feed` dir, it's under integrations, or am I missing something?

src/kratos.ts
rebelchris: Ah I see :) 
Missed the headers specifically as I saw you can manually do custom error status on another call.

PR #1469, opened 2023-10-09
feat: utlize async queue to wait for pending notify calls

capJavert:
We found an issue where async method `notifyGeneratePersonalizedDigest` does not get to be resolved before the stream ends and the instance that is running a cron destroys itself. 

Utlizes async queue to manage notify tasks and awaits for queue to be drained (completed) before exiting.

### DIFF
--- a/src/cron/personalizedDigest.ts
+++ b/src/cron/personalizedDigest.ts
@@ -1,3 +1,4 @@
+import fastq from 'fastq';
 import { notifyGeneratePersonalizedDigest } from '../common';
 import { UserPersonalizedDigest } from '../entity/UserPersonalizedDigest';
 import { Cron } from './cron';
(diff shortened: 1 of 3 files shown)


src/cron/personalizedDigest.ts
```diff
@@ -14,11 +15,18 @@ const cron: Cron = {
       });
 
     const personalizedDigestStream = await personalizedDigestQuery.stream();
+    const notifyQueueConcurrency = 1000;
+    const notifyQueue = fastq.promise(
+      async (personalizedDigest: UserPersonalizedDigest) => {
+        await notifyGeneratePersonalizedDigest(logger, personalizedDigest);
+      },
```
idoshamun, line 22:
```suggestion
      (personalizedDigest: UserPersonalizedDigest) => notifyGeneratePersonalizedDigest(logger, personalizedDigest),
```

idoshamun (approved):
Amazing!

rebelchris (approved):
Nice nothing blocking from my side.
Let those emails come 👋

src/cron/personalizedDigest.ts
capJavert: @idoshamun for some reason with above changes tests timed out, not sure if it's a jest thing, fastq thing https://app.circleci.com/pipelines/github/dailydotdev/daily-api/5895/workflows/db8d60e6-58fc-422b-ac89-dc8f4f0bc8fa/jobs/23036 cc: @rebelchris any clue?

Reverted below and tests passed again.

PR #1470, opened 2023-10-09
fix: add settings default

rebelchris:
Return default settings *(as logged in user) instead of adding query on boot.


omBratteng (changes requested):
Looks good, just need to remove `TypeOrmError` import

vpol (approved):

idoshamun (approved):
Thanks for the quick fix!

omBratteng:
Looks like the default value of `companionExpanded` has changed so tests are failing

omBratteng (approved):

PR #1471, opened 2023-10-09
feat: add generationTimestamp for digest worker

capJavert:
Based on the thread [here](https://dailydotdev.slack.com/archives/C05G8BHHVBR/p1696840091899069?thread_ts=1696581507.796259&cid=C05G8BHHVBR) I added `generationTimestamp`. 

That timestamp can be used to set timestamp for time of generation of digest message and allow us to forward/backward between actual sending days without a need to wait for cron. 

It can also be useful if for any reason digest fails to send we can just provide `generationTimestamp` for previous day and digest will process like it's that day. We would still need to watch out for people timezones but at least it gives us flexibility to do it.

It will also help us with initial testing.

### DIFF
--- a/__tests__/cron/personalizedDigest.ts
+++ b/__tests__/cron/personalizedDigest.ts
@@ -52,6 +52,7 @@ describe('personalizedDigest cron', () => {
       expect(notifyGeneratePersonalizedDigest).toHaveBeenCalledWith(
         expect.anything(),
         personalizedDigest,
+        expect.any(Number),
       );
     });
   });
@@ -81,6 +82,7 @@ describe('personalizedDigest cron', () => {
       expect(notifyGeneratePersonalizedDigest).toHaveBeenCalledWith(
         expect.anything(),
         personalizedDigest,
+        expect.any(Number),
       );
     });
   });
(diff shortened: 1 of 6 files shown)


idoshamun (approved):

rebelchris (approved):

rebelchris (approved):

__tests__/cron/personalizedDigest.ts
```diff
@@ -81,7 +83,28 @@ describe('personalizedDigest cron', () => {
       expect(notifyGeneratePersonalizedDigest).toHaveBeenCalledWith(
         expect.anything(),
         personalizedDigest,
+        expect.any(Number),
       );
     });
   });
+
+  it('should log notify count', async () => {
```
idoshamun, line 91: Usually I don't recommend testing for logs but up to you :)

idoshamun (approved):

__tests__/cron/personalizedDigest.ts
capJavert: I can see what you mean. From my point of view it's a critical part of the feature since we want to make sure logs are there to know how many digests were processed. In the future that could ping some other service (eg. tracking) instead of logging and we would want to test that?

PR #1473, opened 2023-10-11
refactor: use FeedClient for personalized digest

capJavert:
Response was standardized on platform side to match regular feed so we can use default client.

Will sync up with @pnvasanth tomorrow on when we merge.


idoshamun (approved):
Much nicer!

omBratteng (approved):

rebelchris (approved):
Nice!

PR #1478, opened 2023-10-16
feat: add dead letter topic for digest subscription

capJavert:
- added dead letter topic in pulmi
- added configuration for digest subscription (by default subscriptions do not opt into dead letter topic)
- removed try catch from digest worker

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -18,6 +18,8 @@ import {
   Redis,
   detectIsAdhocEnv,
   SqlDatabase,
+  Worker,
+  Stream
 } from '@dailydotdev/pulumi-common';
 
 const isAdhocEnv = detectIsAdhocEnv();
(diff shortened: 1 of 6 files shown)


.infra/workers.ts
```diff
@@ -240,5 +244,20 @@ export const workers: Worker[] = [
   {
     topic: 'api.v1.generate-personalized-digest',
     subscription: 'api.personalized-digest-email',
+    args: {
+      ackDeadlineSeconds: 60,
```
capJavert, line 248: Also updated ack deadline for worker call, default is 10s.

.infra/index.ts
```diff
@@ -83,10 +85,32 @@ const envVars: Record<string, Input<string>> = {
   redisHost,
 };
 
+const deadLetterTopicName = 'api.v1.dead-letter'
+const deadLetterTopic = new Stream(deadLetterTopicName, { isAdhocEnv, name: deadLetterTopicName, })
+
+const addDeadLetterTopicToWorkers = (workers: Worker[]): Worker[] => {
```
idoshamun, line 91: I think we should add the dead letter functionality to pulumi-common. It will be useful to reuse in other projects

.infra/index.ts
```diff
@@ -83,10 +85,32 @@ const envVars: Record<string, Input<string>> = {
...
+  return workers.map(worker => {
+    if (worker.args?.deadLetterPolicy) {
+      return {
+        ...worker,
+        args: {
+            ...worker.args,
+            deadLetterPolicy: {
+              ...worker.args.deadLetterPolicy,
+              deadLetterTopic: deadLetterTopic.name.apply(name => name)
```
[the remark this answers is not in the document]
idoshamun: I'm not sure pubsub emulator supports dead letter

idoshamun, line 100: It's the same. When you do apply without manipulation it has no effect

```suggestion
              deadLetterTopic: deadLetterTopic.name
```

.infra/index.ts
capJavert: Yes please!

idoshamun (approved):

PR #1480, opened 2023-10-17
feat: onboarding v4

capJavert:
Executing API part of [DR](https://dailydotdev.atlassian.net/l/cp/oxreh8j7)

Feed config for preview will be added separately as I work on frontend implementation-


src/schema/tags.ts
```diff
@@ -76,5 +81,22 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
         hits: hits.map((x) => ({ name: x.value })),
       };
     },
+    onboardingTags: async (source, args, ctx): Promise<GQLTagSearchResults> => {
+      const hits = await ctx.getRepository(Keyword).find({
+        select: ['value'],
+        where: {
+          status: 'allow',
```
idoshamun, line 88: Since we set onboarding manually, this feels redundant. Especially when we indexed onboarding flag by itself

src/schema/tags.ts
```diff
@@ -76,5 +81,22 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
         hits: hits.map((x) => ({ name: x.value })),
       };
     },
+    onboardingTags: async (source, args, ctx): Promise<GQLTagSearchResults> => {
```
idoshamun, line 84: Why do we return the response as search results?

capJavert: Wanted to reuse the type, but I guess it does not hurt to add another one.

idoshamun: Yes, I think it's better to define a new type for that. Of course we can keep it 99% compatible

capJavert: Yeah, updated already [48dbfdc](https://github.com/dailydotdev/daily-api/pull/1480/commits/48dbfdcf110a2618671c4d83e91b04ddec2a8103)

src/schema/tags.ts
```diff
@@ -76,5 +94,25 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+          status: 'allow',
+          flags: {
+            onboarding: true,
+          },
+        },
+        order: { value: 'ASC' },
+      });
+
+      return {
+        hits: hits.map((x) => ({ name: x.value })),
```
rebelchris, line 114:
```suggestion
        hits: hits.map((hit) => ({ name: hit.value })),
```

rebelchris (approved):
Nice, thanks for even fixing the seeds 🤯
Assume you'll do the recommendedTags as seperate task?

src/schema/tags.ts
capJavert: I'll do all the API work in this PR I think.

src/schema/tags.ts
capJavert: Added value and status index in 6cbb60a0b26ac82fce03c61396a8064a65c17443

src/migration/1697624128431-TagRecommendation.ts
```diff
@@ -0,0 +1,49 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class TagRecommendation1697624128431 implements MigrationInterface {
+    name = 'TagRecommendation1697624128431'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        const isNonProductionEnv = ['development', 'test'].includes(process.env.NODE_ENV)
+        const createdAtThreshold = isNonProductionEnv ? `'1970-01-01'` : `(current_timestamp - interval '60 day')::date`;
+        const totalThreshold = isNonProductionEnv ? 1 : 100;
```
capJavert, line 9: This is here because in local environment the data conditions can not fulfil thresholds we use in production. This makes sure when we run mutations in development/test we can still have the recommendations view to return something.

Open for better alternatives if anybody has it.

src/schema/tags.ts
```diff
@@ -76,5 +112,46 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+    recommendedTags: async (
+      source,
+      { tags, excludedTags },
+      ctx,
+    ): Promise<GQLTagResults> => {
+      const uniqueTagsToExclude = new Set([...tags, ...excludedTags]);
+
+      if (uniqueTagsToExclude.size > 1000) {
+        throw new ValidationError('Tag limit reached');
+      }
```
capJavert, line 139: Added this as a guard because potentially someone could send really big payloads which would result in big queries to the DB.

src/cron/updateTagRecommendations.ts
```diff
@@ -0,0 +1,20 @@
...
+import { TagRecommendation } from '../entity/TagRecommendation';
+
+const cron: Cron = {
+  name: 'update-tag-recommendations',
+  handler: async (con, logger) => {
+    const materializedViewName =
+      con.getRepository(TagRecommendation).metadata.tableName;
+
+    try {
+      await con.query(`REFRESH MATERIALIZED VIEW ${materializedViewName}`);
```
capJavert, line 11: Materialized view is cached in db until refreshed. This allows us to have fast query for recommendations and we refresh it once a week with new keywords.

src/entity/Keyword.ts
```diff
@@ -8,7 +8,14 @@ import {
...
 export type KeywordStatus = 'pending' | 'allow' | 'deny' | 'synonym';
 
+export type KeywordFlags = Partial<{
+  onboarding: boolean;
+}>;
+
+export type KeywordFlagsPublic = never;
+
 @Entity()
+@Index(['value', 'status'])
```
rebelchris, line 18: Not sure if you want to add a name to it perhaps? (optional)

src/entity/TagRecommendation.ts
```diff
@@ -0,0 +1,44 @@
...
+        select "keywordX", "keywordY", count(*) * 1.0 / min(t.total) as probability
+        from pairs p
+        join totals t on p."keywordX" = t.keyword
+        group by 1, 2
+    )
+    select * from likelihood
+    where probability > 0.05
+  `,
+})
+@Index(['keywordX', 'keywordY', 'probability'])
```
rebelchris, line 34: Tagging if you would like to name the other index :)

rebelchris (approved):
Sorry have nothing smart to say about the limit :(

src/entity/Keyword.ts
capJavert: I don't see we name them usually, name is also handled automatically in that case by typeorm which I think is better honestly?

capJavert: https://github.com/dailydotdev/daily-api/pull/1481#discussion_r1365003145 I just seen this discussion 😄 Let's see.

capJavert: I'll add.

src/cron/updateTagRecommendations.ts
idoshamun: Amazing amazing! Let's not guard it with try catch though. I'd like to employ the retry mechanism

src/entity/Keyword.ts
idoshamun: Why do we need this index?

src/migration/1697624128431-TagRecommendation.ts
idoshamun: I don't think we need it for production, we can just call the refresh function. For local maybe we can utilize seeds? We don't need anything fancy or right, we just need something

src/schema/tags.ts
```diff
@@ -76,5 +112,47 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+
+      const hits = await ctx.getRepository(TagRecommendation).find({
```
idoshamun, line 141: Please add limit here. I think design mentioned 3 or 4 tags

src/cron/updateTagRecommendations.ts
capJavert: Got it, but if we can clear up something. Without dead letter it would retry forever? I am curious how we handle those cases in current production or do they just not happen? 

From my point of view every subscription should have dead letter just for sake of limiting number of retries.

idoshamun: The only case it can happen is when you have a third-party service that you aren't sure of its availability.
Delay is high enough to deal with any production situation basically

src/migration/1697642145710-TagSearchIndex.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class TagSearchIndex1697642145710 implements MigrationInterface {
```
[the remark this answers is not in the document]
capJavert: Isn't the point of having migrations to do one thing? For example adding this index breaks something. We want to revert it, we can do it easily. If materialized view was bundled with it we would also revert that part which would break rest of the codebase.

src/schema/tags.ts
capJavert: Yeah, just checked with product yesterday, it was 5.

src/entity/Keyword.ts
capJavert: Due to this https://github.com/dailydotdev/daily-api/pull/1480#discussion_r1362163808

idoshamun: Oh it's because of the order by. So the right order is first status and then value. Where clause terms should be before order by terms afaik

capJavert: It's not due to ORDER only, status and value are used in WHERE. But since psql used left to right principle it would make sense to put status first yes.

src/cron/updateTagRecommendations.ts
capJavert: Granted above worker is low impact, I'll remove the try catch, just wanted to explain my logic of why I added it.

src/migration/1697624128431-TagRecommendation.ts
capJavert: Not sure what do you mean by not needing it in production? This migration sets up the view for usage. 

As for seeding data locally, we are trying to solve two issues:

> It requires for keyword to be used in more then 100 posts

We could solve it with seeding but we would need more posts and more keywords spread out. I guess we would need to pull some slice of production data into seeds to do it quickly. With lowering the total threshold we will naturally get recommendations from our seed data (I checked it works quite good),

> In production we are showing posts created in last 60 days

With seed posts are created few years ago, and also if we do add dynamic `createdAt` to seed it would again expire in 60 days.

My logic with above is that since we always reset our local env multiple times in a week it was better to do this adjustment to view threshold locally to avoid adding more data or having a dynamic createdAt which would again never satisfy constraints. 

In production this will run once and function as any other migration.

src/entity/Keyword.ts
capJavert: Added in 053c24c744860b38df07347bfbe8b0f014f2d74a

src/schema/tags.ts
capJavert: Added in 053c24c744860b38df07347bfbe8b0f014f2d74a

src/migration/1697642145710-TagSearchIndex.ts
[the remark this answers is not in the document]
idoshamun: Ideally, you should have one migration for PR because PR is what gets deployed eventually. A migration runs in a transaction so you can't have a situation where something succeeded and something failed, one split to multiple migration you can have this situation and then you need to check what migration caused the issue. In case the migration fails in production, the deployment will stop and there will be no impact on production anyway

src/migration/1697624128431-TagRecommendation.ts
idoshamun: Ignore me, I thought it's for something else. Let's keep it as is for now. It's one off

src/schema/tags.ts
```diff
@@ -76,5 +110,48 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+    },
+    recommendedTags: async (
+      source,
+      { tags, excludedTags },
+      ctx,
+    ): Promise<GQLTagResults> => {
+      const uniqueTagsToExclude = new Set([...tags, ...excludedTags]);
+
+      if (uniqueTagsToExclude.size > 1000) {
+        throw new ValidationError('Tag limit reached');
```
DragosIorgulescu, line 136: Is this passed through to the client as is, or does it require special handling? If it does require, we might change this message to a code from `errors.ts`

DragosIorgulescu (approved):
just one small comment from me really, great work 🙌

idoshamun (approved):
Fantastic!

src/schema/tags.ts
capJavert: Good point, added in 16581ff6a1a70cbdf2c814d6523cd64d5444b3bf

PR #1481, opened 2023-10-17
feat: add `Invite` API support

DragosIorgulescu:
The frontend equivalent PR for this: https://github.com/dailydotdev/apps/pull/2277

### Changes
- [x] create the `Invite` table & entity
- [x] update the `referralCampaign` query to load data from the invite table if the origin matches
  - include the referral limit in the response
  - the `invite` record is loaded based on the `referralOrigin` passed into the query arguments
  - if an `invite` is found for the given origin, the returned counts are delegated to it, otherwise the previous implementation is the default
- [x] add a new `acceptFeatureInvite` mutation which both gives access to the given feature & increments the invite count for the referrer
  - checks against the `Feature` table and only updates if the feature does not exist or is not already enabled for the user
  - increments the `count` on the `invite` associated with the referrer
  - saves the `invitedBy` on the `feature` associated with the referred and enables access
- [x] Add cron support to automatically generate invite records for users who have been referred
  - filters out features that are not older than 7 days, that are not blocked, and which don't have an associated invite record already
  - runs daily
- [ ] ~~Add a script that seeds invite records for existing users with access to Search~~
  - there is no need for a script, since the cron script will automatically create invite records for all old features in the db when it first runs. 
- [x] Add `campaignCtaPlacement` to `Settings`

ref [WT-1889](https://dailydotdev.atlassian.net/browse/WT-1889)

[WT-1889]: https://dailydotdev.atlassian.net/browse/WT-1889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/users.ts
```diff
@@ -2252,3 +2294,189 @@ describe('mutation unsubscribePersonalizedDigest', () => {
...
+        query: MUTATION,
+        variables: {
+          token: 'foo',
+          referrerId: 2,
+          feature: CampaignType.Search,
+        },
+      },
+      (errors) => {
+        expect(errors[0].extensions.code).toEqual('GRAPHQL_VALIDATION_FAILED');
+        expect(errors[0].message).toEqual('Invites limit reached');
```
DragosIorgulescu, line 2366: Feels like this will need special handling on the client side, is there an internal enum for error codes which we can use for cases such as this?

src/entity/Feature.ts
```diff
@@ -31,4 +31,10 @@ export class Feature {
 
   @Column({ type: 'smallint', default: FeatureValue.Allow })
   value: FeatureValue = FeatureValue.Allow;
+
+  @Column({ length: 36, nullable: true })
+  invitedById: string;
+
+  @ManyToOne(() => User, { lazy: true, nullable: true })
```
idoshamun, line 38: We need to add a setting here that in case the user gets deleted this will be set to null. I don't remember the exact property but there's something. Remember to regenerate the migration

src/entity/Invite.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+@Entity()
+@Index(['userId', 'campaign'], { unique: true })
+export class Invite {
+  @PrimaryColumn({ type: 'text' })
+  token: string;
+
+  @Column({ type: 'text' })
+  campaign: CampaignType | FeatureType;
```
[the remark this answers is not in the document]
idoshamun: I think you're right but happy to hear others

src/migration/1697545231300-AddInvitedByIdToFeature.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddInvitedByIdToFeature1697545231300
```
idoshamun, line 3: Not a major issue but I'd appreciate one migration per PR. Makes it easy to track all the db changes of a PR

src/schema/users.ts
```diff
@@ -1044,6 +1078,55 @@ export const resolvers: IResolvers<any, Context> = {
...
+      ctx: Context,
+    ): Promise<unknown> => {
+      const featureRepo = ctx.con.getRepository(Feature);
+      const existingFeature = await featureRepo.findOneBy({
+        userId: ctx.userId,
+        feature: args.feature as FeatureType,
+      });
+
+      // silently fail if we already have the feature
+      if (existingFeature?.value === FeatureValue.Allow) return undefined;
```
idoshamun, line 1097: If there's a row in the table, we can immediately return, no need to check if it's allowed. The only other case is if we deliberately blocked someone and in this case we don't want to grant them access

src/schema/users.ts
```diff
@@ -1044,6 +1078,55 @@ export const resolvers: IResolvers<any, Context> = {
...
+      });
+
+      if (referrerInvite.count >= referrerInvite.limit) {
+        throw new ValidationError('Invites limit reached');
+      }
+
+      if (existingFeature) {
+        existingFeature.invitedById = args.referrerId;
+        existingFeature.value = FeatureValue.Allow;
+        await featureRepo.save(existingFeature);
```
idoshamun, line 1113: My previous comment, makes this edge case redundant

src/schema/users.ts
idoshamun: Giving it another thought, maybe even fetching from the feature table is not required. If it already exists when trying to insert the db will return a duplicate entry error and we can catch and silently return as you suggested. It will turn this endpoint to two queries max

src/schema/users.ts
```diff
@@ -1044,6 +1078,55 @@ export const resolvers: IResolvers<any, Context> = {
...
+      } else {
+        await featureRepo.save({
```
idoshamun, line 1115: better to use insert here. save may update existing records

src/schema/users.ts
```diff
@@ -1044,6 +1078,55 @@ export const resolvers: IResolvers<any, Context> = {
...
+          userId: ctx.userId,
+          feature: args.feature as FeatureType,
+          invitedById: args.referrerId,
+          value: FeatureValue.Allow,
+        });
+      }
+
+      await inviteRepo.update(
```
idoshamun, line 1123: This should happen in a transaction together with the insert statement above. We want both of them to fail or succeed as one atom

__tests__/users.ts
```diff
@@ -2252,3 +2294,189 @@ describe('mutation unsubscribePersonalizedDigest', () => {
...
+  });
+
+  it('should raise a validation error if the invite limit has been reached', async () => {
+    loggedUser = '1';
+
+    await con
+      .getRepository(Invite)
+      .update({ token: 'foo' }, { limit: 5, count: 5 });
+
+    await testQueryError(
```
rebelchris, line 2354: I think you can use `testMutationErrorCode` here for simplicity.

__tests__/users.ts
rebelchris: Good point!
@sshanzel are we covering this FE wise?

Dragos: Check the `errors.ts` file, we can leverage that for this part.
Possibly you'll need to introduce a new one.

src/entity/Feature.ts
rebelchris: Good point!
Check `deleteUser` in the user directive please.
There are some examples of author being set to null for instance.

src/entity/Invite.ts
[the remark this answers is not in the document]
rebelchris: My main Idea was every type is a type.
And have a separate map to see if that type is associated with any features.

```
enum CampaignType {
 Search = 'search'
}

campaignHasFeature = {
 [CampaignType.Search] = Features.Search
}
```

src/entity/Invite.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  @Column({ length: 36 })
+  userId: string;
```
rebelchris, line 19: You can define indexes here, it will generate neater once then if you run the creation.
Same for campaign field.

src/migration/1697544817927-Invite.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class Invite1697544817927 implements MigrationInterface {
+  name = 'Invite1697544817927';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "invite" ("token" text NOT NULL, "campaign" text NOT NULL, "userId" character varying(36) NOT NULL, "limit" integer NOT NULL DEFAULT '5', "count" integer NOT NULL DEFAULT '0', CONSTRAINT "PK_83dbe83cb33c3e8468c8045ea7c" PRIMARY KEY ("token"))`,
+    );
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX "IDX_e31046d3d8b9d04432ccdaaf09" ON "invite" ("userId", "campaign") `,
```
rebelchris, line 11: Try and add the index in type, it should generate it based on that.

src/migration/1697545231300-AddInvitedByIdToFeature.ts
rebelchris: Not something we've done in the past, although agree with the sentiment.
Possibly something we can start doing from now on :)

src/schema/users.ts
```diff
@@ -356,6 +361,7 @@ export const typeDefs = /* GraphQL */ `
 
   type ReferralCampaign {
     referredUsersCount: Int!
+    referralCountLimit: Int
```
rebelchris, line 364: This is optional?

src/entity/Feature.ts
sshanzel: I think it should be CASCADE: null. Can't exactly remember the right property though.

src/schema/users.ts
```diff
@@ -1044,6 +1078,55 @@ export const resolvers: IResolvers<any, Context> = {
 
       return { _: true };
     },
+    acceptFeatureInvite: async (
+      _,
+      args: {
```
rebelchris, line 1083: Why do you parse it under `args` you can actually directly add the object with params, it's neater.
Maybe args was some legacy thing inside here.

__tests__/users.ts
sshanzel: We do have some sort of function to get the error code. We can maybe map the error code or use the message.

rebelchris (changes requested):
Nice, really smooth work, hope the DR helped?
Also really nice work on the test cases looks solid.

Some generic small concerns and ideas.

__tests__/users.ts
sshanzel: @rebelchris yes, there is a TODO that says, you should display a toast based on the error message.

src/schema/users.ts
```diff
@@ -1044,6 +1078,55 @@ export const resolvers: IResolvers<any, Context> = {
...
+        token: string;
+        referrerId: string;
+        feature: string;
+      },
+      ctx: Context,
+    ): Promise<unknown> => {
```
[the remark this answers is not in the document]
DragosIorgulescu: I added an empty response to this mutation

src/schema/users.ts
DragosIorgulescu: I went with your suggestion, this definitely simplifies things if we don't care about existing blocked features. However the way to check for a uniqueness error on an insert is not very intuitive, and I am not sure it is the right approach:
https://github.com/dailydotdev/daily-api/blob/1add87fa682f7a9fc9a5459ae1219ad46c9ce1ba/src/schema/users.ts#L1114

__tests__/users.ts
DragosIorgulescu: in this case I am intentionally using `testQueryError` because I want to validate the error message itself as well, apart from the code

.infra/crons.ts
```diff
@@ -42,6 +42,10 @@ export const crons: Cron[] = [
   },
   {
     name: 'personalized-digest',
-    schedule: '15 0 * * *'
-  }
+    schedule: '15 0 * * *',
+  },
+  {
+    name: 'generate-invites',
+    schedule: '15 0 * * *',
```
rebelchris, line 49: We might want to run it on a different schedule to not block to many things at once.

src/cron/generateInvites.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Cron } from './cron';
+import { Feature, FeatureValue, Invite } from '../entity';
+import { generateUUID } from '../ids';
+
+const cron: Cron = {
+  name: 'generate-invites',
+  handler: async (con, logger) => {
+    const featuresQuery = con
+      .createQueryBuilder()
+      .from(Feature, 'f')
```
rebelchris, line 11: I don't think you need the alias actually.

src/cron/generateInvites.ts
```diff
@@ -0,0 +1,55 @@
...
+      .where(
+        'f."createdAt" < now() - interval \'7 days\' AND f."value" = :value',
+        { value: FeatureValue.Allow },
+      ).andWhere(`
```
rebelchris, line 15: I'm not sure if there is an alternative, but think this was pretty costs to run as a query.
Will need to dive into other options for you.

src/cron/generateInvites.ts
rebelchris: Alternative can't we handle it on pubsub or direct?
I can't recall why we choose the cron tbh.

src/schema/users.ts
rebelchris: I see depends on if it's always queried, but in that case it should be fine like this.

src/cron/generateInvites.ts
DragosIorgulescu: an alternative would be to use a `left join` here, but given that we are indexing both `userId` and `campaign`, using a select might actually be more performant. I'm open to other suggestions though

src/cron/generateInvites.ts
DragosIorgulescu: we can delegate this to a topic, similar to how the digest cron is simply publishing events that are then handled async. We would still be using cron, but we would be delegating processing to a worker. Is this what you are referring to?

.infra/crons.ts
idoshamun: Ideally, at least one hour after the digest

src/cron/generateInvites.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const cron: Cron = {
+  name: 'generate-invites',
+  handler: async (con, logger) => {
+    const featuresQuery = con
+      .createQueryBuilder()
+      .from(Feature, 'f')
+      .where(
+        'f."createdAt" < now() - interval \'7 days\' AND f."value" = :value',
+        { value: FeatureValue.Allow },
```
idoshamun, line 14: We also need to make sure we look for search feature, not any feature right? This is specifically for search

src/cron/generateInvites.ts
idoshamun: If it's indexed properly I don't think it should be costly. Another option is not to query for existing invites and on insert just ignore conflicts. I don't mind either, we just need to make sure we have proper indexes

src/cron/generateInvites.ts
```diff
@@ -0,0 +1,55 @@
+import fastq from 'fastq';
+import { Cron } from './cron';
+import { Feature, FeatureValue, Invite } from '../entity';
+import { generateUUID } from '../ids';
+
+const cron: Cron = {
+  name: 'generate-invites',
+  handler: async (con, logger) => {
```
idoshamun, line 8: Instead of querying the data and then processing it on the application level, we can run one insert query that will insert data based on select statement.
Look here for reference: https://dba.stackexchange.com/a/2974
This would make this cron much more simple

src/cron/generateInvites.ts
idoshamun: The invitation system is generic, but this logic is specifically to search. Every feature will have different logic, I assume

src/cron/generateInvites.ts
sshanzel: I think the only concern here is how we're going to generate the token as it currently happens on app level. If we can generate on DB level, then I think we can do select insert.

sshanzel (commented):
Great work overall! Just had a few non-blocking comments 🚀

src/cron/generateInvites.ts
```diff
@@ -0,0 +1,55 @@
...
+          WHERE   i."userId" = f."userId"
+          AND     i."campaign" = f."feature"
+        )
+      `);
+
+    let count = 0;
+    const queueConcurrency = 1000;
+    const queue = fastq.promise(async ({ feature, userId }: Feature) => {
+      await con.getRepository(Invite).insert({
+        token: generateUUID(),
```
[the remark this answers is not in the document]
sshanzel: There should be a directive for the token property if we end up using uuid. I think it is `@PrimaryGeneratedColumn('uuid')`. This would automatically generate the UUID on DB level and would not require app level generation.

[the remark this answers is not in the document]
sshanzel: @idoshamun @rebelchris do we have any preference on how should the token be generated?

src/migration/1697544817927-Invite.ts
rebelchris: Ah missed it on top level I think, missed the regular one.
I do prefer them named myself, but it's maybe personal preference? 

EG: `@Index('IDX_content_image_used_by', ['usedByType', 'usedById'])`

src/cron/generateInvites.ts
rebelchris: Closing as discussed on slack :)

src/cron/generateInvites.ts
rebelchris: That's a good point, we generalised it indeed.
In that case we should probably rename to `generate-search-invites` as well.

src/cron/generateInvites.ts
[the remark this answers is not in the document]
rebelchris: @sshanzel Good point, I actually don't remember why we choose to do the `generateMemberToken` function over database level?
Would love to hear @idoshamun opinion would indeed make things easier if we generate on insert.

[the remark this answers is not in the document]
idoshamun: uuid is a standard thing, it doesn't matter where we generate it. So let's go with the easiest solution which is the insert query I guess

src/migration/1697544817927-Invite.ts
idoshamun: Yes named ones are better I agree

src/schema/users.ts
idoshamun: Yap, this is how we do it. We also have an enum for it, you can check the codebase for some examples. Performance wise it's better than two queries

src/migration/1697544817927-Invite.ts
rebelchris: I prefer them from debugging and logging point of view, easier to identify.
Just bit more explicit to what it indexes for.

capJavert: Yeah, from debugging point of vie we do get a error detail of which column(s) is in violation, which is the same as column would be named. 

But let's have it consistent and enforce naming indexes in guidelines/reviews then.

src/entity/Invite.ts
DragosIorgulescu: Good point, however, this entity has a composite index on both `userId` and `campaign`, does defining indexes inline accomplish the same result? I thought it generates an index just for the respective column when defined here.

src/migration/1697544817927-Invite.ts
DragosIorgulescu: I went ahead with adding a named index here 🙏  but I agree w/ @capJavert , since there are examples of existing indexes which are not named in the project, if this is the way we want to move forward we should start enforcing it everywhere

src/cron/generateInvites.ts
DragosIorgulescu: I changed this to be generated at the db level, and updated the corn job to run one query for everything. Thanks for the suggestion @idoshamun & @sshanzel

src/cron/generateInvites.ts
DragosIorgulescu: The relevant columns are indexed: https://github.com/dailydotdev/daily-api/blob/e1672479f90a7d5978610795fb05e31447c43921/src/migration/1697708253831-Invites.ts#L10-L12 so we should be good here 👍

src/schema/users.ts
```diff
@@ -1042,6 +1083,54 @@ export const resolvers: IResolvers<any, Context> = {
...
+          campaign: feature as CampaignType,
+        });
+
+      if (referrerInvite.count >= referrerInvite.limit) {
+        throw new ValidationError(SubmissionFailErrorKeys.InviteLimitReached);
+      }
+
+      await ctx.con.transaction(async (entityManager): Promise<void> => {
+        try {
+          await entityManager.getRepository(Feature).insert({
```
DragosIorgulescu, line 1115: Do we want an additional check here to make sure that only if if the feature is search we generate the invite? Or do we leave it generic?

src/entity/Settings.ts
```diff
@@ -54,6 +54,9 @@ export class Settings {
   @Column({ default: true })
   autoDismissNotifications: boolean;
 
+  @Column({ type: 'text', default: null })
```
sshanzel, line 57: Our default value should be `header`.
```suggestion
  @Column({ type: 'text', default: 'header` })
```

src/entity/Settings.ts
```diff
@@ -72,4 +75,5 @@ export const SETTINGS_DEFAULT = {
   optOutCompanion: false,
   optOutWeeklyGoal: false,
   sortingEnabled: false,
+  campaignCtaPlacement: null,
```
sshanzel, line 78: Ideally, we should have an enum to hold the values.
```suggestion
  campaignCtaPlacement: 'header',
```

sshanzel (commented):
Looks great overall! However, I think we missed the part where the update settings mutation must verify the values it can accept for the `campaignCtaPlacement` to be from a list/enum.

DragosIorgulescu:
> Looks great overall! However, I think we missed the part where the update settings mutation must verify the values it can accept for the `campaignCtaPlacement` to be from a valid list/enum

thanks @sshanzel , I added support for updating `campaignCtaPlacement` as well, however I am not sure what the purpose of this will be right now since the only possible value here is the default anyway?

src/schema/users.ts
idoshamun: Generic is cool!

idoshamun (approved):
I love how you add comments to your fixtures to understand the test case!
Perfect

src/entity/Invite.ts
```diff
@@ -0,0 +1,40 @@
...
+@Index('IDX_invite_userId_campaign', ['userId', 'campaign'], { unique: true })
+export class Invite {
+  @PrimaryColumn({ type: 'uuid' })
+  @Generated('uuid')
+  token: string;
+
+  @Column({ type: 'text' })
+  campaign: CampaignType | FeatureType;
+
+  @Column({ length: 36 })
```
sshanzel, line 26: I think we are missing the `onDelete` property here. So when the user deletes their action we will set this value to null. The value should be along the lines of `onDelete: 'SET NULL'`. Kindly verify with the docs.

src/entity/Invite.ts
```diff
@@ -0,0 +1,40 @@
...
+  userId: string;
+
+  @ManyToOne(() => User, {
+    lazy: true,
+    onDelete: 'CASCADE',
```
sshanzel, line 31: My mistake, I have mentioned `cascade` before, but the right value should be 'SET NULL' as we wouldn't want to remove this row when the referenced entity is deleted.
```suggestion
    onDelete: 'SET NULL',
```

sshanzel (approved):
Looks great! Just had one non-blocking comment but I believe we should be good to go 🚀

__tests__/workers/cdc.ts
```diff
@@ -1362,6 +1362,7 @@ describe('settings', () => {
     optOutWeeklyGoal: false,
     optOutCompanion: false,
     autoDismissNotifications: true,
+    campaignCtaPlacement: 'header',
```
rebelchris, line 1365: This place should use the enum as well

src/entity/Settings.ts
```diff
@@ -6,6 +6,10 @@ import {
   UpdateDateColumn,
 } from 'typeorm';
 
+export enum CtaPlacement {
+  Header = 'header',
+}
```
rebelchris, line 11: There should be another value here.

src/entity/Invite.ts
DragosIorgulescu: is there any reason we want to keep `invite` records when a `user` is deleted? For reference, the feature table works the same way: https://github.com/dailydotdev/daily-api/blob/5f79df84b49df10824eee083c185270c103a7166/src/entity/Feature.ts#L23-L27

rebelchris (changes requested):
Minor comments, almost there!

src/entity/Invite.ts
DragosIorgulescu: another thing to consider is that if we allow setting `userId` to `NULL`, this will mess up our unique index on `userId` & `campaign`, since the moment a second user is deleted wich has an associated invite with the same campaign as the first, we would get a conflict

src/migration/1697544817927-Invite.ts
sshanzel: Also, there is an ongoing issue with typeorm where it deletes and reintroduces the same index but with a different name from before when generating migrations (the reason why there are a lot of unrelated lines). Having it named will avoid the issue.

src/entity/Invite.ts
sshanzel: Oh no. Nevermind my comment, I thought I was at the `Feature` entity. I checked and everything should be in order now. You are right, we want this deleted when the relevant user delets their account.

rebelchris (approved):
Nice looks alright to me! 
thanks for all the hard work on this one

PR #1484, opened 2023-10-18
feat: install infra dependencies since needed during build

capJavert:
Fixes build crashing after we found hotfix for dead letter https://app.circleci.com/pipelines/github/dailydotdev/daily-api/5966/workflows/23d84678-b5ab-44d9-8afe-ce317d93cf5c/jobs/23325


omBratteng (approved):

idoshamun:
initPubSub is no longer used anywhere. I don't think we need this additional step in the ci pipeline. We install the deps only when using pulumi

omBratteng:
So @capJavert can instead remove that file and mention of it in package.json (`"cloud:init": "ts-node bin/initPubsub.ts",`) (and remove infra deps installation), that will™️ fix the issue.

capJavert:
Even better to remove some code 🚀

omBratteng:
It worked™️

PR #1490, opened 2023-10-20
feat: feed preview

capJavert:
Adding feed config and query for feed preview per [DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/648183809/WT-1888+Onboarding+V4#Feed-preview)

Confirmed with Vas that payload sent is alright.


src/schema/feeds.ts
```diff
@@ -230,6 +230,46 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      Version of the feed algorithm
+      """
+      version: Int = 1
+
+      """
+      Array of supported post types
+      """
+      supportedTypes: [String!]
+    ): PostConnection! @auth
```
capJavert, line 271: I added all the default args that other feeds have to support in future. Was thinking we might want to not allow some of them for preview, WDYT?

src/schema/feeds.ts
```diff
@@ -923,6 +963,21 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       }
       return feedResolverV1(source, args, ctx, info);
     },
+    feedPreview: (source, args: ConfiguredFeedArgs, ctx: Context, info) => {
+      if (process.env.NODE_ENV === 'development') {
+        return feedResolverV1(source, args, ctx, info);
+      }
```
capJavert, line 969: Not sure how else to make `feedResolverV2` work locally other then this, in `feed` query we send `version: 1` from frontend locally but here there is no version really.

src/schema/feeds.ts
```diff
@@ -230,6 +230,46 @@ export const typeDefs = /* GraphQL */ `
...
     ): PostConnection! @auth
 
+    """
+    Get feed preview
+    """
+    feedPreview(
+      """
+      Time the pagination started to ignore new items
+      """
+      now: DateTime
```
idoshamun, line 240: This is legacy property and not used I think

src/schema/feeds.ts
```diff
@@ -230,6 +230,46 @@ export const typeDefs = /* GraphQL */ `
...
+
+      """
+      Paginate after opaque cursor
+      """
+      after: String
+
+      """
+      Paginate first
+      """
+      first: Int
```
idoshamun, line 250: We don't support pagination in the feed preview so it's not needed

src/schema/feeds.ts
```diff
@@ -230,6 +230,46 @@ export const typeDefs = /* GraphQL */ `
...
+
+      """
+      Ranking criteria for the feed
+      """
+      ranking: Ranking = POPULARITY
+
+      """
+      Return only unread posts
+      """
+      unreadOnly: Boolean = false
```
idoshamun, line 260: Not needed

src/schema/feeds.ts
```diff
@@ -230,6 +230,46 @@ export const typeDefs = /* GraphQL */ `
...
+
+      """
+      Version of the feed algorithm
+      """
+      version: Int = 1
```
idoshamun, line 265: We need to use a different version than the regular feed. We may want to experiment differently. So we can refer to it internally and product wise as onboarding feed version. We will have to create dedicated gb flag for that. `onboarding_feed`

src/schema/feeds.ts
```diff
@@ -230,6 +230,46 @@ export const typeDefs = /* GraphQL */ `
       supportedTypes: [String!]
     ): PostConnection! @auth
 
+    """
+    Get feed preview
+    """
+    feedPreview(
```
idoshamun, line 236: We have two options here:
* Use the approach as you present here, where we use the user's tags from the db for the feed
* Accept the tags from the client and forward them downstream to the feed

src/schema/feeds.ts
capJavert: @idoshamun what would be the version, since others are incremental?

capJavert: In actual resolver we do not look at this field at all we always use new onboarding generator.

src/schema/feeds.ts
capJavert: Since the whole flow of the frontend with tags selection is to save the items to user settings it was natural to just load them. Like that we can use existing generators and thats how it was resolved in DR.

Only place where I see the need to forward tags from frontend is that if we want to allow anonymous user to preview feed. Since we don't know if we are going to do that I think this approach is ok and we can adjust down the line if needed.

src/schema/feeds.ts
DragosIorgulescu: can you elaborate on why this is needed? Feels like we should not have such high divergence in code between dev & prod

DragosIorgulescu (commented):
just had one question apart from the ones raised, otherwise looks good 👍

src/schema/feeds.ts
rebelchris: Cant you just use `feedByConfig` as the request method?
(or another existing feed request)

capJavert: Because `feedResolverV2` does not work locally. I agree with divergence that is why I wrote the comment 🤝 . For me true resolving of this issue is to make `feedResolverV2` work locally the same way `feedResolverV1` works.

rebelchris: Ah let's not introduce things just for a local env fix then.
Worst case we can mimic it locally temporary, or indeed as you said make v2 work/mimic on local envs

Ideally we actually get rid of v1 as it's only used locally.

src/schema/feeds.ts
idoshamun: We have two options: either support experimenting right now (from schema pov) or delay it. I don't have strong opinion

capJavert: Removed it for now.

src/schema/feeds.ts
capJavert: Removed it in here 6373e93497ce85df87df73e055fdee5f169efa02 but it is not gonna work locally now. I think we need to find a way to mock resolver v2 locally in near time.

rebelchris: Maybe for the time beeing use something like Mockooon or something like that?
We need to look into spooling it up as a service separate to this ticket, we also need it for auth soon.

src/schema/feeds.ts
```diff
@@ -923,6 +938,22 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       }
       return feedResolverV1(source, args, ctx, info);
     },
+    feedPreview: (
```
rebelchris, line 941: What differentiates this from `feedByConfig` now?

capJavert: configuration and params are different

rebelchris (approved):
Let's go

src/schema/feeds.ts
capJavert: Yeah, don't have the preference, whatever is easiest. Is there a task already for that or should I open it?

idoshamun (approved):
Perfect! Just check my comment before

PR #1491, opened 2023-10-23
feat: add support for collapsing/expanding source pinned posts

DragosIorgulescu:
- controlled via source member flags

ref [WT-1812](https://dailydotdev.atlassian.net/browse/WT-1812)

[WT-1812]: https://dailydotdev.atlassian.net/browse/WT-1812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/sources.ts
```diff
@@ -824,7 +850,7 @@ const updateHideFeedPostsFlag = async (
     { sourceId, userId: ctx.userId },
     {
       flags: updateFlagsStatement<SourceMember>({
-        hideFeedPosts: value,
+        [flag]: value,
       }),
     },
   );
```
capJavert, line 856: Even though code is repeated I don't see much use for this helper anymore. For example if you need to update more flags it would be of no use since we would not want to do multiple queries. 

The original helper was added by me per review feedback of duplicate code but I would just do it inline at this point.

DragosIorgulescu: I agree, what are your thoughts on having one mutation to update all flags to reduce repetition here?

capJavert: We might need different logic validation for different flags so updating them all in one mutation would mean we would also have a lot of if-else depending on which flags are updated.

rebelchris: I think all calls can only update one at the time anyway right?
Just noting as a refactor might be too much work at this point as this would work?

rebelchris (approved):
Nothing from my side.

src/schema/sources.ts
DragosIorgulescu: good points, I also think a more descriptive API is better, even though it can become more verbose. A mutation called `collapsePinnedPosts` is self-documenting vs. something like `editSourceMemberFlags`

capJavert: I would say leave `updateHideFeedPostsFlag` as it is and do the whatever you need directly in your mutations to not change parts of the code not connected to the PR and minimize changes, what do you guys think?

DragosIorgulescu: updated 🙏

capJavert (approved):
Thanks for adjustments!

rebelchris (approved):

idoshamun (approved):

PR #1492, opened 2023-10-23
feat: search squad members through query

sshanzel:
Fetch members through a query.

- New optional parameter on source members query that if it is present, we will return filtered users based on the value.
- The search query here is based on our search query for user mention.


src/schema/sources.ts
```diff
@@ -926,13 +933,23 @@ export const resolvers: IResolvers<any, Context> = {
...
-
             .addOrderBy(
               graphorm.mappings.SourceMember.fields.roleRank.select as string,
               'DESC',
             )
             .addOrderBy(`${builder.alias}."createdAt"`, 'DESC');
 
+          if (query) {
+            builder.queryBuilder = builder.queryBuilder
+              .innerJoin(User, 'u', `${builder.alias}."userId" = u.id`)
```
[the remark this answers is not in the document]
idoshamun: It's not possible at the moment but good catch!

src/schema/sources.ts
```diff
@@ -926,13 +933,23 @@ export const resolvers: IResolvers<any, Context> = {
...
+              .andWhere(
+                `(REPLACE(u.name, ' ', '') ILIKE :name OR  u.username ILIKE :name)`,
```
idoshamun, line 946: My concern is that it is not indexed properly but that's fine for now I guess.
Anyway, found a double space:

```suggestion
                `(REPLACE(u.name, ' ', '') ILIKE :name OR u.username ILIKE :name)`,
```

src/schema/sources.ts
```diff
@@ -926,13 +933,23 @@ export const resolvers: IResolvers<any, Context> = {
...
+                { name: `${query}%` },
+              );
+          }
+
+          console.log(builder.queryBuilder.getQuery());
```
idoshamun, line 951: Can be removed

sshanzel: Oh, right. Debugging 🤦

idoshamun (approved):

src/schema/sources.ts
sshanzel: Yep, I agree. I think it should be fine for now and the pool of users won't be too large as most members don't exceed 1k, yet.

PR #1494, opened 2023-10-24
fix: name can not be null

rebelchris:
name can not be null, so defaulting to use the set `userName` object.

Screenshot 2023-10-24 at 07 41 23


sshanzel (approved):
Good find! Didn't know there are still users without a name.

idoshamun (approved):

PR #1496, opened 2023-10-24
feat: add pinned posts swapping support to `updatePinPost` mutation

DragosIorgulescu:
- allow re-ordering pinned posts based on their `pinnedAt` timestamp

ref WT-1812


src/schema/posts.ts
```diff
@@ -1353,20 +1359,26 @@ export const resolvers: IResolvers<any, Context> = {
...
           SourcePermissions.PostPin,
         );
 
-        await repo.update({ id }, { pinnedAt: pinned ? new Date() : null });
+        if (pinned && swapWithPostId) {
+          const swapWithPost = await repo.findOneBy({ id: swapWithPostId });
+          args.pinnedAt = new Date(swapWithPost.pinnedAt.getTime() + 10 * 1000);
+        }
+
+        await repo.update({ id }, args);
```
capJavert, line 1381: Just generally would it make more sense to create a separate mutation for swapping? It would make the existing mutation more simple.

DragosIorgulescu: I don't have a strong preference, felt like this was minor enough to tag along inside the existing mutation, but we can extract it if you prefer

capJavert: I think it would, but I imagine that it would also take some more changes on frontend to use different mutation for that. Up to you, don't think of it as blocking. 👍

src/schema/posts.ts
```diff
@@ -1353,20 +1361,42 @@ export const resolvers: IResolvers<any, Context> = {
...
+          );
+
+          if (nextPostId) {
+            args.pinnedAt = new Date(swapPinnedTime - 10 * 1000);
+          }
+
+          if (prevPostId) {
+            args.pinnedAt = new Date(swapPinnedTime + 10 * 1000);
+          }
+        }
```
DragosIorgulescu, line 1397: I will extract this into a separate mutation, but I'd like to get some feedback first

idoshamun (approved):

src/schema/posts.ts
```diff
@@ -1371,6 +1391,58 @@ export const resolvers: IResolvers<any, Context> = {
...
+        await manager.query(
+          `
+          UPDATE post
+          SET "pinnedAt" = "pinnedAt" ${operation}
+          WHERE "pinnedAt" IS NOT NULL
+            AND "pinnedAt" ${comparison} $1
+            AND "sourceId" = $2
+        `,
+          [swapPinnedTime, post.sourceId],
+        );
```
capJavert, line 1434: Please use `QueryBuilder`

src/schema/posts.ts
```diff
@@ -1371,6 +1391,58 @@ export const resolvers: IResolvers<any, Context> = {
...
+
+        await repo.update(
+          { id },
+          {
+            pinnedAt: swapPinnedTime,
+          },
+        );
```
capJavert, line 1441: Can probably run this and above query in transaction to perform only single write.

capJavert (changes requested):

src/schema/posts.ts
DragosIorgulescu: are you referring to merging this update into the same query as above? That's a good point, but I think it would make the query a bit more complex and would not necessarily result in not having the equivalent of running 2 queries in performance.

unless you are referring to wrapping this in a transaction, in which case that is already in place 👍

src/schema/posts.ts
DragosIorgulescu: @capJavert for non-standard queries I usually prefer using as much SQL as possible vs ORM functions as it is much more readable. That said, I'm fine with the alternative if you prefer. For context, it would look something like:

```typescript
await manager.createQueryBuilder()
    .update(Post)
    .set({ 
        pinnedAt: () => `"pinnedAt" ${operation}`
    })
    .where('"pinnedAt" IS NOT NULL')
    .andWhere(`"pinnedAt" ${comparison} :swapPinnedTime`, { swapPinnedTime })
    .andWhere('"sourceId" = :sourceId', { sourceId: post.sourceId })
    .execute();
```

src/schema/posts.ts
capJavert: I didn't see it at the top, ignore me you did it perfect! 👌

src/schema/posts.ts
DragosIorgulescu: ok, sure, I'll update then. I may not be very familiar with TypeORM, but I am pretty sure that making use of parameterized arguments (`$1`, `$2`, etc.)  achieves the same result of sanitizing arguments and preventing SQL injection, so the current approach should be safe as well?

capJavert: Honestly not sure as well but it differs from ORM to ORM, that is why I think query builder is again safer and better practice.

capJavert: stuff as `${comparison}` are most dangerous, yes its not user provided now but it could become by mistake, with query builder you can even add different values conditionally with `andWhere`.

src/schema/posts.ts
```diff
@@ -1371,6 +1391,60 @@ export const resolvers: IResolvers<any, Context> = {
...
+          ? `+ interval '1 second'`
+          : `- interval '1 second'`;
+        const comparison = isNextPost ? '>=' : '<=';
+
+        await manager
+          .createQueryBuilder()
+          .update(Post)
+          .set({
+            pinnedAt: () => `"pinnedAt" ${operation}`,
+          })
```
capJavert, line 1430: You can now use if else and conditionally call `set` with different `operation` on query builder to avoid using template string in query.

src/schema/posts.ts
```diff
@@ -1371,6 +1391,60 @@ export const resolvers: IResolvers<any, Context> = {
...
+          .where('"pinnedAt" IS NOT NULL')
+          .andWhere(`"pinnedAt" ${comparison} :swapPinnedTime`, {
+            swapPinnedTime,
+          })
```
capJavert, line 1434: You can now use if else and conditionally call `andWhere` with different `comparison` on query builder to avoid using template string in query.

sshanzel (commented):
Very neat implementation. Don't have anything to add on top of the existing comments. Well done :shipit:

src/schema/posts.ts
DragosIorgulescu: Personally, I feel like this is overly verbose given that we are controlling the value of `comparison` here, it looks more like premature optimization for a possible future where this variable will be passed in and unknown 😅  

In any case, I think this should be good to go, let me know 🙏

capJavert (approved):
🚀

PR #1497, opened 2023-10-24
chore: assert that invites count are not incremented when feature exists

DragosIorgulescu:


sshanzel (approved):
Thanks for the additional test! 🚀

rebelchris (approved):

PR #1499, opened 2023-10-24
fix: add default created at date for users

rebelchris:
Changed `createdAt` to be non nullable and default to now.


sshanzel (approved):
Nice! Very timely! I would need this on the upcoming referral program 🥳

PR #1508, opened 2023-10-26
feat: use text instead of boolean cast index

capJavert:
After deployment of https://github.com/dailydotdev/daily-api/pull/1504 I noticed that onboarding tags were still slow. 

After further debugging I found out why the query is slow. It was not due to index (I mean they help) but the main culprit is how typeorm assembles the query when you use json fields, I logged it under the hood and the query is something like:
```sql
SELECT value FROM keyword WHERE flags = '{ "onboarding": true }' ORDER BY value ASC
```
With the above query we don't actually hit the index since query queries the whole flags field. Optimal query (and what I was expecting `typeorm` was doing) would be:

```sql
SELECT value FROM keyword WHERE (flags->'onboarding') = 'true' ORDER BY value ASC
```

After replacing inefficient query from `typeorm` with the raw one it actually worked and query was much faster.

After adjusting the query I tried the actual query on psql version that was same as production and noticed that index was not used and query was slow again, `EXPLAIN` was saying that it did not use the index:
image

Locally on psql v15 it used it:
image

After some additional debugging with Ido we noticed that on our prod version (psql v11) the index was working only if we used plain text index for json field instead of boolean. So in this PR we add migration to add the plain index.

On general note until we upgrade our psql version we have to make sure we use raw queries when querying per `flags` json columns.

### DIFF
--- a/__tests__/workers/__snapshots__/personalizedDigestEmail.ts.snap
+++ b/__tests__/workers/__snapshots__/personalizedDigestEmail.ts.snap
@@ -46,7 +46,7 @@ Object {
         "source_name": "B",
       },
     ],
-    "preview": "Every Monday, we'll send you five posts you haven't read. Each post was carefully picked based topics you love reading about. Let's get to it!",
+    "preview": "Every Monday, we'll send you five posts you haven't read. Each post was carefully picked based on topics you love reading about. Let's get to it!",
     "title": "Ido, your personal weekly update from daily.dev is ready",
   },
   "from": Object {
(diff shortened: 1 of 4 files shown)


capJavert:
@rebelchris Ido mentioned if you can prioritize checking the existing implementations against this same issue, from my quick check:
- `source_member` table has a multi column index for `flags` that needs to be replaced with plain one
- we have some `find` and `findBy` queries targeting JSON fields which are probably not as efficient because of how typeorm assembles them, they need to be replace with raw ones
- I think we mostly did not see this before because other tables using `flags` are not as big as `keyword` table

rebelchris (approved):
Interesting, thanks for noticing this and doling the extensive research and testing.
Added a ticket for next week :)

idoshamun (approved):

omBratteng (approved):

DragosIorgulescu (approved):
always had a love/hate relationship with ORMs (mostly hate), nice work 🚀

capJavert:
@idoshamun just let me know when you manage to remove the index we added.

idoshamun:
@capJavert I'll do it tomorrow, but that shouldn't block this

capJavert:
@idoshamun I just merge this? Was worried that index delete will hang.

idoshamun:
@capJavert closely monitor the deployment. Let me know if it goes south

PR #1510, opened 2023-10-30
feat: limit search results to 100 and min query to 2 chars

capJavert:
To prevent many search results being returned with too general of a query.

### DIFF
--- a/src/schema/tags.ts
+++ b/src/schema/tags.ts
@@ -20,6 +20,10 @@ type GQLTagResults = Pick<GQLTagSearchResults, 'hits'>;
 
 export const RECOMMENDED_TAGS_LIMIT = 5;
 
+export const MIN_SEARCH_QUERY_LENGTH = 2;
+
+export const SEARCH_TAGS_LIMIT = 100;
+
 export const typeDefs = /* GraphQL */ `
   """
   Post tag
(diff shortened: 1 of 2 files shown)


omBratteng (approved):

src/schema/tags.ts
```diff
@@ -20,6 +20,10 @@ type GQLTagResults = Pick<GQLTagSearchResults, 'hits'>;
 
 export const RECOMMENDED_TAGS_LIMIT = 5;
 
+export const MIN_SEARCH_QUERY_LENGTH = 2;
```
rebelchris, line 23: Should we not handle this from the FE to avoid even sending the query?

rebelchris (commented):
Just one remark.

src/schema/tags.ts
idoshamun: Agree

capJavert: I looked at it, I can do the adjustments but not sure it will be intuitive to not enable the query until 2 chars. We might need some label/error state.

capJavert: With enabling query after 2 chars it would work like:
- write 1 char
- nothing happens and you see all tags
- write 2. char
- then the query goes

omBratteng: Probably doesn't hurt to have it both places, in case front end doesn't actually prevent it.

capJavert: No matter the frontend I think it is valid to protect the query from manual requests with less then 2 char.

idoshamun: Yes, it wasn't to mention we don't need it here just to clear. Just to not forget adjusting the frontend

rebelchris: 100% always nice as a safety fallback, but if possible let's try and block the FE as well to avoid queries

rebelchris (approved):

src/schema/tags.ts
capJavert: https://github.com/dailydotdev/apps/pull/2286/commits/a940df152f2894835fa90e7144d1b5fbf3ff886d

PR #1511, opened 2023-10-30
feat: personalized digest launch adjustments

capJavert:
- [x] enable worker for new users to be enrolled
- [x] adjust import script to apply to all users
- [x] adjust script for variations

### DIFF
--- a/bin/assignUsersToPersonalizedDigestVariation.ts
+++ b/bin/assignUsersToPersonalizedDigestVariation.ts
@@ -1,0 +1,66 @@
+import '../src/config';
+import createOrGetConnection from '../src/db';
+import fs from 'fs';
+import { parse } from 'csv-parse';
+
+(async (): Promise<void> => {
+  const csvFilePath = process.argv[2];
+  const variationArgument = process.argv[3];
+  const splitBy = process.argv[4] || ',';
+
+  if (!csvFilePath) {
+    throw new Error('CSV file path is required');
+  }
+
+  const variation = +variationArgument;
+
+  if (Number.isNaN(variation) || variation < 1) {
+    throw new Error(
+      'variation argument is invalid, it should be positive number',
+    );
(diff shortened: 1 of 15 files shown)


rebelchris (approved):

bin/assignUsersToPersonalizedDigestVariation.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  console.log('running db query');
+
+  const con = await createOrGetConnection();
+
+  await con.transaction(async (manager) => {
+    await manager.query(
+      `
+        UPDATE user_personalized_digest
```
idoshamun, line 54: This will be one big query. I suggest splitting it into batches of 1k users roughly

capJavert: Had the batch logic, I thought I'll just chunk the CSV?

idoshamun (commented):
Regarding the worker, wouldn't it be better if we just create a digest entity together with the user instead of offloading it? I don't see a downside to doing it in one transaction. Doing it in the background may introduce potential race conditions.
Sorry for raising it so late in the process 🙏

bin/assignUsersToPersonalizedDigestVariation.ts
idoshamun: Sure that works!

capJavert: Started to split up CSV.. figured its easier to just batch so.. added batching 0f607d889a4f9e8bf51bbd7a0ff7dac21eb31aaf

capJavert: I tried 10k with initial import and it went instantly so I think it is ok default.

idoshamun (approved):

PR #1513, opened 2023-10-30
fix: id conflicts when signing up

idoshamun:
Due to legacy reasons we generate user id upon first visit and this can cause quite a lot of issues. One of them is id clashes on sign up. This PR handles the id clashes and retries several times before giving up. I also refactored a bit to make it look better 😬

It's not ready to be merged yet as I wait for an heimdall change that will be deployed tomorrow hopefully.

### DIFF
--- a/src/entity/User.ts
+++ b/src/entity/User.ts
@@ -251,6 +251,7 @@ const handleInsertError = async (
           reason: UserFailErrorKeys.UsernameEmailExists,
         };
       }
+
       if (error.message.indexOf('PK_') > -1) {
         if (shouldRetry) {
           data.id = await generateTrackingId(req, 'user creation');
@@ -258,6 +259,7 @@ const handleInsertError = async (
         }
         return { status: 'failed', reason: UserFailErrorKeys.UserExists };
       }
+
       // If it's not username or primary key than it's twitter and github.
       if (shouldRetry) {
         if (error.message.indexOf('users_twitter_unique') > -1) {
--- a/src/entity/User.ts
+++ b/src/entity/User.ts
@@ -2,6 +2,7 @@ import {
   AfterLoad,
   Column,
   DataSource,
+  DeepPartial,
   Entity,
   EntityManager,
   Index,
@@ -11,7 +12,7 @@ import {
 } from 'typeorm';
 import { Post } from './posts';
 import { DevCard } from './DevCard';
-import { FastifyBaseLogger } from 'fastify';
+import { FastifyBaseLogger, FastifyRequest } from 'fastify';
 import {
   TypeOrmError,
   UpdateUserFailErrorKeys,
@@ -27,6 +28,7 @@ import {
   validateRegex,
   ValidateRegex,
 } from '../common/object';
+import { generateTrackingId } from '../ids';
 
 @Entity()
 export class User {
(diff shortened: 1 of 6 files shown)


src/entity/User.ts
```diff
@@ -247,101 +223,118 @@ export const updateUserEmail = async (
...
+      }
+      // If it's not username or primary key than it's twitter and github.
+      if (shouldRetry) {
+        if (error.message.indexOf('users_twitter_unique') > -1) {
+          data.twitter = null;
+        } else if (error.message.indexOf('users_github_unique') > -1) {
+          data.github = null;
+        }
+        return safeInsertUser(req, con, data, maxIterations, iteration + 1);
+      }
```
omBratteng, line 269: The other two if statements does possibly return failed reason, can this one do it as well?

omBratteng, line 269: It's just a personal preference, but I do like a little bit of air between statements like this, to make it faster and easier to differentiate the different logic blocks

```suggestion
      if (error.message.indexOf('users_username_unique') > -1) {
        return {
          status: 'failed',
          reason: UserFailErrorKeys.UsernameEmailExists,
        };
      }

      if (error.message.indexOf('PK_') > -1) {
        if (shouldRetry) {
          data.id = await generateTrackingId(req, 'user creation');
          return safeInsertUser(req, con, data, maxIterations, iteration + 1);
        }
        return { status: 'failed', reason: UserFailErrorKeys.UserExists };
      }

      // If it's not username or primary key than it's twitter and github.
      if (shouldRetry) {
        if (error.message.indexOf('users_twitter_unique') > -1) {
          data.twitter = null;
        } else if (error.message.indexOf('users_github_unique') > -1) {
          data.github = null;
        }
        return safeInsertUser(req, con, data, maxIterations, iteration + 1);
      }
```

src/ids.ts
```diff
@@ -8,13 +8,16 @@ export const generateLongId = customAlphabet(alphabet, 21);
...
+  req: FastifyRequest,
+  origin: string,
+): Promise<string> => {
   if (req.meter) {
     req.meter
       .createCounter('generate_tracking_id', {
         description: 'How many times a tracking id was generated',
       })
-      .add(1);
+      .add(1, { origin });
```
omBratteng, line 20: Good addition

src/entity/User.ts
idoshamun: In this case it will be an unexpected error so I prefer to throw an exception. I hope it makes sense

src/entity/User.ts
```diff
@@ -247,101 +223,120 @@ export const updateUserEmail = async (
...
+          data.id = await generateTrackingId(req, 'user creation');
+          return safeInsertUser(req, con, data, maxIterations, iteration + 1);
+        }
+        return { status: 'failed', reason: UserFailErrorKeys.UserExists };
+      }
+
+      // If it's not username or primary key than it's twitter and github.
+      if (shouldRetry) {
+        if (error.message.indexOf('users_twitter_unique') > -1) {
+          data.twitter = null;
```
rebelchris, line 266: Are we not trying to solve this now in the wrong way then?
It means the user is already existing basically right, so would it not be better to feedback to the FE this is the case?

src/entity/User.ts
```diff
@@ -247,101 +223,120 @@ export const updateUserEmail = async (
...
   if (!checkRequiredFields(data)) {
-    logger.info({ data }, 'missing fields when adding new user');
+    req.log.info({ data }, 'missing fields when adding new user');
     return { status: 'failed', reason: UserFailErrorKeys.MissingFields };
   }
 
-  return con.transaction(async (entityManager) => {
-    const isEmailAvailable = await checkEmail(entityManager, data.email);
-    if (!isEmailAvailable) {
+  const isEmailAvailable = await checkEmail(con, data.email);
```
rebelchris, line 304: Should these not run in a transaction?

src/entity/User.ts
rebelchris: Ah I see we already had this 🤔

rebelchris (approved):
Small comments but nothing blocking

src/entity/User.ts
idoshamun: Yap it was this way all along just refactored a bit. The main fix is the id clashes

src/entity/User.ts
idoshamun: You're right I'll put it back

idoshamun: Thinking about it again, there's no need for a transaction here as it doesn't protect us from anything. It will be an overkill here I guess. Also if the query fails it will rollback the transaction and will not allow me to retry

PR #1515, opened 2023-10-31
feat: move digest subscription from worker to inline during user creation

capJavert:
https://github.com/dailydotdev/daily-api/pull/1511#pullrequestreview-1704204945

> Regarding the worker, wouldn't it be better if we just create a digest entity together with the user instead of offloading it? I don't see a downside to doing it in one transaction. Doing it in the background may introduce potential race conditions.


bin/assignUsersToPersonalizedDigestVariation.ts
```diff
@@ -6,8 +6,8 @@ import { parse } from 'csv-parse';
 (async (): Promise<void> => {
   const csvFilePath = process.argv[2];
   const variationArgument = process.argv[3];
-  const batch = +process.argv[4] || 10000;
-  const splitBy = process.argv[4] || ',';
+  const batch = +process.argv[4] || 100;
+  const splitBy = process.argv[5] || ',';
```
capJavert, line 10: Adjustments I made to script importing variations:
- query adjustment
- argv adjustment (typo)
- batch smaller (when more rows are in the table `IN` queries are much more expensive)

src/routes/private.ts
```diff
@@ -33,6 +34,15 @@ export default async function (fastify: FastifyInstance): Promise<void> {
 
     const body = { ...rest, referralId, referralOrigin };
     const operationResult = await addNewUser(con, body, req.log);
+
+    if (operationResult.status === 'ok') {
```
idoshamun, line 38: Ideally this new insert should be in one transaction with the adding the user. Since `addNewUser` is now a complex process that is not capable of being in a transaction this process may leave us with zombie users without a digest because we don't handle the error. Another approach would be a database trigger on insert. It is more failsafe. @capJavert wdyt? Sorry for being annoying

capJavert: @idoshamun I did not do it in a transaction because the reference integrity was failing due to user not existing yet in the DB even though it does exist in transaction.

Trigger is a good idea but sounds like a hassle if we wanna add any kind of logic on when we subscribe users to digest.

WDYT?

capJavert: With worker we did have automatic retry which was good.

capJavert: When talking about the timezone I am not sure we actually need the timezone on digest then. If we just update that based on `user` table `timezone` column what is the point of it being in `user_personal_digest` table?

capJavert: We would have to do it both ways I guess?

idoshamun: I agree it may be redundant. As a user, you shouldn't manage two different timezones. I should have been more explicit about it during the DR process, I felt it's a bit off

capJavert: Do you want to do both things in this PR? Removing timezone would be a bigger change to handle everywhere.

idoshamun: Your call :)

capJavert: I'd like to solve timezones down the line.

src/migration/1699284661116-UserPersonalizedDigestSubscribeTrigger.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      queryRunner.query(`
+        CREATE OR REPLACE FUNCTION user_personalized_digest_subscribe()
+          RETURNS TRIGGER
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          INSERT INTO user_personalized_digest ("userId", "preferredTimezone", "preferredHour", "preferredDay") VALUES (NEW.id, COALESCE(NULLIF(NEW.timezone, ''), 'Etc/UTC'), 8, 3);
```
idoshamun, line 14: We can add `on conflict do nothing` just in case for testing. I see you have to clean the table every time

capJavert: Yeah, it will not help in tests since its a different issue of being created before in fixtures, but good to add I think nevertheless.

PR #1517, opened 2023-10-31
refactor: some analytics adjustments for Dave

idoshamun:
Dave raised some issues and I'm trying to fix them. Hopefully it will indeed fix the issue

### DIFF
--- a/src/integrations/analytics.ts
+++ b/src/integrations/analytics.ts
@@ -1,4 +1,4 @@
-import { generateLongId } from '../ids';
+import { generateUUID } from '../ids';
 import { retryFetch } from './retry';
 
 const generateEventId = (now) => {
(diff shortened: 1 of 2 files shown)


capJavert (commented):
Why are we changing the format from string id to uuid?

src/integrations/analytics.ts
```diff
@@ -12,8 +12,8 @@ export async function sendAnalyticsEvent<
 >(events: T[]): Promise<void> {
   const now = new Date();
   const [visit_id, session_id] = await Promise.all([
-    generateLongId(),
-    generateLongId(),
+    generateUUID(),
+    generateUUID(),
```
capJavert, line 16: Why are we changing the format from string id to uuid?

idoshamun: Reduce the chances for clashes, UUID should be more failsafe in this case and it's not user facing

PR #1521, opened 2023-11-01
fix: add new alerts columns

rebelchris:
Added new alerts columns as described in [DR](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/659161109/Referral+program+and+profile+UX+improvements#Persistent-data-for-referral-reminder).
had to adjust from Date as it's not supported in JSONB, decided to go for bigint to avoid y2k issues down the line (2038)

WT-1921 #done

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -65,7 +65,7 @@ export type Experimentation = {
 
 export type BaseBoot = {
   visit: { visitId: string; sessionId: string };
-  alerts: Omit<Alerts, 'userId'>;
+  alerts: Omit<Alerts, 'userId' | 'flags'>;
   settings: Omit<Settings, 'userId' | 'updatedAt'>;
   notifications: { unreadNotificationsCount: number };
   squads: BootSquadSource[];
--- a/__tests__/alerts.ts
+++ b/__tests__/alerts.ts
@@ -87,24 +87,6 @@ describe('query userAlerts', () => {
   });
 });
 
-it('should populate database', async () => {
-  const start = new Date('02/05/2020');
-  let loop = new Date(start);
-  for (let i = 0; i < 100000; i++) {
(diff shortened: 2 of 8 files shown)


src/routes/boot.ts
```diff
@@ -65,7 +65,7 @@ export type Experimentation = {
 
 export type BaseBoot = {
   visit: { visitId: string; sessionId: string };
-  alerts: Omit<Alerts, 'userId'>;
+  alerts: Omit<Alerts, 'userId' | 'flags'>;
```
idoshamun, line 68: We have to make sure we don't select this column also

rebelchris: Lee made it a [seperate](https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11?selectedIssue=WT-1924) ticket, if thats ok with you?

idoshamun: The ticket refers to the mutation to update alerts. I talk about the query that fetches alerts from the database for boot. If we introduce it in a different ticket it means there will be a period of time that users get this information even though they shouldn't

rebelchris: @idoshamun I added a function for now, let me know what you think.

idoshamun (approved):
Awesome!

rebelchris:
@idoshamun @capJavert Added the function now as a generic one we can reuse.

Queries should happen like this:
```
SELECT * FROM alerts WHERE dateCastIndex(flags, 'lastReferralReminder') >= '2020-04-20'
```

And we can easily add new ones:
```
CREATE INDEX on alerts ((dateCastIndex(flags, 'lastReferralReminder')))
```

Result using the index fn:
Screenshot 2023-11-02 at 12 56 48

capJavert (approved):
Looks great 🚀

idoshamun (approved):

PR #1522, opened 2023-11-02
fix: add update referral mutation

rebelchris:
Added the new referral mutation flow to support generic update that doesn't follow the regular flow.
(This is to ensure we don't accidentally wrongly set anything)

WT-1924 #done

### DIFF
--- a/src/entity/Alerts.ts
+++ b/src/entity/Alerts.ts
@@ -38,7 +38,7 @@ export class Alerts {
   lastBanner: Date | null;
 
   // Should not be exposed to the client
-  @Column({ type: 'jsonb', default: {}, select: false })
+  @Column({ type: 'jsonb', default: {} })
   flags: AlertsFlags = {};
 
   changelog?: boolean;
--- a/src/schema/alerts.ts
+++ b/src/schema/alerts.ts
@@ -151,7 +151,7 @@ export const updateAlerts = async (
       await repo.save({
         userId,
         ...data,
-        flags: { ...flags, lastReferralReminder: new Date() },
+        flags: flags ? flags : undefined,
       }),
     );
   }
@@ -160,7 +160,9 @@ export const updateAlerts = async (
     await repo.save({
       ...alerts,
       ...data,
-      flags: { ...alerts.flags, lastReferralReminder: new Date() },
+      flags: flags
+        ? { ...alerts.flags, lastReferralReminder: new Date() }
+        : undefined,
     }),
   );
 };
--- a/src/schema/alerts.ts
+++ b/src/schema/alerts.ts
@@ -1,10 +1,17 @@
-import { Alerts, ALERTS_DEFAULT, UserActionType } from '../entity';
+import {
+  Alerts,
+  ALERTS_DEFAULT,
+  SourceMember,
+  UserActionType,
+} from '../entity';
 
 import { IResolvers } from '@graphql-tools/utils';
 import { traceResolvers } from './trace';
 import { Context } from '../Context';
 import { DataSource } from 'typeorm';
 import { insertOrIgnoreAction } from './actions';
+import { GQLEmptyResponse } from './common';
+import { updateFlagsStatement } from '../common';
 
 interface GQLAlerts {
   filter: boolean;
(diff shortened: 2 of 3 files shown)


src/entity/Alerts.ts
```diff
@@ -38,7 +38,7 @@ export class Alerts {
   lastBanner: Date | null;
 
   // Should not be exposed to the client
```
capJavert, line 40: Comment here is not longer needed?

src/schema/alerts.ts
```diff
@@ -167,6 +173,19 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+    updateLastReferralReminder: async (
+      _,
+      __,
+      ctx,
+    ): Promise<GQLEmptyResponse> => {
+      await updateAlerts<Partial<Alerts>>(ctx.con, ctx.userId, {
+        showGenericReferral: false,
+        flags: {
+          lastReferralReminder: new Date(),
+        },
```
capJavert, line 185: We need to make sure to use `updateFlagsStatement` because like this if `alert.flags` had any values already set it would overwrite.

src/entity/Alerts.ts
rebelchris: Hmm still valid I think, more a extra note :)

src/schema/alerts.ts
rebelchris: Ah yes, I was playing with some changes, must have accidentally removed it.
Let me revert back, had some issues with save usage

src/entity/Alerts.ts
capJavert: Thought it was added because of `select` option?

capJavert: Our flags are private by default we use `*Flags` and `*FlagsPublic` to expose them through graphql usually.

capJavert: Ah I looked closely, `getAlerts` is used in `boot` which escapes graphql.

rebelchris:
`save` method doesn't allow our return from `updateFlagsStatement` so did a manual overwrite here as we have access to the old flags basically.

capJavert (approved):

src/schema/alerts.ts
```diff
@@ -146,10 +147,22 @@ export const updateAlerts = async <T = GQLUpdateAlertsInput>(
   }
 
   if (!alerts) {
-    return saveReturnAlerts(await repo.save({ userId, ...data }));
+    return saveReturnAlerts(
+      await repo.save({
+        userId,
+        ...data,
+        flags: { ...flags, lastReferralReminder: new Date() },
```
capJavert: Is this now duplicate, we already pass it as `flags` from `updateLastReferralReminder`?

src/schema/alerts.ts
```diff
@@ -146,10 +147,22 @@ export const updateAlerts = async <T = GQLUpdateAlertsInput>(
...
+      }),
+    );
   }
 
-  return saveReturnAlerts(await repo.save({ ...alerts, ...data }));
+  return saveReturnAlerts(
+    await repo.save({
+      ...alerts,
+      ...data,
+      flags: { ...alerts.flags, lastReferralReminder: new Date() },
```
capJavert: Same as https://github.com/dailydotdev/daily-api/pull/1522/commits/dc9766ecf6d2252d9ae03162a304d9439ff1a8fe#r1380180634

src/schema/alerts.ts
rebelchris: Changed it in last commit :)

src/schema/alerts.ts
```diff
@@ -140,10 +147,22 @@ export const updateAlerts = async (
   }
 
   if (!alerts) {
-    return saveReturnAlerts(await repo.save({ userId, ...data }));
+    return saveReturnAlerts(
```
idoshamun, line 150: General question, why do we keep saving the full object instead of updating only the relevant field?

idoshamun (approved):
LGTM! Just a generic question

src/schema/alerts.ts
rebelchris: Data is only one part of the alerts, not the whole object.
Same with flags, it's one or multiple flags.

PR #1527, opened 2023-11-07
feat: digest de-duplication

capJavert:
Implementing batch id and deduplication logic from [DR](https://dailydotdev.atlassian.net/l/cp/afS2U1Lo)

Part with testing the load issue will be done in separate PRs.

### DIFF
--- a/src/cron/personalizedDigest.ts
+++ b/src/cron/personalizedDigest.ts
@@ -35,6 +35,7 @@ const cron: Cron = {
           logger,
           personalizedDigest,
           timestamp,
+          emailBatchId,
         );
 
         digestCount += 1;
--- a/src/cron/personalizedDigest.ts
+++ b/src/cron/personalizedDigest.ts
@@ -1,11 +1,22 @@
 import fastq from 'fastq';
-import { notifyGeneratePersonalizedDigest } from '../common';
+import {
+  createEmailBatchId,
+  notifyGeneratePersonalizedDigest,
+} from '../common';
 import { UserPersonalizedDigest } from '../entity/UserPersonalizedDigest';
 import { Cron } from './cron';
 
 const cron: Cron = {
   name: 'personalized-digest',
   handler: async (con, logger) => {
+    const emailBatchId = await createEmailBatchId();
+
+    if (!emailBatchId) {
+      throw new Error('failed to create email batch id');
+    }
+
+    logger.info({ emailBatchId }, 'starting personalized digest send');
+
     const nextPreferredDay = (new Date().getDay() + 1) % 7;
     const personalizedDigestQuery = con
       .createQueryBuilder()
@@ -41,7 +52,7 @@ const cron: Cron = {
     await new Promise((resolve, reject) => {
       personalizedDigestStream.on('error', (error) => {
         logger.error(
-          { err: error },
+          { err: error, emailBatchId },
           'streaming personalized digest subscriptions failed',
         );
 
@@ -51,7 +62,7 @@ const cron: Cron = {
     });
     await notifyQueue.drained();
 
-    logger.info({ digestCount }, 'personalized digest sent');
+    logger.info({ digestCount, emailBatchId }, 'personalized digest sent');
   },
 };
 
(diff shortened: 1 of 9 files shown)


src/workers/personalizedDigestEmail.ts
```diff
@@ -213,6 +213,31 @@ const worker: Worker = {
...
     };
 
+    const personalizedDigestWithLastSendDate = await con
+      .getRepository(UserPersonalizedDigest)
+      .findOne({
+        select: ['lastSendDate'],
+        where: {
+          userId: personalizedDigest.userId,
+        },
+      });
```
capJavert, line 223: This might not be enough because workers run in parallel and there could be a case where duplicate is sent so I think we will need some semafor mechanism to fully dedup.

src/cron/personalizedDigest.ts
```diff
@@ -1,11 +1,22 @@
...
+  createEmailBatchId,
+  notifyGeneratePersonalizedDigest,
+} from '../common';
 import { UserPersonalizedDigest } from '../entity/UserPersonalizedDigest';
 import { Cron } from './cron';
 
 const cron: Cron = {
   name: 'personalized-digest',
   handler: async (con, logger) => {
+    const emailBatchId = await createEmailBatchId();
```
rebelchris, line 12: Maybe I'm overlooking this, but where are you using the batchID besides the logs?
It should send through to sendgrid

src/workers/personalizedDigestEmail.ts
rebelchris: Hmm, probably very edge-casey, not sure if we should worry about that, but maybe Ido thinks differently about it.

rebelchris (changes requested):
Nice work, just missing the usage of the batch ID i think.

src/cron/personalizedDigest.ts
capJavert: Yeah, still work in progress. Today I am going to test that sendgrid call works.

capJavert: Implemeneted here 7d286c82cbf2bd0b0fac6cbf53782c6391ec17e5

src/cron/personalizedDigest.ts
```diff
@@ -1,11 +1,22 @@
...
+
+    if (!emailBatchId) {
+      throw new Error('failed to create email batch id');
+    }
```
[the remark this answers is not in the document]
idoshamun: Yes it will. It's better to fail as you did :)

src/workers/personalizedDigestEmail.ts
idoshamun: Achieving exactly once delivery is hard, this implementation already minimizes the chances for dups significantly. Let's leave it as it is now

src/workers/personalizedDigestEmail.ts
```diff
@@ -210,10 +216,36 @@ const worker: Worker = {
...
+
+    if (
+      personalizedDigestWithLastSendDate?.lastSendDate &&
+      isSameWeek(currentDate, personalizedDigestWithLastSendDate.lastSendDate)
```
idoshamun, line 234: Can we use `isSameDay` instead? Pretty sure we would need to support daily digests for some users and this will be easily missed

idoshamun (approved):
One minor comment, other than that ready to go!

rebelchris (approved):
Neat!

idoshamun:
@capJavert I had another idea for the dedup, instead of setting the date at the end, we can set it right after we check for dedup and only if something fails revert it. This way we shorten the window even further. And getting two emails is worse than zero in my opinion. But up to you

capJavert:
> @capJavert I had another idea for the dedup, instead of setting the date at the end, we can set it right after we check for dedup and only if something fails revert it. This way we shorten the window even further. And getting two emails is worse than zero in my opinion. But up to you

yeah, actually had it that way initially, agreed

capJavert:
@idoshamun adjusted send date logic here 879b9b758e58ba9b23cbfc0d41ad9d019f549a63

idoshamun (approved):

PR #1528, opened 2023-11-08
feat: cron to check referral reminder

sshanzel:
Introduced a cron job that will perform the update of whether to display the generic referral popup or not.

There is one caveat though, I noticed that we don't automatically insert an `Alerts` entity whenever a new user is created, for good reasons. The basis of the `UPDATE` query is the `Alerts` table, meaning, there could be users having no row which will not be fetched on the query introduced here.

I thought of doing it in a single query but realized it to be not possible since we will have to do an insert query at the same time as the update. The proposal is, to add a new block of code that will check those users without Alerts entity but has been in the for two weeks now. Thoughts?


src/cron/checkReferralReminder.ts
```diff
@@ -0,0 +1,25 @@
...
+    const { affected } = await con
+      .getRepository(Alerts)
+      .createQueryBuilder()
+      .update()
+      .where({ showGenericReferral: false })
+      .andWhere(
+        `(
+              dateCastIndex(flags, 'lastReferralReminder') <= NOW() - INTERVAL '6 months'
+          OR  (dateCastIndex(flags, 'lastReferralReminder') IS NULL AND (SELECT u."createdAt" FROM public.user AS u WHERE u.id = "userId") <= NOW() - INTERVAL '2 weeks')
+        )`,
```
sshanzel, line 17: Due to the nature of the complexity to fetch the data, have resorted to manual query and subquery for the user as it is [not available through typeorm](https://github.com/typeorm/typeorm/issues/564#issuecomment-310331468)

idoshamun:
@sshanzel I'd rather add a trigger to add an alert for every new user. This way we assume it always exists. Wdyt?

.infra/crons.ts
```diff
@@ -50,6 +50,10 @@ export const crons: Cron[] = [
   //   name: 'generate-search-invites',
   //   schedule: '15 1 * * *',
   // },
+  {
+    name: 'generic-referral-reminder',
+    schedule: '12 1 * * *',
```
idoshamun, line 55: Let's use a different hour. I see we really love 1am :P

sshanzel:
> @sshanzel I'd rather add a trigger to add an alert for every new user. This way we assume it always exists. Wdyt?

Yes, we can do that as well.

__tests__/cron/checkReferralReminder.ts
```diff
@@ -0,0 +1,116 @@
...
+beforeAll(async () => {
+  con = await createOrGetConnection();
+});
+
+beforeEach(async () => {
+  await saveFixtures(con, User, usersFixture);
+  jest.resetAllMocks();
+});
+
+const expectAffectedRows = async (count: number) => {
```
idoshamun, line 21: I wouldn't recommend basing our tests on the logs. It's better to check the db directly

src/cron/checkReferralReminder.ts
rebelchris: Not sure this supports, but could we not do:
- where reminder is null AND createdAt 2 weeks ago
- OR where reminder is 6 months ago?

Maybe it's the subquery efficiency.
If not this is fine from my side to be honest.

rebelchris:
@sshanzel If you add this trigger to ensure alerts exists, let's also add a test case for that scenario please :)

src/cron/checkReferralReminder.ts
sshanzel: Previously, we can't since joining was not possible but since we will introduce a trigger on user creation for alerts, we can probably use the `createdAt` from Alerts entity.

sshanzel:
> @sshanzel If you add this trigger to ensure alerts exists, let's also add a test case for that scenario please :)

Definitely.

Also, I checked the number of users and alert entities we have, they are not equal, I assume we are going to run a migration/script to insert the missing pieces?

idoshamun:
> Also, I checked the number of users and alert entities we have, they are not equal, I assume we are going to run a migration/script to insert the missing pieces?

Yes

__tests__/cron/checkReferralReminder.ts
sshanzel: I see. The part I needed this was to check whether something was not updated, which pulling the value from DB would be the same from start and when finished.

idoshamun: So you can check that nothing changed if that's what you want

sshanzel: Yep, we can do that. Just wanted something more explicit initially 😅

src/cron/checkReferralReminder.ts
sshanzel: After checking, it seems there is no `createdAt` in Alerts table, meaning, we will have to stick with what's introduced here.

idoshamun:
@sshanzel the idea was to utilize database triggers (hence I used the term triggers, sorry for the confusion). Using background worker for this purpose may introduce race conditions and edge cases. Can you please adjust it?

bin/insertMissingAlerts.ts
```diff
@@ -0,0 +1,19 @@
...
+  const users = await con
+    .getRepository(User)
+    .createQueryBuilder()
+    .select('id')
+    .where('id NOT IN (SELECT "userId" FROM public.alerts)')
+    .execute();
+
+  await con
+    .getRepository(Alerts)
+    .save(users.map(({ id }) => ({ ...ALERTS_DEFAULT, userId: id })));
```
idoshamun, line 16: This can be one query, insert from select. We don't need to fetch the users and then create alerts. It will introduce a massive overhead for lots of records

sshanzel:
> @sshanzel the idea was to utilize database triggers (hence I used the term triggers, sorry for the confusion). Using background worker for this purpose may introduce race conditions and edge cases. Can you please adjust it?

Oh, right. I totally missed it. Yes, we've been using triggers recently. Let me revert to that.

__tests__/triggers/userCreateTrigger.ts
```diff
@@ -0,0 +1,27 @@
...
+
+beforeAll(async () => {
+  con = await createOrGetConnection();
+});
+
+beforeEach(async () => {
+  jest.resetAllMocks();
+});
+
+describe('user_create_trigger after insert trigger', () => {
```
idoshamun, line 17: I think we have already some tests on a user trigger, try to check where they are and move it to the right place but I'm not sure. Anyway not blocking

src/migration/1699508350425-UserAlertsTrigger.ts
```diff
@@ -0,0 +1,27 @@
...
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          INSERT INTO public.alerts ("userId") VALUES (NEW."id");
+          RETURN NEW;
+        END;
+        $$
+      `)
+    queryRunner.query('CREATE TRIGGER user_create_trigger AFTER INSERT ON public."user" FOR EACH ROW EXECUTE PROCEDURE insert_user_alerts()')
```
idoshamun, line 19: Let's give this trigger a more explicit name. We may add more triggers in the future.

idoshamun (approved):
Non blocking comments but please address them before you merge :)

src/migration/1699508350425-UserAlertsTrigger.ts
sshanzel: Right. Let me update the naming

__tests__/triggers/userCreateTrigger.ts
sshanzel: I tried to search tests for triggers but only found one suite for the post keyword. I will move them together if I find the similar tests for triggers.

idoshamun: Maybe a better place would be the users tests. For your consideration

sshanzel: That makes sense. There are not a lot of triggers, hosting them together with other suites should not be confusing. Will move this one.

__tests__/users.ts
```diff
@@ -2526,3 +2528,15 @@ describe('mutation acceptFeatureInvite', () => {
...
+
+describe('user_create_alerts_trigger after insert trigger', () => {
+  it('should insert default alerts', async () => {
+    await con.getRepository(User).delete({});
+    const repo = con.getRepository(Alerts);
+    await repo.delete({});
+    const [user] = usersFixture;
+    await saveFixtures(con, User, [user]);
+    const alerts = await repo.findOneBy({ userId: user.id });
+    expect(alerts).toBeTruthy();
```
rebelchris, line 2540: You could even say equal ALERT_DEFAULT, or abstract them inline to validate defaults don't change? (non blocking)

__tests__/workers/insertMissingAlerts.ts
```diff
@@ -0,0 +1,42 @@
...
+});
+
+beforeEach(async () => {
+  jest.clearAllMocks();
+  await saveFixtures(con, User, usersFixture);
+});
+
+afterAll(() => disposeGraphQLTesting(state));
+
+describe('insert missing alerts script', () => {
```
rebelchris, line 29: It's not really a worker, and ideally it's a once-off.
Not urgent, but technically might be better to remove this as it will just add unnecessary time to tests.

bin/insertMissingAlerts.ts
```diff
@@ -0,0 +1,19 @@
...
+  const usersQuery = con
+    .getRepository(User)
+    .createQueryBuilder()
+    .select('id')
+    .where('id NOT IN (SELECT "userId" FROM public.alerts)')
+    .getQuery();
+
+  await con
+    .getRepository(Alerts)
+    .query(`INSERT INTO public.alerts ("userId") ${usersQuery}`);
```
rebelchris, line 16: The default set in SQL are matching our `ALERTS_DEFAULTS`?

rebelchris (approved):
Nice, some minor non blocking comments from me as well

__tests__/users.ts
sshanzel: I checked our ALERT_DEFAULT, the `lastChangelog` has some default value that is not the same with DB defaults. Not sure why though.

bin/insertMissingAlerts.ts
sshanzel: I think it is better to stick with DB defaults (as we also set the default values there). Especially, our trigger is based on DB default values.

__tests__/workers/insertMissingAlerts.ts
sshanzel: Yeah, we can do that.

__tests__/users.ts
idoshamun: So we need to adjust the table default. Chris is right, we better fix it

bin/insertMissingAlerts.ts
rebelchris: I understand, I don't want to change it, mainly want to avoid there being discrepancies between `ALERT_DEFAULTS` and the tables defaults as we are now changing from 1 to the other.
Think the above one solves it, so will resolve this.

__tests__/users.ts
rebelchris: Cool let's make them match, currently users without alerts would get this ALERT_DEFAULTS returned, so we need to make them match indeed

sshanzel: Yeah, I agree with that, not sure though which is correct. The new date or the specific value.

Screenshot 2023-11-09 at 7 44 15 PM

rebelchris: @sshanzel vaguely recall if it wasn't in the past it would cause side re-renders.
But maybe try it locally to see if that happens. *(maybe since now it get's created on user creation it doesn't matter anymore?)

sshanzel: Don't see anything weird on registration. I will update the JS defaults to new Date. Like you mentioned, it might not matter now, also that we require registration for users.

sshanzel: I am still thinking about the previous implementation, maybe we should store the value in the cache for anonymous users (viewing post page) to make it work similarly to how it was before and for existing users. Though not sure if the numbers are significant and if it would be worth it.

sshanzel: From the previous implementation, since we have a static value from the ALERTS_DEFAULT, the value never changes, unlike now, when an anonymous user visits a post page, every visit would then make the `lastBanner` have a new value unlike before. After further digging, the value from alerts.lastBanner is being compared to some value in the cache, which it checks whether the last seen banner (different from `alerts.lastBanner`) is less than to your alerts value, by updating it every time, for anonymous users, the relevant property might return false all the time.

.infra/crons.ts
```diff
@@ -50,6 +50,10 @@ export const crons: Cron[] = [
     name: 'generate-search-invites',
     schedule: '15 1 * * *',
   },
+  // {
+  //   name: 'generic-referral-reminder',
+  //   schedule: '12 3 * * *',
+  // },
```
sshanzel, line 56: Disabled the check for displaying the popup

src/entity/Alerts.ts
```diff
@@ -52,7 +52,7 @@ export const ALERTS_DEFAULT: Omit<Alerts, 'userId' | 'flags'> = {
   myFeed: null,
   companionHelper: true,
   lastChangelog: new Date(),
-  lastBanner: new Date('2023-02-05 12:00:00'), // Has to be in the past to accommodate new users
+  lastBanner: new Date(),
```
idoshamun, line 55: Changing this would cause new users to not see an active banner, wouldn't it?

sshanzel: For anonymous users (viewing post page), I think so, yeah. That's why I was also thinking if we should store some cache value related to it in Redis. https://github.com/dailydotdev/daily-api/pull/1528#discussion_r1388003150

sshanzel: Nevermind, I checked the boot, they always had `banner` property to false, anyway. So I think it would not matter anymore since it is only to be concerned for anonymous users.

rebelchris (approved):

PR #1530, opened 2023-11-09
feat: referred users query

sshanzel:
Fetch all the users that have been referred by the login user.

From the entity itself, the `createdAt` is not nullable, hence, updated the types.

Made the `createdAt` to be always required as it is part of the query for fetching the list which does not necessarily need to be displayed in FE.

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -37,7 +37,7 @@ import {
 } from '../src/entity';
 import { sourcesFixture } from './fixture/source';
 import { getTimezonedStartOfISOWeek } from '../src/common';
-import { DataSource } from 'typeorm';
+import { DataSource, In } from 'typeorm';
 import createOrGetConnection from '../src/db';
 import request from 'supertest';
 import { FastifyInstance } from 'fastify';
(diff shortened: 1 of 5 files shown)


rebelchris (approved):
Looks good to me

idoshamun:
I couldn't find the DR, but do we need the users themselves or just the count?

rebelchris:
@idoshamun We need [list](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/659161109/Referral+program+and+profile+UX+improvements#Query-to-fetch-referred-users) of users.
Screenshot 2023-11-09 at 13 44 49

idoshamun (approved):

__tests__/users.ts
```diff
@@ -310,6 +310,53 @@ describe('query userStats', () => {
...
+          image
+        }
+      }
+    }
+  }`;
+
+  it('should not allow unauthenticated users', () =>
+    testQueryErrorCode(client, { query: QUERY }, 'UNAUTHENTICATED'));
+
+  it('should return users that have been referred by the logged in user', async () => {
```
idoshamun, line 331: Seems like we can merge this and the test below

idoshamun:
@rebelchris thanks!

PR #1532, opened 2023-11-13
refactor: insert alerts through stream

sshanzel:
Refactor the insertion through streams to have better management in terms of load.

### DIFF
--- a/bin/insertMissingAlerts.ts
+++ b/bin/insertMissingAlerts.ts
@@ -28,6 +28,7 @@ import { Alerts, User } from '../src/entity';
     stream.on('end', () => {
       console.log('stream finished');
       resolve(null);
+      process.exit();
     });
   });
 })();
--- a/bin/insertMissingAlerts.ts
+++ b/bin/insertMissingAlerts.ts
@@ -4,14 +4,30 @@ import { Alerts, User } from '../src/entity';
 
 (async () => {
   const con = await createOrGetConnection();
-  const usersQuery = con
+  const stream = await con
     .getRepository(User)
     .createQueryBuilder()
     .select('id')
     .where('id NOT IN (SELECT "userId" FROM public.alerts)')
-    .getQuery();
+    .stream();
 
-  await con
-    .getRepository(Alerts)
-    .query(`INSERT INTO public.alerts ("userId") ${usersQuery}`);
+  stream.on('data', async ({ id }: { id: string }) => {
+    console.log('inserting alerts for: ', id);
+    await con
+      .getRepository(Alerts)
+      .createQueryBuilder()
+      .insert()
+      .values({ userId: id })
+      .orIgnore()
+      .execute();
+    console.log('alerts created for: ', id);
+  });
+
+  await new Promise((resolve, reject) => {
+    stream.on('error', reject);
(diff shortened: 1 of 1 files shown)


rebelchris (approved):

sshanzel:
Also written some local tests (copied from previous PR).
Screenshot 2023-11-13 at 11 11 54 PM

bin/insertMissingAlerts.ts
```diff
@@ -4,14 +4,30 @@ import { Alerts, User } from '../src/entity';
...
+    console.log('alerts created for: ', id);
+  });
+
+  await new Promise((resolve, reject) => {
+    stream.on('error', reject);
+    stream.on('end', () => {
+      console.log('stream finished');
+      resolve(null);
+    });
+  });
```
capJavert, line 32: just add missing `process.exit()` at the end

capJavert (approved):

bin/insertMissingAlerts.ts
sshanzel: Right. To not keep the terminal hanging. Let me add that quickly.

bin/insertMissingAlerts.ts
```diff
@@ -4,14 +4,31 @@ import { Alerts, User } from '../src/entity';
 
 (async () => {
   const con = await createOrGetConnection();
-  const usersQuery = con
+  const stream = await con
```
idoshamun, line 7: This will not work as expected since you have an async function inside a stream. It will introduce race conditions. please look at Ante's cron job for the digest to see how to implement it correctly. It's not a big change but crucial one

sshanzel: I see. Cool. In this case, I assume we will batch the insertion as well, right?

sshanzel: This would mean, we will not need the stream anymore, since we can fetch all the ids in a single query unlike reading from a file which is handled in the stream.

idoshamun: We still stream but we use a library to help us with the async nature of our workload. Take a look on that script and you will see. I'm talking specifically about the cron job for the digest

idoshamun: This one: https://github.com/dailydotdev/daily-api/blob/main/src/cron/personalizedDigest.ts

sshanzel: Ah, but note that this is not a cron job and only a script to run locally while connecting to prod. A one time thing kind of script.

idoshamun: I understand but it will still suffer from the same race condition hence, some users eventually will not have a alerts set properly. The problem is that the script will finish before the async functions are done with their job. Hence it will introduce inconsistency

sshanzel: Got it. Will refactor the process to that.

sshanzel:
Ran with our previous tests as well.

Screenshot 2023-11-14 at 4 52 52 PM

sshanzel:
@idoshamun would love to hear what you think about the changes 🥳

idoshamun (approved):
Perfect!

PR #1533, opened 2023-11-14
feat: disable sendgrid warnings in production

capJavert:
Sendgrid logs warning whenever it detects template data contains some special chars, we don't want it in production since it spams our logs and makes hard to see any other errors.

See: https://github.com/sendgrid/sendgrid-nodejs/blob/main/docs/use-cases/hide-warnings.md


omBratteng (approved):

idoshamun (approved):
Nice catch!

PR #1537, opened 2023-11-15
feat: post relation db and entity changes for collections

capJavert:
Implementing WT-1960, adding entity and db changes so we can push other PRs.


src/entity/posts/Post.ts
```diff
@@ -196,4 +198,9 @@ export class Post {
   @Column({ type: 'uuid', nullable: true })
   @Index('IDX_yggdrasil_id')
   yggdrasilId: string;
+
+  @OneToMany(() => PostRelation, (postRelation) => postRelation.post, {
```
rebelchris, line 202: Is it going to be oneToMany?
Wondering if manyToMany would be better, not sure if for instance we have recommended or something later which could link post-post-post

src/entity/posts/PostRelation.ts
```diff
@@ -0,0 +1,39 @@
+import { Column, Entity, Index, ManyToOne, PrimaryColumn } from 'typeorm';
+import { Post } from './Post';
+
+export enum PostRelationType {
+  Collection = 'collection',
+}
+
+@Entity()
+@Index('IDX_post_relation_post_id_related_post_id_type_created_at', [
```
rebelchris, line 9: Tongue twister 🤣

rebelchris (approved):
Looks good, almost thought we would set the relatedPost on the `postCollection` but see we decided otherwise in the DR.

src/entity/posts/Post.ts
capJavert: It is many to many but `@ManyToMany` in graphql means it will create and manage a table like `PostRelation` by itself and we want to have control over it with our own entity. Basically [this](https://orkhan.gitbook.io/typeorm/docs/many-to-many-relations#many-to-many-relations-with-custom-properties)

rebelchris: Got it thanks for the context :)

src/entity/posts/CollectionPost.ts
```diff
@@ -0,0 +1,23 @@
+import { ChildEntity, Column } from 'typeorm';
+import { Post, PostType } from './Post';
+
+@ChildEntity(PostType.Collection)
+export class CollectionPost extends Post {
```
idoshamun, line 5: Wouldn't it be better to inherit freeform?

idoshamun (approved):
Great!

src/entity/posts/CollectionPost.ts
rebelchris: With @omBratteng DR in mind, perhaps we can create that subclass here? 
Or like you suggest inherit one that already shares all the fields :)

capJavert: Same argument as in that DR, when `FreeformPost` starts to get new fields then `CollectionPost` would get them as well which might not be what we want. For example we add `authorSignature` field to `FreeformPost`, now `CollectionPost` has it as well but that field does not make sense for that post type.

idoshamun: I do believe we need some kind of composition or reusability here to keep it consistent. Any other suggestions?

capJavert: For sake of duplicate code I personally don't mind it. But agree it might be beneficial if we define a better relationship on which type of posts are similar and can extend from each other. The above with always extending from base `Post` is the simplest and allows us to move fast.

rebelchris: My only alternative would be a sub type that's not actively used, but in return might as well use Post :)

capJavert: Yeah, are we good on merging this as is for now?

capJavert: @idoshamun is it ok from your end?

idoshamun: yes

PR #1538, opened 2023-11-16
chore: add logger on update fail

rebelchris:
Add specific log instance if data update fails, else we loose the tracking at this stage.

### DIFF
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -186,6 +186,7 @@ const contentTypeFromPostType: Record<PostType, typeof Post> = {
 };
 
 type UpdatePostProps = {
+  logger: FastifyBaseLogger;
   entityManager: EntityManager;
   data: Partial<ArticlePost>;
   id: string;
@@ -194,6 +195,7 @@ type UpdatePostProps = {
   content_type: PostType;
 };
 const updatePost = async ({
+  logger,
   entityManager,
   data,
   id,
@@ -215,6 +217,10 @@ const updatePost = async ({
     databasePost.metadataChangedAt.toISOString() >=
       data.metadataChangedAt.toISOString()
   ) {
+    logger.info(
+      { data },
+      'post not updated: database entry is newer than received update',
+    );
     return null;
   }
 
@@ -453,6 +459,7 @@ const worker: Worker = {
         } else {
           // Handle update of existing post
           await updatePost({
+            logger,
             entityManager,
             data: fixedData,
             id: post_id,


omBratteng (approved):

src/workers/postUpdated.ts
```diff
@@ -194,101 +205,115 @@ type UpdatePostProps = {
...
+        .createCounter('post_update_date_conflict', {
+          description:
+            'How many posts were not updated because the database entry is newer than the received update',
+        })
+        .add(1);
+      logger.info(
+        { data },
+        'post not updated: database entry is newer than received update',
+      );
+      return null;
```
omBratteng, line 242:
```suggestion
      );
      span.setStatus({
      	code: opentelemetry.SpanStatusCode.ERROR,
      	message: 'post not updated: database entry is newer than received update',
      })
      return null;
```

omBratteng: Could add this too, for a little bit of information and set the span to be errored

omBratteng (changes requested):

src/workers/postUpdated.ts
```diff
@@ -107,64 +108,73 @@ const createPost = async ({
...
-      'url = :url or url = :canonicalUrl or "canonicalUrl" = :url or "canonicalUrl" = :canonicalUrl',
-      { url: data.url, canonicalUrl: data.canonicalUrl },
-    )
-    .getRawOne();
-  if (existingPost) {
-    logger.info({ data }, 'failed creating post because it exists already');
-    return null;
-  }
+  const meter = opentelemetry.metrics.getMeter('api-bg');
+  return runInSpan('createPost', async () => {
```
idoshamun, line 112: We don't necessarily need this span. Tracing is not required but I don't really mind keeping it, your call

src/workers/postUpdated.ts
```diff
@@ -194,101 +205,115 @@ type UpdatePostProps = {
...
-
-  if (
-    !databasePost ||
-    databasePost.metadataChangedAt.toISOString() >=
-      data.metadataChangedAt.toISOString()
-  ) {
-    return null;
-  }
+  const meter = opentelemetry.metrics.getMeter('api-bg');
+  return runInSpan('updatePost', async () => {
```
idoshamun, line 217: Same comments as above here

src/workers/postUpdated.ts
rebelchris: Just this one, or in general? 
As it's only fraction of calls that get traced wasn't sure if the spans adding anything over the counters alone?

idoshamun (changes requested):
Just to be sure, the logic wasn't changed right? I didn't review it.
I don't see we added a metric in either `createPost` or `updatePost`.
There are some examples on how to do it. Not to be confused between tracing and metrics, they're different

src/workers/postUpdated.ts
idoshamun: It's usually good for performance benchmarking and so on. If we care about optimizing the speed of this call or willing to know if anything goes south in networking

rebelchris: I see then the counter alone should do it. We only care about the 2 edge cases returns and how often they happen.

src/workers/postUpdated.ts
```diff
@@ -107,64 +108,73 @@ const createPost = async ({
...
-    .where(
-      'url = :url or url = :canonicalUrl or "canonicalUrl" = :url or "canonicalUrl" = :canonicalUrl',
-      { url: data.url, canonicalUrl: data.canonicalUrl },
-    )
-    .getRawOne();
-  if (existingPost) {
-    logger.info({ data }, 'failed creating post because it exists already');
-    return null;
-  }
+  const meter = opentelemetry.metrics.getMeter('api-bg');
```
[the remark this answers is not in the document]
rebelchris: Coming from the main handler function or from outside in general?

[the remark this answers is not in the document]
idoshamun: main handler function I think. On the API side it usually comes from the `req` object but it doesn't hold here in the bg worker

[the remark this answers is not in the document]
omBratteng: > It’s generally recommended to call `getMeter` in your app when you need it rather than exporting the meter instance to the rest of your app. This helps avoid trickier application load issues when other required dependencies are involved.
> [source](https://opentelemetry.io/docs/instrumentation/js/manual/#acquiring-a-meter)

src/workers/postUpdated.ts
```diff
@@ -215,6 +228,16 @@ const updatePost = async ({
     databasePost.metadataChangedAt.toISOString() >=
       data.metadataChangedAt.toISOString()
   ) {
+    meter
+      .createCounter('post_update_date_conflict', {
+        description:
+          'How many posts were not updated because the database entry is newer than the received update',
+      })
+      .add(1);
```
omBratteng, line 236: Not sure if it is the best way of creating the counter and then increment it.
What we could do, is instantiate the counter in the main worker, and pass it in to `createPost` and `updatePost`.

And instead do something like this.

In worker
```js
const meter = opentelemetry.metrics.getMeter('api-bg');
const counter = meter.createCounter('post_error')
```


In update
```js
counter.add(1, {
	'reason': 'date_conflict',
})
```

And in create
```js
counter.add(1, {
	'reason': 'duplication_conflict',
})
```

That way we have one metric with more cardinality

idoshamun (approved):
Awesome!

PR #1539, opened 2023-11-20
feat: service mock

sshanzel:
Introduced `msw` package - part of the dev dependencies so it won't be included in our build, a mock service that intercepts on network level. Meaning we won't have to spool up an additional server where we would add configurations to make requests point to it. We will just mock the response and the package will check whether it is the same request we are trying to intercept.

This would make mocking super easy, especially in the frontend, we won't have to touch any existing service or logic to make the requests differ from dev/mocking or testing/production. The package is also handy in making test cases.

Do note, even though it says `setupServer` it won't actually host a new server as it only intercepts requests. Just a semantic kind of thing.

How it looks like testing in FE (sample taken from their docs):
```
import React from 'react'
import { http, HttpResponse } from 'msw'
import { setupServer } from 'msw/node'
import { render, screen, waitFor } from '@testing-library/react'
import Dashboard from '../src/components/Dashboard'

const server = setupServer(
  // Describe network behavior with request handlers.
  // Tip: move the handlers into their own module and
  // import it across your browser and Node.js setups!
  http.get('/posts', ({ request, params, cookies }) => {
    return HttpResponse.json([
      {
        id: 'f8dd058f-9006-4174-8d49-e3086bc39c21',
        title: `Avoid Nesting When You're Testing`,
      },
      {
        id: '8ac96078-6434-4959-80ed-cc834e7fef61',
        title: `How I Built A Modern Website In 2021`,
      },
    ])
  }),
)

// Enable request interception.
beforeAll(() => server.listen())

// Reset handlers so that each test could alter them
// without affecting other, unrelated tests.
afterEach(() => server.resetHandlers())

// Don't forget to clean up afterwards.
afterAll(() => server.close())

it('displays the list of recent posts', async () => {
  render(<Dashboard />)

  // 🕗 Wait for the posts request to be finished.
  await waitFor(() => {
    expect(
      screen.getByLabelText('Fetching latest posts...'),
    ).not.toBeInTheDocument()
  })

  // ✅ Assert that the correct posts have loaded.
  expect(
    screen.getByRole('link', { name: /Avoid Nesting When You're Testing/ }),
  ).toBeVisible()

  expect(
    screen.getByRole('link', { name: /How I Built A Modern Website In 2021/ }),
  ).toBeVisible()
})
```

## GitPod relation
Now that we are able to mock the authentication (tested locally), as discussed we will roll this out to support the GitPod setup and make users always authenticated while developing, but it doesn't mean its capacity stops there. Like mentioned above, there are a lot of areas we can utilize this. For example, we can go ahead and develop the email verification flow without having the heimdall ready as we can mock it from the frontend (requesting and submitting the code, not necessarily the email part).

We will just add an environment variable for the GitPod setup of the API service, from there, it should automatically work.

More information about the package here: https://mswjs.io/docs


capJavert (commented):
Just wanted to add I love msw, great pick 👌

idoshamun:
@sshanzel if not mistaken, in gitpod we use our production docker image which means msw will not be available. Am I right?

idoshamun:
Regarding your approach, wouldn't it be easier to override the kratos request altogether? So, instead of mocking the network request, let's just add an if condition in the boot the returns an hard coded user. Sounds much more simple to me

sshanzel:
> @sshanzel if not mistaken, in gitpod we use our production docker image which means msw will not be available. Am I right?

Uses locally, I see the posts we have from our seed last time I checked.

> Regarding your approach, wouldn't it be easier to override the kratos request altogether? So, instead of mocking the network request, let's just add an if condition in the boot the returns an hard coded user. Sounds much more simple to me

We need to generate the cookies and proper JWT so that when they send, let's say, a post-upvote, our GraphQL `@auth` directive will accept it.

idoshamun:
> Oh, but the posts I see is from our seeds, would it still be the production build in that case?

It's not connected to the production but uses a production image locally. It doesn't run the api project itself if not mistaken but you can check the gitpod definition or docker compose of the apps project

> We need to generate the cookies and proper JWT so that when they send, let's say, a post-upvote, our GraphQL @auth directive will accept it.

I understand, my suggestion is to "mock" as if heimdall returns a successful result without actually sending a request or using msw. At the end of the day it's a javascript function so instead of executing the http request we can just return whatever we want. If you want to discuss it through I'm available

sshanzel:
Context: we've decided to postpone the usage of a dedicated mocking server/package to trim down the complexity based on the issue we are trying to resolve here, which is to allow GitPod users to make local development work.

idoshamun (approved):
Perfect!

capJavert (approved):
Still love msw 😄

sshanzel:
> Still love msw 😄

Same here. Pretty slick!

PR #1541, opened 2023-11-20
refactor: notification preference when joining public squad

sshanzel:
When joining a public Squad, the notification preference for `squad_post_added` is default to `muted`. In this case, we updated the `joinSource` mutation though there is the option of creating a DB trigger.

Also updated existing tests for both public and private Squads to reflect what is expected for the preference.

WT-1977 #done


bin/mutePublicSquadNotifications.ts
```diff
@@ -0,0 +1,68 @@
...
+      `
+      s.id NOT IN
+        (
+            SELECT  np."sourceId"
+            FROM    notification_preference np
+            WHERE   np."sourceId" = s.id
+            AND     np."referenceId" = s.id
+            AND     np."userId" = sm."userId"
+            AND     np."notificationType" = 'squad_post_added'
+            AND     np.status = 'muted'
```
rebelchris, line 37: I think we should respect any already set notification preference.
We should only inject those who have it empty right?

Reasoning:
Someone might have muted, then turned back on.

rebelchris (commented):
Pretty neat, not shre the source member inline queries are getting to much, but leave that up to @idoshamun

bin/mutePublicSquadNotifications.ts
sshanzel: When they turn it back on, we remove the row, so we have no reference.

rebelchris: Perhaps you can remove some criteria then?
I think the more you add, the more overhead right? and in those cases we wouldn't need for instance `referenceId` and `status`

sshanzel: The criteria is to reduce the number of rows. If we remove the status we will pull everyone with existing muted status.

rebelchris: I mean because of the np."notificationType" = 'squad_post_added' you only get those, which are only muted no?

rebelchris: I do see it's only 653 rows so not too bad anyway :)

rebelchris (approved):

bin/mutePublicSquadNotifications.ts
sshanzel: We are fetching ones that are **not** muted. The total users to update is around 14k and seeing how the bin script performs, I am trying to lessen the number of rows as much as possible.

bin/mutePublicSquadNotifications.ts
```diff
@@ -0,0 +1,68 @@
...
+} from '../src/notifications/common';
+
+interface SourceMemberQueue {
+  userId: string;
+  sourceId: string;
+}
+
+const QUEUE_CONCURRENCY = 1;
+
+(async () => {
```
idoshamun, line 18: I think we should not apply it retroactively as some users are already used to getting this notifications. Wdyt @sshanzel @rebelchris ?

rebelchris: Hmm that doesn't solve all existing users getting annoyed by it? (including me and you 😂)
Why would we not do it retro?

omBratteng: Do it retroactively, but mention it in a changelog or something, so people know _why_ they might not get the notifications anymore

bin/mutePublicSquadNotifications.ts
```diff
@@ -0,0 +1,68 @@
...
+  NotificationType,
+  saveNotificationPreference,
+} from '../src/notifications/common';
+
+interface SourceMemberQueue {
+  userId: string;
+  sourceId: string;
+}
+
+const QUEUE_CONCURRENCY = 1;
```
idoshamun, line 16: Let's increate the concurrency to 100 otherwise it will be very slow just like the previous script

idoshamun (approved):

bin/mutePublicSquadNotifications.ts
sshanzel: Right. Let me update it quickly.

PR #1542, opened 2023-11-21
fix: add referral link to sendgrid users

rebelchris:
Ensure new parsed people get their referral link added to the new custom field in Sendgrid.
Next step is to do a batch CSV to insert all existing users with this field set.

WT-1978 #done

### DIFF
--- a/src/common/mailing.ts
+++ b/src/common/mailing.ts
@@ -4,6 +4,7 @@ import { MailDataRequired } from '@sendgrid/helpers/classes/mail';
 import { User } from './users';
 import { ChangeObject } from '../types';
 import { FastifyBaseLogger } from 'fastify';
+import { getInviteLink } from './links';
 
 if (process.env.SENDGRID_API_KEY) {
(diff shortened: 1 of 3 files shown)


src/common/mailing.ts
```diff
@@ -90,10 +91,18 @@ interface EmailContact {
...
+    userId: profile.id,
+  });
+
   const contact: EmailContact = {
     id: contactId,
     email: profile.email,
-    custom_fields: { e1_T: profile.id },
+    custom_fields: {
+      e1_T: profile.id,
+      e2_T: genericInviteURL.toString(),
```
idoshamun, line 104: Just to confirm, have you created it on sendgrid already? (the field I mean)

idoshamun (approved):
Looks great! Just check my comment

src/common/mailing.ts
rebelchris: Yep and run the initial CSV already

idoshamun: Perfect!

PR #1545, opened 2023-11-23
test: collections should not be editable

sshanzel:
Since it is explicitly defined which types should only be edited, we just need to add a test to ensure collection type of post should never be allowed.

WT-1960 #done


rebelchris (approved):
Nice easy fix :)

capJavert:
Thanks @sshanzel !

PR #1546, opened 2023-11-23
feat: add related posts query

capJavert:
Implementing WT-1964, adding related posts query that will be used as initial data for post by id query and then as pagination for new widget on post page.

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -58,6 +58,8 @@ import {
   updateSharePost,
   View,
   User,
+  PostRelationType,
+  PostRelation,
 } from '../entity';
 import { GQLEmptyResponse } from './common';
 import {
@@ -187,6 +189,11 @@ interface ReportPostArgs {
   reason: string;
   comment: string;
   tags?: string[];
+}
+
+export interface GQLPostRelationArgs extends ConnectionArguments {
+  id: string;
+  relationType: PostRelationType;
 }
 
 export const typeDefs = /* GraphQL */ `
@@ -572,6 +579,10 @@ export const typeDefs = /* GraphQL */ `
     IRRELEVANT
   }
 
+  enum PostRelationType {
+    COLLECTION
+  }
+
   extend type Query {
     """
     Get post by id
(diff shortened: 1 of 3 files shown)


src/schema/posts.ts
```diff
@@ -1218,6 +1254,46 @@ export const resolvers: IResolvers<any, Context> = {
...
+      args: GQLPostRelationArgs,
+      ctx,
+      info,
+    ): Promise<ConnectionRelay<GQLPost>> => {
+      const post = await ctx.con
+        .getRepository(Post)
+        .findOneByOrFail({ id: args.id });
+      await ensureSourcePermissions(ctx, post.sourceId);
+
+      return queryPaginatedByDate(
```
idoshamun, line 1268: Should we do the permission check and the actual query in parallel to increase performance?

capJavert: I was thinking about it Promise.all would easily do it, but without the fact to actually cancel query not sure it is the best thing to do in regards to abuse. For example someone spamming invalid IDs to our service would have substantially bigger impact vs check ID first and then doing a related query.

I would leave it as is and then optimize if we see its needed, wdyt?

idoshamun: Sure!

src/schema/posts.ts
```diff
@@ -1218,6 +1254,46 @@ export const resolvers: IResolvers<any, Context> = {
...
+        {
+          queryBuilder: (builder) => {
+            builder.queryBuilder = builder.queryBuilder
+              .leftJoin(
+                PostRelation,
+                'pr',
+                `pr."relatedPostId" = ${builder.alias}.id`,
+                {
+                  postId: args.id,
+                },
```
idoshamun, line 1282: Not needed

src/schema/posts.ts
```diff
@@ -1218,6 +1254,46 @@ export const resolvers: IResolvers<any, Context> = {
...
+        info,
+        args,
+        { key: 'createdAt' },
+        {
+          queryBuilder: (builder) => {
+            builder.queryBuilder = builder.queryBuilder
+              .leftJoin(
+                PostRelation,
+                'pr',
+                `pr."relatedPostId" = ${builder.alias}.id`,
```
idoshamun, line 1279: The naming is confusing, so just double-check that this is the right condition and not the other way around. Check the other PR where you actually add it to the database. Confirm that you use it the same way in both PRs

idoshamun (approved):
Check my comments but looks great!

src/schema/posts.ts
capJavert: There is a test that confirms it works.

capJavert: `postId` is owning side so in this case collection id.

idoshamun (approved):

PR #1549, opened 2023-11-23
feat: collections update and worker handling

capJavert:
Implementing collections post handling, also adding sources normalization logic. 

DR section for data normalization logic is [here](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/665124928/WT+Collections#Denormalizing-data)


src/entity/posts/utils.ts
```diff
@@ -389,3 +391,75 @@ export const updateUsedImagesInContent = async (
...
+  postId: Post['id'];
+  entries: string[];
+  relationType: PostRelationType;
+}): Promise<CollectionPost[]> => {
+  if (!entries.length) {
+    return [];
+  }
+
+  const posts = await entityManager.getRepository(CollectionPost).findBy({
+    yggdrasilId: In(entries),
```
rebelchris, line 447: Should we search based on this?
Cant we use our internal relations? (Postid)

Feels a bit odd to rely on external systems data

rebelchris (commented):
Cant we avoid using the yggdrasil ids?

src/entity/posts/utils.ts
capJavert: It is how it comes from platform, I asked in their DR if it is possible and basically issue is they can not get to post ids.

.infra/workers.ts
```diff
@@ -239,6 +239,10 @@ export const workers: Worker[] = [
   {
     topic: 'api.v1.source-created',
     subscription: 'api.source-created-squad-owner-mailing',
+  },
+  {
+    topic: 'api.v1.post-collection-updated',
```
idoshamun, line 244: Just making sure this topic exists

src/entity/posts/utils.ts
```diff
@@ -389,3 +391,75 @@ export const updateUsedImagesInContent = async (
...
+
+export const addRelatedPosts = async ({
+  entityManager,
+  postId,
+  entries,
+  relationType,
+}: {
+  entityManager: EntityManager;
+  postId: Post['id'];
+  entries: string[];
```
idoshamun, line 403: Can we be more explicit here and name it `yggdrasilIds`?

src/entity/posts/utils.ts
```diff
@@ -389,3 +391,75 @@ export const updateUsedImagesInContent = async (
...
+  });
+
+  await entityManager
+    .getRepository(PostRelation)
+    .createQueryBuilder()
+    .insert()
+    .values(
+      posts.map((post) => ({
+        postId,
+        relatedPostId: post.id,
```
idoshamun, line 421: `relatedPostId` is the collection id? Just making sure because I'm always confused by the name

src/entity/posts/utils.ts
```diff
@@ -389,3 +391,75 @@ export const updateUsedImagesInContent = async (
...
+
+export const relatePosts = async ({
+  entityManager,
+  postId,
+  entries,
+  relationType,
+}: {
+  entityManager: EntityManager;
+  postId: Post['id'];
+  entries: string[];
```
idoshamun, line 439: Same comment here

src/migration/1700661785471-PostCollectionSources.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class PostCollectionSources1700661785471 implements MigrationInterface {
+    name = 'PostCollectionSources1700661785471'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "post" ADD "collectionSources" text array DEFAULT '{}'`);
```
idoshamun, line 7: default should be null, isn't it?

src/workers/postCollectionUpdatedNormalizeSources.ts
```diff
@@ -0,0 +1,31 @@
...
+  collection: Pick<CollectionPost, 'id'>;
+}
+
+const worker: Worker = {
+  subscription: 'api.post-collection-updated-normalize-sources',
+  handler: async (message, con): Promise<void> => {
+    const data: Data = messageToJson(message);
+    const { collection } = data;
+
+    const distinctSources = await con
```
idoshamun, line 14: Do we want to order the sources somehow?

src/workers/postUpdated.ts
```diff
@@ -165,12 +175,45 @@ const createPost = async ({
...
+
+  if (isCollectionUpdated) {
+    const collectionPosts = [post, ...relatedCollections];
+
+    await Promise.allSettled(
+      collectionPosts.map((collectionPost) => {
+        return notifyPostCollectionUpdated(logger, collectionPost);
+      }),
+    );
+  }
```
idoshamun, line 215: Why don't we use CDC for this purpose? Wouldn't it be safer?

capJavert: CDC on post relation?

src/migration/1700661785471-PostCollectionSources.ts
capJavert: This is what typeorm pipes out and how we did other array fields.

idoshamun: ok

src/workers/postUpdated.ts
idoshamun: Yes

idoshamun: Another approach is whenever we update the post relation to update also the `metadataUpdatedAt` field and to catch that on a CDC

capJavert: If we did it on post relation it would be multiple CDC messages in the same time. That would not be efficient. 

For CDC on post, sometimes the post is a collection and we add relations to it, and sometimes post is part of multiple collections and then we add it as a relation. So we would be basically updating `metadataUpdatedAt` for multiple posts at the same time again. That field until this point is handled by the platform side which could then lead to missed updates because our `metadataUpdatedAt` would be newer then platform one. 

I hope it makes sense but I think the separate subscription for now is better.

idoshamun: My concern is not the separate subscription but the fact we introduce a distributed transaction without being aware. Meaning we have a function that assumes a database change + pubsub notification are atomic which is not the case. Ideally we want to do only database changes which will trigger CDC which will trigger whatever business logic we need.
Another option is to always update the `metadataUpdatedAt` of the collection on both cases.
Anyway, happy for any ideas to solve the issue I raised

idoshamun: What would be your concrete suggestion in that case? Perhaps it's worth discussing it on Slack, so others can contribute or in the DR.

capJavert: @idoshamun I did a change in 9b287bf3404eb7b0ee584e85866e49b1220ecd11 to use post_relation cdc.

idoshamun (approved):
Handling the markdown content and creation is out of scope for this PR?
If so, this looks fantastic now

capJavert:
@idoshamun yeah, I'll add markdown, that is why I was asking yesterday what the content was. 

What do you mean by creation?

idoshamun:
The creation of the post itself. This part is only `postUpdated`

capJavert:
@idoshamun creation is done inside `postUpdated` worker. That should be ready as well.

idoshamun:
@capJavert right missed that because it was hidden since I reviewed it already, so mostly the MD part. It's quite confusing, probably for next time it's better to add a todo comment in case we need early reviews which I'm perfectly fine with

capJavert:
@idoshamun yeah, it was in draft before, while PR is draft I thought we treat it as not ready for review, but we can add a "WIP" label as well to the PR in the future?

idoshamun:
My bad then. I saw comments so assumed it's ready but you're right

src/migration/1701100478531-PostRelationReplica.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class PostRelationReplica1701100478531 implements MigrationInterface {
```
rebelchris, line 3: This should be together with the above migration :)

src/graphorm/index.ts
```diff
@@ -199,6 +199,22 @@ const obj = new GraphORM({
           return value;
         },
       },
+      collectionSources: {
+        relation: {
+          isMany: true,
+          customRelation: (ctx, parentAlias, childAlias, qb): QueryBuilder => {
+            return qb
+              .where(
+                `${childAlias}.id = ANY(${parentAlias}."collectionSources")`,
```
rebelchris, line 208: Why do we need the `ANY` here?

src/workers/postUpdated.ts
```diff
@@ -472,19 +511,33 @@ const worker: Worker = {
...
-            id: post_id,
+            id: postId,
             mergedKeywords,
             questions,
             content_type,
           });
         }
+
+        if (postId) {
+          await handleCollectionRelations({
```
rebelchris, line 531: This is because it needs to be for both creation/update?
Maybe we can rather have it inside both functions?

Just thinking because it's just a bit harder to follow now.

rebelchris (commented):
Nothing blocking, but some comments from my side.

rebelchris:
@capJavert Be sure to add a task for the markdown so we don't forget :)

src/graphorm/index.ts
capJavert: `collectionSources` is array of source ids

src/workers/postUpdated.ts
idoshamun: @rebelchris I suggest logging the need for a refactor here. I agree we can make it a bit more straightforward

src/migration/1701100478531-PostRelationReplica.ts
capJavert: The issue with migrations was because we thought multiple files could break production due to one migration failing or not. Then we said it does not matter because typeorm [runs them together in transaction](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1697713738231189?thread_ts=1697711684.457159&cid=C02E2C3C13R) right? 

No need to bundle them when they do two separate things.

capJavert:
> @capJavert Be sure to add a task for the markdown so we don't forget :)

I am doing it right now.

src/workers/postUpdated.ts
capJavert: Generally I see the need to handle different post types in the future in structured way because if you look at my changes I had to add `content` and different properties to `fixData` and using union types.

src/migration/1701100478531-PostRelationReplica.ts
rebelchris: CC: @idoshamun

src/graphorm/index.ts
rebelchris: Got it :)

src/workers/postUpdated.ts
rebelchris: Yeah it started as one type, but as we grow it's very messy.
Agreed, we should probably split it up a bit better, and even use different type handlers.

WIll log it, no action needed.

capJavert: 👌 tnx!

capJavert:
Added markdown parsing in 8af256de80572644af2641ea63b8c5bf562f8491

rebelchris (changes requested):
We might want to add a test case to verify markdown parsing happens correctly?
(Mainly if someone accidentally removes/changes it.)

src/workers/postUpdated.ts
```diff
@@ -426,6 +463,11 @@ const fixData = async ({
         sentAnalyticsReport: privacy || !authorId,
       },
       yggdrasilId: data?.id,
+      type: data?.content_type as PostType,
+      content: data?.extra?.content,
+      contentHtml: data?.extra?.content
+        ? markdown.render(data.extra.content, { mentions: [] })
```
capJavert, line 469: Mentions do not exist?

rebelchris: As it's generated they don't, but since it's a shared type I think we made it like this.
So this should do.

__tests__/workers/postUpdated.ts
```diff
@@ -733,3 +737,350 @@ describe('on post update', () => {
...
+      type: 'collection',
+      title: 'New title',
+      sourceId: 'unknown',
+      yggdrasilId: '7ec0bccb-e41f-4c77-a3b4-fe19d20b3874',
+      image: 'http://image.com',
+      content: 'New content',
+      readTime: 10,
+      description: 'New description',
+      summary: 'New summary',
+      contentHtml: '<p>New content</p>\n',
```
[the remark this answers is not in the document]
rebelchris: Ah sorry missed that while reviewing :)

src/workers/postUpdated.ts
rebelchris: Although think this will work as well:
```
markdown.render(data.extra.content)
```

__tests__/workers/postUpdated.ts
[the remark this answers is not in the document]
capJavert: Updated here d623649c880ca66a4d2c3846abdd38711dc2f295

rebelchris (approved):
amazing amazing

idoshamun (approved):

PR #1550, opened 2023-11-23
feat: create entity for YouTubePost and extend AdvancedSettings

omBratteng:
- Created new entity for `YouTubePost`
- Extended `AdvancedSettings` with column `group`

Decided to split up adding the new setting for content types to WT-2000 so that users don't suddenly see a new setting before WT-1985 is done

WT-1983


rebelchris (approved):
Looks good.
As for the actual content, double check if we even run it via a migration, or do it in db directly, cant recall

omBratteng:
Ah, no, we don't actually add the advanced settings via migrations but via the seeds. So probably should update the seed to include that then?

rebelchris:
@omBratteng the seed will only do locally, good to add, but well need to execute on production db once you want it 👍

omBratteng:
Yeah, plan it in WT-2000, either solve it with a migration or directly in database.

src/entity/posts/YouTubePost.ts
```diff
@@ -0,0 +1,20 @@
...
+  description?: string;
+
+  @Column({ type: 'text', nullable: true })
+  summary?: string;
+
+  @Column({ type: 'integer', nullable: true })
+  readTime?: number;
+
+  @Column({ type: 'text' })
+  video_id: string;
```
idoshamun, line 19:
```suggestion
  videoId: string;
```

src/migration/1700735236452-YouTubePost.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class YouTubePost1700735236452 implements MigrationInterface {
```
idoshamun, line 3: Please merge both migrations if possible. Ideally we should strive to have 1 migration per PR to easily review database changes

src/migration/1700735663192-AdvancedSettings.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AdvancedSettings1700735663192 implements MigrationInterface {
+    name = 'AdvancedSettings1700735663192'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "advanced_settings" ADD "group" text NOT NULL DEFAULT 'advanced'`);
+        await queryRunner.query(`ALTER TABLE "advanced_settings" ALTER COLUMN "group" DROP DEFAULT`);
```
idoshamun, line 8: Why?

src/migration/1700735236452-YouTubePost.ts
omBratteng: I don't think the other settings are added via migrations. But I don't mind either way.

If I must add the new setting in this migration, merging will have to wait till the frontend part is completed so we don't show the youtube setting in the wrong place.

src/migration/1700735663192-AdvancedSettings.ts
omBratteng: For me to be able to set the new column as not null, I had to add default value as adding it in a new query fails.

Alternatively set it as nullable, set values, then set column as not null.

Could also leave the default value?

src/entity/posts/YouTubePost.ts
omBratteng: 🫡

src/migration/1700735236452-YouTubePost.ts
rebelchris: Ido means the two migration files, not the actual fields.

Although, we should probably open a discussion around that soon 👍

omBratteng: Ahh, I misunderstood.

I could merge them both, no objections to it.

My personal preference would be keep them separate as they get introduced in two different commits, and I (strive to) try to keep my commits atomic and each commit can be checked out and the application still works fine.

src/migration/1700735663192-AdvancedSettings.ts
idoshamun: Let's leave the default value, makes sense :)

src/migration/1700735236452-YouTubePost.ts
idoshamun: Since we use squash and merge, I prefer the final commit to be easier to understand than individual commits on the side branch. Although I love your approach

src/migration/1700735663192-AdvancedSettings.ts
omBratteng:
```suggestion
```

src/migration/1700735236452-YouTubePost.ts
omBratteng: I never really look at the squashed commit, I usually go directly to see the PR it originated from. Albeit its a bad practice/habit, as it depends on GitHub (or the git host you use) to retain the commits/branch after merging.

Just wish that there was a new option to merge branches. A mix of squash and merge, everything is contained in one squashed commit, but you can "zoom in" and see the commits on the commit.

omBratteng: Will rebase and merge the two migrations tomorrow

idoshamun (approved):
Thanks!

src/entity/AdvancedSettings.ts
```diff
@@ -11,6 +11,9 @@ export class AdvancedSettings {
   @Column({ type: 'text' })
   description: string;
 
+  @Column({ type: 'text' })
```
sshanzel, line 14: I think we should make the default value as part of the entity, so that when generating the migration (even though it is broken 😅) the relevant line should show the accurate result.

```suggestion
  @Column({ type: 'text', default: 'advanced' }) // ideally, if we can take the string to an enum, it would be better
```

PR #1553, opened 2023-11-28
feat: get comment by id query

sshanzel:
In order to lessen the load we request from the server, we will start to not fetch the markdown content on post comments as those values are unused. If need be, the app will send a query to fetch the comment's details which includes what was missed in the post comments, which is the markdown content.

### DIFF
--- a/src/schema/comments.ts
+++ b/src/schema/comments.ts
@@ -279,6 +279,11 @@ export const typeDefs = /* GraphQL */ `
     Markdown equivalent of the user's comment
     """
     commentPreview(content: String!, sourceId: String): String @auth
+
+    """
+    Fetch a comment by id
+    """
+    comment(id: ID!): Comment @auth
   }
 
   extend type Mutation {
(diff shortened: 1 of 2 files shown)


src/schema/comments.ts
```diff
@@ -665,6 +670,19 @@ export const resolvers: IResolvers<any, Context> = {
...
+      ctx,
+      info,
+    ): Promise<GQLComment> =>
+      graphorm.queryOneOrFail<GQLComment>(ctx, info, (builder) => ({
+        queryBuilder: builder.queryBuilder.where(
+          `"${builder.alias}"."id" = :id`,
+          { id },
+        ),
+        ...builder,
+      })),
```
capJavert, line 685: Should we check that user is author of the comment or has access (for example in private squads)?

rebelchris (commented):
No further comments from my side :)

rebelchris (approved):

src/schema/comments.ts
```diff
@@ -665,6 +670,30 @@ export const resolvers: IResolvers<any, Context> = {
 
       return markdown.render(trimmed, { mentions });
     },
+    comment: async (
+      _,
+      { id }: { id: string },
+      ctx,
+      info,
+    ): Promise<GQLComment> => {
+      const comment = await ctx.con.getRepository(Comment).findOneOrFail({
```
idoshamun, line 679: Optimization suggestion: instead of querying the comment, query it via graphorm like you do at the end. Keep the result and use the returned id to fetch the post and ensure permissions and then just return this object. Will save one call to the database

sshanzel: Awesome. Will do that. I just thought that the GraphORM might not work for the expected result if not called from return. Let me update it quickly.

idoshamun (approved):
Perfect!

PR #1554, opened 2023-11-28
fix: shortened URL for referral

rebelchris:
Add shortened URL for sendgrid.

WT-1980

@idoshamun We only need to worry about the data loader in graphQL requests? (thus can omit like in this PR?)

### DIFF
--- a/src/common/mailing.ts
+++ b/src/common/mailing.ts
@@ -106,11 +106,7 @@ const profileToContact = async (
     referralOrigin: 'generic',
     userId: profile.id,
   });
-
-  const [genericInviteURL] = await Promise.all([
-    getShortUrl(rawInviteURL.toString(), log),
-  ]);
-
+  const genericInviteURL = await getShortUrl(rawInviteURL.toString(), log);
   const contact: EmailContact = {
     id: contactId,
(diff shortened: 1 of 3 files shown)


src/common/mailing.ts
```diff
@@ -97,12 +97,20 @@ interface EmailContact {
...
+  log: FastifyBaseLogger,
+) => {
+  const rawInviteURL = getInviteLink({
     referralOrigin: 'generic',
     userId: profile.id,
   });
 
+  const [genericInviteURL] = await Promise.all([
+    getShortUrl(rawInviteURL.toString(), log),
+  ]);
```
capJavert, line 112: Why promise.all here?

capJavert (commented):
Nice, no need for data loader here since you are not requesting same short url multiple times in worker context. Also I don't think we can use it since we only added support in request context for now.

src/common/mailing.ts
rebelchris: Ah I actually had 2 items there before, let me adjust it :)

capJavert (approved):

idoshamun (approved):
Looks good!

PR #1555, opened 2023-11-28
feat: add updatedAt field for post

capJavert:
We need updated at value to show in collections post page. `metadataUpdatedAt` is `updated_at` from platform so it should suffice and we can always adjust it with other field thus i called it `updatedAt` generically. 

image


rebelchris (approved):
Looks good to me :)

idoshamun (approved):

PR #1560, opened 2023-11-30
refactor: type safety pubsub implementation

idoshamun:
**This PR is a proposal nothing that should go to production at the moment.**

The main goal is to ensure type safety when using pubsub both on the publisher and consumer side and keep them consistent.

My approach uses a centralized schema type that is enforced by a single function for sending events (`triggerTypedEvent`) and on the worker side by a new `TypedWorker` interface. The DX is significantly improved as you get auto complete and validation for topic names and payload as long as we have the schema up to date.

In this PR, I migrated some topics and a single worker to show how we can make it work if we agree on this approach. I wanted to demo how it's gonna look before suggesting an action plan so if you don't like it at all, say it explicitly.

### DIFF
--- a/src/workers/sourceRequestApprovedRep.ts
+++ b/src/workers/sourceRequestApprovedRep.ts
@@ -1,18 +1,12 @@
 import { ReputationEvent, ReputationReason, ReputationType } from '../entity';
-import { messageToJson, Worker } from './worker';
-import { SourceRequest } from '../entity';
+import { TypedWorker } from './worker';
 import { NotificationReason } from '../common';
 
-interface Data {
-  reason: NotificationReason;
-  sourceRequest: Pick<SourceRequest, 'id' | 'userId'>;
-}
-
-const worker: Worker = {
+const worker: TypedWorker<'pub-request'> = {
   subscription: 'pub-request-rep',
   handler: async (message, con, logger): Promise<void> => {
-    const data: Data = messageToJson(message);
-    const { reason, sourceRequest }: Data = messageToJson(message);
+    const { data } = message;
+    const { reason, sourceRequest } = data;
 
     if (reason !== NotificationReason.Publish) {
       return;
--- a/__tests__/helpers.ts
+++ b/__tests__/helpers.ts
@@ -8,8 +8,8 @@ import { DataSource, DeepPartial, ObjectType } from 'typeorm';
 import request from 'supertest';
 import { GraphQLFormattedError } from 'graphql';
 import { Context } from '../src/Context';
-import { Message, Worker } from '../src/workers/worker';
-import { base64 } from '../src/common';
+import { Message, TypedWorker, Worker } from '../src/workers/worker';
+import { base64, PubSubSchema } from '../src/common';
 import { Roles } from '../src/roles';
 import { Cron } from '../src/cron/cron';
 import { ChangeMessage, ChangeObject } from '../src/types';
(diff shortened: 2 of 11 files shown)


src/workers/cdc.ts
```diff
@@ -108,35 +108,31 @@ const onSourceRequestChange = async (
 ): Promise<void> => {
   if (data.payload.op === 'c') {
     // New source request
-    await notifySourceRequest(
-      logger,
-      NotificationReason.New,
-      data.payload.after,
-    );
+    await triggerTypedEvent(logger, 'pub-request', {
```
idoshamun, line 111: Here you get autocomplete when typing the topic and also when setting the object + type validation

src/workers/sourceRequestApprovedRep.ts
```diff
@@ -1,18 +1,12 @@
...
-  sourceRequest: Pick<SourceRequest, 'id' | 'userId'>;
-}
-
-const worker: Worker = {
+const worker: TypedWorker<'pub-request'> = {
   subscription: 'pub-request-rep',
   handler: async (message, con, logger): Promise<void> => {
-    const data: Data = messageToJson(message);
-    const { reason, sourceRequest }: Data = messageToJson(message);
+    const { data } = message;
```
idoshamun, line 8: type of `data` is automatically inferred thanks to `TypedWorker<'pub-request'>`

capJavert:
I like it and appreciate the auto complete additions, those can be merged either way 👌. What I am missing is the part where we enforce specific payload for events eg. what Chris said.. does the event get change object or just id or something else. For example we could extend the type to always require full entity. So not to just solve type safety but also rule what is allowed more strictly.

Either way if we are going in this direction we might also want to take a look at some more full fledged schema solution like zod or valibot across the api.

idoshamun:
> What I am missing is the part where we enforce specific payload for events eg. what Chris said.. does the event get change object or just id or something else.

That's a solid point, but do you think it should be enforced by a type system/tooling or on review? I think it's more of a convention but we should keep it flexible. I can imagine some exception use cases.

> Either way if we are going in this direction we might also want to take a look at some more full fledged schema solution like zod or valibot across the api.

What would be the advantage of zod in this case? We don't need to really validate anything, just make sure we comply with the types. Zod is more suited towards user input, isn't it? I never used it so trying to understand how you envision it.

kopancek:
I think I agree with @capJavert on enforcing the type of the payload for events. If we are using typescript, let's use it for everything. We can still have flexibility with type system, e.g. by using a different type for event payloads, when it's needed. IMO it's easier to see on a PR that "Oh hey, this code is using a new type for the event payload, I should pay more attention now".

Other than that, I like the improved DX a lot, even though I have not been working in this code at all so far.

rebelchris (approved):
Very happy with this initial approach as mentioned by the others.
I would take a mix.

We should by default accept the ChangeObject of the type, but there are some exceptions as I recall where we have topics for custom things (eg kratos integrations) they don't match our own.

So in that case if we somehow can type-cast responses that would be great.
On top of that let's document somewhere by default it should publish changeObject including the whole object.

sshanzel (commented):
This definitely improves the DX having generics. Would also greatly avoid mistakes in the past where we should have passed the whole object itself (alerts) but instead passed an object with `alerts` property. I'm up for it regardless if take 3rd party package vs homemade in terms of the validation mentioned.

PR #1561, opened 2023-11-30
feat: add subscribe notification preference mutation

capJavert:
Implementing WT-2025 to support collection post subscription.

### DIFF
--- a/__tests__/notifications.ts
+++ b/__tests__/notifications.ts
@@ -306,7 +306,11 @@ describe('query notifications', () => {
   });
 });
 
-const prepareNotificationPreferences = async () => {
+const prepareNotificationPreferences = async ({
+  status,
+}: {
+  status: NotificationPreferenceStatus;
+}) => {
   await saveFixtures(con, User, usersFixture);
   await saveFixtures(con, Source, sourcesFixture);
   await saveFixtures(con, Post, postsFixture);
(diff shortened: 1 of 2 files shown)


src/notifications/common.ts
```diff
@@ -148,7 +148,7 @@ export const saveNotificationPreference = async (
       .createQueryBuilder()
       .insert()
       .values(params)
-      .orIgnore()
+      .orUpdate(['status'], ['referenceId', 'userId', 'notificationType'])
```
capJavert, line 151: @sshanzel I would like your eyes here specifically, so I noticed now that we have new status `subscribed` it's possible to (due to frontend being out of sync for any reason) to send subscribe when you are muted and vice versa. 

So I updated to instead of ignore always overwrite `status` on conflict. Nothing change functionality wise we only additionally support this case.

__tests__/notifications.ts
```diff
@@ -750,3 +819,238 @@ describe('mutation clearNotificationPreference', () => {
...
+      .getRepository(NotificationPreference)
+      .findOneBy(params);
+
+    expect(notificationPreference).toBeTruthy();
+    expect(notificationPreference!.status).toEqual(
+      NotificationPreferenceStatus.Subscribed,
+    );
+  });
+
+  it('should update status when different preference is set', async () => {
```
capJavert, line 1006: This is the test that confirms https://github.com/dailydotdev/daily-api/pull/1561/files#r1411843132

idoshamun (approved):

src/notifications/common.ts
sshanzel: Yeah, this makes sense. Currently, we remove the row when the user unsubscribe, so the expected behavior as of the moment is to always insert a new row.

Just not sure how the `orUpdate` works, but if it has been verified that it applies the correct condition of which entity to update and it sets the right values, we should be good to go 🚢

sshanzel (approved):
Looks great! 🥳 

Just checked whether it is possible to add another safety expect statement, but if verified it does not, this is non-blocking.

src/notifications/common.ts
sshanzel: Nice, in that case we can be sure that this is the only one that updates 🚀

PR #1563, opened 2023-12-01
chore: improve boot time for local development ⚡️

idoshamun:
Following the discussion on slack (https://dailydotdev.slack.com/archives/C02E2C3C13R/p1701438309729029), we found that the bottleneck of the boot time is actually ts-node. After some research we found out we can easily utilize swc to gain a massive performance boost.

To many more DX PRs 🍻

### DIFF
--- a/src/entity/Keyword.ts
+++ b/src/entity/Keyword.ts
@@ -23,6 +23,7 @@ export class Keyword {
   value: string;
 
   @Column({
+    type: 'text',
     default: 'pending',
   })
   @Index('IDX_keyword_status')
--- a/src/entity/User.ts
+++ b/src/entity/User.ts
@@ -91,7 +91,7 @@ export class User {
   @Column({ type: 'text', nullable: true })
   timezone?: string;
 
-  @Column({ default: false })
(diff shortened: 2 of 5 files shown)


src/entity/Keyword.ts
```diff
@@ -23,6 +23,7 @@ export class Keyword {
   value: string;
 
   @Column({
+    type: 'text',
```
idoshamun, line 26: swc couldn't guess the type by itself hence it failed to boot

src/entity/User.ts
```diff
@@ -91,7 +91,7 @@ export class User {
   @Column({ type: 'text', nullable: true })
   timezone?: string;
 
-  @Column({ default: false })
+  @Column({ type: 'boolean', default: false })
```
idoshamun, line 94: Same here

src/entity/Keyword.ts
omBratteng: Doesn't hurt to explicitly set the type

capJavert (approved):
need for speed

PR #1569, opened 2023-12-06
feat: collection updated notification email

capJavert:
Implementing email notification. Thread [here](https://dailydotdev.slack.com/archives/C0601MRBJJU/p1701702997935609)


src/workers/newNotificationMail.ts
```diff
@@ -501,8 +504,59 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
...
+    const latestRelatedPost = await con.getRepository(ArticlePost).findOne({
+      where: {
+        id: latestPostRelation.relatedPostId,
+      },
+      relations: ['source'],
+    });
+
+    if (!latestRelatedPost) {
+      return;
+    }
```
kopancek, line 541: Doing 3 SQL queries seems wasteful to me. It also seems like we now need to handle the case of not having any data 3 times. Is it possible to do it in only 1 query? Or are we worried about the query complexity and future readability?

capJavert: Valid point, since this is done in the worker I don't think this is too impactful. I could optimize to two queries with second done in join but personally prefer readability in this case. Let's see what others think and I'll adjust.

capJavert: On the second thought, this does not look too bad, using `relations` to do the join on orm side which saves us one query. 6caf7e96cbe7b89b0ca5db5eb6ff84dd108c417f

sshanzel (approved):

rebelchris (approved):
Nice work

PR #1571, opened 2023-12-08
feat: notification v2 schema

idoshamun:
Apply schema changes for notification v2 as described here: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/701267969/WT-1936+Optimize+notifications#Database-Schema

WT-2042 #done

### DIFF
--- a/src/migration/1702045361925-NotificationV2.ts
+++ b/src/migration/1702045361925-NotificationV2.ts
@@ -1,0 +1,36 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class NotificationV21702045361925 implements MigrationInterface {
+    name = 'NotificationV21702045361925'
+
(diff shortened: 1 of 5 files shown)


src/migration/1702045361925-NotificationV2.ts
```diff
@@ -0,0 +1,36 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class NotificationV21702045361925 implements MigrationInterface {
+    name = 'NotificationV21702045361925'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
```
capJavert, line 6: Missing replica identity full for new table?

idoshamun: Good catch!

PR #1573, opened 2023-12-10
feat: notification v2 triggers

idoshamun:
Prerequisite for: https://github.com/dailydotdev/daily-api/pull/1572

New triggers to update avatars when needed and delete relevant notifications.

I don't delete avatars and attachments, I feel like after the optimization the hassle doesn't worth it. I will have to index these fields in the notifications table and update the array too much effort for too little value.


sshanzel (approved):
Nice. Good thing we are using triggers. I thought there will be new workers to watch the changes. It would have been ironic we are doing the optimization if that was the case 😂

PR #1575, opened 2023-12-11
feat: map video data to correct fields

omBratteng:


src/workers/postUpdated.ts
```diff
@@ -52,6 +52,7 @@ interface Data {
   collections?: string[];
   extra?: {
     keywords?: string[];
+    keywords_native?: string[];
```
rebelchris, line 55: What exactly is this?
Not sure where it comes from/or why we need it?

rebelchris (commented):
Just the keywords I'm not sure why it's introduced/needed

src/workers/postUpdated.ts
omBratteng: It comes from Yggdrasil.
`keywords` is generated from the captions, and `keywords_native` comes from video description.

Pulled from the Yggdrasil model:
https://github.com/dailydotdev/yggdrasil/blob/befaa6c92e3e1dc598e6fc2cdb8368c50bbb83f8/services/content_publisher/types.go#L35-L36

omBratteng:
```suggestion
    keywords_native?: string[];  // Extra keywords field that comes from YouTube videos
```

Could add a comment to it like this?

src/workers/postUpdated.ts
```diff
@@ -414,7 +419,7 @@ const fixData = async ({
 
   const { allowedKeywords, mergedKeywords } = await mergeKeywords(
     entityManager,
-    data?.extra?.keywords,
+    [...(data?.extra?.keywords || []), ...(data?.extra?.keywords_native || [])],
```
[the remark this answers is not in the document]
capJavert: Perfect. 👌 

also, can `keywords_native` contain duplicates that were already in `keywords`. Might be worth a check.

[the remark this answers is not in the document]
omBratteng: The `mergeKeywords` function should already handle that. But I'll add a separate test for that.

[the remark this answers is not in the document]
omBratteng: Done

capJavert (approved):

src/entity/posts/YouTubePost.ts
```diff
@@ -15,6 +15,14 @@ export class YouTubePost extends Post {
   @Column({ type: 'integer', nullable: true })
   readTime?: number;
 
+  @Column({ type: 'text' })
```
idoshamun, line 18: Don't we miss a migration for the new fields?

src/workers/postUpdated.ts
```diff
@@ -412,9 +417,17 @@ const fixData = async ({
     data,
   });
 
+  let keywords: string[] = [];
+  if (data?.extra?.keywords) {
+    keywords = keywords.concat(data.extra.keywords);
+  }
+  if (data?.extra?.keywords_native) {
+    keywords = keywords.concat(data.extra.keywords_native);
+  }
```
idoshamun, line 426: We should not change the existing logic of the keywords. We should keep this business logic in yggdrasil

omBratteng: That would be preferred to handle it in Yggdrasil.
Guess a change would need to be done there then?
cc @denisb0

idoshamun: No change is required there actually :)
It already follows the right logic

omBratteng: In the sample content I got from Denis, there's different keywords in `keywords` and `keywords_native`. And the way I interpret the type in [Yggdrasil](https://github.com/dailydotdev/yggdrasil/blob/befaa6c92e3e1dc598e6fc2cdb8368c50bbb83f8/services/content_publisher/types.go#L35-L36) those keywords are gathered from different places, and [never merged](https://github.com/dailydotdev/yggdrasil/blob/befaa6c92e3e1dc598e6fc2cdb8368c50bbb83f8/workflows/entry_engines/youtube/workflow.go#L341-L342)?

Or am I misunderstanding?

idoshamun: That's right but we don't want to merge them also. Native ones can be quite a garbage to our experience

idoshamun: Yap

omBratteng: Removed

rebelchris (approved):
Looks good to me :)

idoshamun (approved):

PR #1577, opened 2023-12-11
feat: use notifications v2 as the source of truth

idoshamun:
Read notifications from the notification v2 tables. This involves some GraphORM magic to fetch the attachments and avatars from the right place. 🪄

I also noticed we don't use the unread notifications count worker, so I removed it

See: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/701267969/WT-1936+Optimize+notifications#Read-from-the-new-schema

Prerequisite:
- [x] https://github.com/dailydotdev/daily-api/pull/1572
- [x] Merge old data

WT-2046 #done


src/common/pubsub.ts
```diff
@@ -47,7 +47,6 @@ const userDeletedTopic = pubsub.topic('user-deleted');
 const userUpdatedTopic = pubsub.topic('user-updated');
 const usernameChangedTopic = pubsub.topic('username-changed');
 const settingsUpdatedTopic = pubsub.topic('settings-updated');
-const notificationsReadTopic = pubsub.topic('api.v1.notifications-read');
```
rebelchris, line 50: You also removed this from streams?

src/schema/notifications.ts
```diff
@@ -368,9 +367,9 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
           notificationsPageGenerator.nodeToCursor(page, args, node, index),
         (builder) => {
           builder.queryBuilder
-            .andWhere(`${builder.alias}."userId" = :user`, { user: ctx.userId })
-            .andWhere(`${builder.alias}."public" = true`)
-            .addOrderBy(`${builder.alias}."createdAt"`, 'DESC');
+            .andWhere(`un."userId" = :user`, { user: ctx.userId })
```
rebelchris, line 370: Does this scale?
Just wondering if it doesn't break

rebelchris (approved):
Nice nothing blocking from my side.

src/common/pubsub.ts
idoshamun: Good point! 🙈

src/schema/notifications.ts
idoshamun: I did everything I could to make it fast. I believe it should be ok.
Basically pagination, ordering, filtering is done on a single table `UserNotification`. It should reduce the load significantly

src/common/pubsub.ts
idoshamun: https://github.com/dailydotdev/streams/pull/24

PR #1578, opened 2023-12-11
fix: update the post type on youtube video when we get it from Yggdrasil

omBratteng:
When we share a YouTube link in a Squad, it is saved as an `article`. So we need to update it to `video:youtube` when we get the updated data from Yggdrasil.

### DIFF
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -261,11 +261,11 @@ const updatePost = async ({
 
   // If we don't find the post, we need to check if it's a youtube video and
   // try to find it again as an article
-  if (!databasePost && content_type === PostType.VideoYouTube) {
+  if (!databasePost) {
     await entityManager
       .createQueryBuilder()
       .update(Post)
-      .set({ type: PostType.VideoYouTube })
+      .set({ type: content_type })
       .where('id = :id', { id })
       .execute();
 
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -217,6 +217,7 @@ const createPost = async ({
 const allowedFieldsMapping = {
   [PostType.VideoYouTube]: [
     'type',
+    'videoId',
     'contentCuration',
     'description',
     'metadataChangedAt',
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -215,16 +215,6 @@ const createPost = async ({
 };
 
 const allowedFieldsMapping = {
-  [PostType.VideoYouTube]: [
-    'type',
-    'videoId',
-    'contentCuration',
-    'description',
-    'metadataChangedAt',
-    'readTime',
-    'summary',
-    'tagsStr',
-  ],
   freeform: [
     'contentCuration',
     'description',
@@ -264,7 +254,7 @@ const updatePost = async ({
   questions,
   content_type = PostType.Article,
 }: UpdatePostProps) => {
-  let postType = contentTypeFromPostType[content_type];
+  const postType = contentTypeFromPostType[content_type];
   let databasePost = await entityManager
     .getRepository(postType)
     .findOneBy({ id });
@@ -272,7 +262,13 @@ const updatePost = async ({
   // If we don't find the post, we need to check if it's a youtube video and
   // try to find it again as an article
   if (!databasePost && content_type === PostType.VideoYouTube) {
-    postType = ArticlePost;
+    await entityManager.getRepository(ArticlePost).update(
+      { id },
+      {
+        type: PostType.VideoYouTube,
+      },
+    );
+
     databasePost = await entityManager
       .getRepository(postType)
       .findOneBy({ id });
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -215,6 +215,15 @@ const createPost = async ({
 };
 
 const allowedFieldsMapping = {
+  [PostType.VideoYouTube]: [
+    'type',
+    'contentCuration',
+    'description',
+    'metadataChangedAt',
+    'readTime',
+    'summary',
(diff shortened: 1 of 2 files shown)


src/workers/postUpdated.ts
```diff
@@ -254,11 +263,20 @@ const updatePost = async ({
...
     .findOneBy({ id });
 
+  // If we don't find the post, we need to check if it's a youtube video and
+  // try to find it again as an article
+  if (!databasePost && content_type === PostType.VideoYouTube) {
+    postType = ArticlePost;
+    databasePost = await entityManager
+      .getRepository(postType)
+      .findOneBy({ id });
+  }
```
omBratteng, line 278: I don't particularly like this part, but not sure if there's any better methods to re-fetch it with a different repository selected.

capJavert: This indeed is limitation on typeorm side. Maybe we can do a generic sql query for just the type in the db and then do a proper query when we know what is in the DB?

rebelchris: Can't we do something like `getRepository(Post)` and see if the type matches or not?

I think I'm also missing the actual part where you change the type right? (save it as such type)

omBratteng: The change of type happens in [allowedFieldsMapping](https://github.com/dailydotdev/daily-api/blob/db94e7a148551f90fdb1eb4db9bd0b8f5c65f8e6/src/workers/postUpdated.ts). I added `type` (and others) [to the mapping](https://github.com/dailydotdev/daily-api/blob/db94e7a148551f90fdb1eb4db9bd0b8f5c65f8e6/src/workers/postUpdated.ts#L219).

omBratteng: So I think it might need to be a bit more advanced than what I wrote here.
Since `ArticlePost` doesn't have `videoId`, updating it fails.

src/workers/postUpdated.ts
```diff
@@ -255,10 +255,25 @@ const updatePost = async ({
...
+      {
+        type: PostType.VideoYouTube,
+      },
+    );
+
+    databasePost = await entityManager
+      .getRepository(postType)
+      .findOneBy({ id });
+  }
+
```
omBratteng, line 276: @capJavert @rebelchris I changed it to this, as it failed trying to set `videoId` on `ArticlePost`, I have to update the post type, then re-fetch it.

Could also make this a bit more generic. So before we get the databasePost in the first place, we just do a query like @capJavert suggested, and update the type (it it is allowed for that type)

rebelchris: This is where I wonder if we can't get the type in the initial one if we change to `Post` and then use post.type === the one we expect?
Might be a little bit more stable?

capJavert: We can but we can't write it since the columns would be just the ones existing in `Post`. Raw query is more memory efficient since no need to assemble entity but it would do the same thing. No strong preferences.

omBratteng:
```ts
await entityManager
  .createQueryBuilder()
  .update(Post)
  .set({ type: PostType.VideoYouTube })
  .where('id = :id', { id })
  .execute();
```

omBratteng: Like this ☝️ @idoshamun?

src/workers/postUpdated.ts
```diff
@@ -255,10 +255,25 @@ const updatePost = async ({
...
     .getRepository(postType)
     .findOneBy({ id });
 
+  // If we don't find the post, we need to check if it's a youtube video and
+  // try to find it again as an article
+  if (!databasePost && content_type === PostType.VideoYouTube) {
+    await entityManager
+      .createQueryBuilder()
+      .update(Post)
+      .set({ type: PostType.VideoYouTube })
```
idoshamun, line 268: Why don't we set it to every content type which is not article?

src/workers/postUpdated.ts
omBratteng: Updated to use query builder for updating

src/workers/postUpdated.ts
omBratteng: Allow it for all posts, not just YouTube?
Was thinking to just do it for YouTube for now, to not introduce any unintended side-effects (if there are any).

```ts
if (!databasePost) {
  await entityManager
    .createQueryBuilder()
    .update(Post)
    .set({ type: content_type })
    .where('id = :id', { id })
    .execute();

  databasePost = await entityManager
    .getRepository(postType)
    .findOneBy({ id });
}
```

rebelchris: Yeah think this is a good proposal, simply because we might introduce more types soon.
even know type might be article, but render as collection or whatever we might want to update it.

rebelchris (approved):
Agree with Ido, but otherwise happy

src/workers/postUpdated.ts
idoshamun: yes

omBratteng: Updated

idoshamun (approved):

src/workers/postUpdated.ts
```diff
@@ -255,10 +255,25 @@ const updatePost = async ({
   content_type = PostType.Article,
 }: UpdatePostProps) => {
   const postType = contentTypeFromPostType[content_type];
-  const databasePost = await entityManager
+  let databasePost = await entityManager
     .getRepository(postType)
     .findOneBy({ id });
 
+  // If we don't find the post, we need to check if it's a youtube video and
+  // try to find it again as an article
```
capJavert, line 263: This comment is outdated now since we do it in generic way?

omBratteng: Yes it is 😬

PR #1579, opened 2023-12-11
feat: feed side caching support

idoshamun:
We gonna run an internal test with v20 to move the responsibility of the caching to the feed side.

This means we have to respect the cursor returned by the feed and also provide it on every call.

I hacked my way around for the testing and once we confirm and I can remove all the legacy stuff, I'll make it prettier. I promise.

### DIFF
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -467,8 +467,8 @@ describe('query feed', () => {
   };
 
   const QUERY = `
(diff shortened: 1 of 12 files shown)


rebelchris (approved):

capJavert (approved):
Nothing caught my eye.

pnvasanth (approved):

PR #1582, opened 2023-12-12
feat: notification v2 real time worker

idoshamun:
In an effort to reduce the complexity of the PRs, I'm not plugging this one to production messages, but I'd like it to be available in the codebase so I won't have to create one gigantic PR.

This is the implementation of v2 for the real time notifications.

WT-2045


src/workers/newNotificationV2RealTime.ts
```diff
@@ -0,0 +1,55 @@
...
+            redisPubSub.publish(`events.notifications.${userId}.new`, {
+              ...notification,
+              attachments,
+              avatars,
+            }),
+          QUEUE_CONCURRENCY,
+        );
+      }
+    } catch (err) {
+      // Don't throw the error because we send the notification to multiple users, retrying may spam some users
```
rebelchris, line 42: Would this give us issues down the line if it errors often basically means people will not get realtime?

src/entity/notifications/NotificationV2.ts
```diff
@@ -52,3 +62,42 @@ export class NotificationV2 {
   @Column({ type: 'uuid', array: true, default: null })
   avatars: string[];
 }
+
+export const getNotificationV2AndChildren = (
```
rebelchris, line 66: I always wonder if we should include this in the entity file actually?
(Feels a bit messy)

rebelchris (approved):
Nothing breaking from my side, but maybe best to get another set of eyes.

src/entity/notifications/NotificationV2.ts
idoshamun: Agreed. Just stuck to existing practices. We have a similar function in `Notification`. Happy to move it elsewhere

rebelchris: Maybe let's set the right tone and move it away from the entity folder.
`src/notification/common` might even work?

idoshamun: Sure let me do that

rebelchris (approved):
Nothing more from my side.

PR #1583, opened 2023-12-12
feat: notification v2 push worker

idoshamun:
In an effort to reduce the complexity of the PRs, I'm not plugging this one to production messages, but I'd like it to be available in the codebase so I won't have to create one gigantic PR.

This is the implementation of v2 for the push notifications.

[WT-2045](https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11?selectedIssue=WT-2045)

[WT-2045]: https://dailydotdev.atlassian.net/browse/WT-2045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/workers/newNotificationV2Push.ts
```diff
@@ -0,0 +1,60 @@
...
+              if (users.length) {
+                await sendPushNotification(users, notification, avatars?.[0]);
+              }
+            },
+            QUEUE_CONCURRENCY,
+            BATCH_SIZE,
+          );
+        } catch (err) {
+          // Don't throw the error because we send the notification to multiple users, retrying may spam some users
+          logger.error(
```
capJavert, line 46: Would splitting this in more smaller jobs fix the issue with spam eg. this main one can publish a message to send each one. It would increase number of messages but if we moved it to separate deployment it would scale separately?

idoshamun: https://github.com/dailydotdev/daily-api/pull/1582#discussion_r1423899170
To elaborate, on a nutshell you're right. But the tradeoff is consistency vs latency. So introducing another layer of complexity will increase the lag between the event and push and ideally I'd like to minimize that as much as possible

capJavert: Yeah got it, the above would also help with personalizing notifications. I am fine with leaving it for now, don't think it would be much work to split if needed down the line 👍

capJavert (approved):

src/workers/newNotificationPush.ts
```diff
@@ -20,7 +20,7 @@ const worker: Worker = {
           notificationId: notification.id,
           order: 0,
         });
-        await sendPushNotification(notification, avatar);
+        await sendPushNotification([notification.userId], notification, avatar);
```
rebelchris, line 23: Did you not change this to userIds as in plurar and multiple?
Or does the main thing already split them (can't really recall)

rebelchris (approved):
Just some questions, but nothing breaking.

src/workers/newNotificationPush.ts
idoshamun: This is worker v1 so user id is still embedded in the notification object

PR #1585, opened 2023-12-12
fix: support sql escaping

idoshamun:
The previous implementation was sensitive to sql injections and escaping issues. I use parameterized query now to fix it

### DIFF
--- a/src/notifications/index.ts
+++ b/src/notifications/index.ts
@@ -125,33 +125,45 @@ const upsertAndReturnIds = async <Entity>(
   const table = entityManager.getRepository(entity).metadata.tableName;
   const colsStr = cols.map((col) => `"${col.toString()}"`).join(',');
   const uniqueStr = unique.map((col) => `"${col.toString()}"`).join(',');
-  const values = records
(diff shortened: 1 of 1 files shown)


rebelchris (approved):

idoshamun:
@capJavert I don't like it either but TypeORM query builder for such queries is lacking. The part that is sensitive to injections would still need to be written as raw query.
The other option is to drop this optimization, and use multiple queries to achieve the same thing. I tried to make it work in one query. Essentially creating whatever doesn't exist and then fetch everything.
I can change it to two queries. As it's a background worker, I don't mind the performance so much. Would you prefer that?

Sidenote: I enabled auto merge (sorry 🙏 ), whatever we discuss and agree I can apply on a separate PR.

rebelchris:
Hmm maybe the 2 separate using queryBuilder would be best then?

capJavert:
@idoshamun yeah, no worries. Yes I would prefer 2 queries too. 👌

idoshamun:
@capJavert @rebelchris coming right at ya

PR #1590, opened 2023-12-13
feat: notification v2 email worker

idoshamun:
Same drill, this worker is not connected to production. The next PR will cover that.

As mentioned in the DR, I utilized the `personalizations` that allows sending multiple emails in one sendgrid call. Pretty rad!
I tested that personalizations actually works as expected, I sent the same email to both my accounts.

There was lots of good old copy-paste because v1 worker will be deleted soon.

I also copied the tests and adjusted them accordingly.

WT-2045


__tests__/workers/newNotificationV2Mail.ts
DragosIorgulescu, line 1: Since these specs are covering the sendgrid API integration, would running the original `newNotificationMail` tests not be covering the V2 version as well? 

In other words, the specs seem to be testing that the API is called with the expected args, so it should work regardless of how the internal implementation is changed -- so do we need a separate test file with the same tests here?

email.ts
DragosIorgulescu, line 1: testing script? do we want this committed to the repo?

src/workers/newNotificationV2Mail.ts
```diff
@@ -0,0 +1,702 @@
...
+            notification.type
+          ](con, users, notification, attachments, avatars);
+          if (!templateData) {
+            return;
+          }
+          const formattedData = formatTemplateDate(templateData.static);
+          const personalized =
+            templateData.personalized?.map(formatTemplateDate);
+          const templateId = notificationToTemplateId[notification.type];
+          await sendEmail({
```
DragosIorgulescu, line 678: not sure how relevant it is here, but do we get any info about which emails failed since we are processing these in batches? or do we even care about that here?

DragosIorgulescu (approved):
looks good to me 👌 just added a couple non-blocking comments / questions

src/workers/newNotificationV2Mail.ts
```diff
@@ -0,0 +1,702 @@
...
+    };
+  },
+  article_report_approved: async (con, users, notification) => {
+    const post = await con
+      .getRepository(Post)
+      .findOneBy({ id: notification.referenceId });
+    return {
+      static: {
+        post_title: truncatePostToTweet(post),
+        post_image: (post as ArticlePost).image || pickImageUrl(post),
```
rebelchris, line 194: I know it's existing, wondering if this could also have side effect for video/collection etc?

rebelchris (approved):
Nothing major. have some concerns about `ArticlePost` usage but it's unrelated to these changes

email.ts
idoshamun: Good catch! Hell no

__tests__/workers/newNotificationV2Mail.ts
idoshamun: Not sure I follow. Sorry 🙏

src/workers/newNotificationV2Mail.ts
idoshamun: As long as image is defined, it's be ok. It's just a synthetic sugar for typing issue

src/workers/newNotificationV2Mail.ts
idoshamun: Good point. See this: https://github.com/dailydotdev/daily-api/pull/1583#discussion_r1424015711

PR #1594, opened 2023-12-13
feat: switch workers to notifications v2

idoshamun:
Delete old workers and use the new ones.
I validated that switching to existing topic to v2 will not have any significant impact (potentially some notifications may not be delivered during deployment but that's fine given the alternative).

I also added metrics 😎

Sidenote: switched to `toHaveBeenCalledTimes` as the other function is deprecated

WT-2045 #done


src/common/metrics.ts
```diff
@@ -0,0 +1,11 @@
+import { opentelemetry } from '../telemetry/opentelemetry';
+
+export function getNotificationFailedCounter(): opentelemetry.Counter<{
+  channel: string;
+}> {
+  const meter = opentelemetry.metrics.getMeter('api-bg');
```
rebelchris, line 6: You love to see it

src/common/metrics.ts
```diff
@@ -0,0 +1,11 @@
+import { opentelemetry } from '../telemetry/opentelemetry';
+
+export function getNotificationFailedCounter(): opentelemetry.Counter<{
+  channel: string;
```
rebelchris, line 4: Don't you want to use the channel somewhere in the counter?

src/workers/newNotificationV2Mail.ts
```diff
@@ -695,6 +697,7 @@ const worker: Worker = {
         },
         'failed to send notification email',
       );
+      counter.add(1, { channel: 'email' });
```
rebelchris, line 700: Oh I see you export the counter :) 
That makes more sense now.

rebelchris (approved):

src/workers/newNotificationV2Mail.ts
idoshamun: Type safety rules!!

PR #1595, opened 2023-12-13
feat: add collections source to seed so that local logic can have it

capJavert:
Just adding one source at the end:
```json
{"id":"collections","active":true,"name":"Collections","image":"https://daily-now-res.cloudinary.com/image/upload/s--iK6zGJCz--/f_auto,t_logo/v1698841319/logos/collections.jpg","private":false,"handle":"collections"}
```


rebelchris (approved):
Nice to have! 👍

PR #1596, opened 2023-12-13
fix: redirector and submitExternalLink on YouTube posts

omBratteng:


src/routes/redirector.ts
```diff
@@ -9,10 +9,21 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
-        where: [{ id: req.params.postId }, { shortId: req.params.postId }],
-      });
+      const q = con
+        .createQueryBuilder()
+        .select([
+          'post.id AS id',
+          'post.url AS url',
+          'post.tagsStr AS "tagsStr"',
+        ])
+        .from('post', 'post')
```
rebelchris, line 19: Why don't we use `Post, 'post'` notation here?
I think with that the select can also be cleaned.

src/schema/posts.ts
```diff
@@ -1793,10 +1793,13 @@ export const resolvers: IResolvers<any, Context> = {
         }
 
         const existingPost: Pick<ArticlePost, 'id' | 'deleted' | 'visible'> =
-          await manager.getRepository(ArticlePost).findOne({
-            select: ['id', 'deleted', 'visible'],
-            where: [{ url: cleanUrl }, { canonicalUrl: cleanUrl }],
-          });
+          await manager
+            .createQueryBuilder(Post, 'post')
+            .select(['post.id', 'post.deleted', 'post.visible'])
```
rebelchris, line 1798: I think you can omit `post` everywhere here right?

rebelchris (approved):
Nice, just some non blocking questions.

src/routes/redirector.ts
capJavert: Just `Post` is enough I think as well.

src/schema/posts.ts
omBratteng: I never get it to work without `post` or the alias that is set

capJavert (approved):

src/routes/redirector.ts
```diff
@@ -9,10 +9,21 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
-      const post = await con.getRepository(ArticlePost).findOne({
-        select: ['id', 'url', 'tagsStr'],
-        where: [{ id: req.params.postId }, { shortId: req.params.postId }],
-      });
+      const q = con
+        .createQueryBuilder()
+        .select([
+          'post.id AS id',
+          'post.url AS url',
+          'post.tagsStr AS "tagsStr"',
```
idoshamun, line 17: I believe we don't need the `AS` part and also the `post.` prefix

src/routes/redirector.ts
omBratteng: Ah, yes, we can. In earlier tests it didn't work because it would then add the WHERE type, or complain about url property not existing

src/routes/redirector.ts
omBratteng: With out `AS` it returns them as `post_id` etc

idoshamun (approved):

src/routes/redirector.ts
omBratteng: Changed it to `Post, 'post'`. But also without `AS` it returns them as post_id etc, so I added `AS` to "throw" them back to their correct name.

PR #1597, opened 2023-12-13
refactor: stop using notification v1

idoshamun:
Remove all usage of notification v1, it's a prerequisite for deleting the actual tables.

WT-2047


DragosIorgulescu (approved):
No comment here 👌

__tests__/workers/postUpdated.ts
```diff
@@ -459,7 +459,11 @@ it('should save a new post with content curation', async () => {
   });
   const posts = await con.getRepository(Post).find();
   expect(posts.length).toEqual(3);
-  expect(posts[2].contentCuration).toStrictEqual(['news', 'story', 'release']);
+  expect(posts[2].contentCuration).toIncludeSameMembers([
```
rebelchris, line 462: Leaving it here, but did you mean to remove so many test and test files?

rebelchris (approved):
Overall no concerns, was just wondering why we deleted the updater workers.
Wouldn't we still want to update notifications in those cases?

__tests__/workers/postUpdated.ts
idoshamun: Double checked them and they're relevant only for v1 so yes

PR #1599, opened 2023-12-13
feat: change the message key from `msg` to `message`

omBratteng:
This lets Google Cloud Logs Explorer to automagically show the log message instead of the entire payload

Docs: https://getpino.io/#/docs/api?id=messagekey-string

### DIFF
--- a/__tests__/cron/personalizedDigest.ts
+++ b/__tests__/cron/personalizedDigest.ts
@@ -107,7 +107,9 @@ describe('personalizedDigest cron', () => {
       })),
     );
 
-    const logger = pino();
+    const logger = pino({
+      messageKey: 'message',
+    });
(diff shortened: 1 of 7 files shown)


idoshamun (approved):

rebelchris (approved):

__tests__/cron/personalizedDigest.ts
```diff
@@ -130,7 +132,9 @@ describe('personalizedDigest cron', () => {
       })),
     );
 
-    const logger = pino();
+    const logger = pino({
+      messageKey: 'message',
+    });
```
kopancek, line 137: Do we need multiple instances of logger or can we create a static one per file or maybe even per app? That way, we could centralise this configuration.

kopancek (approved):

PR #1600, opened 2023-12-14
feat: fetch similar posts from feed service

idoshamun:
Feed service supports now returning similar posts which is way better than our current implementation. Not all posts support that, hence I'm keeping the old implementation as well for now.

### DIFF
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -1276,7 +1276,41 @@ describe('query randomSimilarPostsByTags', () => {
     }
   }`;
 
-  it('should return random similar posts by tags', async () => {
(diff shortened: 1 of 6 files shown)


rebelchris (approved):
Looks good to me at first glance.
I expect old posts cant resolve like this?
Does it mean we have to support it forever?

idoshamun:
@rebelchris 

> I expect old posts cant resolve like this?
Once we confirm it works, Viktor will import quite a significant amount of old data

> Does it mean we have to support it forever?
We can simplify the algorithm to something much simpler, or we can add this logic to the feed service which will be better imo

PR #1604, opened 2023-12-14
chore: early return on worker to avoid logs

sshanzel:
In local environment, when you delete an account, the background worker would run endlessly trying to execute the subscription below if there is no sendgrid api key.

It is proving difficult to keep track of the logs due to the error prompting from the worker.


idoshamun (approved):

capJavert (approved):
🙏 awesome!

rebelchris (approved):
Amazing amazing

PR #1605, opened 2023-12-14
refactor: notification attachment to support video

sshanzel:
The nature of our notification attachment is, after generating it, the entity won’t be attached to the relevant entity anymore. Meaning, when rendering it, we won’t know whether the relevant entity came from a video in order for us to add a play button overlay. We need to have a column to tell us that the attachment was from a video-related entity.

WT-2055


src/entity/notifications/NotificationAttachmentV2.ts
```diff
@@ -20,4 +20,7 @@ export class NotificationAttachmentV2 {
 
   @Column({ type: 'text' })
   referenceId: string;
+
+  @Column({ type: 'bool', default: false })
+  isAttachmentVideo: boolean;
```
idoshamun, line 25: May I suggest utilizing the attachment type instead? It already exists
`isAttachmentVideo` is very specific

sshanzel: We currently use the type whether it is a `post` or something, but we can probably do that. Seeing we haven't really utilized the attachment type in the FE.

sshanzel: @rebelchris @omBratteng thoughts?

idoshamun: The problem with the current approach is that it's very specific. Tomorrow we would want to add a podcast for example, you will have to introduce another boolean field.

sshanzel: Agreed, let me refactor the part we generate post attachments and identify whether it was a video or not.

src/entity/notifications/index.ts
```diff
@@ -8,7 +8,4 @@ export * from './NotificationPreference';
 export * from './NotificationPreferencePost';
 export * from './NotificationPreferenceSource';
 export * from './NotificationPreferenceComment';
-export * from './NotificationV2';
-export * from './NotificationAvatarV2';
-export * from './NotificationAttachmentV2';
```
sshanzel, line 13: I noticed this was exported twice. Duplicate of line 4-6.

rebelchris:
We also need to ensure the mailing as well as frontend notifications have the extras render

sshanzel:
> We also need to ensure the mailing as well as frontend notifications have the extras render

Good point! I will check whether the change must be on the email template's end. For the FE, yes, I will raise a separate PR for that once this is concluded.

idoshamun (commented):
Looks good! Can you please add tes?

sshanzel:
> Looks good! Can you please add tes?

Oh, right. Let me spin up one, quickly.

idoshamun (commented):
When I think about it, I believe we have a db trigger based on the type which will need to be adjusted. Can you please verify and also check for other things that may break?

rebelchris:
@omBratteng Can you maybe see if you can open this PR to someone else with Lee being away for a week.

It won't block release of video, but would definitely be a nice to have.
(And think we are almost there)

capJavert:
> When I think about it, I believe we have a db trigger based on the type which will need to be adjusted. Can you please verify and also check for other things that may break?

@idoshamun triggers wise I think we are good, the type referred in `delete_post_notifications` trigger refers to `referenceType` not attachment type.

src/notifications/builder.ts
```diff
@@ -168,8 +174,11 @@ export class NotificationBuilder {
   }
 
   attachmentPost(post: Reference<Post>): NotificationBuilder {
+    const type =
+      postTypeToAttachmentType[post.type] ?? NotificationAttachmentType.Post;
```
capJavert, line 178: We might sync this after the notification creation with trigger because shared post do not get video type immediately.

idoshamun (approved):

src/notifications/builder.ts
capJavert: Added here 9c3fd2642e5e5ce17b629df8a88b15b58353ecb0

__tests__/notifications/index.ts
```diff
@@ -326,6 +329,90 @@ describe('generateNotification', () => {
     ]);
   });
 
+  it('should update notification attachment when post type changes', async () => {
```
idoshamun, line 331: We have triggers test on a separate file if not mistaken

idoshamun (approved):
Non blocking comment

omBratteng (approved):

rebelchris (approved):
Looks good to me, nothing breaking :)

__tests__/notifications/index.ts
capJavert: Sorted 478b4a0180c97ba39240250aeaca676ba8ab00c8

sshanzel:
@dailydotdev/web-team thanks for covering! 🥳

PR #1603, opened 2023-12-14
chore: upgrade postgres to v15 in CircleCI pipelines

omBratteng:
As a preparation for the upgrade to psql 15 in production, we update the circleci version to 15 too.

PR #1608, opened 2023-12-14
fix: do not flip visible condition when post is already visible

capJavert:
Basically a bug when post was already visible the condition was flipped with becomes visible. Everything else logic wise looks fine and platform DOES send whole payload during updates. 

Requirements for post visible updates:
- visible is once off, once post is visible it can't become invisible
- post becomes visible if it has the title
- post visibleAt is updated when post becomes visible (default null during creation)

### DIFF
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -459,7 +459,14 @@ it('should save a new post with content curation', async () => {
   });
   const posts = await con.getRepository(Post).find();
   expect(posts.length).toEqual(3);
-  expect(posts[2].contentCuration).toStrictEqual(['news', 'story', 'release']);
+  const curatedPost = posts.find(
+    (item) => item.yggdrasilId === 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+  );
+  expect(curatedPost!.contentCuration).toStrictEqual([
+    'news',
+    'story',
+    'release',
+  ]);
 });
 
 it('save a post as public if source is public', async () => {
@@ -761,6 +768,31 @@ describe('on post update', () => {
       id: postId,
     });
     expect(updatedPost!.title).toEqual('New title 2');
+  });
+
+  it('should retain visible state when title is present', async () => {
+    const postId = 'p1';
+
+    const existingPost = await con.getRepository(ArticlePost).save({
+      id: postId,
+      title: 'Post title',
+      yggdrasilId: 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+      visible: true,
+    });
+
+    expect(existingPost).not.toBeNull();
+    expect(existingPost.visible).toEqual(true);
+
+    await expectSuccessfulBackground(worker, {
+      id: 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+      post_id: undefined,
+      title: 'New title 2',
+    });
+
+    const updatedPost = await con.getRepository(ArticlePost).findOneBy({
+      id: postId,
+    });
+    expect(updatedPost!.visible).toEqual(true);
   });
 });
 
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -323,6 +323,7 @@ it('should update freeform post and only modify allowed fields', async () => {
   expect(post.metadataChangedAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
   expect(post.visible).toEqual(true);
   expect(post.flags.visible).toEqual(true);
+  expect(post.visibleAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
   expect(post.contentCuration).toEqual(['news', 'story', 'release']);
   expect(post.yggdrasilId).toEqual('f99a445f-e2fb-48e8-959c-e02a17f5e816');
   expect(post.title).toEqual('freeform post');
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -738,6 +738,30 @@ describe('on post update', () => {
       );
     });
   });
+
+  it('should resolve post id from yggdrasil id when post id is missing', async () => {
+    const postId = 'p1';
+
+    const existingPost = await con.getRepository(ArticlePost).save({
+      id: postId,
+      title: 'Post title',
+      yggdrasilId: 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+    });
+
+    expect(existingPost).not.toBeNull();
+    expect(existingPost.title).toEqual('Post title');
+
+    await expectSuccessfulBackground(worker, {
+      id: 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+      post_id: undefined,
+      title: 'New title 2',
+    });
+
+    const updatedPost = await con.getRepository(ArticlePost).findOneBy({
+      id: postId,
+    });
+    expect(updatedPost!.title).toEqual('New title 2');
+  });
 });
 
 describe('on youtube post', () => {
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -323,7 +323,6 @@ it('should update freeform post and only modify allowed fields', async () => {
   expect(post.metadataChangedAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
   expect(post.visible).toEqual(true);
   expect(post.flags.visible).toEqual(true);
-  expect(post.visibleAt).toEqual(new Date('2023-01-05T12:00:00.000Z'));
   expect(post.contentCuration).toEqual(['news', 'story', 'release']);
   expect(post.yggdrasilId).toEqual('f99a445f-e2fb-48e8-959c-e02a17f5e816');
   expect(post.title).toEqual('freeform post');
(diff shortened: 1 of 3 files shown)


__tests__/workers/postUpdated.ts
```diff
@@ -459,7 +459,14 @@ it('should save a new post with content curation', async () => {
...
   expect(posts.length).toEqual(3);
-  expect(posts[2].contentCuration).toStrictEqual(['news', 'story', 'release']);
+  const curatedPost = posts.find(
+    (item) => item.yggdrasilId === 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+  );
+  expect(curatedPost!.contentCuration).toStrictEqual([
+    'news',
+    'story',
+    'release',
+  ]);
```
capJavert, line 469: This is flaky and kept failing while I was fixing the issue so fixed it as well. Basically db does not guarantee order of rows.

src/workers/postUpdated.ts
```diff
@@ -301,7 +301,6 @@ const updatePost = async ({
 
   data.id = databasePost.id;
   data.title = title;
-  data.visible = updateBecameVisible;
```
rebelchris, line 304: But deleting this wont actually set community picks to visible so not sure this is right?

capJavert: Can you specify exact case where it would break?

rebelchris: Just from my head:
User adds community pick
Think we add as not visible?
Then yggdrasil send update
Becomes visible

capJavert: `should update the post and make it visible if title is available` test confirms that case works. `p1` is `visible: false`, then update with title is sent, post becomes visible.

rebelchris: Ah ok if thats the case then its fine, but then what did this evaluation do?
Do we set it somewhere else as well?

__tests__/workers/postUpdated.ts
idoshamun: I fixed it on one of my PRs, you can have a look

src/workers/postUpdated.ts
idoshamun: I think the right fix is actually changing the condition of updateBecameVisible. The bug there is that we flip the visibility but actually what we need is to define it as true if title is defined. Like Chris I'm not sure in the current implementation how it ever becomes true 🥵

capJavert: @rebelchris we set it in `fixData`, that is where `data.visible` comes from and it is set.

src/workers/postUpdated.ts
```diff
@@ -301,7 +301,6 @@ const updatePost = async ({
...
   data.visibleAt = updateBecameVisible
```
capJavert, line 304: Checking it one more time, @idoshamun how should `visibleAt` function, needs to be set each time post becomes visible (changes from false to true) or only the first time? Also if post is not visible it should be `null`?

idoshamun: I'm ooo, so can't review completely but let me clarify the business logic:
So if a post doesn't have a title, it should have visible set to false and visible at null. Once it has a title, we set visible to true and visible at accordingly. Visible is a one way street. Once a post is visible it can't become invisible

capJavert: > Visible is a one way street. Once a post is visible it can't become invisible

This is the crucial info I was missing, and also missing from our worker.

rebelchris: I do think now the title always persist from db if exist so technically couldnt ever be flipped afaik.
But indeed its not explicitly known.

__tests__/workers/postUpdated.ts
capJavert: I'll revert then.

omBratteng: I also fixed that it one of my PRs 😬

omBratteng: In #1603, running the tests 2 more times to see that nothing else is flaky with psql15

src/workers/postUpdated.ts
capJavert: We can continue at https://github.com/dailydotdev/daily-api/pull/1608#discussion_r1427751683

src/workers/postUpdated.ts
```diff
@@ -294,19 +300,25 @@ const updatePost = async ({
...
-    content_type === PostType.Freeform
-      ? databasePost.visible
-      : !databasePost.visible && !!title?.length;
 
   data.id = databasePost.id;
   data.title = title;
-  data.visible = updateBecameVisible;
-  data.visibleAt = updateBecameVisible
-    ? databasePost.visibleAt ?? data.metadataChangedAt
-    : null;
```
capJavert, line 307: Removed it from `fixData` because we need to know if current post exist to properly deduce current visible status.

src/workers/postUpdated.ts
```diff
@@ -203,6 +204,12 @@ const createPost = async ({
   data.origin = data?.scoutId
     ? PostOrigin.CommunityPicks
     : data.origin ?? PostOrigin.Crawler;
+  data.visible = getPostVisible({ post: data });
+  data.visibleAt = data.visible ? postCreatedAt : null;
+  data.flags = {
+    ...data.flags,
+    visible: data.visible,
+  };
```
capJavert, line 212: Because of removal of https://github.com/dailydotdev/daily-api/pull/1608/files#r1427897526 we set the default visible status here during creation.

src/workers/postUpdated.ts
```diff
@@ -485,7 +495,6 @@ const fixData = async ({
       showOnFeed: !data?.order,
       flags: {
         private: privacy,
-        visible: becomesVisible,
```
capJavert, line 488: Removed it from `fixData` because we need to know if current post exist to properly deduce current visible status.

src/entity/posts/utils.ts
```diff
@@ -504,3 +504,7 @@ export const normalizeCollectionPostSources = async ({
     collectionSources: distinctSources.map((item) => item.id),
   });
 };
+
+export const getPostVisible = ({ post }: { post: Pick<Post, 'title'> }) => {
+  return !!post?.title?.length;
+};
```
capJavert, line 510: Added common util so we can reuse and not check title all over the place.

src/workers/postUpdated.ts
```diff
@@ -294,19 +301,25 @@ const updatePost = async ({
   }
 
   const title = data?.title || databasePost.title;
-  const updateBecameVisible =
-    content_type === PostType.Freeform
-      ? databasePost.visible
```
capJavert, line 299: Not sure why this was here but `should update freeform post and only modify allowed fields` confirms this works still.

src/workers/postUpdated.ts
```diff
@@ -294,19 +301,25 @@ const updatePost = async ({
...
+    updateBecameVisible = getPostVisible({ post: { title } });
+  }
+
+  if (updateBecameVisible) {
+    data.visible = true;
+    data.visibleAt = data.metadataChangedAt;
+  } else {
+    data.visible = databasePost.visible;
+    data.visibleAt = databasePost.visibleAt ?? data.metadataChangedAt;
+  }
```
capJavert, line 321: Post can become visible only once, if it became visible we set the state to `true` and set `visibleAt`. In all other cases (if already visible or still `false` we just use the current values from DB.

src/workers/postUpdated.ts
```diff
@@ -294,19 +301,27 @@ const updatePost = async ({
...
+
+  if (updateBecameVisible) {
+    data.visible = true;
+  } else {
+    data.visible = databasePost.visible;
+  }
+
+  if (data.visible && !databasePost.visibleAt) {
+    data.visibleAt = data.metadataChangedAt;
+  }
```
capJavert, line 323: if post is visible but missing `visibleAt` set it here. This fixes current issue in production where we have a bunch of posts that are `visible: true` but `visibleAt: null`.

Check:
```
select count(*) from post p where p.visible = true and p."visibleAt" is null
```

src/workers/postUpdated.ts
```diff
@@ -294,19 +301,27 @@ const updatePost = async ({
...
+
+  if (!databasePost.visible) {
+    updateBecameVisible = getPostVisible({ post: { title } });
+  }
+
+  if (updateBecameVisible) {
+    data.visible = true;
+  } else {
+    data.visible = databasePost.visible;
+  }
```
capJavert, line 319: Post can become visible only once, In all other cases (if already visible or still `false` we just use the current values from DB.

src/workers/postUpdated.ts
capJavert: @rebelchris @idoshamun I refactored the logic, hopefully is more streamlined. All tests pass and I added to additional to validate cases Ido metnioned: 
- visible is once off, once post is visible it can't become invisible
- post becomes visible if it has the title
- post visibleAt is updated when post becomes visible (default null during creation)

src/workers/postUpdated.ts
rebelchris: I think it had to do with freeform are always visible.

rebelchris (approved):
Evaluation looks good to me now though!
Thanks for checking this out and fixing it.

src/workers/postUpdated.ts
capJavert: Yeah I think so too, but now it will be handled in `createPost` since freeform has the title during creation.

idoshamun (approved):
Thanks a lot for addressing all the issues and requirements

PR #1610, opened 2023-12-15
feat: exclude post content types from feed

omBratteng:
Discovered that it actually didn't filter out posts with type `video:youtube` as the methods that already existed only took into account sources.

This updates the AdvancedSettings table by adding a new column `options` that takes jsonb content, which lets me avoid having to hard code a mapping between settings based on title or something else.

The `feedToFilters` has been updated to returned any disabled content types based on feed advanced settings, which means we actually can add options to disable any content type by adding new rows in the database.

WT-1986


idoshamun (approved):

rebelchris (approved):
Thanks for fixing the seeds as well 🙏

PR #1614, opened 2023-12-18
refactor: split notifications to a new cdc handler

idoshamun:
To overcome the ordering constraint of CDC, we separate the notifications to a new topic to allow much faster handling with spikes

WT-2048 #done

### DIFF
--- a/.infra/application.properties
+++ b/.infra/application.properties
@@ -13,10 +13,13 @@ debezium.source.plugin.name=pgoutput
 debezium.source.heartbeat.interval.ms=60000
 debezium.source.topic.prefix=t
 debezium.source.tombstones.on.delete=false
-debezium.transforms=Reroute,ReadOperationFilter,PostsFilter
+debezium.transforms=Reroute,Notifications,ReadOperationFilter,PostsFilter
 debezium.transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
-debezium.transforms.Reroute.topic.regex=(.*)
+debezium.transforms.Reroute.topic.regex=^((?!\.notification_v2).)*$
 debezium.transforms.Reroute.topic.replacement=%topic%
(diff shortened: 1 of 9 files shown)


rebelchris (approved):
Perhaps wait for a second set of eyes.
Looks ok, but not 100% sure on the debezium config myself.

idoshamun:
@capJavert can you have another look please? you also fought debezium quite a lot

capJavert (approved):
Looks alright, one question, if we deploy this and for example have X notifications to process they would be lost due to original cdc no longer handling it or would Debezium re-route?

.infra/application.properties
```diff
@@ -13,10 +13,13 @@ debezium.source.plugin.name=pgoutput
 debezium.source.heartbeat.interval.ms=60000
 debezium.source.topic.prefix=t
 debezium.source.tombstones.on.delete=false
-debezium.transforms=Reroute,ReadOperationFilter,PostsFilter
+debezium.transforms=Reroute,Notifications,ReadOperationFilter,PostsFilter
 debezium.transforms.Reroute.type=io.debezium.transforms.ByLogicalTableRouter
-debezium.transforms.Reroute.topic.regex=(.*)
+debezium.transforms.Reroute.topic.regex=^((?!\.notification_v2).)*$
```
capJavert, line 18: Additionally, would this match, our topic does not have `_v2` in its name?

capJavert: Or is this a table name match?

idoshamun: This table name match. It took me a while to find the right regex but I tested everything locally. It was a tough fight. We no longer have v1 tables btw

capJavert: roger, roger! good then 👍

capJavert (approved):

PR #1615, opened 2023-12-19
fix: ignore dates before unix time

capJavert:
Platform can send date strings before UNIX timestamp `0001-01-01T00:00:00Z` as empty values, those get parsed as valid Dates in Node but in DB they are nulls. Thread [here](https://dailydotdev.slack.com/archives/C0601MRBJJU/p1702974255027129).

Current issue is avoided during creation because TypeORM filters out extra columns with `create` but on update it tries to write it out. 

Since those values are empty I added the new `parseDate` which handles the case for all date parsing in `fixData` function in our updated worker. It also ignores any falsy or invalid date values. 

Alternative was to add `allowedFieldsMapping` for collection type but seemed wasteful since we don really need publishedAt at the first place since its empty.

### DIFF
--- a/__tests__/commons/utils.ts
+++ b/__tests__/commons/utils.ts
@@ -2,6 +2,7 @@ import { DataSource } from 'typeorm';
 import createOrGetConnection from '../../src/db';
 import { ArticlePost, Post, PostTag, Source } from '../../src/entity';
 import {
+  parseDate,
   removeEmptyValues,
   removeSpecialCharacters,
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Looks good to me 🗡️

omBratteng (approved):

kopancek (approved):
🚢

idoshamun (approved):

PR #1618, opened 2023-12-20
feat: add db:seed:upsert npm script

kopancek:
Adds `--upsert` option to `import.ts` script and new npm script `db:seed:upsert`.

This allows the import script to try to upsert all of the seeds, even if there are errors/duplicates on some of the insert calls. In case there's a conflict, it updates the row instead.

Without this option the script fails on the first error/duplicate row it finds and never inserts the rest of the entities.

### DIFF
--- a/bin/import.ts
+++ b/bin/import.ts
@@ -3,25 +3,40 @@ import createOrGetConnection from '../src/db';
 import { DataSource } from 'typeorm';
 import { TagRecommendation } from '../src/entity/TagRecommendation';
 
-const importEntity = async (con: DataSource, name: string): Promise<void> => {
+const importEntity = async (
+  con: DataSource,
+  name: string,
+  throwOnError = true,
+): Promise<void> => {
   console.log(`importing ${name}`);
   const entities = JSON.parse(readFileSync(`./seeds/${name}.json`).toString());
   const repository = con.getRepository(name);
+
   // Batch insert with dirty hack
   for (let i = 0; i < entities.length; i += 1000) {
-    await repository.insert(entities.slice(i, i + 1000));
+    await repository.insert(entities.slice(i, i + 1000)).catch((err) => {
+      if (throwOnError) {
+        throw err;
+      } else {
+        // Swallow the error if we're ignoring them
+        // this will allow the next importEntity call to run
+        console.error(err);
+      }
+    });
   }
 };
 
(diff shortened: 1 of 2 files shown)


idoshamun (approved):
💪

capJavert (commented):
In which case is this needed, only when you already have some data in the db?

kopancek:
> In which case is this needed, only when you already have some data in the db?

Yes, exactly. But I figured out how to do upsert properly, so will update this PR...

capJavert (commented):
To me personally this seems like too much logic for little gains, its a rare case where you actually need to keep local data and also when we modify the seed. It is easier to just do a schema reset with `db:migrate:reset` and then seed again. 

Maybe as alternative would just `repository.save` be a better option if we want to support the multiple calls of seed?

rebelchris:
I personally also prefer to simply redo the seed all together.

kopancek:
> To me personally this seems like too much logic for little gains 🙏 , its a rare case where you actually need to keep local data and also when we modify the seed. It is easier to just do a schema reset with `db:migrate:reset` and then seed again.
> 
> Maybe as alternative would just `repository.save` be a better option if we want to support the multiple calls of seed?

I don't know, I find it quite useful for local development to keep the test data intact and just seed new things. Especially if someone else upgraded the DB schema or seeds and I don't really care about those changes, just want everything to work nicely.

It also already happened to me more than once that I ran the import seed script and it failed with an error and I didn't understand how to reset the DB state.

If we decide to scrap this, at least I now know that `db:migrate:reset` resets the DB. And I know how to work with `upsert` from typeorm, so learned something 😊

capJavert:
> It also already happened to me more than once that I ran the import seed script and it failed with an error and I didn't understand how to reset the DB state.

yeah sorry for that, this was due to bug that was fixed https://github.com/dailydotdev/daily-api/commit/b682d04feacf628b00661883e4a7aa7524215044, now seed will go through

sshanzel (approved):
Since it is in a separate command, I am cool with it as it has proven beneficial to some. However, personally, I would probably be sticking with the original commands to do a clean slate whenever I want to import data.

nensidosari (approved):
Should we merge or close this PR? It would be good if we would just take a decision.
Personally I think since it improves lives for some people, and has no downsides, why not have it?

idoshamun (approved):

PR #1627, opened 2023-12-27
feat: support for profile v2

idoshamun:
* User's readme support including rendering the readme to markdown
* Cover image support including uploading to cloudinary
* Expose stats to everyone

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -334,6 +334,31 @@ const getExperimentation = async (
   };
 };
 
+const getUser = (con: DataSource, userId: string): Promise<User> =>
+  con.getRepository(User).findOne({
+    where: { id: userId },
+    select: [
+      'id',
+      'username',
+      'name',
+      'email',
+      'image',
+      'company',
+      'title',
+      'infoConfirmed',
+      'notificationEmail',
+      'acceptedMarketing',
+      'reputation',
+      'bio',
+      'twitter',
+      'github',
+      'portfolio',
+      'hashnode',
+      'timezone',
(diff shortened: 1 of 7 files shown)


src/routes/boot.ts
```diff
@@ -334,6 +334,31 @@ const getExperimentation = async (
...
+      'id',
+      'username',
+      'name',
+      'email',
+      'image',
+      'company',
+      'title',
+      'infoConfirmed',
+      'notificationEmail',
+      'acceptedMarketing',
```
rebelchris, line 350: Do we need these in the fe from boot?

src/schema/users.ts
```diff
@@ -661,7 +709,6 @@ export const resolvers: IResolvers<any, Context> = {
       { id }: { id: string },
       ctx: Context,
     ): Promise<GQLUserStats | null> => {
-      const isSameUser = ctx.userId === id;
```
rebelchris, line 664: We no longer want this eval?

rebelchris (commented):
Small questions

src/schema/users.ts
idoshamun: Nope, we gonna show these metrics publicly for now

src/routes/boot.ts
idoshamun: So, I didn't want to take risks in this PR, but I definitely know we don't need the new fields I added, which can be quite big. Hence, I manually selected what we had before. I think we should optimize it and see what's needed, but that's out of this scope. The previous implementation just selected everything

PR #1631, opened 2024-01-03
fix: sources should not be deduplicated

rebelchris:
Previously we ensured sources where distinct de-duplicated, however according to new insights we should allow all sources including duplicates.

Secondary effort for this task will be to run a query to update existing post-relations and once-off update the `collectionSources` to reflect this change. (My suggestion is a query, but please let me know if you have different ideas)


idoshamun (approved):
Thanks for the quick fix!

PR #1632, opened 2024-01-05
feat: added a profile picture redirector

rebelchris:
Added a public redirector endpoint where based on `userId` you can get redirected to user's profile image.

Usage as:
```
api.local.com/{userId}/profile-image
```

it will fallback to the placeholder image if user doesn't exist or has no image set in DB.

### DIFF
--- a/__tests__/redirector.ts
+++ b/__tests__/redirector.ts
@@ -1,13 +1,14 @@
 import appFunc from '../src';
 import { FastifyInstance } from 'fastify';
 import { saveFixtures, TEST_UA } from './helpers';
-import { ArticlePost, Source, YouTubePost } from '../src/entity';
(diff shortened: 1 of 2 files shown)


omBratteng (approved):

capJavert (approved):
Looks good!

For what do we/want to use this for?

Also not sure for fastify, which status code 3xx is default for redirect, we probably want 307 temp since user can change profile picture.

rebelchris:
@capJavert The default is 302 which also indicates it's [temporary](https://www.baeldung.com/cs/redirection-status-codes#302-redirection).
Is that alright?

capJavert:
Yeah, fine as well in this case 👍

idoshamun (approved):
Nice!

PR #1633, opened 2024-01-05
fix: turn of changelog in favor of squad promotion

rebelchris:
This should disable the changelog as we want to favor it being a squad now to promote more organic.
I left the code if the experiment has negative impact and we want to revert it back quickly.


capJavert (commented):
I think just remove the commented code 🔪 , revert exists so we can always get the code back that way. You can even link the commit/pr to be reverted in the comment you already added in `boot.ts`?

idoshamun (approved):

capJavert (approved):

idoshamun (approved):

PR #1638, opened 2024-01-11
fix: create rules for protecting the ghost user from accidental deletion or modification

omBratteng:
As discovered while debugging deletion of users, I noticed that the 404 / ghost user had disappeared.
So this migration creates a rule to prevent `DELETE` and `UPDATE` of the row.

### DIFF
--- a/src/migration/1704948666033-GhostUserProtect.ts
+++ b/src/migration/1704948666033-GhostUserProtect.ts
@@ -4,11 +4,13 @@ export class GhostUserProtect1704948666033 implements MigrationInterface {
     name = 'GhostUserProtect1704948666033'
 
     public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`UPDATE "user" SET "infoConfirmed" = true WHERE "id" = '404';`);
       await queryRunner.query(`CREATE RULE prototect_ghostuser_deletion AS ON DELETE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
       await queryRunner.query(`CREATE RULE prototect_ghostuser_update AS ON UPDATE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
     }
 
     public async down(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`UPDATE "user" SET "infoConfirmed" = false WHERE "id" = '404';`);
       await queryRunner.query(`DROP RULE prototect_ghostuser_deletion on "user";`);
       await queryRunner.query(`DROP RULE prototect_ghostuser_update on "user";`);
     }
--- a/src/migration/1704948666033-GhostUserProtect.ts
+++ b/src/migration/1704948666033-GhostUserProtect.ts
@@ -6,12 +6,10 @@ export class GhostUserProtect1704948666033 implements MigrationInterface {
     public async up(queryRunner: QueryRunner): Promise<void> {
       await queryRunner.query(`UPDATE "user" SET "infoConfirmed" = true WHERE "id" = '404';`);
       await queryRunner.query(`CREATE RULE prototect_ghostuser_deletion AS ON DELETE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
-      await queryRunner.query(`CREATE RULE prototect_ghostuser_update AS ON UPDATE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
     }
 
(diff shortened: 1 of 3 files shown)


src/migration/1704948666033-GhostUserProtect.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class GhostUserProtect1704948666033 implements MigrationInterface {
+    name = 'GhostUserProtect1704948666033'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`CREATE RULE prototect_ghostuser_deletion AS ON DELETE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
+      await queryRunner.query(`CREATE RULE prototect_ghostuser_update AS ON UPDATE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
```
omBratteng, line 8: The `UPDATE` rule is optional, as it might cause some headaches in future migrations if we want to modify something on all users, and forget about the rule.

src/migration/1704948666033-GhostUserProtect.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class GhostUserProtect1704948666033 implements MigrationInterface {
+    name = 'GhostUserProtect1704948666033'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`CREATE RULE prototect_ghostuser_deletion AS ON DELETE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
```
omBratteng, line 7: Can replaced `IN ('404')` with `= '404'`. Just wanted to check if there were any other special users we want to protect.

src/migration/1704948666033-GhostUserProtect.ts
rebelchris: Yeah update is fine to remove it think, we sometimes change default image etc.

rebelchris (approved):
Nice addition, maybe remove update but lets see what ido says 👍

src/migration/1704948666033-GhostUserProtect.ts
omBratteng:
```suggestion
```

src/migration/1704948666033-GhostUserProtect.ts
```diff
@@ -0,0 +1,15 @@
...
+      await queryRunner.query(`CREATE RULE prototect_ghostuser_update AS ON UPDATE TO "user" WHERE old.id IN ('404') DO INSTEAD nothing;`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`DROP RULE prototect_ghostuser_deletion on "user";`);
+      await queryRunner.query(`DROP RULE prototect_ghostuser_update on "user";`);
```
omBratteng, line 13:
```suggestion
```

src/migration/1704948666033-GhostUserProtect.ts
idoshamun: It's ok to use IN for this purpose

src/migration/1704948666033-GhostUserProtect.ts
idoshamun: Agree, let's remove it. Especially if we want to make it a little bit more funny like you suggested

idoshamun (approved):
Let's just remove the update rule

src/migration/1704948666033-GhostUserProtect.ts
omBratteng: Removed it.

src/migration/1704948666033-GhostUserProtect.ts
omBratteng: Think so too. Also makes it easier (like 2 seconds saved ⏲️) in the future if we want to expand this rule

sshanzel:
Now this makes me think that it is probably related to the regular deleting of account. Transferring the ownership of the post doesn't work which ultimately breaks the process.

omBratteng:
@sshanzel yeah, regular deletion of users failed because it tried to transfer ownership of posts and/or comments to the ghost user which did not exist anymore, hence it failed.
If a user had no posts and/or comments, user deletion worked just fine.

PR #1639, opened 2024-01-11
fix: do nothing in `updateMailingList` if user is ghost user

omBratteng:
This fixes a long hanging unacked pubsub message on the `user-updated-api-mailing` subscription because we had to re-add the ghost user to the database.


capJavert (commented):
I think just missing a test for this, otherwise 👍

omBratteng:
Added test for this case

capJavert (approved):

idoshamun (approved):

PR #1641, opened 2024-01-11
fix: give the ghost user a ghostly username 👻

omBratteng:
<table>
<tr>
 <td>Before
 <td>
<tr>
 <td>After
 <td>
</table>

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -5,7 +5,8 @@ const REMOVE_SPECIAL_CHARACTERS_REGEX = /[^a-zA-Z0-9-_#.]/g;
 
 export const ghostUser = {
   id: '404',
-  username: 'inactive_user',
+  username: 'ghost',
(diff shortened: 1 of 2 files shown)


capJavert (commented):
I think there was a [discussion before](https://dailydotdev.slack.com/archives/C03QZBF46JC/p1683629548726909) on how we should call the user so not sure this rename is needed. No strong feelings about it though but might want to align with product and everyone else.

omBratteng:
There are ongoing [discussions here](https://dailydotdev.slack.com/archives/C02QDJFKUJ1/p1704951468237789), also raised the rename

omBratteng:
Converted it to draft while we wait for readme copy and an updated avatar 👻

rebelchris (approved):
Cool, nothing blocking from my side.

idoshamun (approved):

PR #1642, opened 2024-01-15
refactor: remove feed cache implementation

idoshamun:
Now that feed service is in charge of caching, we can remove it completely from the api as promised :)


rebelchris (approved):
Nice!

omBratteng (approved):

PR #1643, opened 2024-01-15
test: trying to fix snapshot issue

idoshamun:


omBratteng:
So, the snapshot it is getting caught in is search.ts.snap, but search doesn't actually test those.

Added in #1640, not sure why the search.ts.snap contains those snapshots

omBratteng:
oooh, I see why...

rebelchris (approved):
Gotta hate snapshots

omBratteng (approved):

PR #1646, opened 2024-01-17
fix: hotfix too many posts returned from feed

idoshamun:
Since we no longer cache locally the feed, we need to limit the total pages to 1 not 40 :P


omBratteng (approved):

rebelchris (approved):
This is fine 🥶😂

PR #1649, opened 2024-01-19
fix: latest package bumps.

rebelchris:
Added some package bumps.
Remaining outstanding we can't do because of moves to ESM only.

Something to keep in mind:
- graphql-upload changed to graphql-upload-minimal in mercurius upload, we might want to also do this?
- EDIT: above i reverted as it was clearly not inline with what we use for upload types, need to look into it
- pubsub major bump


omBratteng (approved):

package.json
```diff
@@ -28,124 +28,124 @@
...
+    "@swc/core": "^1.3.104",
     "@types/express": "^4.17.21",
-    "@types/jest": "^29.5.8",
+    "@types/jest": "^29.5.11",
     "@types/jsonwebtoken": "^9.0.5",
-    "@types/lodash": "^4.14.201",
+    "@types/lodash": "^4.14.202",
     "@types/node": "^18.16.17",
-    "@types/node-fetch": "^2.6.9",
+    "@types/node-fetch": "^2.6.11",
```
capJavert, line 123: I don't think we really need node-fetch anymore since Node 18.x has native fetch, one thing to check are the workers, but I think they run on the same version.

capJavert (approved):

package.json
rebelchris: Hmm agreed, but i'll leave it as seperate from the bumps to avoid complications

idoshamun (approved):
Perfect!

PR #1651, opened 2024-01-23
feat: streak table

sshanzel:
Note: triggers will be for a different PR.

Entity for UserStreak purely from the ticket's desired structure.

Apologies for the many file changes, I just organized `User` entities into a single folder. To make the review easier, changes related to the new entity should be found here: https://github.com/dailydotdev/daily-api/commit/7217c399df69b27822ee615c94e0a3cd875d4f64 and from this file https://github.com/dailydotdev/daily-api/blob/ab7756ca027b9a339cebf255a339b1ebc329bbf9/src/entity/user/UserStreak.ts

MI-68 #done


src/entity/user/UserStreak.ts
```diff
@@ -0,0 +1,23 @@
+import { Index, Column, PrimaryColumn, Entity } from 'typeorm';
+
+@Entity()
+export class UserStreak {
+  @PrimaryColumn({ type: 'text' })
+  @Index()
```
idoshamun, line 6: PrimaryColumn is already indexed, we don't need that

src/entity/user/UserStreak.ts
```diff
@@ -0,0 +1,23 @@
...
+  userId: string;
```
idoshamun, line 7: You're missing a foreign key definition here

src/entity/user/UserStreak.ts
sshanzel: Should be fixed now 🚀

src/entity/user/UserStreak.ts
sshanzel: Applied the definition 🥳

src/entity/user/UserStreak.ts
```diff
@@ -0,0 +1,29 @@
+import { Column, PrimaryColumn, Entity, OneToOne } from 'typeorm';
+import { User } from './User';
+
+@Entity()
+export class UserStreak {
+  @PrimaryColumn({ type: 'text' })
+  userId: string;
+
+  @OneToOne(() => User, {
```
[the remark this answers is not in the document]
sshanzel: Yeah, was wondering about that too. Thought it was normal for PK FK. Let me check again.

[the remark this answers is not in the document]
idoshamun: Check the tests, there are some errors maybe it's related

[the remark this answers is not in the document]
sshanzel: After further investigation, when having `@OneToOne` relation, you also need to have the `@JoinColumn` decorator. Should be added now.

idoshamun (approved):

sshanzel:
@rebelchris have now covered Ido's comments. Will merge in a few minutes since you mentioned nothing more is blocking.

PR #1652, opened 2024-01-23
feat: user streak trigger on user creation

sshanzel:
The trigger to create user_streak when a row is created on the user's table.

~~Question: what about existing users? Do we introduce a bin script to insert existing users?~~

- Tested the script locally, works as intended.

The retro-fix on missing streaks will be tackled in a different PR as it needs calculation, etc.

MI-69 #done


idoshamun:
@sshanzel yes, we will create it retroactively :)

idoshamun (commented):
Please add a test for the trigger

idoshamun (approved):

src/migration/1706110501504-UserStreakTrigger.ts
```diff
@@ -0,0 +1,45 @@
...
+    await queryRunner.query(
+      `ALTER TABLE "user_streak" ALTER COLUMN "lastViewAt" DROP NOT NULL`,
+    );
+
+    // Create the trigger function
+    await queryRunner.query(`
+      CREATE OR REPLACE FUNCTION create_user_streak()
+      RETURNS TRIGGER AS $$
+      BEGIN
+          INSERT INTO user_streak ("userId")
```
kopancek, line 16: AFAIK, this is how `user_streak` table was created:
```
CREATE TABLE "user_streak" ("userId" character varying(36) NOT NULL, "currentStreak" integer NOT NULL DEFAULT '0', "totalStreak" integer NOT NULL DEFAULT '0', "maxStreak" integer NOT NULL DEFAULT '0', "lastViewAt" TIMESTAMP WITH TIME ZONE NOT NULL, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(), CONSTRAINT "PK_3d8ec0aea914605bd0455034673" PRIMARY KEY ("userId"))
```

the `lastViewAt` column is `NOT NULL`, but we do not specify a default value. Here we are not inserting a value either. Is that expected to work?

kopancek (commented):
Added a question, because I'm not sure if the trigger actually works. Maybe I'm missing something :)

src/migration/1706110501504-UserStreakTrigger.ts
sshanzel: Also, the test should show how the trigger works.

kopancek: Ahh, didn't see that :) thanks!

kopancek (approved):

PR #1653, opened 2024-01-26
feat: extend newView worker to count user streaks

kopancek:
Adds an SQL query to increase or reset the `user_streaks` counters for the user. This query is executed on every newView worker execution, unless the newView worker does not write the view.

Because of this requirement, the code does just 1 SQL query. Splitting it into multiple queries would make it more readable/understandable, but less performant.

Extensive suite of unit tests added to make sure that it actually works as required.


src/workers/newView.ts
```diff
@@ -24,12 +24,104 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
   return false;
 };
 
+const DEFAULT_TIMEZONE = 'UTC'; // in case user doesn't have a timezone set
+const INC_STREAK_QUERY = `
+WITH u AS (
+  SELECT "id", COALESCE("timezone", $3) AS timezone
+  FROM public.user
+  WHERE public.user.id = $1
+),
```
capJavert, line 33: From my point of view since this is running in worker it is fine to run it in multiple queries but appreciate you did it in one 🤘 

It just looks like it will be hard to swallow like this for anyone trying to change it down the road, even know while looking briefly I am not really sure if it does what it needs to do. Maybe as other alternatives we can comment it out a bit to make it more readable.

kopancek: I was thinking about adding more comments and/or making the query more parametrized. But wanted to first get some feedback if we want to go down this path in the first place :)

idoshamun: we have lots of view events per day, i'd rather have this worker highly optimized to reduce the load on the database. Have you considered a new worker for that actually? Not sure it's needed but just surfacing the idea.

My advise would be to go with 1 optimized query and well commented so we can read it properly. I agree that now it's very intimidating

src/workers/newView.ts
```diff
@@ -24,12 +24,104 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
...
+vars AS (
+  SELECT (DATE($2 AT TIME ZONE u.timezone)) AS today FROM u
+),
+d AS (
+  SELECT "userId", "currentStreak", today, (today - DATE("lastViewAt" AT TIME ZONE u.timezone)) AS dateDiff, EXTRACT(DOW FROM today) AS dayOfWeek
```
idoshamun, line 38: `DATE("lastViewAt" AT TIME ZONE u.timezone)` this part simply add a timezone to a non timezone time, or actually move from UTC to the right timezone? If it's the first one I think it's wrong

src/workers/newView.ts
idoshamun: Also looking at the logic, I think it may be more complicated than what I would expect. Shouldn't the logic be as simple as: If it's your first read for today then increase the counter (except on the weekend)?
Assuming we have a cron job that resets the streak, then this should be ok, right? Am I missing anything?

src/workers/newView.ts
kopancek: The `lastViewAt` column in `user_streak` table is a `timestamptz` column. So `AT TIME ZONE` gives us the time in the timezone of the user. Postgres always stores times in the timezone that the DB is configured with, but since we give the timezone at the time of writing as well, this should be correct. E.g. imagine a following scenario:

1. User opens an article at 9:00 AM CET. 
2. This is stored as 8:00+00 in Postgres, since Postgres timezone is configured for UTC. 
3. the `"lastViewAt" AT TIME ZONE 'Europe/Vienna'` returns 9:00 AM CET.

So all in all, I think this is what we want. I can add some unit tests for timezones as well, if you think it's worth doing.

idoshamun: Yes I agree. This is the expected behavior. Perfect

src/workers/newView.ts
kopancek: There was this discussion on slack: https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1706260253838599

Where we decided that we will also increase the counter on weekends, if you read an article on a weekend.
Also, if you read an article on Monday (it's the first day of the week), you need to look if the last view was >= 3 days ago. 

Resetting the streak with the DB call is just a precaution and it is also still just 1 DB write. I don't think that part is making the query much more complicated. With a cron job, you have the problem of when do you actually do the reset. At midnight UTC? Or at midnight of user's time zone? What happens if the user is continously reading articles at night (a night owl type of a user)? With eventual consistency, there might be edge cases, that resetting the state in the SQL query takes care of.

kopancek: > we have lots of view events per day, i'd rather have this worker highly optimized to reduce the load on the database. Have you considered a new worker for that actually? Not sure it's needed but just surfacing the idea.

I thought about this as well. Decided not to do it for now. The new/separate reading streaks worker would be triggered on every post view anyway (which would mean 1 more message needs to be sent & processed per each view).

idoshamun: > Where we decided that we will also increase the counter on weekends, if you read an article on a weekend.
Also, if you read an article on Monday (it's the first day of the week), you need to look if the last view was >= 3 days ago.

Noted!

> Resetting the streak with the DB call is just a precaution and it is also still just 1 DB write. I don't think that part is making the query much more complicated. With a cron job, you have the problem of when do you actually do the reset. At midnight UTC? Or at midnight of user's time zone? What happens if the user is continously reading articles at night (a night owl type of a user)? With eventual consistency, there might be edge cases, that resetting the state in the SQL query takes care of.

At midnight of the user timezone. The cron should run every hour and reset based on the user's midnight.
The reason I'm highlighting that is that under the assumption that streaks are always up to date, only thing we need to do is check whether the user read today a post, if not increase the streak. Do you agree with this logic?

> I thought about this as well. Decided not to do it for now. The new/separate reading streaks worker would be triggered on every post view anyway (which would mean 1 more message needs to be sent & processed per each view).

Sure thing!

src/workers/newView.ts
rebelchris: Not 100% sure from the query, but are we storing it under a specific timezone already now?
(think that's what we decided right?) (disregard if i'm wrong here)

src/workers/newView.ts
kopancek: > The reason I'm highlighting that is that under the assumption that streaks are always up to date, only thing we need to do is check whether the user read today a post, if not increase the streak. Do you agree with this logic?

Yes, I agree. Just that in my career I learned not to rely on such assumptions too much, especially in distributed systems and eventual consistency 😊 It always works, until it doesn't :trollface: 

Removing the reset functionality would make the query 3 lines of SQL and 3 lines of comments less. So I can definitely remove it for the sake of simplicity. But I think the query does not become much more readable as a result. Providing it below as an example. 

Let me know how you would prefer to proceed here, I don't have a super strong preference.

```sql
-- Get the timezone of the user, defaulting to DEFAULT_TIMEZONE if not set
WITH u AS (
  SELECT "id", COALESCE("timezone", $3) AS timezone
  FROM public.user
  WHERE public.user.id = $1
),
-- Create a today variable in the timezone we got above
vars AS (
  SELECT (DATE($2 AT TIME ZONE u.timezone)) AS today FROM u
),
/* Store some more variables based on previous results
   * Get the date difference between the lastViewAt and today and store it in dateDiff
   * Extract current day of week from today
   * Return other columns needed below
*/
d AS (
  SELECT "userId", "currentStreak", today, (today - DATE("lastViewAt" AT TIME ZONE u.timezone)) AS dateDiff, EXTRACT(DOW FROM today) AS dayOfWeek
  FROM user_streak
  INNER JOIN u ON "userId" = u.id
  JOIN vars ON TRUE
),
cond AS (
/* shouldIncrement is set to true if
   a) currentStreak is 0
   b) dateDiff is NULL - this is to cover new users who never had a view before
   c) last view was yesterday (the date difference between today and lastViewAt is 1
   d) today is Monday and last view was Friday, Saturday or Sunday
      (the date difference is more than 1 and less than 3 - assuming 2 day weekend on Sat and Sun)
*/
  SELECT (
    "currentStreak" = 0 OR
      (dateDiff IS NULL OR
        dateDiff = 1 OR
        (dayOfWeek = 1 AND dateDiff > 1 AND dateDiff <= 3) OR
        (dayOfWeek = 0 AND dateDiff = 2)
    )) AS shouldIncrement,
    "userId"
  FROM d
),
/* The update query to run if shouldIncrement is true
   * Increment currentStreak by 1
   * Increment total streak by 1
   * Increment max streak if it's equal to current streak
   * set lastViewAt to the time of the event
   * set updateAt to now
*/
incrementCurrent AS (
  UPDATE user_streak AS us
  SET
    "lastViewAt" = $2,
    "updatedAt" = now(),
    "currentStreak" = us."currentStreak" + 1,
    "totalStreak" = us."totalStreak" + 1,
    "maxStreak" = CASE WHEN us."maxStreak" = us."currentStreak" THEN us."maxStreak" + 1 ELSE us."maxStreak" END
  FROM cond
  WHERE us."userId" = cond."userId" AND shouldIncrement
  RETURNING TRUE
),
/* The update query to run if shouldIncrement is false
   * set lastViewAt to the time of the event
   * set updatedAt to now
   * set currentStreak to 1 if shouldReset is true
*/
noIncrementCurrent AS (
  UPDATE user_streak AS us
  SET
    "lastViewAt" = $2,
    "updatedAt" = now()
  FROM cond
  WHERE us."userId" = cond."userId" AND NOT shouldIncrement
  RETURNING FALSE
)
/* Run both update queries, only one of them will actually update,
   depending on the shouldIncrement value
*/
SELECT * FROM incrementCurrent
UNION ALL
SELECT * FROM noIncrementCurrent
```

idoshamun: Still I think there's a lot more functionality here than I'd expect. And if that's the case than as Ante suggested maybe we should fall back to some JS magic.
This is how I see in pseudo code:
```
todayView = fetchViewFromToday(userTimezone)
if todayView exists do nothing
else increment streak by one (make sure to update lastView and max streak accordingly)
```

Looking at this comment:
```
/* shouldIncrement is set to true if
   a) currentStreak is 0
   b) dateDiff is NULL - this is to cover new users who never had a view before
   c) last view was yesterday (the date difference between today and lastViewAt is 1
   d) today is Monday and last view was Friday, Saturday or Sunday
      (the date difference is more than 1 and less than 3 - assuming 2 day weekend on Sat and Sun)
*/
```
I still think you have quite logic around the freezes and resetting the streak which adds an unexpected complexity to me. I agree with your notion about distributed systems, but this feels like a premature optimization and since we have raw data, in case we have an issue we can recalculate and fix.
Anyway, happy to have a session if needed

rebelchris: The new_streak could even just focus on ONLY adding the view for singular days.
Another CDC alternative could then handle the streak record? (just putting this option out here to simply logic)

kopancek:
```
todayView = fetchViewFromToday(userTimezone)
if todayView exists do nothing
else increment streak by one (make sure to update lastView and max streak accordingly)
```

This pseudo code does not really work. It only covers the case when you already read an article today and increments in all other cases.
If you've read an article on 24.12.2023, it would still try to increment the currentStreak.

What we are trying to do can actually be split into 3 workflows
1. Starting the streak - this should be done when currentStreak is 0 or when lastViewAt is NULL - this cannot be handled by a cron job. Imagine a situation when a user account is created and the user immediately reads an article. The streak must start at that point, even if a scheduled cron job to reset streaks would run in ~1 hour.
2. Continuing the streak - this is the increment. It should only be done if: 
  a) `lastViewAt` was the day before
  b) `lastViewAt` was on a freeze day
  c) `lastViewAt` was done on the last day before the freeze day
3. Ending the streak - this is the few lines of code that I removed above. Can be done by a cron job as discussed above. But does not really improve the SQL query complexity.

So to simplify, the pseudocode should look like this:
```
lastView = fetchCurrentView(currentTime, userTimezone)
if lastView is today 
  updateTimestamps()
else if lastView does not exist
  startStreak()

dayBeforeView = fetchViewFromDayBefore(currentTime, userTimezone) // needs to understand when is a streak freeze
if dayBeforeView does not exist
  endStreak() // or do nothing if we rely solely on the cron job
else
  incrementStreak() 
```

I can try to rewrite the SQL statement so that it reads more naturally, but SQL is declarative, not imperative, so not sure I can achieve that without introducing a psql function.

We could theoretically split this into 2 queries - one to read the current state and users' timezone. Another one to do the actual update. Would make the code much easier to understand, since it would much more closely follow the pseudocode above. If we go down this path, it might make more sense to do the update in another worker. So the newView worker would only check if it should create an event for reading streaks update, the update worker would do the update itself.

Let me know which idea do you like more. I still like the single SQL query approach more, since all of the logic is in one place.

kopancek: Actually, scratch that. You're right, if the cron job updates the streaks correctly, it is much simpler, as you say above. Since the freeze is restarted every day, you can always just increment if lastViewAt was not today.

But then you put all of the logic from here into the cron job. Which I'm fine with, but we didn't write it yet, which would block merging of this change.

idoshamun: > Starting the streak - this should be done when currentStreak is 0 or when lastViewAt is NULL - this cannot be handled by a cron job. Imagine a situation when a user account is created and the user immediately reads an article. The streak must start at that point, even if a scheduled cron job to reset streaks would run in ~1 hour.

We already have a trigger for that on user creation :)

> But then you put all of the logic from here into the cron job. Which I'm fine with, but we didn't write it yet, which would block merging of this change. Unless we have some feature toggle on the API side that I am not aware off, that would allow us to toggle this functionality in runtime.

We will recalculate the streaks anyway before deployment so I wouldn't worry about that. If we really want, we can block this PR as you suggested proceed with the cron job and then merge. But I don't think it really matters

kopancek: PR updated based on the conversation here, please take a look again 🙏  cc @idoshamun

src/workers/newView.ts
```diff
@@ -24,12 +24,78 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
...
+UPDATE user_streak AS us SET
+  "lastViewAt" = $2,
+  "updatedAt" = now(),
+  "currentStreak" = CASE WHEN shouldIncrementStreak THEN us."currentStreak" + 1 ELSE us."currentStreak" END,
+  "totalStreak" = CASE WHEN shouldIncrementStreak THEN us."totalStreak" + 1 ELSE us."totalStreak" END,
+  "maxStreak" = CASE WHEN shouldIncrementStreak AND shouldUpdateMaxStreak THEN us."maxStreak" + 1 ELSE us."maxStreak" END
+FROM u
+WHERE us."userId" = u.id
+RETURNING shouldIncrementStreak
+`;
```
DragosIorgulescu, line 56: I tend to agree that this is not the most easy to read query, but if we have proper test cases covering it, I don't mind it.

src/workers/newView.ts
```diff
@@ -24,12 +24,78 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
...
+  data: DeepPartial<View>,
+): Promise<boolean> => {
+  const repo = con.getRepository(UserStreak);
+  const { userId, timestamp } = data;
+
+  if (!userId) {
+    return false;
+  }
+
+  const viewTime = timestamp ? new Date(timestamp as string) : new Date();
```
DragosIorgulescu, line 69: So in this case, the incoming timestamp is stored in the db with the timezone the database is in, correct? Thus we are casting it to the proper timezone in the query above

DragosIorgulescu (approved):
Looks good, nice work on this, I think I'll add a follow-up ticket here to double check the timezone question & maybe do some tests on this query w/ live data to see how well it performs, since we are calling it on every view event.

src/workers/newView.ts
capJavert: I agree, would this really be that inefficient if it was two queries select + update and created with query builder or ORM directly (we can also use transaction)? We are talking about querying and updating only 1 row here.

I can't seem to get over it. 🙏

capJavert (commented):
I still think that the query can be split without any performance hit, so raising again. Comment below.

src/workers/newView.ts
```diff
@@ -24,12 +24,78 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
...
+        b) the currentStreak is 0 - we should start a new streak
+        c) we didn't do it today already
+      */
+      ("lastViewAt" is NULL OR
+       "currentStreak" = 0 OR
+       DATE("lastViewAt" AT TIME ZONE COALESCE("timezone", $3)) <> DATE($2 AT TIME ZONE COALESCE("timezone", $3))
+      ) AS shouldIncrementStreak,
+
+      -- Increment maxStreak if it's the same as currentStreak
+      "currentStreak" = "maxStreak" AS shouldUpdateMaxStreak
```
idoshamun, line 42: Just in case 😉 

```suggestion
      "currentStreak" >= "maxStreak" AS shouldUpdateMaxStreak
```

idoshamun: Actually thinking about it, it's not even needed here. We can move this logic to the update section below

src/workers/newView.ts
```diff
@@ -24,12 +24,78 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
...
+  FROM public.user AS users
+  INNER JOIN user_streak ON user_streak."userId" = users.id
+  WHERE users.id = $1
+)
+UPDATE user_streak AS us SET
+  "lastViewAt" = $2,
+  "updatedAt" = now(),
+  "currentStreak" = CASE WHEN shouldIncrementStreak THEN us."currentStreak" + 1 ELSE us."currentStreak" END,
+  "totalStreak" = CASE WHEN shouldIncrementStreak THEN us."totalStreak" + 1 ELSE us."totalStreak" END,
+  "maxStreak" = CASE WHEN shouldIncrementStreak AND shouldUpdateMaxStreak THEN us."maxStreak" + 1 ELSE us."maxStreak" END
```
idoshamun, line 52: Just thinking this will be more consistent:
```suggestion
  "maxStreak" = GREATEST("maxStreak", us."currentStreak" + 1)
```

idoshamun (approved):
Looks way way better! See my final suggestion for making it even simpler

src/workers/newView.ts
idoshamun: After the simplification + additional comments i'm ok with that

__tests__/workers/newView.ts
```diff
@@ -45,44 +51,181 @@ it('should save a new view with the provided timestamp', async () => {
   const views = await con.getRepository(View).find();
   expect(views.length).toEqual(1);
   expect(views[0]).toMatchSnapshot();
+
+  const streak = await con
+    .getRepository(UserStreak)
+    .findOne({ where: { userId: 'u1', lastViewAt: timestamp } });
+  expect(streak).toMatchSnapshot({
```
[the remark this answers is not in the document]
kopancek: Didn't know about that. I can rewrite these, no problem 👍

src/workers/newView.ts
kopancek: Changed. We cannot use GREATEST like you suggest unfortunately, because we only want to increment maxStreak if we are doing the currentStreak increment as well.

idoshamun: You're right! We can add `CASE WHEN shouldIncrementStreak` in this case. My biggest concern is to have drifts. This is why I prefer to have the logic sticked to the query and reduce the assumptions here. Does that make sense?

__tests__/workers/newView.ts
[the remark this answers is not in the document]
rebelchris: That be great yes :) 
Snapshots sooner of later always come back to bite us 🤣

rebelchris (approved):
Nice way cleaner indeed!
And pretty clean to maintain both writes in the worker.

src/workers/newView.ts
kopancek: I tested locally with nodejs runtime timezone set to PST and it all works as it's supposed to: https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1706631354151889?thread_ts=1706625076.569029&cid=C05TQL6HWTH

sshanzel (approved):
You are definitely one of the SQL kung-fu masters 🥋

PR #1655, opened 2024-01-28
feat: get user streak

sshanzel:
Introduce a new query to fetch the user's reading streak.

Note: this PR is purely retrieval. The additional logic that resets the user's current streak based on some checks will be for a different ticket/PR.

MI-72 #done

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -74,6 +74,13 @@ const obj = new GraphORM({
   CommentUpvote: {
     requiredColumns: ['createdAt'],
   },
+  UserStreak: {
+    fields: {
+      max: { select: 'maxStreak' },
+      total: { select: 'totalStreak' },
+      current: { select: 'currentStreak' },
+    },
+  },
   Post: {
     additionalQuery: (ctx, alias, qb) =>
       qb
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -304,6 +304,26 @@ describe('query userStats', () => {
   });
 });
 
+describe('query userStreaks', () => {
+  const QUERY = `query UserStreak {
+    userStreak {
+      max
+      total
+      current
+      lastViewAt
+    }
+  }`;
+
+  it('should not allow unauthenticated users', () =>
+    testQueryErrorCode(client, { query: QUERY }, 'UNAUTHENTICATED'));
+
+  it('should return the user stats', async () => {
+    loggedUser = '1';
+    const res = await client.query(QUERY);
+    expect(res.errors).toBeFalsy();
+  });
+});
+
 describe('query referredUsers', () => {
   const QUERY = `query ReferredUsers {
     referredUsers {
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -97,6 +97,13 @@ export interface GQLUser {
   readmeHtml?: string;
 }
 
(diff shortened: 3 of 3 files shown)


src/graphorm/index.ts
```diff
@@ -74,6 +74,13 @@ const obj = new GraphORM({
   CommentUpvote: {
     requiredColumns: ['createdAt'],
   },
+  UserStreak: {
+    fields: {
+      max: { select: 'maxStreak' },
+      total: { select: 'totalStreak' },
+      current: { select: 'currentStreak' },
```
idoshamun, line 81: Why do we have separate names for the db columns and the graphql ones? Why don't we keep the same name?

sshanzel: Took it as how it was desired from the DR. But we can probably keep the same names to reduce complexity. cc @DragosIorgulescu wdyt?

idoshamun (approved):
I'm approving but just double check with Dragos. I'm ok either way

src/graphorm/index.ts
rebelchris: Yeah you query `userStreak.something` right?
Would make sense to me to keep them inline.

DragosIorgulescu: we can change the naming here to be the same, up to you if you want to use single word variations @sshanzel

DragosIorgulescu (commented):
> Note: this PR is purely retrieval. The additional logic that resets the user's current streak based on some checks will be for a different ticket/PR.

Could you point to the ticket where this would be covered? I think the ticket for this includes the clearing of the `currentStreak` functionality as well.

src/graphorm/index.ts
sshanzel: Alright. Let's do single-word names since we also had that on our RFC too 🥳

sshanzel:
> Could you point to the ticket where this would be covered? I think the ticket for this includes the clearing of the currentStreak functionality as well.

This PR is for the MI-72 ticket. I've thought of that as well. But then there is a separate ticket specifically for that (modifying this query). https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31/backlog?epics=visible&issueParent=14432&selectedIssue=MI-108

Which is why I also had to ask it in the [relevant ticket](https://dailydotdev.atlassian.net/browse/MI-72?focusedCommentId=11331).

DragosIorgulescu:
> > Could you point to the ticket where this would be covered? I think the ticket for this includes the clearing of the currentStreak functionality as well.
> 
> This PR is for the [MI-72](https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31?selectedIssue=MI-72) ticket. I've thought of that as well. But then there is a separate ticket specifically for that (modifying this query). https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31/backlog?epics=visible&issueParent=14432&selectedIssue=MI-108
> 
> Which is why I also had to ask it in the [relevant ticket](https://dailydotdev.atlassian.net/browse/MI-72?focusedCommentId=11331).

@sshanzel you're right, I created a separate ticket for that (and forgot I did 🤦‍♂️ ) -- we will tackle it separately, this PR should be good to go 🙏

DragosIorgulescu (approved):
🚀

PR #1656, opened 2024-01-29
feat: verify flow changes

rebelchris:
Changes needed to accomodate the verification flow.


src/kratos.ts
```diff
@@ -118,12 +126,16 @@ export const dispatchWhoami = async (
       req,
       `${heimdallOrigin}/api/whoami`,
     );
-    if (whoami?.identity?.traits?.userId) {
+    const { session, verified } = whoami;
```
idoshamun, line 129: We need to check backwards compatability

__tests__/boot.ts
```diff
@@ -131,8 +131,11 @@ const mockWhoami = (expected: unknown, statusCode = 200) => {
 
 const mockLoggedIn = (userId = '1') =>
   mockWhoami({
-    identity: { traits: { userId } },
-    expires_at: KRATOS_EXPIRATION,
+    session: {
```
idoshamun, line 134: Remember to add test for the backwards compat

src/kratos.ts
idoshamun: We should remember to return `verified: true` for the old case

src/routes/boot.ts
```diff
@@ -462,11 +465,15 @@ export const getBootData = async (
   ) {
     return loggedInBoot(con, req, res, false, middleware);
   }
+
   const whoami = await dispatchWhoami(req);
   if (whoami.valid) {
     if (whoami.cookie) {
       setRawCookie(res, whoami.cookie);
     }
```
capJavert, line 473: Should we set cookie before we know if user is verified, I am not sure for what is it for?

rebelchris: Yeah verified doesn't mean your not logged in basically.
It's just that you still need to verify your email if that makes sense.

capJavert: 👍

capJavert (approved):

idoshamun (approved):
Perfect!

PR #1657, opened 2024-01-30
fix: reduce the freeform minimum content length to 50 chars

omBratteng:
Reduces the amount required amount of characters for triggered content requested on freeform.


rebelchris (approved):

idoshamun (changes requested):
I'm not approving yet since we wanted combined title and content and we said we gonna sync on that

rebelchris (approved):

PR #1658, opened 2024-01-30
refactor: query user streak - clear when possible

sshanzel:
- Update the `userStreak` query to clear the user's `currentStreak` whenever applicable.
- The test cases have covered all the possible cases which involve weekends. Feel free to go through it.

MI-108 #done

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -421,7 +421,7 @@ describe('query userStreaks', () => {
     await expectStreak(5, 5, lastViewAt);
   });
 
-  it('should not reset streak on if last view is the same day today', async () => {
+  it('should not reset streak if last view is the same day today', async () => {
     loggedUser = '1';
     const fakeToday = new Date(2024, 0, 9); // Tuesday
     const lastViewAt = subDays(fakeToday, 0); // Tuesday
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -36,6 +36,7 @@ import {
   User,
   View,
   UserPersonalizedDigest,
+  UserStreak,
 } from '../src/entity';
 import { sourcesFixture } from './fixture/source';
 import { getTimezonedStartOfISOWeek } from '../src/common';
@@ -314,13 +315,119 @@ describe('query userStreaks', () => {
     }
   }`;
 
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
   it('should not allow unauthenticated users', () =>
     testQueryErrorCode(client, { query: QUERY }, 'UNAUTHENTICATED'));
 
-  it('should return the user stats', async () => {
+  it('should return the user streaks', async () => {
     loggedUser = '1';
     const res = await client.query(QUERY);
     expect(res.errors).toBeFalsy();
+  });
+
+  it('should return the user streaks when last view is null', async () => {
+    loggedUser = '1';
+    const res = await client.query(QUERY);
+    expect(res.errors).toBeFalsy();
+  });
+
+  const expectStreak = async (
+    currentStreak: number,
+    expectedCurrentStreak: number,
+    lastViewAt: Date,
+  ) => {
+    const repo = con.getRepository(UserStreak);
+    await repo.update({ userId: loggedUser }, { currentStreak, lastViewAt });
+
+    const res = await client.query(QUERY);
+    expect(res.errors).toBeFalsy();
+
+    const streak = await repo.findOneBy({ userId: loggedUser });
+    expect(streak.currentStreak).toEqual(expectedCurrentStreak);
+  };
+
+  it('should reset streak on Saturday when last read is Thursday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 6); // Saturday
+    const lastViewAt = subDays(fakeToday, 2); // Thursday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 0, lastViewAt);
+  });
+
+  it('should reset streak on Sunday when last read is Thursday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 7); // Sunday
+    const lastViewAt = subDays(fakeToday, 3); // Thursday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 0, lastViewAt);
+  });
+
+  it('should not reset streak on Saturday when last read is Friday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 6); // Saturday
+    const lastViewAt = subDays(fakeToday, 1); // Friday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 5, lastViewAt);
+  });
+
+  it('should not reset streak on Sunday when last read is Friday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 7); // Sunday
+    const lastViewAt = subDays(fakeToday, 2); // Friday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 5, lastViewAt);
+  });
+
+  it('should reset streak on Monday when last read was Thursday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 8); // Monday
+    const lastViewAt = subDays(fakeToday, 4); // Thursday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 0, lastViewAt);
+  });
+
+  it('should not reset streak on Monday when last read was Friday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 8); // Monday
+    const lastViewAt = subDays(fakeToday, 3); // Friday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 5, lastViewAt);
+  });
+
+  it('should reset streak on Tuesday when last read was Friday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 9); // Tuesday
+    const lastViewAt = subDays(fakeToday, 4); // Friday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 0, lastViewAt);
+  });
+
+  it('should not reset streak on Tuesday when last read was Monday', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 9); // Tuesday
+    const lastViewAt = subDays(fakeToday, 1); // Monday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 5, lastViewAt);
+  });
+
+  it('should not reset streak on if last view is the same day today', async () => {
+    loggedUser = '1';
+    const fakeToday = new Date(2024, 0, 9); // Tuesday
+    const lastViewAt = subDays(fakeToday, 0); // Tuesday
+
+    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    await expectStreak(5, 5, lastViewAt);
   });
 });
 
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -435,12 +435,13 @@ describe('query userStreaks', () => {
     const tz = 'America/Tijuana';
     await con.getRepository(User).update({ id: loggedUser }, { timezone: tz });
     const fakeToday = new Date(2024, 0, 6); // Saturday
+    const fakeTodayTz = addHours(fakeToday, 16); // To ensure UTC offset would not make today as Friday
     const lastViewAt = subDays(fakeToday, 2); // Thursday
     const lastViewAtTz = addHours(lastViewAt, 22); // by UTC time, this should be Friday
     // No reset should happen if we are not considering timezone
     // but here, it should reset
 
-    jest.useFakeTimers({ advanceTimers: true, now: fakeToday });
+    jest.useFakeTimers({ advanceTimers: true, now: fakeTodayTz });
     await expectStreak(5, 0, lastViewAtTz);
   });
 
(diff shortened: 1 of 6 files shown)


src/common/users.ts
```diff
@@ -339,3 +350,51 @@ export const getUserReadingRank = async (
...
+
+export const clearThenGetUserStreak = (
+  ctx: Context,
+  info: GraphQLResolveInfo,
+  userId: string,
+) => {
+  return ctx.con.transaction(async (manager) => {
+    await manager
+      .getRepository(UserStreak)
+      .update({ userId }, { currentStreak: 0 });
```
kopancek, line 390: We should also update the `updatedAt` column with the current date.

kopancek (changes requested):
Looks great mostly, thanks for the exhaustive set of tests, really helpful :) 

I think we need to be careful with the timezone of the user, I don't think the current code would work well with that in mind. 

Also, suggested to refactor the condition a little bit. To me it's more readable that way, but that is personal preference.

src/schema/users.ts
```diff
@@ -850,14 +850,61 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const dayToday = today.getDay();
+      const difference = differenceInDays(today, lastView);
+
+      if (difference === 0) {
+        return streak;
+      }
+
+      // Even though it is the weekend, we should still clear the streak for when the user's last read was Thursday
+      // Due to the fact that when Monday comes, we will clear it anyway when we notice the gap in Friday
+      if (WEEKENDS.includes(dayToday)) {
```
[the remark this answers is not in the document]
sshanzel: > You could then argue, that we could also do everything in a simple if/else condition as well. Much quicker to read and I think readability is actually improved, since you don't need to read so many lines:
```
if ((today === Saturday && difference > 1) ||
     (today === Sunday && difference > 2) || 
     (today === Monday && difference > 3) ||
     difference > 1) {
     return clearThenGetUserStreak(ctx, info, ctx.userId);
}

return streak; 
```
Yeah, this makes sense and much simpler. Let's apply it.

src/schema/users.ts
```diff
@@ -850,14 +850,61 @@ export const resolvers: IResolvers<any, Context> = {
...
+        ctx,
+        info,
+        (builder) => ({
+          queryBuilder: builder.queryBuilder.where(
+            `"${builder.alias}"."userId" = :id`,
+            { id: ctx.userId },
+          ),
+          ...builder,
+        }),
+      );
```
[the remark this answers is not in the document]
sshanzel: The `lastViewAt` has timezone as part of its data. We can probably just turn it to UTC on return so the value would match with `new Date()`.

src/common/users.ts
```diff
@@ -339,3 +350,51 @@ export const getUserReadingRank = async (
...
   };
 };
+
+export const getStreakLastView = async (
+  con: DataSource | EntityManager,
+  streak: GQLUserStreak,
+): Promise<Date> => {
+  if (streak.lastViewAt) {
+    return streak.lastViewAt;
+  }
```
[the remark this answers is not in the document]
sshanzel: We don't have yet the infrastructure that updates this field. This will always return `null` for now until we have the worker updated.

src/schema/users.ts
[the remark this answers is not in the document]
sshanzel: > This condition is unnecessary. I would flatten the conditions to:
> 
> ```js
> const clearTheStreak = () => clearThenGetUserStreak(ctx, info, ctx.userId)
> 
> if (today === Saturday && difference > 1) {
>   return clearTheStreak();
> }
> if (today === Sunday && difference > 2) {
>   return clearTheStreak();
> }
> if (today === Monday && difference > 3) {
>   return clearTheStreak();
> }
> if (difference > 1) {
>   return clearTheStreak();
> }
> 
> return streak;
> ```
> 
> You could then argue, that we could also do everything in a simple if/else condition as well. Much quicker to read and I think readability is actually improved, since you don't need to read so many lines:
> 
> ```js
> if ((today === Saturday && difference > 1) ||
>      (today === Sunday && difference > 2) || 
>      (today === Monday && difference > 3) ||
>      difference > 1) {
>      return clearThenGetUserStreak(ctx, info, ctx.userId);
> }
> 
> return streak;    
> ```

This looks great but it won't work when it is a Weekend the user's last read was Friday where it should not reset.

src/common/users.ts
sshanzel: This should have happened automatically. I instead introduced a decorator for when the entity gets updated, this column should get updated too.

https://github.com/dailydotdev/daily-api/pull/1658/commits/05fc9c296358b0875c38536a4b9804a1f313bc88

src/schema/users.ts
```diff
@@ -850,14 +850,61 @@ export const resolvers: IResolvers<any, Context> = {
...
+
+        return streak;
+      }
+
+      if (dayToday === Day.Monday) {
+        if (difference - FREEZE_DAYS_IN_A_WEEK < 2) {
+          return streak;
+        }
+
+        return clearThenGetUserStreak(ctx, info, ctx.userId);
```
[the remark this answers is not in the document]
sshanzel: The tests should serve as an easy check to validate whether the conditions are correct.

With the suggestion, this would fail the test where it `should not reset streak on Monday when last read was Friday`.

src/common/users.ts
```diff
@@ -339,3 +350,51 @@ export const getUserReadingRank = async (
...
+      .getRepository(UserStreak)
+      .update({ userId }, { currentStreak: 0 });
+
+    return graphorm.queryOneOrFail<GQLUserStreak>(ctx, info, (builder) => ({
+      queryBuilder: builder.queryBuilder.where(
+        `"${builder.alias}"."userId" = :id`,
+        { id: userId },
+      ),
+      ...builder,
+    }));
```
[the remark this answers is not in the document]
sshanzel: I haven't tried this using GraphORM (our internal library), but I can try and see if it works with that.

src/schema/users.ts
```diff
@@ -850,14 +852,56 @@ export const resolvers: IResolvers<any, Context> = {
...
-          { id: ctx.userId },
-        ),
-        ...builder,
-      })),
+    userStreak: async (_, __, ctx: Context, info): Promise<GQLUserStreak> => {
+      const streak = await ctx.con
+        .getRepository(UserStreak)
+        .createQueryBuilder()
+        .select('"lastViewAt"') // needs to consider user's timezone
+        .addSelect('"userId"')
```
rebelchris, line 860: Cant we just do this?
```suggestion
        .select('lastViewAt, userId')
```

src/schema/users.ts
```diff
@@ -850,14 +852,56 @@ export const resolvers: IResolvers<any, Context> = {
...
+      const today = new Date();
+      const dayToday = today.getDay();
+      const difference = differenceInDays(today, lastView);
+
+      if (difference === 0) {
+        return getStreak();
+      }
+
+      // Even though it is the weekend, we should still clear the streak for when the user's last read was Thursday
+      // Due to the fact that when Monday comes, we will clear it anyway when we notice the gap in Friday
```
rebelchris, line 878: Is it not "free" but could still show as "Read" if you decided to read?
(not really sure where this part is used, so more of a question at this point)

rebelchris (commented):
Only did a high level review, nothing blocking, but have some questions (more to validate)

src/common/users.ts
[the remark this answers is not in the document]
idoshamun: But we will not use this data as well right? I appreciate the workaround but it's not needed until it is which in that point we will have the worker and data

src/common/users.ts
idoshamun: > This should have happened automatically. I instead introduced a decorator for when the entity gets updated, this column should get updated too.

@sshanzel I don't think this applies on such updates. Can you please confirm?

src/schema/users.ts
sshanzel: I intentionally separated them. Once we get apply all the timezone fetch and conversion on the `lastViewAt`, it is going to be a mess to look at, so to easily see what are the other fields selected, I moved them to the next line.

src/schema/users.ts
sshanzel: Yes, if you read anything, it will count. The concern here is that, let's say today is Saturday, and the user's last read was Thursday. We should clear it since there was a gap, which is the Friday.

src/schema/users.ts
```diff
@@ -850,14 +852,56 @@ export const resolvers: IResolvers<any, Context> = {
...
     },
-    userStreak: (_, __, ctx: Context, info): Promise<GQLUserStreak> =>
-      graphorm.queryOneOrFail<GQLUserStreak>(ctx, info, (builder) => ({
-        queryBuilder: builder.queryBuilder.where(
-          `"${builder.alias}"."userId" = :id`,
-          { id: ctx.userId },
-        ),
-        ...builder,
-      })),
+    userStreak: async (_, __, ctx: Context, info): Promise<GQLUserStreak> => {
```
idoshamun, line 855: This logic is quite complex to follow since we're trying to do several things here:
* Workaround our missing data - Should be ignored as once we roll out data should be migrated
* Fetch the streak - This is great but we should do it only in one place, at the beginning of this function. Currently, we do it in different places. Ideally we can add `lastViewAt` as required column in graphorm. Fetch the graphorm in the beginning and do all the checks later.
* Reset the streak - So once we have the streak from the DB, we can check if it requires a reset. Ideally this should be encapsulated in a function. If we need to reset it, we can update the DB without refetching again and just a merged object of the db streak with some adjusted fields.

The above should make the function much more readable imo

src/common/users.ts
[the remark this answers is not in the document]
sshanzel: Cool. Let me remove this and just utilize what is stored in `lastViewAt`.

src/common/users.ts
sshanzel: Found some issue related to this https://github.com/typeorm/typeorm/issues/5385

It is probably not best to rely on the decorator.

src/schema/users.ts
sshanzel: Cool. Let me refactor it a little bit. Thanks for the points.

src/schema/users.ts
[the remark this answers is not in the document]
kopancek: [See my comment on slack](https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1706629059323989?thread_ts=1706625076.569029&cid=C05TQL6HWTH), it's not the way we thought :) 

On top of that, the RFC mentions that we want to do all of the operations on these dates in user's timezone. I think clearing should take that into account as well, otherwise we would inconsistently clear the streaks.

src/schema/users.ts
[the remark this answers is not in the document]
kopancek: Good catch. But it will work, with a slight modification:
```javascript
if ((dayToday === Sunday && difference > 2) || 
     (dayToday === Monday && difference > 3) ||
     (dayToday > Monday && dayToday <= Saturday && difference > 1)) {
     return clearThenGetUserStreak(ctx, info, ctx.userId);
}

return streak; 
```

So now, if it's Sunday, the difference must be more than 2 days to clear.
If it's Monday, the difference must be more than 3 days to clear.
If it's any other day, it can be more than 1 day to clear.

I think we are covered by this. We can split it into multiple conditions as well, if you find it more readable, but I like this concise version more, me being me :)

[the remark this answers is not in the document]
sshanzel: Probably a matter of preference, but I think your suggestion is much clearer, and I'd prefer it more. I applied the suggestion then just modified some of the magic numbers just to make it easier to read.

src/schema/users.ts
idoshamun: It looks great now!!

src/common/users.ts
```diff
@@ -339,3 +349,56 @@ export const getUserReadingRank = async (
...
+  info: GraphQLResolveInfo,
+  streak: GQLUserStreak,
+): Promise<boolean> => {
+  const { lastViewAt } = streak;
+
+  if (!lastViewAt) {
+    return false;
+  }
+
+  const today = new Date();
```
idoshamun, line 389: We need to convert it to the user's timezone don't we?

__tests__/users.ts
```diff
@@ -314,14 +315,120 @@ describe('query userStreaks', () => {
     }
   }`;
 
+  afterEach(() => {
```
idoshamun, line 318: I love the test coverage!

src/common/users.ts
kopancek: Yup, we do

kopancek: This comment probably got missed before :) 
https://github.com/dailydotdev/daily-api/pull/1658#discussion_r1471146921

src/schema/users.ts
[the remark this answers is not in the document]
kopancek: The conditions for clearing are now 👨‍🍳 💋 , love it ❤️

We still need to fix the user timezone issue on the date comparison though. https://github.com/dailydotdev/daily-api/pull/1658#discussion_r1471548004

src/common/users.ts
sshanzel: Still WIP with the timezones, I am still working on making the query convert the `lastViewAt` to the user preferred timezone. After that, I will match this Date instance.

src/schema/users.ts
```diff
@@ -850,14 +850,61 @@ export const resolvers: IResolvers<any, Context> = {
...
+        }),
+      );
+
+      const lastView = await getStreakLastView(ctx.con, streak);
+
+      if (!lastView) {
+        return streak;
+      }
+
+      const today = new Date();
```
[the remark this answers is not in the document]
sshanzel: Should now be accurate with the latest changes.

src/schema/users.ts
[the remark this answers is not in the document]
sshanzel: Should now take into account user's timezone.

rebelchris (approved):
The tests are amazing, nicely done 👍
Also lot more cleaner code then last time I reviewed it, well done Lee!

src/common/users.ts
```diff
@@ -339,3 +354,58 @@ export const getUserReadingRank = async (
...
+): Promise<boolean> => {
+  const { lastViewAtTz: lastViewAt, timezone: timeZone } = streak;
+
+  if (!lastViewAt) {
+    return false;
+  }
+
+  const now = new Date();
+  const timeZonedToday = now.toLocaleDateString('en', { timeZone });
+  const today = new Date(timeZonedToday);
```
kopancek, line 396: Potential improvement - move this to a utility function.

kopancek (approved):
Looks good to me now timezone wise. Love the test coverage for that as well 🙇 . Added non-blocking comment.

idoshamun (approved):
Amazing!

src/common/users.ts
sshanzel: Yeah was thinking of that too. I'll move it today.

DragosIorgulescu (approved):
🙏 🚀

PR #1659, opened 2024-01-30
feat: feed v23

idoshamun:


src/integrations/feed/generators.ts
```diff
@@ -91,19 +91,19 @@ export const feedGenerators: Record<FeedVersion, FeedGenerator> = Object.freeze(
...
       }),
     ),
-    '22': new FeedGenerator(
+    '23': new FeedGenerator(
       feedClient,
       new FeedUserStateConfigGenerator(snotraClient, {
         personalised: new FeedPreferencesConfigGenerator(
           {
-            feed_config_name: FeedConfigName.VectorV22,
+            feed_config_name: FeedConfigName.VectorV21,
```
capJavert, line 99: Just sanity check this should really be `21` and not `23`?

capJavert (approved):

PR #1660, opened 2024-01-31
fix: set upload limit to match FE limit

rebelchris:
After checking we actually set the FE limit to 5MB:
```
export const imageSizeLimitMB = 5;
```

So decided to set this one to match so we will catch it on both sides correctly.

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -10,6 +10,7 @@ import {
   SourceMember,
   User,
   UserPost,
+  UserNotification,
 } from '../entity';
 import {
   SourceMemberRoles,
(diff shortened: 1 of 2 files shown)


src/graphorm/index.ts
```diff
@@ -23,7 +24,6 @@ import { GQLBookmarkList } from '../schema/bookmarks';
 import { base64 } from '../common';
 import { GQLComment } from '../schema/comments';
 import { GQLUserPost } from '../schema/posts';
-import { UserNotification } from '../entity/notifications/UserNotification';
```
rebelchris, line 26: Just a cleaner import I noticed.
See above the `schema` ones, we should probably add a linter here for cleaner imports as we have on apps project.

src/index.ts
```diff
@@ -129,7 +129,7 @@ export default async function app(
   // });
 
   app.register(MercuriusGQLUpload, {
-    maxFileSize: 1024 * 1024 * 2,
+    maxFileSize: 1024 * 1024 * 5,
```
idoshamun, line 132: Let's do something bigger even

```suggestion
    maxFileSize: 1024 * 1024 * 20,
```

idoshamun (approved):

src/index.ts
rebelchris: I thought about that, but the FE is set to 5 anyway?
We can submit this regardless and see if we want to adjust it there?

idoshamun: yes, probably good idea to change there as well :)

PR #1662, opened 2024-01-31
fix: do not update reading streak for non-existing user

kopancek:
On anonymous events, the newView worker receives a message with `userId`, but there is no such user with the userId. Our temporary code for inserting rows into the `user_streak` table when there is none didn't work correctly for such cases.

Rewritten to first get the user and only try to insert if the user exists.

### DIFF
--- a/__tests__/workers/newView.ts
+++ b/__tests__/workers/newView.ts
@@ -14,6 +14,8 @@ beforeAll(async () => {
 });
 
 beforeEach(async () => {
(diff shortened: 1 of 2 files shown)


idoshamun (approved):

DragosIorgulescu (approved):
Nice 👌

nensidosari (approved):

omBratteng (approved):

rebelchris (approved):

PR #1664, opened 2024-02-01
fix: shared post trigger

idoshamun:
For some odd reason PG not equal operator doesn't care about null 😅

This means if a value was null and then changed it will not be captured by `<>`

#FML


capJavert (approved):
#FML indeed, I learned today.

PR #1665, opened 2024-02-01
fix: popular feed

idoshamun:
The popular feed configuration was broken for a while but the issue was surfaced when we introduced caching.

This PR fixes it and also includes blocking the tags and sources on this feed.

### DIFF
--- a/src/integrations/feed/generators.ts
+++ b/src/integrations/feed/generators.ts
@@ -11,7 +11,6 @@ import { FeedClient } from './clients';
 import {
   FeedPreferencesConfigGenerator,
   FeedUserStateConfigGenerator,
-  SimpleFeedConfigGenerator,
 } from './configs';
 import { SnotraClient } from '../snotra';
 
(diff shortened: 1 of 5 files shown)


omBratteng (approved):

src/integrations/feed/generators.ts
```diff
@@ -111,22 +110,15 @@ export const feedGenerators: Record<FeedVersion, FeedGenerator> = Object.freeze(
...
-            remove_engaged_posts: true,
-            page_size_fraction: 1,
-            fallback_provider: 'fresh',
-          },
+      new FeedClient(process.env.POPULAR_FEED),
+      new FeedPreferencesConfigGenerator(
+        {},
+        {
+          includePostTypes: true,
+          includeBlockedSources: true,
```
rebelchris, line 118: Does this mean we can more easily actually decide to block them now as well?
What was the reason we dont block them here?

Feel if you said you really hate a source and block it, it should not show anywhere basically 👀

idoshamun: When it wasn't part of the feed service originally it was harder to do it.
Now with feed service everything is encapsulated and turned into config, so for api it's just turning on a flag :P
Pretty slick ah?

rebelchris: 100% but we want to keep exposing blocked stuff there?

rebelchris: Oh amazing! Misread it 🤧
Pretty cool 🙏

PR #1666, opened 2024-02-02
feat: automation infra

idoshamun:
New generic endpoint for working with automation services (e.g Retool).

The idea is to provide as much flexibility as possible so it just passes through any parameters from the client. It adds the `userId` to any call and also checks for authentication.

This way we just need to add new automation to the object and viola!

I also have some ideas on how to store the automations in the DB, but that's too early for that.


src/integrations/automation/types.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * An interface for an automation service (Retool, Zapier, etc)
+ */
+export interface IAutomationService<Args, Ret> {
+  run(args: Args): Promise<Ret>;
+}
+
+export type Automation = 'roaster';
```
capJavert, line 8: This should be union type in the future? eg. when we add another one it will be `'roaster' | 'fylla'`?

src/routes/automations.ts
```diff
@@ -0,0 +1,34 @@
...
+    async (req, res) => {
+      const { body, userId } = req;
+      const name = req.params.name;
+      const auto = automations[name];
+      if (!auto) {
+        return res.status(404).send();
+      }
+      if (!userId) {
+        return res.status(401).send();
+      }
```
capJavert, line 21: Might be blind at this point, but what prevents anyone for just supplying their userId? Should this be limited only to team members?

capJavert (commented):
Nice idea I love these kind of 3party integrations in the API. Nice job on abstracting it as well. 👍 I have few questions.

src/integrations/automation/types.ts
idoshamun: Exactly

src/routes/automations.ts
idoshamun: `userId` is extracted from the JWT so you can't trick it :)

src/integrations/automation/types.ts
capJavert: We usually use enums more so might be better to just use enum, or is there some other reason?

src/routes/automations.ts
capJavert: And our JWT knows to limit this endpoint just for team members or this supposed to be "open" to any registered user?

src/integrations/automation/types.ts
idoshamun: Just copied my feed version implementation. Can do enum if you prefer, I don't mind tbh

src/routes/automations.ts
idoshamun: Should be opened to any registered user. So the idea is to offload some of the work to automation services. Since we currently only have one in place I wasn't 100% how to generalize. Maybe once we get the second one, we'll see if all of them should be public or just a subset. Anyway we can fix this endpoint according to future requirements. For now we have only one

capJavert: Ok, then its good, I was not sure if even this one should be open. LGTM!

src/integrations/automation/types.ts
capJavert: I would prefer enum yes, since you can also use it as value for different checks vs union is just type.

idoshamun: Done!

omBratteng (approved):

rebelchris (approved):
No code ido 😈

PR #1668, opened 2024-02-02
feat: implement script to seed user_streak data based on view table

kopancek:
Related PR for adding the stream topic: 
- https://github.com/dailydotdev/streams/pull/27

Batch size is configurable directly in the script with `BATCH_SIZE` const variable.


idoshamun:
@kopancek usually run this kind of migration with a script in production, but I guess this one is heavier than usual?

rebelchris:
Do we need it after the one off batch?
Like ido said script is normally cleaner if its once off.

kopancek:
The decision to run it as a cron was done in this thread: https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1706779286403069?thread_ts=1706779172.790499&cid=C05TQL6HWTH

I'm not against rewriting it to a simple script that we run once, but I was a little concerned about the write cadence - we need to run this for every user that had a view in the end to calculate the reading streak. We have ~380k users in the DB for which we would write a row in the database. The write is batched, depending on the batch size, but still a lot of new rows written.

Let me know what you both think @rebelchris @idoshamun

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,179 @@
...
+  createdAt?: string;
+}
+
+const addNextBatch = async (
+  con: DataSource,
+  insertQueue: fastq.queueAsPromised<Partial<User>[], void>,
+  userPointer?: UserPointer,
+  limit: number = BATCH_SIZE,
+): Promise<UserPointer> => {
+  const qb = con
```
idoshamun, line 24: Do we need to check that the user doesn't have a streak calculated already?
May be a good guard in case something breaks mid-run

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,179 @@
...
+  const timeZone = user.timezone ?? 'UTC';
+
+  const views: ViewWithDateTz[] = await con
+    .getRepository(View)
+    .createQueryBuilder()
+    .select([])
+    .addSelect('MAX(timestamp)', 'maxTimestamp')
+    .addSelect('DATE(timestamp::timestamptz AT TIME ZONE :timezone)', 'date')
+    .where({ userId: user.id, timestamp: Not(IsNull()) })
+    .innerJoin(User, 'u', `"userId" = u.id`)
```
idoshamun, line 90: We don't really need to join if we already validated that the user exist right?
May reduce the load on the db

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,179 @@
...
+const FREEZE_DAYS_IN_A_WEEK = 2;
+const MISSED_LIMIT = 1;
+
+const computeReadingStreaksData = async (
+  con: DataSource,
+  user: Partial<User>,
+): Promise<DeepPartial<UserStreak>> => {
+  const timeZone = user.timezone ?? 'UTC';
+
+  const views: ViewWithDateTz[] = await con
```
idoshamun, line 83: This can potentially have tens of thousands records in some cases. We should guard ourselves from it. Please check my assumption though

bin/seedUserStreaks.ts
idoshamun, line 90: Do we need this join? I'm not sure

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,179 @@
...
+      throw err;
+    }
+    console.log('user streaks created for: ', ids);
+
+    insertCount += ids.length;
+  }, QUEUE_CONCURRENCY);
+
+  let nextUser = undefined;
+  let i = 0;
+  while (nextUser !== null && i++ < 6) {
```
idoshamun, line 171: why `i < 6`?

bin/seedUserStreaks.ts
kopancek: Currently there's no way to tell. Unless we want to store the ids of processed users in some intermediary table/redis. 

Even if there is a row in the `user_streak` table for the user, we don't know if it's correct numbers or not. Because since we released the newView worker changes, these numbers are incremented all the time, just not from initial values.

idoshamun: I see. So let's just print the pointer after every iteration so we can start from that point

bin/seedUserStreaks.ts
kopancek: This is my debugging getting into the PR, good catch :) removing

bin/seedUserStreaks.ts
kopancek: Not really. Originally it was there because of the date/timezone conversion. But since we supply the timezone with the user as a parameter now, no need for a join anymore.

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  const users = await qb.execute();
+
+  // nothing to do anymore, all the users are processed
+  if (users.length === 0) {
+    console.log('no users to process');
+    return null;
+  }
+
+  console.log(
```
rebelchris, line 56: Just checking you meant to keep all the logs here?
It's not blocking, just wanting to make sure.

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,177 @@
...
+  };
+};
+
+interface ViewWithDateTz extends View {
+  date: Date;
+  maxTimestamp: Date;
+}
+
+const FREEZE_DAYS_IN_A_WEEK = 2;
+const MISSED_LIMIT = 1;
```
rebelchris, line 75: Didn't we have this somewhere already?
Let's make it shared common so we avoid later on only changing one spot :)

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,177 @@
...
+const computeReadingStreaksData = async (
+  con: DataSource,
+  user: Partial<User>,
+): Promise<DeepPartial<UserStreak>> => {
+  const timeZone = user.timezone ?? 'UTC';
+
+  const views: ViewWithDateTz[] = await con
+    .getRepository(View)
+    .createQueryBuilder()
+    .select([])
```
rebelchris, line 86: Why is the empty select here?

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,177 @@
...
+
+      // restart streak if difference is greater than 1, taking freezes into account
+      if (
+        (day === Day.Sunday && difference > FREEZE_DAYS_IN_A_WEEK) ||
+        (day === Day.Monday &&
+          difference > FREEZE_DAYS_IN_A_WEEK + MISSED_LIMIT) ||
+        (day > Day.Monday && difference > MISSED_LIMIT)
+      ) {
+        currentStreak = 1;
+      }
```
rebelchris, line 123: This logic is the same as the one in @sshanzel function (right?)
If it's the case can we uniform it

rebelchris (commented):
Bit out of date with the project, so mostly questions I can't fully recall 😅

bin/seedUserStreaks.ts
kopancek: No idea anymore 😊

bin/seedUserStreaks.ts
kopancek: Yep, let's keep the log statements, at least we know what was happening if something fails.

capJavert (commented):
Generally looks good, few questions usually in mind with migrations like this:
- new user streak is still not used in production?
- what happens if we screw up the streaks, how can we now which ones do we need to remove/update again since we already have the records inside for new users?

kopancek:
> * new user streak is still not used in production?
The streaks are being counted, but since we didn't seed the data, the numbers are only correct for new users who joined after the newView worker was updated to count the streaks. Hence why we need this script

> * what happens if we screw up the streaks, how can we now which ones do we need to remove/update again since we already have the records inside for new users?

In a previous version there was a unit test, which tested all possible combinations, so I am reasonably sure that it works just fine. But even if it does not, the current numbers are incorrect for majority of our users, so we will not make it worse. We can always rerun the script for all users to recalculate. If we ever need to do that, we can modify the update part to do an upsert,  so that it only updates if the streak numbers differ.

If there is an error during running of the script, we can continue where we left of by supplying the last `userPointer`, which is logged in the console.

bin/seedUserStreaks.ts
```diff
@@ -0,0 +1,168 @@
...
+  con: DataSource,
+  insertQueue: fastq.queueAsPromised<Partial<User>[], void>,
+  userPointer?: UserPointer,
+  limit: number = BATCH_SIZE,
+): Promise<UserPointer> => {
+  const qb = con
+    .getRepository(User)
+    .createQueryBuilder('u')
+    // fetching createdAt as a string, because Date object in node is milisecond precision, but postgres is microsecond
+    .select(['id', 'timezone', '"createdAt"::text'])
```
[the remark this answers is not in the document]
kopancek: The devil is in the microsecond precision part. 

If we run the query without the text conversion, [typeorm messes up the microsecond precision and rounds up to the nearest milisecond instead](https://github.com/typeorm/typeorm/issues/4964). This is caused by the pg driver that typeorm uses. 

Which on my local env means, that the script keeps running forever, because:
1. user has `createdAt` set to `2024-02-15 19:24:19.474025`
2. returned value for `createdAt` is `2024-02-15 19:24:19.474Z` -> the `025` portion is rounded down
3. When running the where condition, `createdAt` > `2024-02-15 19:24:19.474Z` it succeeds, since `2024-02-15 19:24:19.474025 > 2024-02-15 19:24:19.474`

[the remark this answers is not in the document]
capJavert: Actually good that you caught it even. 👍

capJavert (approved):
From my end I don't see anything breaking. I think we can try it when the time is right. Worst case we will have some cleanup to do.

PR #1669, opened 2024-02-06
feat: store devcard 2.0 fields in DB

kopancek:
Adds new fields in the `dev_card` table:
- `theme`
- `isProfileCover`
- `showBorder`

MI-159 #done

### DIFF
--- a/src/entity/DevCard.ts
+++ b/src/entity/DevCard.ts
@@ -36,7 +36,7 @@ export class DevCard {
   @Column({ type: 'enum', enum: DevCardTheme, default: DevCardTheme.Default })
   theme: DevCardTheme;
 
-  @Column({ type: 'bool' })
+  @Column({ type: 'bool', default: false })
   isProfileCover: boolean;
 
   @Column({ type: 'bool', default: true })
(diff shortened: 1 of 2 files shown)


src/entity/DevCard.ts
```diff
@@ -22,6 +33,15 @@ export class DevCard {
   @Column({ type: 'text', nullable: true })
   background: string | null;
 
+  @Column({ type: 'enum', enum: DevCardTheme, default: DevCardTheme.Default })
```
idoshamun, line 36: Usually we store enums as string in the DB so we will not have to change the schema

src/entity/DevCard.ts
```diff
@@ -22,6 +33,15 @@ export class DevCard {
   @Column({ type: 'text', nullable: true })
   background: string | null;
 
+  @Column({ type: 'text', default: DevCardTheme.Default })
+  theme: DevCardTheme;
+
+  @Column({ type: 'bool', default: false })
+  isProfileCover: boolean;
```
nensidosari, line 40: Just adding it here in case we want to do it that way, future proof - can we make this field `devCardCover` and it stores a string: either `default`, `profile cover` or `....url...` so we have the option for a new cover if we want to later to allow users to upload a new img.

But non blocking.

nensidosari (approved):
One comment to consider, but non blocking, trust your judgment.

idoshamun (approved):

rebelchris (approved):

PR #1670, opened 2024-02-07
feat: get users favorite sources

sshanzel:
Due to having different variations of the devcard, we are rendering it as an actual HTML element. With that, we need to pull the user's favorite sources.

Instead of just returning the actual logos, we are returning the sources since we need more information for redirection when users click the logos.

Note: the existing query doesn't include returning private source reads

### DIFF
--- a/src/common/devcard.ts
+++ b/src/common/devcard.ts
@@ -25,13 +25,14 @@ export const getMostReadTags = async (
   }));
 };
 
-const getFavoriteSourcesLogos = async (
+export const generateGetFavoriteSources = (
   con: DataSource,
   userId: string,
-): Promise<string[]> => {
-  const sources = await con
+  alias = 'source',
+) => {
+  return con
     .createQueryBuilder()
-    .select('min(source.image)', 'image')
+    .select(`p."sourceId"`, 'id')
     .from(View, 'v')
     .innerJoin(Post, 'p', 'v."postId" = p.id')
     .innerJoin(
@@ -46,15 +47,37 @@ const getFavoriteSourcesLogos = async (
     )
     .innerJoin(
       Source,
-      'source',
-      'source.id = p."sourceId" and source.active = true and source.private = false',
+      alias,
+      `"${alias}".id = p."sourceId" and "${alias}".active = true and "${alias}".private = false`,
     )
     .where('v."userId" = :id', { id: userId })
     .andWhere(`s.count > 10`)
     .groupBy('p."sourceId"')
     .orderBy('count(*) * 1.0 / min(s.count)', 'DESC')
-    .limit(5)
-    .getRawMany();
+    .limit(5);
+};
+
+export const getFavoriteSourcesIds = async (
+  con: DataSource,
+  userId: string,
+): Promise<string[]> => {
+  const query = generateGetFavoriteSources(con, userId);
+  const sources = await query.getRawMany();
+
+  return sources.map((source) => source.id);
+};
+
+export const getFavoriteSourcesLogos = async (
+  con: DataSource,
+  userId: string,
+): Promise<string[]> => {
+  const alias = 'source';
+  const query = generateGetFavoriteSources(con, userId, alias).addSelect(
+    `min("${alias}".image)`,
+    'image',
+  );
+  const sources = await query.getRawMany();
+
   return sources.map((source) => source.image);
 };
 
--- a/src/common/devcard.ts
+++ b/src/common/devcard.ts
@@ -30,7 +30,7 @@ interface FavoriteSourcesProps {
   limit: number;
 }
 
-export const generateGetFavoriteSources = (
+export const generateFavoriteSourcesQuery = (
   con: DataSource,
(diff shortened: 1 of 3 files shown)


src/schema/sources.ts
```diff
@@ -283,6 +289,11 @@ export const typeDefs = /* GraphQL */ `
     """
     sourceByFeed(feed: String!): Source @auth
 
+    """
+    Get user's favorite sources
+    """
+    favoriteSources(userId: ID!, limit: Int): [Source]
```
capJavert, line 295: Missing `@auth`?

src/common/devcard.ts
```diff
@@ -25,13 +25,19 @@ export const getMostReadTags = async (
...
+  alias: string;
+  limit: number;
+}
+
+export const generateGetFavoriteSources = (
   con: DataSource,
   userId: string,
-): Promise<string[]> => {
-  const sources = await con
+  { alias = 'source', limit = 5 }: Partial<FavoriteSourcesProps> = {},
```
capJavert, line 36: Maybe name it `generateFavoriteSourcesQuery` just so its clear it returns a query builder object? I had to look at the usage to figure it does that?

src/schema/sources.ts
sshanzel: Should be intended not to have `@auth` for the tooltip part when viewing other users.

capJavert: But you could have auth and see other users still, its just so that someone who is not logged in the platform can scrape and find favorites from user ids. That feels like it should not be possible?

src/common/devcard.ts
```diff
@@ -25,13 +25,19 @@ export const getMostReadTags = async (
...
+  alias: string;
+  limit: number;
+}
+
+export const generateFavoriteSourcesQuery = (
   con: DataSource,
   userId: string,
-): Promise<string[]> => {
-  const sources = await con
+  { alias = 'source', limit = 5 }: Partial<FavoriteSourcesProps> = {},
```
rebelchris, line 36: What alternative `alias` do we use?

src/common/devcard.ts
```diff
@@ -46,15 +52,38 @@ const getFavoriteSourcesLogos = async (
     )
     .innerJoin(
       Source,
-      'source',
-      'source.id = p."sourceId" and source.active = true and source.private = false',
+      alias,
+      `"${alias}".id = p."sourceId" and "${alias}".active = true and "${alias}".private = false`,
```
rebelchris, line 56: For devCard can we not remove the private omit?
It seems "ok" to expose private squads basically.

src/common/devcard.ts
```diff
@@ -46,15 +52,38 @@ const getFavoriteSourcesLogos = async (
...
+): Promise<string[]> => {
+  const alias = 'source';
+  const query = generateFavoriteSourcesQuery(con, userId, {
+    alias,
+    limit: 5,
+  }).addSelect(`min("${alias}".image)`, 'image');
+  const sources = await query.getRawMany();
+
   return sources.map((source) => source.image);
 };
```
rebelchris, line 88: Do we query these seperate or always same time?
(Feels like it's just one query with different graphQL props?

rebelchris (commented):
Minor questions

src/common/devcard.ts
capJavert: I think its just when you want to modify the query since the function returns the query builder vs the data.

src/common/devcard.ts
idoshamun: We'd rather not to though. Just in case

src/schema/sources.ts
idoshamun: Anonymous users should have access to it. Lee is right. Imagine someone landing on a post page hovering over a profile, they expected to see the devcard

src/schema/sources.ts
```diff
@@ -983,6 +994,26 @@ export const resolvers: IResolvers<any, Context> = {
...
+        return [];
+      }
+
+      return graphorm.query<GQLSource>(ctx, info, (builder) => {
+        builder.queryBuilder = builder.queryBuilder.andWhere(
+          `${builder.alias}.id IN (:...ids)`,
+          { ids: sourceIds },
+        );
+        return builder;
+      });
```
idoshamun, line 1015: Instead of doing that as two queries we can merge it into one.
We have the query builder so we can generate a query out of it and use it in the `IN` operator.
This is a suggestion, I understand the hackathon pressure

idoshamun (approved):

src/schema/sources.ts
capJavert: Got it 👍

src/common/devcard.ts
sshanzel: Yes, this is to make it easier to add queries and avoid issues of not knowing the relevant entity. However, we will now close this PR.

nensidosari:
Closing as this was duplicated in another PR

PR #1671, opened 2024-02-07
feat: add devcards route to generate devcard images

kopancek:
# Changes

- moved graphql mutation and query to it's own resolver
- removed svg implementation, since it's not needed
- changed read path for devcard image URL to redirect to the new URL for backwards compatibility for existing URLs that are out there already
- unit test changes

MI-157 #done


src/routes/devcards.ts
```diff
@@ -1,63 +1,16 @@
...
-          .send(svgString);
-      } catch (err) {
-        if (err.code === '22P02') {
-          return res.status(404).send();
-        }
-        throw err;
-      }
+      const [id] = req.params.id?.split('.') ?? [];
+      const query = qs.stringify(req.query);
+      return res.redirect(
```
idoshamun, line 10: This will break production basically right? We'll get ugly images

kopancek: This PR is merging to `devcard2-hackathon` branch, so these changes are not going to production yet.

src/migration/1707227186596-DevCard20Settings.ts
```diff
@@ -0,0 +1,25 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DevCard20Settings1707227186596 implements MigrationInterface {
+  name = 'DevCard20Settings1707227186596';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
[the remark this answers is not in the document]
kopancek: hmm, not sure what happened there 😢

nensidosari (approved):
🚀

src/schema/devcards.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      // Avoid caching issues when devcard is generated again
+      const randomStr = Math.random().toString(36).substring(2, 5);
+      const queryStr = qs.stringify({
+        type,
+        r: randomStr,
+      });
+
+      return {
+        imageUrl: `${process.env.OG_URL}/devcards/${devCard.id}.png?${queryStr}`,
```
rebelchris, line 181: Probably neater to use URL builder + searchParams, but non blocking

src/schema/devcards.ts
```diff
@@ -0,0 +1,185 @@
...
+      theme: DevCardTheme
+      type: DevCardType
+      isProfileCover: Boolean
+      showBorder: Boolean
+    ): DevCardImage @auth
+  }
+`;
+
+interface GenerateDevCardInput {
+  theme: DevCard['theme'];
```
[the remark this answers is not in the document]
kopancek: I started with just the theme and then added more parameters. Will switch it to `extends Pick<DevCard ...` in a followup PR.

As for the enum type - this string union type is only used to parse the graphql query, where we get the data as an uppercase string and then pass it on to the URL.

The URL is [parsed in the og repo and there we use a proper enum type](https://github.com/dailydotdev/og/pull/2/files#diff-f2fdc3d96f51f3df3808d3d5d1ce10301abe8c715922645f1087c2af50aaec17R15-R19). I don't know an easy way to share the same definition in both repositories, otherwise I would.

src/schema/devcards.ts
capJavert: I don't think you need omitBy and isEmpty. We just check if type is defined and then add to URL.searchParams with append. 

```js
const url = new URL(`/devcards/${devCard.id}.png`, process.env.OG_URL)
url.searchParams.append('r', randomStr)

if (type) {
  url.searchParams.append('type', type)
}

return { imageUrl: url.toString() };
```

PR #1673, opened 2024-02-08
Mi 153 dev card notification

nensidosari:
- adding new notification for devcard unlock
- added action to make sure this notification is only sent one when user passes the reputation threshold.
- added tests for the new notification

Build is failing from the template id missing - waiting for it from product. Will update once ready


src/notifications/generate.ts
```diff
@@ -296,4 +298,11 @@ export const generateNotificationMap: Record<
       .avatarManySources(ctx.sources)
       .numTotalAvatars(ctx.total)
       .uniqueKey(ctx.post.metadataChangedAt?.toString()),
+  dev_card_unlocked: (builder) =>
+    builder
+      .icon(NotificationIcon.Bell)
+      .description(
+        'You can now generate your own DevCard to showcase your daily.dev achievements.',
+      )
+      .targetUrl(generateDevCard),
```
idoshamun, line 307: We need to add `.uniqueKey(ctx.userIds[0])` here. Otherwise after the first notification, we will get duplicate error issue

src/workers/notifications/devCardUnlocked.ts
```diff
@@ -0,0 +1,37 @@
...
+    const userAction = await con
+      .getRepository(UserAction)
+      .findOneBy({ userId: data.userId, type: UserActionType.DevCardUnlocked });
+
+    if (userAction) return null;
+
+    await con.getRepository(UserAction).insert({
+      userId: data.userId,
+      type: UserActionType.DevCardUnlocked,
+    });
```
idoshamun, line 27: Why do we need the action? It's not really needed as you guarded yourself in the cdc condition and same notification for the same user can't be triggered more than once

src/workers/cdc/primary.ts
```diff
@@ -300,6 +301,13 @@ const onUserChange = async (
         }
       }
     }
+    if (data.payload.after.reputation > data.payload.before.reputation) {
```
sshanzel, line 304: Nonblocking but I think we should do the check of the threshold here rather than the worker. This is to avoid sending a pubsub event in every reputation increase which is not necessary on some cases.

sshanzel (approved):
Just had one non-blocking comment, but other than that, this looks great! :shipit:

src/workers/cdc/primary.ts
nensidosari: This is now notification for reputation increase, so not directly related to the devcard threshold

src/workers/notifications/devCardUnlocked.ts
```diff
@@ -0,0 +1,30 @@
...
+interface Data {
+  user: ChangeObject<User>;
+  userAfter: ChangeObject<User>;
+}
+
+const worker: NotificationWorker = {
+  subscription: 'api.user-reputation-updated-notification',
+  handler: async (message) => {
+    const data: Data = messageToJson(message);
+    if (data.userAfter.reputation < DEFAULT_DEV_CARD_UNLOCKED_THRESHOLD)
```
idoshamun, line 19: Shouldn't this be better?

```suggestion
    if (data.user.reputation > DEFAULT_DEV_CARD_UNLOCKED_THRESHOLD || data.userAfter.reputation < DEFAULT_DEV_CARD_UNLOCKED_THRESHOLD)
```

nensidosari: Yup, thank u

idoshamun (approved):

PR #1674, opened 2024-02-08
feat: add more fields to devcard sources list

kopancek:
Changed the query to return more fields on the sources, not just the URL string.

Also added feedback from Related PR: 
- #1671

### DIFF
--- a/src/schema/devcards.ts
+++ b/src/schema/devcards.ts
@@ -1,5 +1,5 @@
 import { IResolvers } from '@graphql-tools/utils';
-import qs from 'qs';
+import { omitBy, isEmpty } from 'lodash';
 
 import { Context } from '../Context';
 import { traceResolverObject } from './trace';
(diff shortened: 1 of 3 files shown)


src/schema/devcards.ts
```diff
@@ -172,13 +190,20 @@ export const resolvers: IResolvers<any, Context> = {
 
       // Avoid caching issues when devcard is generated again
       const randomStr = Math.random().toString(36).substring(2, 5);
-      const queryStr = qs.stringify({
-        type,
-        r: randomStr,
-      });
+      const queryStr = new URLSearchParams(
```
rebelchris, line 193: We can just use the plain URL one, no need to use the lodash ones I think.

sshanzel (approved):
Looks great! 🚀

src/schema/devcards.ts
kopancek: If we do not add the `omitBy`, we end up with `type=undefined` in the query string. I'd rather not have empty parameters there :)

nensidosari (approved):
LGTM

src/schema/devcards.ts
```diff
@@ -40,6 +40,26 @@ export const typeDefs = /* GraphQL */ `
     X
   }
 
+  """
+  Source for dev card frequent sources
+  """
+  type DevCardSource {
```
idoshamun, line 46: Can we utilize the existing source type please? I know it'll be somewhat broken but it's better for consistency. We can always fix fact that it's broken after the hackathon

PR #1681, opened 2024-02-11
feat: posts sitemap

idoshamun:


omBratteng (approved):

src/routes/sitemaps.ts
```diff
@@ -0,0 +1,28 @@
...
+      .createQueryBuilder()
+      .select(`concat('${process.env.COMMENTS_PREFIX}', '/posts/', id)`, 'url')
+      .from(Post, 'p')
+      .where('type NOT IN (:...types)', { types: [PostType.Welcome] })
+      .andWhere('NOT private')
+      .andWhere('NOT banned')
+      .andWhere('NOT deleted')
+      .andWhere('"createdAt" > current_timestamp - interval \'90 day\'')
+      .orderBy('"createdAt"', 'DESC')
+      .limit(50_000);
```
rebelchris, line 18: We might have to consider paginated sitemaps for the amount we have.

Think most parsers struggle with big amounts in root, making the expectef behaviour not accomplished.

Should be a testing tool for this

idoshamun: Yes, I checked 50k is the max

PR #1683, opened 2024-02-12
fix: revert changes to generateDevCard mutation

kopancek:
Introduces the generateDevCardV2 mutation instead. 
Also revived the tests.


idoshamun (approved):
Awesome! thanks for the quick fix

PR #1687, opened 2024-02-13
fix: object used for query params

sshanzel:
The `Params` and `Querystring` are two different objects. The former is for required values and the latter is for optional, the type is for the latter.

Should fix (and verified locally) that the correct type is passed.

### DIFF
--- a/src/routes/devcards.ts
+++ b/src/routes/devcards.ts
@@ -79,11 +79,10 @@ export default async function (fastify: FastifyInstance): Promise<void> {
   //   },
   // );
 
-  fastify.get<{ Params: { name: string; type: string } }>(
+  fastify.get<{ Params: { name: string }; Querystring: { type: string } }>(
     '/v2/:name',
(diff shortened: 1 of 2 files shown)


src/routes/devcards.ts
```diff
@@ -79,11 +79,10 @@ export default async function (fastify: FastifyInstance): Promise<void> {
   //   },
   // );
 
-  fastify.get<{ Params: { name: string; type: string } }>(
+  fastify.get<{ Params: { name: string }; Querystring: { type: string } }>(
```
kopancek, line 82: Good catch!

kopancek (approved):
Would be great to add a unit test for this as well...

nensidosari (approved):

idoshamun (approved):

sshanzel:
> Would be great to add a unit test for this as well...

Totally agree with that. Should now be present with the recent commit.

PR #1690, opened 2024-02-13
fix: create devcard if not exists and remove 404 checks

kopancek:
- Removed the check from the REST API `/devcards/v2/` route that checks if the devcard exists. No 404 returned there anymore.
- Modified the logic, so that when we query a devcard with the graphql query, it is created if it not exists, instead of returning a 404.


idoshamun (commented):
We should probably still check the user exists right?

kopancek:
> We should probably still check the user exists right?

There's a FK relationship, so if the user does not exist, it would fail on the DB call. Should we return a 404 in that case?

idoshamun (approved):

nensidosari (approved):

PR #1691, opened 2024-02-13
fix: bring back post similarity provider

rebelchris:
Since we solved some memory issues, maybe a good try to test this again.


omBratteng (approved):
I guess... I am still monitoring

PR #1692, opened 2024-02-13
fix: fallback to empty string if feedId does not exist

omBratteng:


capJavert (commented):
In which cases is it empty?

omBratteng:
If you do an unauthenticated query to anon feed (popular feed). Pretty sure that's what's been causing some memory runaways.

If feedId (or userId for that matter) is empty, this select query doesn't have a where statement, and selects 53 million rows, totalling about 4GiB of data 😬

Thread here

https://dailydotdev.slack.com/archives/C014VT1J59V/p1707832311957059

src/schema/feeds.ts
```diff
@@ -935,7 +935,11 @@ const legacySimilarPostsResolver = randomPostsResolver(
 export const resolvers: IResolvers<any, Context> = traceResolvers({
   Query: {
     anonymousFeed: (source, args: AnonymousFeedArgs, ctx: Context, info) => {
-      if (args.version >= 2 && args.ranking === Ranking.POPULARITY) {
+      if (
+        args.version >= 2 &&
+        args.ranking === Ranking.POPULARITY &&
+        ctx.userId
```
rebelchris, line 941: But this is already in annonymousfeed no?
Also think we dknt really surface these anymore in favor of feedservice (but could have some use still)

rebelchris (approved):
Just one remark, not against it more reason to fully move to feed service feeds

rebelchris:
Hmm, but annon cant even get to popular feed? 🫤

src/schema/feeds.ts
idoshamun: It's used for the popular feed. it's possible that some users have an old version. I think it's a valid solution as a quick hack for now

idoshamun (approved):

capJavert (approved):

omBratteng:
> Hmm, but annon cant even get to popular feed? 🫤

Not via webapp, no. There we require login for any feed. But if I were to call api directly, I can still access it. 

E.g. this command which is a cURL of the popular feed.

After you run this, you'll see one of the pods get OOMKilled and you don't get a response.

```console
curl --location 'https://app.daily.dev/api/graphql' \
--header 'Content-Type: application/json' \
--data-raw '{"query":"\n  query AnonymousFeed(\n    $loggedIn: Boolean! = false\n    $first: Int\n    $after: String\n    $ranking: Ranking\n    $version: Int\n    $supportedTypes: [String!] = [\"article\",\"share\",\"freeform\",\"video:youtube\",\"collection\"]\n  ) {\n    page: anonymousFeed(\n      first: $first\n      after: $after\n      ranking: $ranking\n      version: $version\n      supportedTypes: $supportedTypes\n    ) {\n      ...FeedPostConnection\n    }\n  }\n  \n  fragment FeedPostConnection on PostConnection {\n    pageInfo {\n      hasNextPage\n      endCursor\n    }\n    edges {\n      node {\n        ...FeedPost\n        contentHtml\n        ...UserPost @include(if: $loggedIn)\n      }\n    }\n  }\n  \n  fragment FeedPost on Post {\n    ...SharedPostInfo\n    sharedPost {\n      ...SharedPostInfo\n    }\n    trending\n    feedMeta\n    collectionSources {\n      handle\n      image\n    }\n    numCollectionSources\n    updatedAt\n  }\n  \n  fragment SharedPostInfo on Post {\n    id\n    title\n    titleHtml\n    image\n    readTime\n    permalink\n    commentsPermalink\n    summary\n    createdAt\n    private\n    upvoted\n    commented\n    bookmarked\n    views\n    numUpvotes\n    numComments\n    videoId\n    scout {\n      ...UserShortInfo\n    }\n    author {\n      ...UserShortInfo\n    }\n    type\n    tags\n    source {\n      ...SourceBaseInfo\n    }\n    downvoted\n    flags {\n      promoteToPublic\n    }\n    userState {\n      vote\n      flags {\n        feedbackDismiss\n      }\n    }\n  }\n  \n  fragment SourceBaseInfo on Source {\n    id\n    active\n    handle\n    name\n    permalink\n    public\n    type\n    description\n    image\n    membersCount\n    privilegedMembers {\n      user {\n        id\n      }\n      role\n    }\n    currentMember {\n      ...CurrentMember\n    }\n    memberPostingRole\n    memberInviteRole\n  }\n  \n  fragment CurrentMember on SourceMember {\n    user {\n      id\n    }\n    permissions\n    role\n    referralToken\n    flags {\n      hideFeedPosts\n      collapsePinnedPosts\n    }\n  }\n\n\n  \n  fragment UserShortInfo on User {\n    id\n    name\n    image\n    permalink\n    username\n    bio\n  }\n\n\n\n  \n  fragment UserPost on Post {\n    read\n    upvoted\n    commented\n    bookmarked\n    downvoted\n  }\n\n","variables":{"version":24,"first":13,"loggedIn":false}}'
```

src/schema/feeds.ts
omBratteng: Probably don't need it actually, the other fix should already take care of the rouge query.

capJavert: Ideally it should be controlled on entry level eg. routes that need or do not need auth, but as a quick fix to confirm it works I agree its fine.

haiderRizvii (approved):

PR #1694, opened 2024-02-14
refactor: reduce load from the devcard query

idoshamun:
I noticed the query for sources is quite heavy so I reduced the logic a bit to make it easier on the db. DevCard is gonna be more popular now that it's shown on tooltip and everywhere essentially


rebelchris (approved):

omBratteng (approved):

kopancek (approved):
Nice!

PR #1695, opened 2024-02-14
feat: dynamic robots.txt

idoshamun:
* Introduce bots table (will be filled with data using a retool app for now)
* Dynamically generate a robots.txt file based on the bots table and fresh users

### DIFF
--- a/src/entity/user/BotUser.ts
+++ b/src/entity/user/BotUser.ts
@@ -1,0 +1,37 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class BotUser {
+  @PrimaryColumn({ length: 36 })
+  id: string;
+
+  @Column({ type: 'text', nullable: true })
+  name: string;
+
+  @Column({ type: 'text', nullable: true })
+  email: string;
+
+  @Column({ type: 'text', nullable: true })
+  company?: string;
+
+  @Column({ type: 'text', nullable: true })
+  title?: string;
+
+  @Column({ length: 39, nullable: true })
+  username?: string;
+
+  @Column({ type: 'text', nullable: true })
+  bio?: string;
+
+  @Column({ type: 'text', nullable: true })
+  portfolio?: string;
+
+  @Column({ type: 'text', nullable: true })
+  timezone?: string;
+
+  @Column({ nullable: false })
+  createdAt: Date;
+
+  @Column({ type: 'text', nullable: true })
+  readme?: string;
+}
--- a/src/routes/sitemaps.ts
+++ b/src/routes/sitemaps.ts
@@ -31,7 +31,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
 
     return res
       .type('text/plain')
-      .header('cache-control', 'public, max-age=14400')
+      .header('cache-control', 'public, max-age=14400, s-max-age=14400')
       .send(stream);
   });
 
@@ -76,7 +76,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
 
     return res
       .type('text/plain')
-      .header('cache-control', 'public, max-age=3600')
+      .header('cache-control', 'public, max-age=3600, s-max-age=3600')
       .send(stream);
   });
 }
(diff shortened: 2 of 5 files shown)


src/routes/sitemaps.ts
```diff
@@ -25,4 +34,49 @@ export default async function (fastify: FastifyInstance): Promise<void> {
       .header('cache-control', 'public, max-age=14400')
       .send(stream);
   });
+
+  fastify.get('/robots.txt', async (req, res) => {
+    const con = await createOrGetConnection();
+    const bots = con
+      .createQueryBuilder()
+      .select('username')
+      .from(BotUser, 'b')
```
capJavert, line 43: How is `bot_user` table filled?

src/routes/sitemaps.ts
```diff
@@ -25,4 +34,49 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
+    baseRobotsStream.push(BASE_ROBOTS);
+    baseRobotsStream.push(null);
+
+    let stream = new PassThrough();
+    stream = baseRobotsStream.pipe(stream, { end: false });
+    stream = usersStream.pipe(stream, { end: true });
+
+    return res
+      .type('text/plain')
+      .header('cache-control', 'public, max-age=3600')
```
capJavert, line 79: Does this actually caches on the CDN? We don't have one for API, thinking since it will be rewrote in vercel we could add s-maxage so their CDN picks it up and caches for a while?

src/routes/sitemaps.ts
```diff
@@ -25,4 +34,49 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
+      .orderBy('"createdAt"', 'DESC')
+      .where('"createdAt" > now() - interval \'14 day\'')
+      .andWhere('reputation <= 10')
+      .limit(10_000)
+      .getQuery();
+
+    const queryStream = await con
+      .createQueryBuilder()
+      .select('*')
+      .from(`((${bots}) UNION (${freshUsers}))`, 'a')
```
capJavert, line 61: Basically omit new users and bots?

src/routes/sitemaps.ts
idoshamun: We do have CDN enabled for the API side. Can add also s-maxage if needed, no problem

src/routes/sitemaps.ts
idoshamun: Yes

src/routes/sitemaps.ts
idoshamun: Retool app currently that is manually reviewed by me. Later I'll do something more automatic

src/routes/sitemaps.ts
capJavert: Nice 👌

capJavert: Yeah, s-maxage is mostly used for server side expiration so I would add it if CDN knows to pick it up.

idoshamun: Added it everywhere

capJavert (commented):
Let's try it 🚀

capJavert (approved):

src/entity/user/BotUser.ts
```diff
@@ -0,0 +1,37 @@
...
+  portfolio?: string;
+
+  @Column({ type: 'text', nullable: true })
+  timezone?: string;
+
+  @Column({ nullable: false })
+  createdAt: Date;
+
+  @Column({ type: 'text', nullable: true })
+  readme?: string;
```
rebelchris, line 36: Just wondering do we need all these fields for anything?

src/routes/sitemaps.ts
```diff
@@ -22,7 +31,52 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
+    const baseRobotsStream = new Readable();
+    baseRobotsStream._read = () => {};
+    baseRobotsStream.push(BASE_ROBOTS);
+    baseRobotsStream.push(null);
+
+    let stream = new PassThrough();
+    stream = baseRobotsStream.pipe(stream, { end: false });
+    stream = usersStream.pipe(stream, { end: true });
+
+    return res
```
rebelchris, line 77: What about the couple robot entries I did today already? 👀

PR #1697, opened 2024-02-14
fix: devcard svg embedded png url not correct in prod

kopancek:
With the original code, we ended up with `https://app.daily.dev/devcards/v2/userId.png` instead of `https://app.daily.dev/api/devcards/v2/userId.png` in the end. The `/api` portion got swallowed somewhere on the way.

With this code, we point to `https://api.daily.dev/devcards/userId.png` instead.


idoshamun (approved):
I don't remember exactly the value of this env var but let's try :)

PR #1698, opened 2024-02-15
chore: add optimization indexes

rebelchris:
With some of the latest additions we are actually not leveraging all indexes needed.
Thus added the missing ones.


idoshamun (commented):
You forgot the migration though 😬

idoshamun (approved):

PR #1700, opened 2024-02-15
chore: remove indexes

rebelchris:
Removed the unused indexes:

Screenshot 2024-02-15 at 12 56 14

### DIFF
--- a/src/entity/SourceMember.ts
+++ b/src/entity/SourceMember.ts
@@ -18,7 +18,6 @@ export type SourceMemberFlagsPublic = Pick<
 >;
 
 @Entity()
-@Index('IDX_source_member_userId_flags_hideFeedPosts', { synchronize: false })
 @Index('IDX_source_member_userId_role', ['userId', 'role'])
 @Index('IDX_source_member_sourceId_role', ['sourceId', 'role'])
 export class SourceMember {
--- a/src/entity/SourceMember.ts
+++ b/src/entity/SourceMember.ts
@@ -20,6 +20,7 @@ export type SourceMemberFlagsPublic = Pick<
 @Entity()
 @Index('IDX_source_member_userId_flags_hideFeedPosts', { synchronize: false })
 @Index('IDX_source_member_userId_role', ['userId', 'role'])
+@Index('IDX_source_member_sourceId_role', ['sourceId', 'role'])
 export class SourceMember {
   @PrimaryColumn({ type: 'text' })
   @Index('IDX_source_member_sourceId')
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -44,6 +44,12 @@ export type PostFlags = Partial<{
 export type PostFlagsPublic = Pick<PostFlags, 'private' | 'promoteToPublic'>;
 
 @Entity()
+@Index('IDX_post_deleted_visible_type_views', [
+  'deleted',
+  'visible',
+  'type',
+  'views',
+])
 @Index('IDX_post_source_id_pinned_at_created_at', [
   'sourceId',
   'pinnedAt',
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -101,7 +101,6 @@ export class Post {
   metadataChangedAt: Date;
 
   @Column({ type: 'text', default: UNKNOWN_SOURCE })
-  @Index()
   sourceId: string;
 
   @ManyToOne(() => Source, (source) => source.posts, {
@@ -169,7 +168,6 @@ export class Post {
   trending?: number;
 
   @Column({ nullable: true })
-  @Index('IDX_post_last_trending')
   lastTrending?: Date;
 
   @Column({ type: 'integer', nullable: true })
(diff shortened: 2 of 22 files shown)


src/entity/posts/Post.ts
```diff
@@ -215,10 +211,6 @@ export class Post {
   downvotes: number;
 
   @Column({ type: 'jsonb', default: {} })
-  @Index('IDX_post_flags_sentAnalyticsReport', { synchronize: false })
-  @Index('IDX_post_flags_banned', { synchronize: false })
-  @Index('IDX_post_flags_deleted', { synchronize: false })
-  @Index('IDX_post_flags_promoteToPublic', { synchronize: false })
```
capJavert, line 221: I think these should not be deleted, its just a matter of we still not using the flags instead of columns.

src/entity/user/User.ts
```diff
@@ -78,18 +78,15 @@ export class User {
   bio?: string;
 
   @Column({ length: 15, nullable: true })
-  @Index('users_twitter_unique', { unique: true })
```
capJavert, line 81: Would this not allow adding duplicate twitter handles? We even check this error in `updateProfile` mutation?

src/entity/posts/Post.ts
```diff
@@ -101,7 +101,6 @@ export class Post {
   metadataChangedAt: Date;
 
   @Column({ type: 'text', default: UNKNOWN_SOURCE })
-  @Index()
```
capJavert, line 104: Curious how this does not get hit for example in `sourceFeed` where we have `builder.andWhere('s.sourceId = :sourceId', { sourceId });`

vpol:
I'm not typeorm expert, but looks good. 🚀

src/entity/posts/Post.ts
rebelchris: At this point they don't ever hit according to the query though, so either they wrong or not used.

src/entity/user/User.ts
rebelchris: Can revert the unique ones to be safe rather.

vpol: I'd go over unique once and double check which ones we need.

src/entity/posts/Post.ts
capJavert: They are not used. We have to make not to return them when we start using `flags.deleted` instead of `deleted` column (whenever that happens).

rebelchris: Ah gotcha :)

rebelchris: Let's reintroduce separately then once we do

src/entity/posts/Post.ts
rebelchris: Not actually sure tbh, anyway we can verify it further? (maybe it was a duplicate?)

capJavert: Maybe quickly try with EXPLAIN ANALYSE and see if there are differences for those queries locally with or without index? Just to validate, we might get some good insight.

rebelchris: @capJavert :

With index on source_id:
Screenshot 2024-02-15 at 15 58 50

Without index on source_id:
Screenshot 2024-02-15 at 15 59 20

capJavert: Ah yes, so it reuses `IDX_post_source_id_pinned_at_created_at` if the column index is not there. So basically what could happen is that we remove `IDX_post_source_id_pinned_at_created_at` and then you end up with slow queries.

vpol: it does not have to use specific index. if it constantly has some index in memory (because it's used for other things) - it will stick to using it even if there is potentially more "narrow" alternative index.

rebelchris: Ah didn't realize that's how it would prefer indexes, always thought it would go for the narrowest one.
Anyway we have a tool to tell us if we MISS indexes, not the other way around, so let's just remove it for now 😅

src/entity/SourceMember.ts
```diff
@@ -18,7 +18,6 @@ export type SourceMemberFlagsPublic = Pick<
 >;
 
 @Entity()
-@Index('IDX_source_member_userId_flags_hideFeedPosts', { synchronize: false })
```
capJavert, line 21: This one might be invalid if it was not hit, I would expect it hit from `feedToFilters` so just keep a lookout on the tool if it reports anything for this flag. Maybe got broken after PG migration.

capJavert (approved):
Let's try it but keep an eye out for any slowness or missed indexes.

vpol (approved):

src/entity/posts/Post.ts
idoshamun: if we know we gonna use them in the near future, can we keep them please? Introducing indexes on the posts table can be expensive

rebelchris: We can, but think they'll be different anyway.

Lets leave them existing for now 👍

idoshamun: Your call then. If you want remove them

vpol: > Introducing indexes on the posts table can be expensive

with `CONCURRENCLY` shouldn't be.

rebelchris:
@idoshamun Decided to bring the post flags back, let me know if it's good to go

idoshamun (approved):

PR #1702, opened 2024-02-16
fix: missing meta tag for twitter preview

sshanzel:
I checked and on most social media platforms our API link works, but fails on Twitter.

After further investigation, Twitter requires a bit more information by adding meta tags.

Reference: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image
image

I have also utilize some online tool to troubleshoot the issue we are encountering, as can be seen from below, on a regular post, we are getting the correct meta tags, but with the API link, we are missing it.

Post link (meta tags found):
Screenshot 2024-02-16 at 5 48 48 PM

Screenshot 2024-02-16 at 5 50 44 PM

DevCard svg link (no meta tags found):
Screenshot 2024-02-16 at 5 48 39 PM

Screenshot 2024-02-16 at 5 51 23 PM

### After merging

We can again, easily test whether the changes would work through the online tool.

### DIFF
--- a/src/routes/devcards.ts
+++ b/src/routes/devcards.ts
@@ -96,7 +96,12 @@ export default async function (fastify: FastifyInstance): Promise<void> {
 
         // for svg, return the same image as png, wrapped in svg tag
         if (format === 'svg') {
-          const svgString = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><image xlink:href="${req.protocol}://${req.hostname}${req.originalUrl.replace('.svg', '.png')}" /></svg>`;
+          const svgString = `
+            <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+              <meta name="twitter:card" content="summary_large_image">
+              <meta name="twitter:title" content="DevCard of @${user.username}">
+              <image xlink:href="${req.protocol}://${req.hostname}${req.originalUrl.replace('.svg', '.png')}" />
+            </svg>`;
 
           return res
             .type('image/svg+xml')
--- a/src/schema/devcards.ts
+++ b/src/schema/devcards.ts
@@ -200,7 +200,7 @@ export const resolvers: IResolvers<any, Context> = {
         imageUrl: `${process.env.URL_PREFIX}/devcards/${devCard.id.replace(
           /-/g,
           '',
-        )}.png?r=${randomStr}`,
+        )}.svg?r=${randomStr}`,
       };
     },
     generateDevCardV2: async (


src/schema/devcards.ts
```diff
@@ -200,7 +200,7 @@ export const resolvers: IResolvers<any, Context> = {
         imageUrl: `${process.env.URL_PREFIX}/devcards/${devCard.id.replace(
           /-/g,
           '',
-        )}.png?r=${randomStr}`,
+        )}.svg?r=${randomStr}`,
```
sshanzel, line 203: We might need to turn the response to SVG so that when the user copies the link, they will have the SVG at hand.

nensidosari (approved):
Looks good to me, not sure if we had any requirements to actually have this as a png though? 🤔

src/routes/devcards.ts
```diff
@@ -99,7 +99,12 @@ export default async function (fastify: FastifyInstance): Promise<void> {
         if (format === 'svg') {
           const pngUrl = `${process.env.URL_PREFIX}${req.originalUrl.replace('.svg', '.png')}`;
           const encodedUrl = encode(pngUrl, { useNamedReferences: true });
-          const svgString = `<svg xmlns="http://www.w3.org/2000/svg"><image href="${encodedUrl}" /></svg>`;
+          const svgString = `
+            <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+              <meta name="twitter:card" content="summary_large_image">
+              <meta name="twitter:title" content="DevCard of @${user.username}">
```
rebelchris, line 105: Meta works inside svg? 👀
Happy to try, but wondering if it's supported like this.

Is there maybe some other tool we can test hardcoded HTML in?

rebelchris (approved):
Not 100% sure, but doesn't seem too breaking.
Let's also ask @capJavert

idoshamun:
Isn't possible to actually share an image via the twitter share params instead of changing the api project?

nensidosari:
> Isn't possible to actually share an image via the twitter share params instead of changing the api project?
So append these kind of info in the app you mean where we share on twitter? Can check

idoshamun:
@nensidosari exactly

src/routes/devcards.ts
```diff
@@ -99,7 +99,12 @@ export default async function (fastify: FastifyInstance): Promise<void> {
         if (format === 'svg') {
           const pngUrl = `${process.env.URL_PREFIX}${req.originalUrl.replace('.svg', '.png')}`;
           const encodedUrl = encode(pngUrl, { useNamedReferences: true });
-          const svgString = `<svg xmlns="http://www.w3.org/2000/svg"><image href="${encodedUrl}" /></svg>`;
+          const svgString = `
+            <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
kopancek, line 103: We should not need the xlink namespace. It's not used anywhere in the svg from what I can see.

src/routes/devcards.ts
kopancek: This is my question as well. Should we instead return html from the OG link with meta tags embedded in? Then we don't need to modify the API...

kopancek: Even the spec says, it should be an HTML page instead: https://ogp.me/

src/schema/devcards.ts
kopancek: I think it's much clearer to return png here and rely on HTML generation for the opengraph link.

kopancek (commented):
I'm not sure this is the right direction

src/routes/devcards.ts
capJavert: Agree with what @kopancek and @idoshamun said, never seen this approach personally.

omBratteng: We can do what those with fake ads on Twitter does 😈
They check the user-agent header, if it's a browser, just return the image, otherwise wrap it in a html block
image

sshanzel:
Closing the PR as discussed, our intention is to have an image shared rather than having it displayed through OG image (best-case scenario).

PR #1703, opened 2024-02-16
feat(digest): load dynamic email meta from gb feature

capJavert:
AS-119
Also AS-120


src/workers/personalizedDigestEmail.ts
```diff
@@ -215,28 +274,21 @@ const worker: Worker = {
...
+    });
     const emailPayload: MailDataRequired = {
       ...baseNotificationEmailData,
-      to: {
-        email: user.email,
-        name: userName,
-      },
       sendAt: Math.floor(emailSendDate.getTime() / 1000),
-      templateId: emailTemplateId,
+      templateId: digestFeature.templateId,
```
capJavert, line 287: Potentially fallback to `features.personalizedDigest.defaultValue.templateId` in case of partial object in default. Don't think growthbook deep merges default values.

src/workers/personalizedDigestEmail.ts
```diff
@@ -166,6 +201,18 @@ const worker: Worker = {
       return;
     }
 
+    const growthbookClient = getUserGrowthBookInstace(user.id, {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      trackingCallback: (experiment, result) => {
+        // TODO notify allocation
+      },
```
capJavert, line 208: I have to test locally that callback is triggered. Need to setup some dev experiment.

src/workers/personalizedDigestEmail.ts
```diff
@@ -91,45 +96,75 @@ const getPostsTemplateData = ({ posts }: { posts: TemplatePostData[] }) => {
...
     day_name: dayName,
-    first_name: firstName,
+    first_name: userName,
     posts: getPostsTemplateData({ posts }),
+    date: format(currentDate, 'MMM d, yyyy'),
+    user: {
+      username: user.username,
+      image: user.image,
+      reputation: user.reputation,
+    },
```
capJavert, line 136: Using nested object for `user` entity properties. We can probably also apply it to `post` to avoid whole `post_` schema but for backward compatibility let's leave that part as is for now.

capJavert:
Open for some early review before e2e testing everything with GB and then proceeding to making a retool app for product.

src/workers/personalizedDigestEmail.ts
idoshamun: I'd suggest putting the tracking callback in `getUserGrowthBookInstace` to make it fault tolerant for future uses as well

src/workers/personalizedDigestEmail.ts
idoshamun: Good point about deep merge. Double check that. You can connect to the dev env of GB

src/workers/personalizedDigestEmail.ts
```diff
@@ -91,45 +96,75 @@ const getPostsTemplateData = ({ posts }: { posts: TemplatePostData[] }) => {
...
     dynamicTemplateData: {
       ...data,
-      title: `${firstName}, your personal weekly update from daily.dev is ready`,
-      preview: defaultPreview,
+      title: `${userName}, your personal weekly update from daily.dev is ready`,
+      preview: `Every ${dayName}, we'll send you five posts you haven't read. Each post was carefully picked based on topics you love reading about. Let's get to it!`,
     },
   };
+
+  if (personalizedDigest.variation === 2) {
```
idoshamun, line 155: If this variation lost, we can delete it

src/workers/personalizedDigestEmail.ts
```diff
@@ -91,45 +96,75 @@ const getPostsTemplateData = ({ posts }: { posts: TemplatePostData[] }) => {
...
+    to: {
+      email: user.email,
+      name: userName,
+    },
     dynamicTemplateData: {
       ...data,
-      title: `${firstName}, your personal weekly update from daily.dev is ready`,
-      preview: defaultPreview,
+      title: `${userName}, your personal weekly update from daily.dev is ready`,
+      preview: `Every ${dayName}, we'll send you five posts you haven't read. Each post was carefully picked based on topics you love reading about. Let's get to it!`,
```
idoshamun, line 151: Didn't we want to give them the ability to experiment on that?

capJavert: Yes. Keeping this property for backward compatibility. 

They can experiment by adjusting and using all other variables in templates directly. Both email subject and preview (which is a div inside template body) support variables. This [thread](https://dailydotdev.slack.com/archives/C05G8BHHVBR/p1708006426646369)

src/workers/personalizedDigestEmail.ts
capJavert: Yep, I imagine variation will not be needed anymore now that we have GB feature hooked up.

src/workers/personalizedDigestEmail.ts
capJavert: It does not, I utilized `deepmerge`, we already had it as peer dependency so just added it as depndency now.

src/workers/personalizedDigestEmail.ts
capJavert: Removed.

src/workers/personalizedDigestEmail.ts
```diff
@@ -166,6 +192,55 @@ const worker: Worker = {
...
+        } & AnalyticsEvent,
+      ) => {
+        await sendAnalyticsEvent([data]);
+      },
+      1,
+    );
+
+    const growthbookClient = getUserGrowthBookInstace(user.id, {
+      enableDevMode: process.env.NODE_ENV !== 'production',
+      subscribeToChanges: false,
```
capJavert, line 211: Recommended for request/expiring runtimes like worker/request by [gb docs](https://docs.growthbook.io/lib/js#:~:text=Set%20this%20to%20%60false%60%20to%20improve%20performance%20in%20server%2Dside%20environments).

src/workers/personalizedDigestEmail.ts
```diff
@@ -166,6 +192,55 @@ const worker: Worker = {
...
+    });
+
+    try {
+      await growthbookClient.loadFeatures();
+    } catch (error) {
+      logger.error(
+        { personalizedDigest, err: error },
+        'Failed to load features',
+      );
+    }
```
capJavert, line 231: Loading features manually, don't think there is a better option since each worker has a new context. Logging for now just so we can see what is happening in first few tries. 

Basically if this fails user would fallback to default value eg. control.

src/workers/personalizedDigestEmail.ts
```diff
@@ -166,6 +192,55 @@ const worker: Worker = {
...
+      enableDevMode: process.env.NODE_ENV !== 'production',
+      subscribeToChanges: false,
+      trackingCallback: async (experiment, result) => {
+        analyticsQueue.push({
+          event_name: 'allocation',
+          event_timestamp: new Date(),
+          user_id: user.id,
+          device_id: 'api',
+          experiment_id: experiment.key,
+          variation_id: result.variationId,
```
capJavert, line 219: @idoshamun not sure what `device_id` and `event_name` should be. We don't send it from frontend. Let me know.

src/workers/personalizedDigestEmail.ts
rebelchris: Hmm should we not still also make this dynamic here?
Looking at upcoming multiple sending days it might be easier for us to support then in sendgrid?

src/growthbook.ts
```diff
@@ -53,11 +55,44 @@ export function getEncryptedFeatures(): string {
...
+
+  readonly defaultValue?: T;
+
+  constructor(id: string, defaultValue?: T) {
+    this.id = id;
+    this.defaultValue = defaultValue;
+  }
+}
+
+export const features = {
```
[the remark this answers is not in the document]
capJavert: Basically it is the same interface we have in `apps`, it is not just default.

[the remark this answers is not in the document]
rebelchris: Hmm I see, let's keep it for now, one might just be confused as to what these values are.
It's not blocking.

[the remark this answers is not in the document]
capJavert: Yeah, my idea is that basically as we start to use different features in the API side as well we can have them enumerated here so they can be used for different checks.

__tests__/workers/personalizedDigestEmail.ts
```diff
@@ -9,13 +9,50 @@ import { sourcesFixture } from '../fixture/source';
...
 
+jest.mock('../../src/integrations/analytics', () => ({
+  ...(jest.requireActual('../../src/integrations/analytics') as Record<
+    string,
+    unknown
+  >),
+  sendAnalyticsEvent: jest.fn(),
+}));
+
+getUserGrowthBookInstace;
```
idoshamun, line 29: What's this?

__tests__/workers/personalizedDigestEmail.ts
```diff
@@ -9,13 +9,50 @@ import { sourcesFixture } from '../fixture/source';
...
+
+jest.mock('../../src/growthbook', () => ({
```
idoshamun, line 31: Nice! Will become handy in the future

src/workers/personalizedDigestEmail.ts
```diff
@@ -166,6 +192,55 @@ const worker: Worker = {
...
+      ) => {
+        await sendAnalyticsEvent([data]);
+      },
+      1,
+    );
+
+    const growthbookClient = getUserGrowthBookInstace(user.id, {
+      enableDevMode: process.env.NODE_ENV !== 'production',
+      subscribeToChanges: false,
+      trackingCallback: async (experiment, result) => {
```
idoshamun, line 212: * I think we should have the `trackingCallback` support in `getUserGrowthBookInstace`. This will be very useful for future development. For your consideration though
* We don't need the queue here imo, we just send the request because each user is a single request. Again for your consideration
* The request you send here is actually the wrong one. We don't need analytics event, we need experiment allocation. It's the request that goes to `/e/x`. It doesn't require `event_name`. For device_id, we better just generate uuid

src/workers/personalizedDigestEmail.ts
idoshamun: We can load it in the entry point of the background worker. This way no one else will have to think about it

__tests__/workers/personalizedDigestEmail.ts
capJavert: No clue, most have been autocomplete artefact.

src/workers/personalizedDigestEmail.ts
capJavert: As for analytics event endpoint/payload I updated in 7d90a0a93331df435d77cf7a9b301180c5fbcbd1

src/integrations/analytics.ts
```diff
@@ -39,3 +39,31 @@ export async function sendAnalyticsEvent<T extends AnalyticsEvent>(
...
 }
+
+export type ExperimentAnalyticsEvent = {
+  event_timestamp: Date;
+  user_id: string;
+  experiment_id: string;
+  variation_id: number;
+};
+
+export async function sendExperimentAnalyticsEvent<
```
idoshamun, line 50: instead of experiment analytics, can we use the term experiment allocation please?

src/integrations/analytics.ts
```diff
@@ -39,3 +39,31 @@ export async function sendAnalyticsEvent<T extends AnalyticsEvent>(
...
+  T extends ExperimentAnalyticsEvent,
+>(event: T): Promise<void> {
+  const device_id = generateUUID();
+  const transformed = {
+    device_id,
```
idoshamun, line 55: I saw that device id is optional in case we want to remove it. sorry for not checking earlier

idoshamun (approved):
Looks great some synthetic comments, not blocking

src/workers/personalizedDigestEmail.ts
capJavert: Nice, like this? e6f1c814768021831e091610f69369bab2c61ac0

src/integrations/analytics.ts
capJavert: Is it only allocation, thinking it might need something else down the line so just wanted to be abstract?

idoshamun: The official term we use everywhere (including the table name in BigQuery) is experiment allocation. It's a single request with the experiment information that is stored in BQ and then we know how to split the user base

src/integrations/analytics.ts
capJavert: Ok, removing then.

src/workers/personalizedDigestEmail.ts
capJavert: I am not sure, would prefer if we use the Sendgrid since it supports it. I will talk to product to see how they feel about usage and then we can adjust if it slows them down by much. WDYT?

rebelchris: Sounds like a plan :)

rebelchris (approved):

PR #1704, opened 2024-02-16
feat: streak milestone alert computation

kopancek:
- Adds new `column` `showStreakMilestone` boolean in `alerts` table.
- Sets value in `newView` worker to `true` if currentStreak is a number from the fibonacci sequence.
- Fixed an issue where graphQL always returned `currentStreak` as 0, because we missed an `await` keyword on a promise.
- Added unit tests, so pretty sure this works. Maybe 🤞

### DIFF
--- a/src/common/fibonacci.ts
+++ b/src/common/fibonacci.ts
@@ -1,0 +1,9 @@
+export const isPerfectSquare = (n: number) => {
+  return Number.isInteger(Math.sqrt(n));
+};
+
+// Number is Fibonacci if (5*n^2 + 4) or (5*n^2 - 4) is a perfect square
+export const isFibonacci = (num: number) => {
+  const x = 5 * Math.pow(num, 2);
+  return isPerfectSquare(x + 4) || isPerfectSquare(x - 4);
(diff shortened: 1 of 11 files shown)


src/common/fibonacci.ts
```diff
@@ -0,0 +1,9 @@
+export const isPerfectSquare = (n: number) => {
+  return Number.isInteger(Math.sqrt(n));
+};
+
+// Number is Fibonacci if (5*n^2 + 4) or (5*n^2 - 4) is a perfect square
```
idoshamun, line 5: Nice!! Didn't know that

idoshamun (approved):
Looks good!!!

sshanzel (approved):
Looks great! 🥳

src/common/fibonacci.ts
capJavert: TIL

kopancek: I also learned it when googling how to test if a number belongs to the fibonacci sequence :)

PR #1705, opened 2024-02-19
fix: user constraint fail fix

rebelchris:
There was a edge-case where the user constraint would fail if the user got deleted before subscription finishes.
We said to only listen to this specific constraint for now.


idoshamun (approved):
awesome!

PR #1706, opened 2024-02-19
refactor: search low hanging improvements

idoshamun:
* Add index (can't believe we didn't have it before 😅)
* Change ordering to be based on the search rank and then created at to make it more predictable
* Change how we build the query to make only the last word partial search

Hopefully it will be better now

### DIFF
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -182,6 +182,7 @@ export class Post {
   deleted: boolean;
 
   @Column({ nullable: true, type: 'tsvector', select: false })
+  @Index('IDX_post_tsv', { synchronize: false })
   tsv: unknown;
 
   @Column({ default: false })
--- a/src/schema/common.ts
+++ b/src/schema/common.ts
@@ -142,11 +142,11 @@ export interface PageGenerator<
   ) => TReturn[];
 }
 
-export const getSearchQuery = (
-  param: string,
-): string => `SELECT to_tsquery('english',
-                                string_agg(lexeme || ':*', ' & ' order by positions)) AS query
-              FROM unnest(to_tsvector('english', process_text(${param})))`;
+export const getSearchQuery = (param: string): string =>
+  `SELECT to_tsquery('english', ${param}) AS query`;
+
+export const processSearchQuery = (query: string): string =>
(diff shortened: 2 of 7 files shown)


src/entity/posts/Post.ts
```diff
@@ -182,6 +182,7 @@ export class Post {
   deleted: boolean;
 
   @Column({ nullable: true, type: 'tsvector', select: false })
+  @Index('IDX_post_tsv', { synchronize: false })
```
rebelchris, line 185: Weird dont recall this popping up in eversql so far 👀

rebelchris: Unless its one i just removed? Cant recall 😅

src/schema/common.ts
```diff
@@ -142,11 +142,11 @@ export interface PageGenerator<
   ) => TReturn[];
 }
 
-export const getSearchQuery = (
-  param: string,
-): string => `SELECT to_tsquery('english',
-                                string_agg(lexeme || ':*', ' & ' order by positions)) AS query
-              FROM unnest(to_tsvector('english', process_text(${param})))`;
+export const getSearchQuery = (param: string): string =>
+  `SELECT to_tsquery('english', ${param}) AS query`;
```
rebelchris, line 146: Did you check speed and hits on partials with this?
Ref: https://dailydotdev.slack.com/archives/C03C3JZPRU4/p1705389107962379?thread_ts=1705387675.301509&channel=C03C3JZPRU4&message_ts=1705389107.962379

rebelchris (commented):
Not against this, but please check my slack messages on this some time ago 👍

src/schema/common.ts
idoshamun: the production implementation does partials on all words, now i changed it to do only for the last word.
haven't checked speed because in your message you refer to `like`, here i use the same mechanism just changed a bit the query. I did check hits though.
I'm not 100% sure hence i want to try in production and worst case i'll revert

rebelchris: I tried 3 methods there. Anyway happy to try, you asked me there if i recall to not go as it would slow down, so just wanted to check 🙏

rebelchris (approved):

PR #1710, opened 2024-02-22
revert: post notification muted for public squads

sshanzel:
As mentioned in the Slack conversation, public Squads are more like sources now, which means it is less noisy than it was before.

With that, we are bringing back the default to notify members when a new post is added.

MI-148 #done


rebelchris (approved):

rebelchris:
Just sad for this feature, not the code 🤣

capJavert (approved):

idoshamun (approved):

PR #1711, opened 2024-02-23
feat: digest email schedule endpoint

capJavert:
AS-122, will be used in retool app (AS-123) to trigger emails on demand

To support it I needed to refactor the worker a bit:
- cron now calculates and passes the dates (instead of giving timestamp then is that calculated to date in the worker)
  - this allows us schedule emails from different parts of the system without being tied to recurring logic of digest
- added deduplicate param so we can send emails multiple times aside from digest schedule
- moved some stuff to utils to make it more readable
- also made batch id optional for worker (it is still guarded in cron to avoid send if batch id fails to create for any reason)


src/routes/index.ts
```diff
@@ -61,4 +68,58 @@ Disallow: /`);
     req.log.debug({ body: req.body }, 'allocation received');
     return res.status(204).send();
   });
+
+  fastify.post<{ Body: { userIds: string[] } }>(
+    '/digest/send',
+    async (req, res) => {
+      // TODO AS-122-email-schedule-endpoint authenticate here
+
```
capJavert, line 76: Not sure if we already authenticate somehow from retool? I know we have automations part but that is more on exposing automation stuff from API, this would be pinging API endpoint from retool. As mentioned in RFC I though of just adding a secret that we can send from retool?

capJavert: As addition we can also limit by IP ranges https://docs.retool.com/data-sources/concepts/ip-allowlist-cloud-orgs, again not sure if we connected something already from retool.

rebelchris (approved):
Don't think we have any checks on retool now, the only other endpoints is Ido's

I think maybe we secret for keepsake is good for now.

src/routes/index.ts
capJavert: I added bearer auth here 6ec050347f5759a4a60116d63999fd31ee80f4df

rebelchris (approved):
cool!

idoshamun:
@capJavert @rebelchris fyi for next time, we can utilize this agent to call internal services for a more secured approach
https://github.com/dailydotdev/retool-agent

capJavert:
@idoshamun nice 👌 I think refactoring to this will be pretty easy as well. With bearer we just wanted to have something.

PR #1712, opened 2024-02-23
fix: add post downvote rep events

rebelchris:
Added reputation events for post downvoting.
We realized we forgot cancelling/removing downvotes, but handled it directly here.
Also applied the slightly different rule of only granting this neg rep to author if no scout is set.

We had a debate if we should create one generic postVote topic, but decided to leave it as 4 separate ones to maintain flexibility/clarity.

AS-133 #done


src/workers/postDownvoteCanceledRep.ts
```diff
@@ -0,0 +1,61 @@
...
+  userId: string;
+  postId: string;
+}
+
+const worker: Worker = {
+  subscription: 'api.post-downvote-canceled-rep',
+  handler: async (message, con, logger): Promise<void> => {
+    const data: Data = messageToJson(message);
+    try {
+      const post = await con.getRepository(Post).findOneBy({ id: data.postId });
```
capJavert, line 20: Check that post exists, edge case but could happen with squad posts for example?

capJavert: I know you are checking below with `?.` but feels better to stop the workflow early if post can not be found.

src/workers/postDownvotedRep.ts
```diff
@@ -0,0 +1,94 @@
...
+              grantToId: post.scoutId,
+              ...props,
+            }),
+          );
+        }
+
+        /**
+         * Business logic states that we only decrease reputation of the author if there is no scout associated with the post
+         */
+        if (post.authorId && post.authorId !== data.userId && !events.length) {
```
capJavert, line 64: Maybe explicitly check scout instead of `events.length` since in the future some more events could be added before this logic.

src/workers/cdc/primary.ts
```diff
@@ -205,6 +197,78 @@ const onPostUpvoteChange = async (
...
+  switch (data.payload.op) {
+    case 'c':
+      await notifyPostDownvoted(
+        logger,
+        data.payload.after.postId,
+        data.payload.after.userId,
+      );
+      break;
+    case 'u': {
+      const isUpvoteCanceled = data.payload.after.vote === UserPostVote.None;
```
capJavert, line 214: isDownvoteCancelled?

capJavert (commented):
Awesome 🚀 Tnx for taking care of the cancelation as well, just a few ideas for logic but ready to go. Let's wait with merging to check with product if they want to have this active before frontend part.

src/workers/index.ts
```diff
@@ -64,6 +66,8 @@ export const workers: Worker[] = [
   postCommentedSlackMessage,
   postUpvotedRep,
   postUpvoteCanceledRep,
+  postDownvotedRep,
+  postDownvoteCanceledRep,
```
capJavert, line 70: And maybe also add `should be registered` into tests so we make sure they are exported, as in `collectionUpdated` worker for example.

rebelchris:
@capJavert Great review, thanks adjusted all feedback :)
Good point on the worker test, we should do that by default indeed.

rebelchris:
@capJavert This one is good to merge or are we waiting for product to give go?

PR #1714, opened 2024-02-26
feat: add language column to posts

omBratteng:


__tests__/workers/postUpdated.ts
```diff
@@ -422,6 +422,34 @@ it('should save a new post with basic information', async () => {
...
+    id: 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+    title: 'Title',
+    url: 'https://post.com',
+    source_id: 'a',
+    order: 0,
+    language: 'nb',
+  });
+  const posts = await con.getRepository(Post).find();
+  expect(posts.length).toEqual(3);
+  expect(posts[2]).toMatchSnapshot({
```
rebelchris, line 440: If we can lets remove snapshot and just make it inline evaluate.
We slowly want to get rid of them.

src/workers/postUpdated.ts
```diff
@@ -506,6 +507,7 @@ const fixData = async ({
         ? markdown.render(data.extra.content)
         : undefined,
       videoId: data?.extra?.video_id,
+      language: data?.language,
```
rebelchris, line 510: Just check with @denisb0 if its under direct or extra prop?

rebelchris (commented):
Minor comment on snapshot if possible and double check on attr.

src/workers/postUpdated.ts
omBratteng: Direct. Already checked :)

__tests__/workers/postUpdated.ts
omBratteng: Removed snapshot

rebelchris (approved):
Nice! Thanks 🥳

__tests__/workers/postUpdated.ts
```diff
@@ -422,6 +422,26 @@ it('should save a new post with basic information', async () => {
...
+    language: 'nb',
+  });
+  const posts = await con.getRepository(Post).find();
+  expect(posts.length).toEqual(3);
+  expect(posts[2]).toMatchObject({
+    sourceId: 'a',
+    title: 'Title',
+    showOnFeed: true,
+    language: 'nb',
   });
```
capJavert, line 445: Might be flaky without sort.

capJavert (approved):

__tests__/workers/postUpdated.ts
omBratteng: I don't think `toMatchObject` does care about the order, as it only matches partials

omBratteng: Took [the test from here](https://jestjs.io/docs/expect#tomatchobjectobject) and moved  the key `kitchen` in `desiredHouse`, and the test succeeded.

capJavert: Not `toMatchObject` but `posts[2]` might not always be the same row since order is not guaranteed.

omBratteng: Ah, yeah, that one 😬

PR #1715, opened 2024-02-27
refactor: user entity to add acquisition channel column

sshanzel:
Additional column to track the acquisition channel.

MI-176 #done

### DIFF
--- a/src/entity/user/User.ts
+++ b/src/entity/user/User.ts
@@ -122,6 +122,9 @@ export class User {
   @Column({ type: 'text', nullable: true })
   readmeHtml?: string;
 
+  @Column({ type: 'text', nullable: true })
+  acquisitionChannel: string;
+
   @ManyToOne(() => User, {
     lazy: true,
     onDelete: 'SET NULL',
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -2681,3 +2681,43 @@ describe('user_create_alerts_trigger after insert trigger', () => {
     expect(alerts).toBeTruthy();
   });
 });
+
+describe('addUserAcquisitionChannel mutation', () => {
+  const MUTATION = `
+    mutation AddUserAcquisitionChannel($acquisitionChannel: AcquisitionChannel!) {
+      addUserAcquisitionChannel(acquisitionChannel: $acquisitionChannel) {
+        _
+      }
+    }
+  `;
+
+  it('should not allow unauthenticated users', () =>
+    testMutationErrorCode(
+      client,
+      { mutation: MUTATION, variables: { acquisitionChannel: 'friend' } },
+      'UNAUTHENTICATED',
+    ));
+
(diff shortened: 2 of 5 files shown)


rebelchris (approved):

src/entity/user/User.ts
```diff
@@ -122,6 +122,9 @@ export class User {
   @Column({ type: 'text', nullable: true })
   readmeHtml?: string;
 
+  @Column({ type: 'text', nullable: true })
+  acquisitionChannel: string;
```
capJavert, line 126: From RFC this should accept enum values eg. `AcquisitionChannel` from https://github.com/dailydotdev/daily-api/pull/1716

capJavert, line 126: From RFC this should accept enum values eg. `AcquisitionChannel` from https://github.com/dailydotdev/daily-api/pull/1716?

rebelchris: It does in the mutation, we try to keep column text afaik for future additions/changes.
(correct me if i'm wrong here)

capJavert: It will still be text in SQL but its good to set as enum for synthetic sugar in TS

capJavert: eg. `UserAction`

sshanzel: The JS enum itself is not used anywhere, I thought it might just litter codebase.

sshanzel: Yeah, I thought of that too, initially. But I was pushing the simplest thing as possible. As far as I understood, the value of code cleanliness is not the top priority anymore. We should probably draw a line on how much time we want to spend on like you said, a syntactic sugar similar to this.

capJavert: Does adding an enum here complicate your implementation in any way? It should be just one line update setting a type to entity property we already.

sshanzel: Not really, I'm just saying, I just pushed the least code necessary that is acceptable for the requirements. So I didn't bother having an additional code footprint.

kopancek: What is the value of that enum though? The schema already validates the possible values. This would be just another code to change in case we add more values to the schema later on. I would prefer keeping it as a string. We already follow that pattern elsewhere in the code.

kopancek (approved):

src/entity/user/User.ts
nensidosari: Screenshot 2024-02-27 at 10 05 43 AM

Just to link this as it was already [discussed](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/797376528?focusedCommentId=798720013#comment-798720013) in the RFC

capJavert: @sshanzel and I just synced. Good to go 🚀

capJavert (approved):

PR #1716, opened 2024-02-27
feat: acquisition mutation

sshanzel:
Update user information about acquisition channel.

MI-177 #done

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -2684,8 +2684,8 @@ describe('user_create_alerts_trigger after insert trigger', () => {
 
 describe('addUserAcquisitionChannel mutation', () => {
   const MUTATION = `
-    mutation AddUserAcquisitionChannel($source: AcquisitionChannel!) {
-      addUserAcquisitionChannel(source: $source) {
+    mutation AddUserAcquisitionChannel($acquisitionChannel: AcquisitionChannel!) {
+      addUserAcquisitionChannel(acquisitionChannel: $acquisitionChannel) {
         _
       }
     }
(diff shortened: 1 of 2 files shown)


rebelchris (approved):

src/schema/users.ts
```diff
@@ -413,6 +413,16 @@ export const typeDefs = /* GraphQL */ `
     lastViewAt: DateTime
   }
 
+  enum AcquisitionChannel {
```
nensidosari, line 416: There's also form_dismissed mentioned in RFC, do we want to add here?

__tests__/users.ts
```diff
@@ -2681,3 +2681,41 @@ describe('user_create_alerts_trigger after insert trigger', () => {
...
+  `;
+
+  it('should not allow unauthenticated users', () =>
+    testMutationErrorCode(
+      client,
+      { mutation: MUTATION, variables: { source: 'friend' } },
+      'UNAUTHENTICATED',
+    ));
+
+  it('should not throw an error when value is not found', () =>
```
nensidosari, line 2701: Should this be `should throw` not `should not throw` ?

src/schema/users.ts
```diff
@@ -1256,6 +1271,17 @@ export const resolvers: IResolvers<any, Context> = {
         );
       return getCurrentUser(ctx, info);
     },
+    addUserAcquisitionChannel: async (
+      _,
+      { source }: { source: string },
+      ctx,
+    ): Promise<GQLEmptyResponse> => {
+      await ctx.con
+        .getRepository(User)
```
capJavert, line 1280: I would validate that `source` here is one of keys of `AcquisitionChannel`. We don't want any garbage there right?

capJavert: Also maybe name the argument `acquisitionChannel` because `source` means something else (eg. our entity) in different mutations.

capJavert (commented):
Minor comments, looks good

__tests__/users.ts
```diff
@@ -2681,3 +2681,41 @@ describe('user_create_alerts_trigger after insert trigger', () => {
...
+    testMutationErrorCode(
+      client,
+      { mutation: MUTATION, variables: { source: 'random' } },
+      'UNEXPECTED',
+    ));
+
+  it('should not throw an error when value is not found', async () => {
```
nensidosari, line 2708: Should this be when `value is found` ?

src/schema/users.ts
capJavert: Ah, source is validate through graphql schema, so just renaming potentially.

src/schema/users.ts
sshanzel: I think that was actually a mistake. I am not sure we want to record that, as it won't help. Thoughts? cc @dailydotdev/web-team

src/schema/users.ts
sshanzel: Yes. I just used `source` as it was defined in the RFC GraphQL schema as that.

sshanzel: This actually makes me think whether we can just utilize the updateUserProfile, but I went and did what RFC have written.

capJavert: Yeah I think it can still be changed, just so it does not look as if it should be our source (as our entity) as this is something else.

capJavert (approved):

nensidosari (approved):

src/schema/users.ts
sshanzel: Good point. Let's probably check with the product/analytics team whether they want to monitor the amount of users that ignored and dismissed. Wdyt?

nensidosari: Has product requested this? Let's confirm with them better.

kopancek: Based on latest conversation, we will rely on tracking event for the analytics, so no need to store it in the DB.

src/schema/users.ts
kopancek: We could use the `updateUserProfile`, but IMO this value has nothing to do with user profile, hence why I suggested a separate mutation.

PR #1719, opened 2024-02-27
fix: limit the jest idle worker memory

omBratteng:


jest.config.js
```diff
@@ -13,4 +13,5 @@ module.exports = {
     escapeString: true,
     printBasicPrototype: true,
   },
+  workerIdleMemoryLimit: '512MB',
```
capJavert, line 16: We can try with 1GB, or 2GB as well to see if it improves the time but still makes the memory in check? Since we have 8GB instance.

omBratteng: Set it to 1GiB just to check, plus added heap usage logging just for testing

rebelchris (approved):
Fine to try

omBratteng:
Looks like 2GiB works fine. Ran 3 tests on both.
image

capJavert (approved):
Let's go!

PR #1721, opened 2024-02-27
feat: changing the fallback images

nensidosari:
Changing the fallback images as discussed in thread: https://dailydotdev.slack.com/archives/C013QLM6GAD/p1708515326688139?thread_ts=1704905223.059279&cid=C013QLM6GAD

There are now 12 images instead of 8


.infra/Pulumi.prod.yaml
```diff
@@ -38,7 +38,7 @@ config:
     defaultImageRatio:
       secure: AAABAHpff5rGIhX2EmH5QBZx1br4VVzK+JHKbbAUjYRT
     defaultImageUrl:
```
nensidosari, line 40: Should I also change these values in adhoc file?

omBratteng: Yeah, we want new pretty images in adhoc too

nensidosari: Done

.infra/Pulumi.prod.yaml
```diff
@@ -38,7 +38,7 @@ config:
...
...P5R5ueT1tbstTeKTaNaIKwQNmucP5ikB+X5b94MSvQlheX7AgahQn1GuOooQJkN1gwYrPBRvdhtJwvkYyNwWhKf+lQCU0Tnfh92EfrZWsM9lOHdna0zeww0YEglJgbS1A904ciM85Uud+mgZRklrTE5BJCCXcj7i+4UUgjXygbAJqU8BIkUkaTDPydUOuoUgE7nu997vKM8t/0iNx9KlR7g9EQdWyv2QOFLO1bvBy3TIy5JRQJ4YkTOJD7RxbFU/O4K2rBJQLhF8xcSyO5ONuKS2S+xARu4XhMxaxqxZZ8ybzCmKZlKREFWDUvadtZvwDBMgoaayuwVECPfH3koEJYbYVyu1Xm1EozTX3K7RADVTjR1bMG1SGPNjWy9zeR3phPK+KK1AjmaVcWCUKqjo5XSgTks9EsNUHKBS52GMpsSQ0ff6d5GBWBv6c5zSYuPSamgNsQyy4Lqta3rvZNQgqY2n45X/G0M0aZq0SY/28bdcY4nTO4kg75L843kzIomTlz2f6lsA1lF2ZIJyTDjRvEp8w0WUTtSmdOR4oetxBDaBY53XqseMq5PvT7Vu8fQBGVG5IshOodGwmoE3WLG9nmLdnjF6iIwOtGj1n13F38NWDUI1Y707N4ohvM67vvhbmi14sEZsQNPVNRkzRRAw84PMv6mCsZ1MtYC0ApYnmG7e1SmTO+ihu1tzILjusX6W3p+6UIAipiPgSjrR5RgvCCYCi5nzHmO0h362YafpwJ35X00ZAwEur1Pq17ioSlX0C4015mSx5iheXh57aqDoxfs2gzlbTVywJSiU44rcVI3tg5ApqDz2RwTksn2FvtYNYlf3X9va+hCAkVeJjnNRSjrYNpBHYM4kdu9LBoxBNb5mdZSerDXtLw5ZJJHLj+INb6hVhp7T0X0plyeZkeUoxWYOzOil5RJaVB1pXPOlzJQnHJxBXjQ0/0ic8Qh8CF3xkTjHGWRIMlRVNig3aRAogDSPbXXhSv5skWUNIHsLa0CN3JFkGSr5oRIr3Q8n9IgSLgoGc8x4uDhvoQJ68Kczq1wfc4K19PdXAhfo34492g+uvw5qqZBbtwTG1Ul5pEwGL6M9qJgZfdUVfBtMkVB8ml/rVJR6wUMSd8vpf/yN1835gmno3EcAk1TGy3j2kQDphAwbfIRobpYC0Pc9VswNnHI0Qg2wZF7a/TturnCeNI005A==
```
omBratteng, line 41: Doubt that this value needs to be stored as secure, especially since the same value is in adhoc 😬

Same goes with other non-secret "secrets" really. Thoughts @capJavert @idoshamun?

rebelchris: Probably not, but wont matter much right?

rebelchris (approved):
Can you check if we did any fallback on apps as well? (Can't recall)

nensidosari:
> Can you check if we did any fallback on apps as well? (Can't recall)

Yup we do add just the first link there, can add the first link of the new ones.

.infra/Pulumi.prod.yaml
nensidosari: Nope, doesn't really matter. If we decide we want to change these we can change anytime.

omBratteng:
> Can you check if we did any fallback on apps as well? (Can't recall)

Yeah, there's a fallback that is used if we fail to load the image (onerror)
https://github.com/dailydotdev/apps/blob/8d161b4e6ef5bfe657e4c931263d5ef26d630d51/packages/shared/src/lib/image.ts#L14-L17

.infra/Pulumi.prod.yaml
idoshamun: Nice to have not a blocker :)

nensidosari:
Change in apps here: https://github.com/dailydotdev/apps/pull/2693

omBratteng (approved):

PR #1723, opened 2024-02-28
fix: map pino log levels to Google Cloud Logging severity levels

omBratteng:
I noticed that logs from api doesn't really have any severity attached to it in Logs Explorer. Found out there's a [recommended mapping by Pino](https://getpino.io/#/docs/help?id=mapping-pino-log-levels-to-google-cloud-logging-stackdriver-severity-levels) to map level to the severity field used by gcloud. We have already done that mapping from `msg` to `message`.

image

By default, pino only logs to a numbered level (which also doesn't match what Google is used). So added a `severity` field to the outputted json log.

Also went ahed and used a singular logger instance, instead of copying the same config X amount of places.


capJavert (approved):
👌

PR #1724, opened 2024-02-28
feat: add graphql query for URL shortener

kopancek:
Adds a graphql query so that we can request shortened URLs from the frontend.

Related: 
- MI-188

### DIFF
--- a/__tests__/urlShortener.ts
+++ b/__tests__/urlShortener.ts
@@ -1,0 +1,76 @@
+import nock from 'nock';
+import { DataSource } from 'typeorm';
+import { v4 as uuidv4 } from 'uuid';
+
+import createOrGetConnection from '../src/db';
+import {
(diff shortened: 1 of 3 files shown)


sshanzel (approved):
Looks great! This unblocks the work for copy link in FE 🚀 

I was thinking the `urlShortener` having its own schema feels a bit overkill? Maybe we can introduce a more general schema that will house this query and everything similar in the future. Probably something along the lines of `integration` schema? Since this involves integration with another service that we have and for sure we will do more integrations in the future.

sshanzel:
> For now I'm OK with where it is. It's small and self contained. If in the future we add more utility functions like this, we can for sure rename it to something more generic.

This is my concern, in the future, the one introducing a new utility function won't know about this history, that we should rename the existing one and might introduce a new one.

PR #1727, opened 2024-02-29
fix: return 0 streak when there is no record in the database

omBratteng:
In the case the user does not have an entry in the database for their user streak, we just return an empty one.

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -846,34 +846,22 @@ export const resolvers: IResolvers<any, Context> = {
         .getRawMany();
     },
     userStreak: async (_, __, ctx: Context, info): Promise<GQLUserStreak> => {
-      try {
-        const streak = await graphorm.queryOneOrFail<GQLUserStreakTz>(
-          ctx,
-          info,
-          (builder) => ({
-            queryBuilder: builder.queryBuilder
-              .addSelect(
-                `(date_trunc('day', "${builder.alias}"."lastViewAt" at time zone COALESCE(u.timezone, 'utc'))::date) AS "lastViewAtTz"`,
-              )
-              .addSelect('u.timezone', 'timezone')
-              .innerJoin(User, 'u', `"${builder.alias}"."userId" = u.id`)
-              .where(`"${builder.alias}"."userId" = :id`, { id: ctx.userId }),
-            ...builder,
-          }),
-        );
+      const streak = await graphorm.query<GQLUserStreakTz>(
+        ctx,
+        info,
+        (builder) => ({
+          queryBuilder: builder.queryBuilder
+            .addSelect(
+              `(date_trunc('day', "${builder.alias}"."lastViewAt" at time zone COALESCE(u.timezone, 'utc'))::date) AS "lastViewAtTz"`,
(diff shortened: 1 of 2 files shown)


rebelchris:
Didn't we add these with pg function? (not sure maybe we decided otherwise)

omBratteng:
We have `93468` entries in `user_streak` and `404491` entries in `user`. So not everyone has an entry for user streaks.
In that case the query would return an error, and in frontend it would keep retrying and not render anything in the header.

omBratteng:
So this would return "empty" streak for those users who have no entries already

src/schema/users.ts
```diff
@@ -846,27 +846,42 @@ export const resolvers: IResolvers<any, Context> = {
...
+        return streak;
+      } catch (err) {
+        ctx.log.error('Error fetching user streak', err);
+        return {
+          max: 0,
+          total: 0,
+          current: 0,
+          lastViewAt: null,
+          userId: ctx.userId,
+        };
```
capJavert, line 883: I would not do it with catch because some legit error would then also return empty streak. I would rather early return if no streak in DB and return this empty stub.

omBratteng: Removed the try...catch and just used same logic as inside the queryOneOrFail except without throwing the error.

src/schema/users.ts
```diff
@@ -846,7 +846,7 @@ export const resolvers: IResolvers<any, Context> = {
         .getRawMany();
     },
     userStreak: async (_, __, ctx: Context, info): Promise<GQLUserStreak> => {
-      const streak = await graphorm.queryOneOrFail<GQLUserStreakTz>(
+      const streak = await graphorm.query<GQLUserStreakTz>(
```
rebelchris, line 849: Can still use queryOne right? Withouth the fail part

rebelchris (approved):
Non blocking comment

src/schema/users.ts
omBratteng: Don't think there is a queryOne, at least my autocomplete didn't suggest it.

nensidosari (approved):

DragosIorgulescu:
> Didn't we add these with pg function? (not sure maybe we decided otherwise)

@rebelchris we added a function which creates the streak record on user create, but we didn't retroactively generate them for existing users

DragosIorgulescu (approved):

src/schema/users.ts
capJavert: We can add one, its part of graphorm, but this is fine as well yes.

capJavert: We can add one in graphorm, it would be the same as `queryOneOrFail` just without `not found` throw. Non blocking yes.

capJavert (approved):

src/schema/users.ts
omBratteng: I'll open a new PR for that. Do you think we should return null when nothing is found? Because `res[0]` itself might cause an issue if nothing is found.

omBratteng: Opened #1728

PR #1729, opened 2024-03-01
fix: don't trigger alerts multiple times on the same milestone / day

DragosIorgulescu:
- wrapping the reading streak increment in a db transaction
- update the conditions for which alerts are generated
  - regular milestone alerts now take into account if the last view at on the user streak is different than the current day
  - max streak milestones check both if this is a new max streak and that the last view at column is different than the current day

_this approach only works if we update the Alert before we update the streak record with new values, hence the need to wrap these queries in a transaction_

ref MI-200

### DIFF
--- a/src/workers/newView.ts
+++ b/src/workers/newView.ts
@@ -56,56 +56,63 @@ RETURNING "currentStreak", "maxStreak"
 const incrementReadingStreak = async (
   con: DataSource,
   data: DeepPartial<View>,
-): Promise<boolean> => {
-  const users = con.getRepository(User);
-  const repo = con.getRepository(UserStreak);
-  const { userId, timestamp } = data;
+): Promise<boolean> =>
+  con.transaction(async (entityManager) => {
+    const users = entityManager.getRepository(User);
+    const repo = entityManager.getRepository(UserStreak);
+    const { userId, timestamp } = data;
 
-  if (!userId) {
-    return false;
-  }
-
-  const viewTime = timestamp ? new Date(timestamp as string) : new Date();
-
-  // TODO: This code is temporary here for now because we didn't run the migration yet,
-  // so not every user is going to have a row in the user_streak table and we need to create it.
-  // We can get rid of it once/if we implement the migration in https://dailydotdev.atlassian.net/browse/MI-70
-  const user = await users.findOne({
-    where: { id: userId },
-    relations: ['streak'],
-  });
-  const streak = await user?.streak;
-  if (user && !streak) {
-    await repo.save(
-      repo.create({
-        userId,
-        currentStreak: 1,
-        totalStreak: 1,
-        maxStreak: 1,
-        lastViewAt: viewTime,
-      }),
-    );
-  } else {
-    const results = await repo.query(INC_STREAK_QUERY, [
-      userId,
-      viewTime,
-      DEFAULT_TIMEZONE,
-    ]);
-
-    const { currentStreak, maxStreak } = results?.[0]?.[0] ?? {};
-
-    if (currentStreak > 0) {
-      // milestones are currently defined on fibonacci sequence
-      const showStreakMilestone =
-        isFibonacci(currentStreak) || currentStreak >= maxStreak;
-      await con.getRepository(Alerts).save({
-        userId,
-        showStreakMilestone,
-      });
+    if (!userId) {
+      return false;
     }
-  }
-  return true;
-};
+
+    const viewTime = timestamp ? new Date(timestamp as string) : new Date();
+
+    // TODO: This code is temporary here for now because we didn't run the migration yet,
+    // so not every user is going to have a row in the user_streak table and we need to create it.
+    // We can get rid of it once/if we implement the migration in https://dailydotdev.atlassian.net/browse/MI-70
+    const user = await users.findOne({
+      where: { id: userId },
+      relations: ['streak'],
+    });
+
+    const streak = await user?.streak;
+    // we need to check before we actually update the max streak
+    // otherwise we end up in a spot where we create an alert record
+    // every time a post is viewed, even if we already alerted the user
+    const { currentStreak, maxStreak } = streak || {
+      currentStreak: 0,
+      maxStreak: 0,
+    };
+
+    // milestones are currently defined on fibonacci sequence
+    const showStreakMilestone =
+      isFibonacci(currentStreak + 1) &&
+      streak?.lastViewAt?.getDate() !== viewTime.getDate();
+
+    const showMaxStreakMilestone =
+      showStreakMilestone && currentStreak + 1 > maxStreak;
+
+    await entityManager.getRepository(Alerts).save({
+      userId,
+      showStreakMilestone: showStreakMilestone || showMaxStreakMilestone,
+    });
+
+    if (user && !streak) {
+      await repo.save(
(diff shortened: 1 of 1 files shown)


src/workers/newView.ts
```diff
@@ -56,56 +56,62 @@ RETURNING "currentStreak", "maxStreak"
...
+    const { currentStreak, maxStreak } = streak || {};
+
+    // milestones are currently defined on fibonacci sequence
+    const showStreakMilestone =
+      isFibonacci(currentStreak) &&
+      streak?.lastViewAt?.getDate() !== viewTime.getDate();
+
+    const showMaxStreakMilestone =
+      streak &&
+      currentStreak + 1 > maxStreak &&
```
nensidosari, line 92: Why is this current streak + 1?

DragosIorgulescu: We are checking if the next increment is a new max streak here because this check is happening before we actually perform the query to increment the value in the db

src/workers/newView.ts
```diff
@@ -56,56 +56,62 @@ RETURNING "currentStreak", "maxStreak"
...
+
+    const streak = await user?.streak;
+    // we need to check before we actually update the max streak
+    // otherwise we end up in a spot where we create an alert record
+    // every time a post is viewed, even if we already alerted the user
+    const { currentStreak, maxStreak } = streak || {};
+
+    // milestones are currently defined on fibonacci sequence
+    const showStreakMilestone =
+      isFibonacci(currentStreak) &&
```
nensidosari, line 87: Then based on [this](https://github.com/dailydotdev/daily-api/pull/1729/files#r1508980994), would we need to check for currentStreak + 1 here as well?

src/workers/newView.ts
```diff
@@ -56,56 +56,62 @@ RETURNING "currentStreak", "maxStreak"
...
+      streak?.lastViewAt?.getDate() !== viewTime.getDate();
+
+    const showMaxStreakMilestone =
+      streak &&
+      currentStreak + 1 > maxStreak &&
+      streak.lastViewAt.getDate() !== viewTime.getDate();
+
+    await entityManager.getRepository(Alerts).save({
```
rebelchris, line 95: Correct that we always insert something?
Should it not only insert if something is true?

rebelchris (commented):
Just one question, but maybe I'm missing some context that happens.

src/workers/newView.ts
DragosIorgulescu: We are using the same `Alert` record here, and instead of creating a new one every time, we toggle the `showStreakMilestone` flag on the existing one

src/workers/newView.ts
```diff
@@ -56,56 +56,63 @@ RETURNING "currentStreak", "maxStreak"
...
-        totalStreak: 1,
-        maxStreak: 1,
-        lastViewAt: viewTime,
-      }),
-    );
-  } else {
-    const results = await repo.query(INC_STREAK_QUERY, [
+    // TODO: This code is temporary here for now because we didn't run the migration yet,
+    // so not every user is going to have a row in the user_streak table and we need to create it.
+    // We can get rid of it once/if we implement the migration in https://dailydotdev.atlassian.net/browse/MI-70
```
nensidosari, line 73: This ticket is already done, does this mean we have implemented the migration? meaning we can get rid of this code?

src/workers/newView.ts
rebelchris: Hmm but all we save is the type, so how do you base the trigger on this?

Feel like i miss why this is easier then the alternative.

src/workers/newView.ts
nensidosari: Just mentioned as it was referring to the done ticket.

capJavert: I think the part of the code that this was referring to was for `repo.save` right? Because that was creating the record for old users (the stuff script also needed to create). So we have to make sure the comments is attached properly.

src/workers/newView.ts
capJavert: I am ok to storing it in alerts to avoid changing the scope as you mentioned on slack 👍 . I also just miss on why we write every time, from my understanding of the logic it should be something like:
```js
if (showStreakMilestone || showMaxStreakMilestone) {
  await entityManager.getRepository(Alerts).save({
    userId,
    showStreakMilestone: showStreakMilestone || showMaxStreakMilestone,
  });
}
```

src/workers/newView.ts
```diff
@@ -56,56 +56,63 @@ RETURNING "currentStreak", "maxStreak"
...
+    // otherwise we end up in a spot where we create an alert record
+    // every time a post is viewed, even if we already alerted the user
+    const { currentStreak, maxStreak } = streak || {
+      currentStreak: 0,
+      maxStreak: 0,
+    };
+
+    // milestones are currently defined on fibonacci sequence
+    const showStreakMilestone =
+      isFibonacci(currentStreak + 1) &&
```
capJavert, line 90: Probably can extract `currentStreak + 1` as `nextStreak` to avoid multiple assignments below.

src/workers/newView.ts
```diff
@@ -56,56 +56,63 @@ RETURNING "currentStreak", "maxStreak"
...
+      streak?.lastViewAt?.getDate() !== viewTime.getDate();
```
kopancek, line 91: This is not going to work nice with user timezones. The backend runs on UTC, but the streak calculations run on whatever timezone the user has setup.

Imagine the following scenario:
User is in PST timezone (UTC - 8h). The user reads an article at `2024-03-05 22:12:34-08:00`, which is `2024-03-06T06:12:34Z`. Last view at is `2024-03-05 14:00:52-08:00`, which is `2024-03-05T22:00:52Z`. So `showStreakMilestone` would be true here, even though we are not actually increasing the streak in the database with the `INC_STREAK_QUERY` in  the end.

I think we should instead rely on the previous logic. We can return `shouldIncrementStreak` from the `INC_STREAK_QUERY`. That way we know if the reading streak was updated on this view or not. If not, we don't update the alerts either.

Another possibility would be to modify the `INC_STREAK_QUERY` to also do the alert table modification. That way you can rely on DB logic completely.

src/workers/newView.ts
```diff
@@ -56,56 +56,63 @@ RETURNING "currentStreak", "maxStreak"
 const incrementReadingStreak = async (
   con: DataSource,
   data: DeepPartial<View>,
-): Promise<boolean> => {
-  const users = con.getRepository(User);
-  const repo = con.getRepository(UserStreak);
-  const { userId, timestamp } = data;
+): Promise<boolean> =>
+  con.transaction(async (entityManager) => {
```
kopancek, line 60: Why do we need a transaction here?

kopancek (changes requested):
I don't think the current approach is going to work well with timezones.

src/workers/newView.ts
kopancek: To put money where my mouth is, created an alternative approach PR: https://github.com/dailydotdev/daily-api/pull/1735

src/workers/newView.ts
DragosIorgulescu: I moved the alert update before the increment query, so I wanted this to rollback if for some reason the increment query failed

PR #1735, opened 2024-03-05
fix: don't trigger alerts multiple times on the same day

kopancek:
Alternative approach to 
- https://github.com/dailydotdev/daily-api/pull/1729

Here we update the SQL query to return if the streak was incremented or not and based on that continue setting the alert value if needed.

Doing it via DB logic avoids the problem with timezones.

### DIFF
--- a/__tests__/workers/newView.ts
+++ b/__tests__/workers/newView.ts
@@ -324,7 +324,7 @@ describe('reading streaks', () => {
       await con.getRepository(Alerts).clear();
     });
 
-    it('should set showStreakMilestone to true if current streak is same as max streak and not a fibonacci number', async () => {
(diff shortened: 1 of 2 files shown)


DragosIorgulescu (approved):
nice approach! I think we can merge this PR as is, and I'll update my PR to take your changes into account 🙏

capJavert (approved):
Ok, lets see where it goes.

kopancek:
> nice approach! I think we can merge this PR as is, and I'll update my PR to take your changes into account 🙏

Cool. I updated the PR to take [Sab's latest comment into account as well](https://dailydotdev.slack.com/archives/C05TQL6HWTH/p1709634303156989?thread_ts=1709567408.417039&cid=C05TQL6HWTH).

PR #1736, opened 2024-03-05
fix: add force refresh arg

rebelchris:
We push a `refresh` key from frontend, needs to propagate to feed itself.

### DIFF
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -213,6 +213,11 @@ export const typeDefs = /* GraphQL */ `
       unreadOnly: Boolean = false
 
       """
+      Force refresh the feed
+      """
+      refresh: Boolean = false
+
+      """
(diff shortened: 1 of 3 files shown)


omBratteng (approved):

src/schema/feeds.ts
```diff
@@ -212,6 +212,11 @@ export const typeDefs = /* GraphQL */ `
       """
       unreadOnly: Boolean = false
 
+      """
+      Force refresh the feed
+      """
+      refresh: Boolean = false
```
capJavert, line 218: As we synced on slack, we should add a metric to keep this usage in check. Ideally we would have a grace period internally in feed service to avoid cache overrides to often.

idoshamun (approved):

capJavert (approved):

nensidosari (approved):

PR #1739, opened 2024-03-06
feat: add new source post added notification type and notify logic

capJavert:
AS-154

- add new notification type
  - allow it for frontend subscribe
- update `postAdded` worker to notify for `machine` sources
- tested notification dispatch locally
- added trigger to unsubscribe user when he blocks a source automatically so we don't have to run requests from frontend, also added [automatic cache invalidation on frontend](https://github.com/dailydotdev/apps/pull/2750/commits/3c7d32fbeb4a69bb076345a4161b8096bd47bfd4)

image

### DIFF
--- a/src/migration/1709739543179-FeedSourceBlockUnsubscribe.ts
+++ b/src/migration/1709739543179-FeedSourceBlockUnsubscribe.ts
@@ -1,0 +1,32 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class FeedSourceBlockUnsubscribe1709739543179 implements MigrationInterface {
+    name = 'FeedSourceBlockUnsubscribe1709739543179'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      queryRunner.query(`
+        CREATE OR REPLACE FUNCTION unsubscribe_notification_source_post_added()
+          RETURNS TRIGGER
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          DELETE FROM notification_preference WHERE
+            "sourceId" = NEW."sourceId"
+             AND "userId" = NEW."feedId"
+             AND "notificationType" = 'source_post_added'
+             AND type = 'source'
+             AND status = 'subscribed';
+          RETURN NEW;
+        END;
+        $$
+      `)
+      queryRunner.query('CREATE TRIGGER unsubscribe_notification_source_post_added_trigger AFTER INSERT ON "feed_source" FOR EACH ROW EXECUTE PROCEDURE unsubscribe_notification_source_post_added()')
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+      queryRunner.query('DROP TRIGGER IF EXISTS unsubscribe_notification_source_post_added_trigger ON feed_source')
+      queryRunner.query('DROP FUNCTION IF EXISTS unsubscribe_notification_source_post_added')
+    }
+
+}
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -315,9 +315,7 @@ export const generateNotificationMap: Record<
     ctx: NotificationPostContext & NotificationDoneByContext,
   ) =>
     builder
-      .referenceSource(ctx.source)
       .icon(NotificationIcon.Bell)
       .avatarSource(ctx.source)
-      .targetSource(ctx.source)
-      .uniqueKey(ctx.post.id),
+      .objectPost(ctx.post, ctx.source, ctx.sharedPost),
 };
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -96,7 +96,7 @@ export const notificationTitleMap: Record<
   dev_card_unlocked: () => 'DevCard unlocked!',
   source_post_added: (
     ctx: NotificationPostContext & NotificationDoneByContext,
-  ) => `New posts from <b>${ctx.source.name}</b>, check them out now!`,
+  ) => `New post from <b>${ctx.source.name}</b>, check it out now!`,
 };
 
 export const generateNotificationMap: Record<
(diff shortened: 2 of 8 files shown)


src/notifications/generate.ts
```diff
@@ -307,4 +310,14 @@ export const generateNotificationMap: Record<
...
+  source_post_added: (
+    builder,
+    ctx: NotificationPostContext & NotificationDoneByContext,
+  ) =>
+    builder
+      .referenceSource(ctx.source)
+      .icon(NotificationIcon.Bell)
+      .avatarSource(ctx.source)
+      .targetSource(ctx.source)
+      .uniqueKey(ctx.post.id),
```
capJavert, line 322: Unique key so that user gets notified whenever new post is added. If not this user would be notified only once.

src/workers/newNotificationV2Mail.ts
```diff
@@ -637,6 +638,9 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
       },
     };
   },
+  source_post_added: async () => {
+    return null;
+  },
```
capJavert, line 643: Email notification is currently out of scope.

omBratteng (approved):

src/notifications/generate.ts
```diff
@@ -94,6 +94,9 @@ export const notificationTitleMap: Record<
   collection_updated: (ctx: NotificationPostContext) =>
     `The collection "<b>${ctx.post.title}</b>" just got updated with new details`,
   dev_card_unlocked: () => 'DevCard unlocked!',
+  source_post_added: (
```
rebelchris, line 97: This type of message might be weird UX if source adds more then 1 article in one go.
You would get multiple entries that all are the same (I know different post, but used doesn't know this)

capJavert: Agree, we can easily show title/attachment of the post. Would that make more sense? If yes I can ask product.

src/migration/1709739543179-FeedSourceBlockUnsubscribe.ts
```diff
@@ -0,0 +1,32 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class FeedSourceBlockUnsubscribe1709739543179 implements MigrationInterface {
+    name = 'FeedSourceBlockUnsubscribe1709739543179'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      queryRunner.query(`
+        CREATE OR REPLACE FUNCTION unsubscribe_notification_source_post_added()
```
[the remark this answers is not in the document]
capJavert: We have a separate triggers test folder that can be checked. In my opinion U would say triggers are not hidden any more then workers. As an idea, maybe we can have separate migrations folder for triggers to give them more visibility? 

I would love for ORM to support some kind of trigger builder that would be visible a bit more in code as well but it does not currently.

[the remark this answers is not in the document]
rebelchris: Hmm let's leave it for now, not 100% sure what i'd expect to make it more clear 🙏

src/notifications/generate.ts
rebelchris: Maybe, if you can double check if that's an option?

Maybe propose something like this?

Screenshot 2024-03-07 at 09 55 41

capJavert: Yes, just like squads, will raise 👌

nensidosari (approved):

src/notifications/generate.ts
capJavert: Not needed anymore because now we reference post for attachment.

src/notifications/generate.ts
capJavert: Updated in 6e664cf3c9a46537d6c1fcc4e158073aa8cff9d4, added updated notification look in PR description.

src/notifications/generate.ts
```diff
@@ -94,6 +94,9 @@ export const notificationTitleMap: Record<
...
+    ctx: NotificationPostContext & NotificationDoneByContext,
+  ) => `New posts from <b>${ctx.source.name}</b>, check them out now!`,
```
rebelchris, line 99: Oh just one minor thing, not sure who provided the copy, but it's a singular post right?
`New post from {source}, check it out now!` ?

rebelchris (approved):
Non blocking, but maybe good to check with Sab/Nimrod

src/notifications/generate.ts
capJavert: This is what it was in figma by Sab but yeah, I'll just do singular.

capJavert: Now that we started using GB on API we can basically load this from features also, but yes agree.

src/workers/newNotificationV2Mail.ts
idoshamun: Are you sure email is out of scope? It's weird we should include it

src/migration/1709739543179-FeedSourceBlockUnsubscribe.ts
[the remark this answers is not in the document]
idoshamun: > In my opinion I would say triggers are not hidden any more then workers
I don't agree as workers are versioned control and can be easily changed and reviewed. Whereas our current support for triggers version management is lacking, not our fault but the lack of ORM support. It's much harder to understand what's the latest version of triggers we run. With that said, it's out of scope to solve this issue as part of this PR. Something we should address at some point

src/workers/newNotificationV2Mail.ts
capJavert: Yes, it was mentioned in the spec and confirmed [here](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/801374248/Subscribe+to+source?focusedCommentId=818151448)

idoshamun: Ok, I asked for email. Without an email it will have almost no impact

PR #1743, opened 2024-03-12
feat: add entities for marketing CTAs

omBratteng:
AS-167

### DIFF
--- a/__tests__/boot.ts
+++ b/__tests__/boot.ts
@@ -91,6 +91,7 @@ const LOGGED_IN_BODY = {
     company: null,
     cover: null,
   },
+  marketingCta: null,
 };
 
 const ANONYMOUS_BODY = {
--- a/src/entity/MarketingCta.ts
+++ b/src/entity/MarketingCta.ts
@@ -15,13 +15,13 @@ export type MarketingCtaFlags = Partial<{
 @Entity()
 export class MarketingCta {
   @PrimaryColumn({ type: 'text' })
-  campaignId?: string;
+  campaignId: string;
 
   @Column({ default: () => 'now()' })
   createdAt: Date;
 
   @Column({ type: 'text' })
-  variant?: MarketingCtaVariant;
+  variant: MarketingCtaVariant;
 
   @Column({ type: 'jsonb', default: {} })
   flags: MarketingCtaFlags;
--- a/src/entity/MarketingCta.ts
+++ b/src/entity/MarketingCta.ts
@@ -1,0 +1,28 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+export type MarketingCtaVariant = 'card' | 'popover';
+
+export type MarketingCtaFlags = Partial<{
+  tagText: string;
+  tagColor: string;
+  ctaUrl: string;
+  ctaText: string;
+  description: string;
+  image: string;
+  title: string;
+}>;
+
+@Entity()
+export class MarketingCta {
+  @PrimaryColumn({ type: 'text' })
+  campaignId?: string;
+
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ type: 'text' })
+  variant?: MarketingCtaVariant;
+
+  @Column({ type: 'jsonb', default: {} })
+  flags: MarketingCtaFlags;
+}
--- a/src/entity/MarketingCta.ts
+++ b/src/entity/MarketingCta.ts
@@ -2,15 +2,15 @@ import { Column, Entity, PrimaryColumn } from 'typeorm';
 
 export type MarketingCtaVariant = 'card' | 'popover';
 
-export type MarketingCtaFlags = Partial<{
-  tagText: string;
-  tagColor: string;
+export type MarketingCtaFlags = {
+  title: string;
+  description?: string;
+  image?: string;
+  tagText?: string;
+  tagColor?: string;
   ctaUrl: string;
   ctaText: string;
-  description: string;
-  image: string;
-  title: string;
-}>;
+};
 
 @Entity()
 export class MarketingCta {
--- a/src/entity/user/UserMarketingCta.ts
+++ b/src/entity/user/UserMarketingCta.ts
@@ -1,0 +1,30 @@
+import { Column, Entity, ManyToOne, PrimaryColumn } from 'typeorm';
+import { User } from './User';
+import { MarketingCta } from '../MarketingCTA';
+
+@Entity()
+export class UserMarketingCta {
+  @PrimaryColumn({ type: 'text' })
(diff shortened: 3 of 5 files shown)


src/entity/MarketingCta.ts
```diff
@@ -0,0 +1,28 @@
...
+  ctaText: string;
+  description: string;
+  image: string;
+  title: string;
+}>;
+
+@Entity()
+export class MarketingCta {
+  @PrimaryColumn({ type: 'text' })
+  campaignId?: string;
```
omBratteng, line 18: @capJavert I was thinking, as this is going to be unique, we don't need a specific ID row.
What do you think?

src/routes/boot.ts
```diff
@@ -90,6 +90,7 @@ export type LoggedInBoot = BaseBoot & {
     canSubmitArticle: boolean;
   };
   accessToken: AccessToken;
+  marketingCta: null;
```
capJavert, line 93: This should be of type `marketingCta?: MarketingCta` right? Same below.

src/migration/1710235892939-MarketingCta.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class MarketingCta1710235892939 implements MigrationInterface {
+    name = 'MarketingCta1710235892939'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE TABLE "marketing_cta" ("campaignId" text NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "variant" text NOT NULL, "flags" jsonb NOT NULL DEFAULT '{}', CONSTRAINT "PK_28d00a5e83072d15724c50f8520" PRIMARY KEY ("campaignId"))`);
```
capJavert, line 7: Missing Replica identity full since it will be used with workers in the future.

src/entity/user/UserMarketingCta.ts
```diff
@@ -0,0 +1,30 @@
+import { Column, Entity, ManyToOne, PrimaryColumn } from 'typeorm';
+import { User } from './User';
+import { MarketingCta } from '../MarketingCta';
+
+@Entity()
+export class UserMarketingCta {
+  @PrimaryColumn({ type: 'text' })
+  marketingCtaId: string;
+
+  @PrimaryColumn({ length: 36 })
```
capJavert, line 10: We can just use `text`, it both uses the same [underlying data structure](https://www.depesz.com/index.php/2010/03/02/charx-vs-varcharx-vs-varchar-vs-text/).

src/entity/MarketingCta.ts
capJavert: If that is the case I agree.

capJavert (commented):
Minor comments, maybe think about merging this into epic branch to merge with mutation/boot implementation there. Just want to avoid any migrations due to nuances we sometimes miss while implementing, your call.

src/migration/1710235892939-MarketingCta.ts
omBratteng: Does that have to be added manually?

src/routes/boot.ts
omBratteng: It should, just haven't been implemented yet. Just added it to "reserve" the name for it. Shouldn't be an issue for now if I just merge it into a epic branch

omBratteng:
I created an epic branch and pointed this PR to that

src/migration/1710235892939-MarketingCta.ts
capJavert: Yes, there are examples in codebase.

src/routes/boot.ts
capJavert: Yes, that works 👍

src/migration/1710235892939-MarketingCta.ts
omBratteng: On `marketing_cta`, `user_marketing_cta` or both?

capJavert: We can add both.

omBratteng: done

capJavert (approved):

__tests__/boot.ts
```diff
@@ -91,6 +91,7 @@ const LOGGED_IN_BODY = {
     company: null,
     cover: null,
   },
+  marketingCta: null,
```
rebelchris, line 94: Could be seperate PR, but maybe let's add a test for testing that it adds one full object as well?

src/entity/MarketingCta.ts
rebelchris: In that case it shouldn't be optional right :)

src/entity/MarketingCta.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ type: 'text' })
+  variant?: MarketingCtaVariant;
```
rebelchris, line 24: Should not be optional.
Do we go for variant or type?
(I see [type](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/808878142/RFC-020+-+Idea+Validation+Tooling+-+marketing+posts#Arguments) here)

src/entity/MarketingCta.ts
```diff
@@ -0,0 +1,28 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+export type MarketingCtaVariant = 'card' | 'popover';
+
+export type MarketingCtaFlags = Partial<{
```
rebelchris, line 5: Do we maybe want to define which ones are required/not?

rebelchris (changes requested):
Minor comment on optionality.

src/entity/MarketingCta.ts
omBratteng: Changed to variant in RFC

__tests__/boot.ts
omBratteng: It'll come in the next PR. Just added it to "reserve" the name for it. So I also had to add it to the tests to satisfy the tests 😬

src/entity/MarketingCta.ts
omBratteng: Yeah, I've updated this type now

src/entity/MarketingCta.ts
omBratteng: Yeah, it shouldn't. Fixed it.

nensidosari (approved):

rebelchris (approved):

PR #1745, opened 2024-03-12
fix: introducing feature flagged meilisearch

rebelchris:
Running meilisearch as a option beyond feature flags. (Would say just for internal testing for now)
Not fully optimized, but leveraging all things we already have in place 🙏

### DIFF
--- a/.infra/Pulumi.prod.yaml
+++ b/.infra/Pulumi.prod.yaml
@@ -62,6 +62,12 @@ config:
       secure: AAABAAAcNFaMFyerA4UVMKYeQYf69uQ0mMfl2I0DIiKOBa6HwtbnTXoA1N4OwidUgnIRlGzdYrgQWzQTMsimcE/Yarpceg==
     magniOrigin:
       secure: AAABAHxH1dQ0e4nHwnseCd65qMd7qz3vSsKV/H0vMQIKWjxS4H7BemSLMpPGEH7YsDLUM9hEMw==
+    meiliIndex:
+      secure: AAABADypa9PiyaCl1GQdpBI+5tcJRA0G894+ZL/lqfW3iWEetKlYQg==
+    meiliOrigin:
+      secure: AAABAEueTZvzwG4iG4mZtiCnTO/NJafye14NBJQUaOWcRSs5+kymF+LMYk/Muqragwku/+7ZMuohDZ2lFQZD4sES6HiNVfBVTIkqjFh53lk=
+    meiliToken:
+      secure: AAABAIb8OOQKe7BopTHOYpKey+GCet0D+d0PKoTQBLZv6w1f6C0CjSL8SReV3toI1PVkMc9K65/WscTCZKJPO4PAIjKSrx5G
     nodeEnv: production
     ogUrl:
       secure: AAABAALHhEXgjaHbmM/389iGJ+mAweDwlWup8z1We+zmalk8DXIGAg+jjEC/58kl5VeBJw==
--- a/src/integrations/meilisearch.ts
+++ b/src/integrations/meilisearch.ts
@@ -8,6 +8,7 @@ const meiliIndex = process.env.MEILI_INDEX;
 
 type Meili = {
   post_id: string;
+  title?: string;
 };
 interface MeiliResponse {
   hits: Meili[];
--- a/src/integrations/meilisearch.ts
+++ b/src/integrations/meilisearch.ts
@@ -1,0 +1,31 @@
+import { retryFetchParse } from './retry';
+import { fetchOptions } from '../http';
+import { Headers } from 'node-fetch';
+
+const meiliOrigin = process.env.MEILI_ORIGIN;
+const meiliToken = process.env.MEILI_TOKEN;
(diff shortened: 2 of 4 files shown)


src/schema/search.ts
```diff
@@ -228,6 +246,29 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
       ctx,
       info,
     ): Promise<ConnectionRelay<GQLPost> & { query: string }> => {
+      const gbClient = getUserGrowthBookInstace(ctx.userId);
```
idoshamun, line 249: Our server side feature flag doesn't support the team attribute which will force us to use hardcoded user ids and bloat the boot endpoint. This is something we want to consider here. Can we perhaps send a search version from the client to `searchPosts` and `searchPostSuggestions` like we do with the feed? Will be much easier I believe

idoshamun (changes requested):
Someone is really keen to have new search 😉 
Overall looks great I had a suggestion and also `searchPostSuggestions` implementation is missing. We need to check their autocomplete support

src/schema/search.ts
```diff
@@ -201,9 +224,21 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
       getSession(ctx.userId, id),
     searchPostSuggestions: async (
       source,
-      { query }: { query: string },
+      { query, version }: { query: string; version: number },
       ctx,
     ): Promise<GQLSearchPostSuggestionsResults> => {
+      if (version === 2) {
+        const hits = await searchMeili(
+          `q=${query}&attributesToRetrieve=post_id,title`,
```
idoshamun, line 232: It's fine for now but pretty sure they have something dedicated for auto complete

idoshamun (approved):

src/schema/search.ts
rebelchris: Yeah they do, not sure if we just want to try it for now and see what we want to use?
They have a million and one options basically 😅

Just keeping it lean to evaluate for now (seems a tad slow still)

idoshamun: No worries let's get it started and optimize later

rebelchris:
Deploying for test drive.
Feel free to still leave feedback

.infra/Pulumi.prod.yaml
```diff
@@ -62,6 +62,12 @@ config:
       secure: AAABAAAcNFaMFyerA4UVMKYeQYf69uQ0mMfl2I0DIiKOBa6HwtbnTXoA1N4OwidUgnIRlGzdYrgQWzQTMsimcE/Yarpceg==
     magniOrigin:
       secure: AAABAHxH1dQ0e4nHwnseCd65qMd7qz3vSsKV/H0vMQIKWjxS4H7BemSLMpPGEH7YsDLUM9hEMw==
+    meiliIndex:
+      secure: AAABADypa9PiyaCl1GQdpBI+5tcJRA0G894+ZL/lqfW3iWEetKlYQg==
```
capJavert, line 66: I think we can manage indexes as enums right? If we have more in the future and they are not exactly secret?

src/integrations/meilisearch.ts
```diff
@@ -0,0 +1,32 @@
...
+  hits: Meili[];
+}
+export const searchMeili = async (params: string): Promise<Meili[]> => {
+  const headers = new Headers({
+    Authorization: `Bearer ${meiliToken}`,
+  });
+  const requestOptions = {
+    method: 'GET',
+    headers,
+  };
```
capJavert, line 23: I would advise using `POST` endpoint in the future, it has a lot more options and capabilities. https://www.meilisearch.com/docs/reference/api/search#search-in-an-index-with-post

src/schema/search.ts
```diff
@@ -224,10 +259,27 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
...
     ): Promise<ConnectionRelay<GQLPost> & { query: string }> => {
+      if (args.version === 2) {
+        const meilieSearchRes = await searchMeili(
+          `q=${args.query}&attributesToRetrieve=post_id`,
+        );
+
+        const meilieArgs: FeedArgs & { ids: string[] } = {
+          ...args,
+          ids: meilieSearchRes.map((x) => x.post_id),
+        };
```
capJavert, line 274: What fields are available in our documents? Also, we should also have dev only API key and index so we can connect to it and use in development locally.

capJavert (commented):
Awesome work on bringing it online fast 🚀

.infra/Pulumi.prod.yaml
rebelchris: Potentially yes, for now we only have one just added one more later of security 😂

src/integrations/meilisearch.ts
rebelchris: Nice both seem to have many options already 🤯
Lets see how poc goes if good we can switch 👍

src/schema/search.ts
rebelchris: Not many think top of head
- time
- post_id
- title
- content 

In this query we only need id as we do internal feed resolver anyway 🙏

PR #1746, opened 2024-03-12
fix: the relation column name in UserMarketingCta

omBratteng:
Didn't realise that typeorm had added a extra column for foreign key

### DIFF
--- a/src/entity/user/UserMarketingCta.ts
+++ b/src/entity/user/UserMarketingCta.ts
@@ -17,7 +17,7 @@ export class UserMarketingCta {
   readAt: Date;
 
   @ManyToOne(() => MarketingCta, {
-    lazy: true,
+    eager: true,
     onDelete: 'CASCADE',
   })
   @JoinColumn({ name: 'marketingCtaId' })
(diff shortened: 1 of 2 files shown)


src/entity/user/UserMarketingCta.ts
```diff
@@ -17,9 +17,10 @@ export class UserMarketingCta {
   readAt: Date;
 
   @ManyToOne(() => MarketingCta, {
-    lazy: true,
+    eager: true,
```
capJavert, line 20: Why do we need this? It means relation will be loaded each time we query the table through `find`. Which should not be the case.

capJavert (changes requested):
Pesky typeorm 😄 Good thing we added epic branch 👌

src/entity/user/UserMarketingCta.ts
omBratteng: Wasn't supposed to add eager to it

omBratteng: Intended to use `relations: ['marketingCta'],` in the findOne but was testing to see what was the difference and forgot to revert it

capJavert (approved):

PR #1747, opened 2024-03-12
feat: fetch marketing CTA for user if it exists and store it in redis

omBratteng:


src/routes/boot.ts
```diff
@@ -328,6 +330,47 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+  );
+  const rawRedisValue = await getRedisObject(redisKey);
+
+  if (rawRedisValue === 'SLEEPING') {
+    // TODO: remove log or replace with proper logging
+    console.log('sleeping, not fetching from db');
+    return null;
+  } else {
+    marketingCta = JSON.parse(rawRedisValue);
+  }
```
omBratteng, line 348: My reasoning behind this is to reduce the amount of database queries when the user have no marketing CTAs available

src/routes/boot.ts
```diff
@@ -328,6 +330,48 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+const getMarketingCta = async (con: DataSource, userId: string) => {
+  let marketingCta: MarketingCta | null = null;
+  const redisKey = generateStorageKey(
+    StorageTopic.Boot,
+    'marketing_cta',
+    userId,
+  );
+  const rawRedisValue = await getRedisObject(redisKey);
+
+  if (rawRedisValue === 'SLEEPING') {
```
rebelchris, line 342: Why do we write sleeping?
Should it not just be there or not?

omBratteng: Commented earlier about it
https://github.com/dailydotdev/daily-api/pull/1747#discussion_r1522265543

src/routes/boot.ts
rebelchris: Alright, not against it 👍
Maybe just make it a const :)

omBratteng: Set `SLEEPING` to be a const. Could also make it a global const that can be re-used too maybe?

rebelchris: Yeah sorry global const indeed so we can use it more often.
Maybe could be in redis.ts or something

omBratteng: Put it into a enum and called it RedisMagicValues 🪄  Didn't know of of a better name for it. Thought of using a enum so it could possibly be extended in the future

src/routes/boot.ts
```diff
@@ -328,6 +331,48 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+  const redisKey = generateStorageKey(
+    StorageTopic.Boot,
+    'marketing_cta',
+    userId,
+  );
+  const rawRedisValue = await getRedisObject(redisKey);
+
+  if (rawRedisValue === RedisMagicValues.SLEEPING) {
+    // TODO: remove log or replace with proper logging
+    console.log('sleeping, not fetching from db');
```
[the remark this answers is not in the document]
omBratteng: Same goes for the two others in this file I added

src/redis.ts
```diff
@@ -45,8 +45,13 @@ export function deleteKeysByPattern(pattern: string): Promise<void> {
 }
 
 export const ONE_MINUTE_IN_SECONDS = 60;
+export const ONE_HOUR_IN_SECONDS = 60 * 60;
 export const ONE_DAY_IN_SECONDS = 60 * 60 * 24;
 
+export enum RedisMagicValues {
+  SLEEPING = 'SLEEPING',
+}
```
omBratteng, line 53: Used enum in case we might want to add any magic values in the future

src/routes/boot.ts
```diff
@@ -328,6 +331,48 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+      relations: ['marketingCta'],
+    });
+
+    if (userMarketingCta) {
+      marketingCta = userMarketingCta.marketingCta;
+      await setRedisObject(redisKey, JSON.stringify(marketingCta));
+    } else {
+      marketingCta = null;
+      // TODO: replace 10 with ONE_HOUR_IN_SECONDS
+      await setRedisObjectWithExpiry(redisKey, RedisMagicValues.SLEEPING, 10);
```
omBratteng, line 366: Is 1h enough? Or maybe do a couple of hours? And when we do the customer.io integration, this redis key will be written over anyways, so don't think it hurts to have a long TTL 🤔

rebelchris: Could probably even be indefinite right?
Since whatever will set it, will simply overwrite the value 👀

src/routes/boot.ts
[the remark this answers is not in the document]
rebelchris: Hmm this one doesn't seem usefll tbh

src/routes/boot.ts
```diff
@@ -328,6 +331,48 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+    // TODO: remove log or replace with proper logging
+    console.log('sleeping, not fetching from db');
+    return null;
+  } else {
+    marketingCta = JSON.parse(rawRedisValue);
+  }
+
+  if (!marketingCta) {
+    // TODO: remove log or replace with proper logging
+    console.log('fetching from db');
```
rebelchris, line 353: This one you can remove, kind of normal behavior

src/routes/boot.ts
```diff
@@ -328,6 +331,48 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+      marketingCta = userMarketingCta.marketingCta;
+      await setRedisObject(redisKey, JSON.stringify(marketingCta));
+    } else {
+      marketingCta = null;
+      // TODO: replace 10 with ONE_HOUR_IN_SECONDS
+      await setRedisObjectWithExpiry(redisKey, RedisMagicValues.SLEEPING, 10);
+    }
+  } else {
+    // TODO: remove log or replace with proper logging
+    console.log('fetched from redis');
```
rebelchris, line 369: Probably the whole else can be removed.

src/routes/boot.ts
omBratteng: Set it to infinity

rebelchris (commented):
Think we don't need the logs here actually.
Only thing interesting would be try...catch around the query (*but seems overkill)

Maybe add a early return is no userId is set, and logger on that? (that should cover enough logging)

src/routes/boot.ts
omBratteng: Will remove, but keeping it for now as debugging

src/routes/boot.ts
omBratteng: Will remove, but keeping it for now as debugging

src/routes/boot.ts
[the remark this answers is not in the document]
omBratteng: Will remove, but keeping it for now as debugging

omBratteng:
Added early return and logging

omBratteng:
The early return is probably very redundant, as `getMarketingCta` is only ran inside `loggedInBoot`, which would assume there is a user.

__tests__/boot.ts
```diff
@@ -16,6 +16,8 @@ import {
   Feature,
   FeatureType,
   MachineSource,
+  MarketingCta,
```
rebelchris, line 19: Just something I want to streamline between apps and api 
I called it `MarketingCTA` as CTA is basically shortcut already.
But let me know which one we prefer 😅

__tests__/boot.ts
```diff
@@ -399,6 +407,105 @@ describe('logged in boot', () => {
   });
 });
 
+describe('boot marketing cta', () => {
```
rebelchris, line 410: Do you maybe want to add one with SLEEPING from start?

src/routes/boot.ts
```diff
@@ -328,6 +336,49 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+) => {
+  if (!userId) {
+    log.info('no userId provided when fetching marketing cta');
+    return null;
+  }
+
+  let marketingCta: MarketingCta | null = null;
+  const redisKey = generateStorageKey(
+    StorageTopic.Boot,
+    'marketing_cta',
```
rebelchris, line 352: Not sure it exists, but don't we have enum for storage keys? (maybe it's only on apps though)

src/routes/boot.ts
```diff
@@ -328,6 +336,49 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+    userId,
+  );
+  const rawRedisValue = await getRedisObject(redisKey);
+
+  if (rawRedisValue === RedisMagicValues.SLEEPING) {
+    return null;
+  } else {
+    marketingCta = JSON.parse(rawRedisValue);
```
rebelchris, line 360: Should we not simply return here already?

rebelchris (approved):
Minor comments, non blocking

src/routes/boot.ts
rebelchris: Yeah maybe, it's not for you, but something we should have done a while ago 😂

__tests__/boot.ts
omBratteng: I started with `MarketingCTA`, then went to `MarketingCta`. Mostly because we write `ID` as `Id`.


Personally I like the use the right capitalisation, but it looks so wrong if you need to write things like `MarketingCTAId`, like where is the abbreviation starting or ending 😬  So it only works when there's spaces separation the words

__tests__/boot.ts
omBratteng: Added one

src/routes/boot.ts
omBratteng: Added it just to have done it

src/routes/boot.ts
omBratteng: No 😿
It's for when there is no redis key available, the redis response will return `null`, and in that case, we want to query the database and fill it

src/routes/boot.ts
```diff
@@ -328,6 +341,49 @@ const getUser = (con: DataSource, userId: string): Promise<User> =>
...
+      order: { createdAt: 'ASC' },
+      relations: ['marketingCta'],
+    });
+
+    if (userMarketingCta) {
+      marketingCta = userMarketingCta.marketingCta;
+      await setRedisObject(redisKey, JSON.stringify(marketingCta));
+    } else {
+      marketingCta = null;
+      await setRedisObject(redisKey, RedisMagicValues.SLEEPING);
```
capJavert, line 380: Why do we need this `SLEEPING` value? Is it not the same as no value?

rebelchris: AFAIK to avoid having to re-query when empty.
But if we somehow can detect just if redis obj exist we could bypass this?

(Open to discussion)

src/routes/boot.ts
idoshamun: let's not set TTL for infinity on redis. We have to remember users are churning. Worst case we just query again the DB in real time and filling the cache

omBratteng: How long should we set TTL to? 24 hours? More?

idoshamun: Whatever works for you. Can be even days if you want. Just something

src/routes/boot.ts
omBratteng: Empty redis key returns `''`, and `null` if it doesn't exist.

My idea was for it to indicate that there is no CTA available, and no need to actually query the database on future requests (and when there is a CTA available, we'll just override the redis key directly).

There is [`exists`](https://redis.github.io/ioredis/classes/Redis.html#exists) that can be used instead perhaps. It returns `1` when the key exists and `0` if not.

capJavert: You would have to remove SLEEPING key when user gets a new CTA to avoid it not showing due to key being there?

capJavert: Sounds good to me, better to read once then to do exist and then read the value (2 redis queries).

@idoshamun what do you think?

capJavert (approved):

src/routes/boot.ts
omBratteng: There's now a TTL of 7 days on the redis key, so `SLEEPING` would expire by itself and then re-fetched after expiry.

Another way that `SLEEPING` would "expire", would be in the integration with customer.io when we assign a CTA for the user, we would just (over)write the redis key when we write to `user_marketing_cta`

PR #1748, opened 2024-03-13
feat: add separate meter reason for deleted users

omBratteng:
With the surge in new users, it would be interesting to see how many of them deletes their accounts.
Saw a surge of manual logouts in the metrics, but it doesn't differentiate between actual logout or deletion of the user.
This adds a new reason, `deletion logout`, to the counter
image

### DIFF
--- a/src/kratos.ts
+++ b/src/kratos.ts
@@ -164,12 +164,9 @@ export const logout = async (
   isDeletion = false,
 ): Promise<FastifyReply> => {
   const query = req.query as { reason?: LogoutReason };
-  const reason =
-    LogoutReason[
-      isDeletion
-        ? LogoutReason.UserDeleted
-        : query?.reason || LogoutReason.ManualLogout
-    ];
+  const reason = Object.values(LogoutReason).includes(query?.reason)
+    ? query.reason
+    : LogoutReason.ManualLogout;
 
   try {
     const { res: logoutFlow } = await fetchKratos(
@@ -186,6 +183,11 @@ export const logout = async (
       req.log.warn({ err: e }, 'unexpected error while logging out');
     }
   }
-  await clearAuthentication(req, res, reason);
+
+  await clearAuthentication(
+    req,
+    res,
+    isDeletion ? LogoutReason.UserDeleted : reason,
+  );
   return res.status(204).send();
 };
--- a/src/kratos.ts
+++ b/src/kratos.ts
@@ -160,8 +160,13 @@ export const dispatchWhoami = async (
 export const logout = async (
   req: FastifyRequest,
   res: FastifyReply,
-  isDeleted = false,
+  isDeletion = false,
 ): Promise<FastifyReply> => {
+  const query = req.query as { reason?: string };
+  const reason = isDeletion
+    ? 'deletion logout'
+    : query?.reason || 'manual logout';
+
   try {
     const { res: logoutFlow } = await fetchKratos(
       req,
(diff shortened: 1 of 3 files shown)


src/kratos.ts
```diff
@@ -176,6 +177,10 @@ export const logout = async (
       req.log.warn({ err: e }, 'unexpected error while logging out');
     }
   }
-  await clearAuthentication(req, res, 'manual logout');
+  await clearAuthentication(
+    req,
+    res,
+    isDeleted ? 'deletion logout' : 'manual logout',
```
rebelchris, line 183: It can actually still be split up more as from the frontend this can get triggered due to 2 reasons.
Maybe we want to enable that level of detail?

rebelchris (approved):
Non blocking, but maybe good to know 👍 (this nuance is what added the surge imo)

idoshamun (approved):

src/kratos.ts
omBratteng: Doesn't hurt. How do we tell them apart? Maybe add something in the frontend that differentiate the calls?

capJavert: Maybe also better tracked on the frontend since we have all the info there, api can stay as global counter.

rebelchris: Yep we'd need to send something from fe to the logout call basically.
Can differentiate between actual manual logout and existing_session_logout or something

Not super urgent though

rebelchris: @capJavert Frontend already tracks these as events. (there are cases where api will do logic though, that's why it's in 2 places now)

capJavert: Not sure what all the different cases are as well.

capJavert: @rebelchris nice, from my point of view if frontend tracka it no need to also do it here.

capJavert: Github is slow with showing comments to day... So we have these to cases that Ole-Martin added and two are tracked on the frontend?

omBratteng: Send this error ID as an argument inside `logout`
https://github.com/dailydotdev/apps/blob/5c716e2c6669e994dbac12e197a47f3e51b0cc57/packages/shared/src/hooks/useRegistration.ts#L85-L90

And inside the `logout` function, we add that argument as a `?reason=`
https://github.com/dailydotdev/apps/blob/5c716e2c6669e994dbac12e197a47f3e51b0cc57/packages/shared/src/lib/user.ts#L99-L104

which we then override here?

capJavert (approved):

src/kratos.ts
capJavert: Only thing that comes to mind is that it would be possible to fake it on frontend and send `reason=whatever` but not blocking logic if it helps us track this better.

omBratteng: Added reasoning on frontend https://github.com/dailydotdev/apps/pull/2805

src/kratos.ts
```diff
@@ -160,7 +160,13 @@ export const dispatchWhoami = async (
 export const logout = async (
   req: FastifyRequest,
   res: FastifyReply,
+  isDeletion = false,
 ): Promise<FastifyReply> => {
+  const query = req.query as { reason?: string };
+  const reason = isDeletion
+    ? 'deletion logout'
+    : query?.reason || 'manual logout';
```
capJavert, line 168: We should validate `reason`, enum could help as mentioned on frontend.

omBratteng:
```ts
export enum LogoutReason {
  ManualLogout = 'manual logout',
  UserDeleted = 'user deleted',
  KratosSessionAlreadyAvailable = 'kratos session already available',
}

const reason =
    LogoutReason[
      isDeletion
        ? LogoutReason.UserDeleted
        : query?.reason || LogoutReason.ManualLogout
    ];
```

Did it like that. Haven't tested it yet, so not 100% if it actually works

capJavert (approved):
Good. Just if you can validate requests without reason fallback as before.We could use tests for these kratos routes in the future.

PR #1766, opened 2024-03-21
feat: replace send type feature with daily digest config

capJavert:
Idea to [solve](https://dailydotdev.slack.com/archives/C06L6UB03U0/p1711028492991389), discussion and enable experimentation on both digest types.

Also request [here](https://dailydotdev.slack.com/archives/C06L6UB03U0/p1711097204217649?thread_ts=1711096239.730069&channel=C06L6UB03U0&message_ts=1711097204.217649)

- added new property `newUserSendType` to enrol new users to either send type (replaces current `personalized_digest_send_type` feature flag)
- weekly send type uses current personalized_digest config for weekly digest
- workdays send type uses new daily_personalized_digest config for daily digest

### DIFF
--- a/src/workers/personalizedDigestEmail.ts
+++ b/src/workers/personalizedDigestEmail.ts
@@ -22,6 +22,7 @@ interface Data {
   previousSendTimestamp: number;
   emailBatchId?: string;
   deduplicate?: boolean;
+  config?: PersonalizedDigestFeatureConfig;
 }
 
 type SetEmailSendDateProps = Pick<
@@ -75,6 +76,7 @@ const worker: Worker = workerToExperimentWorker({
       previousSendTimestamp,
       emailBatchId,
       deduplicate = true,
+      config,
     } = data;
     const emailSendDate = new Date(emailSendTimestamp);
     const previousSendDate = new Date(previousSendTimestamp);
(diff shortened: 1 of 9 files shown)


src/workers/personalizedDigestEmail.ts
```diff
@@ -43,6 +52,14 @@ const setEmailSendDate = async ({
...
 };
 
+const sendTypeToFeatureMap: Record<
+  UserPersonalizedDigestSendType,
+  Feature<PersonalizedDigestFeatureConfig>
+> = {
+  [UserPersonalizedDigestSendType.weekly]: features.personalizedDigest,
+  [UserPersonalizedDigestSendType.workdays]: features.dailyDigest,
+};
+
```
capJavert, line 62: Basically we have to do this matching since in RFC we opted to save send state in our db sp we cant just rely on feature flag to know which config to load. This will do for initial daily digest test but if we introduce more send types we will have to move to just evaluating GB feature in cron and then pass the value of digest config to worker directly.

rebelchris (approved):
Looks good to me.
Definitely more dynamic to have 2 for now.

Additional experiments on daily could also run faster compared to weekly.

capJavert:
Gonna merge this so I can test on retool. Nothing will happen until Monday EOD so we can go through setup on GB.

PR #1767, opened 2024-03-22
fix: additional feed for all comments with context

rebelchris:
Added a comment feed endpoint.
Rules set are: 
- No welcome post
- Only people with over 10 reputation

Not the biggest fan of the 3 joins, but for now it should be ok to experiment quickly with.
Also needed to add the parent context since we want to showcase if you replied to set person.

AS-202

### DIFF
--- a/src/schema/comments.ts
+++ b/src/schema/comments.ts
@@ -225,7 +225,7 @@ export const typeDefs = /* GraphQL */ `
       Paginate first
       """
       first: Int
-    ): CommentConnection!
+    ): CommentConnection! @auth
 
     """
     Get the comments of a post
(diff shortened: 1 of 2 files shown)


src/schema/comments.ts
```diff
@@ -206,6 +212,21 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      Paginate after opaque cursor
+      """
+      after: String
+
+      """
+      Paginate first
+      """
+      first: Int
+    ): CommentConnection!
```
capJavert, line 228: Should this be only for auth users?

capJavert (approved):

DragosIorgulescu (approved):

src/schema/comments.ts
rebelchris: Hmm at this point it is yes, let me add that quickly :)

omBratteng (approved):

PR #1769, opened 2024-03-25
ci: split tests

idoshamun:
Use CircleCI to run tests in parallel to reduce the CI time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,8 +62,8 @@ jobs:
       - run:
           name: Test
           command: |
-            TEST=$(find __tests__ -name "*.ts" -not -path "*/fixture/*")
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Does it also tackle the issue with rerunning only failed tests?
(Seeing you already use circleci runner)

.circleci/config.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           - POSTGRES_DB: api_test
           - POSTGRES_PASSWORD: 12345
       - image: redis:6
+    parallelism: 4
```
capJavert, line 37: Does this also create 4 databases for each run?

idoshamun: Yes. Basically it runs the same job 4 times with the full config and just split the tests using some circleci utils

idoshamun:
@rebelchris not sure tbh, haven't tried

.circleci/config.yml
capJavert: Perfect!

PR #1770, opened 2024-03-25
feat: comment downvote entities

capJavert:
AS-210

- adding new user_comment entity
- adding downvotes to comments
- refactor UserPostVote to UserVote
- migrations & test updates

Mutations and data move will be in separate PRs.

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -524,6 +524,14 @@ const obj = new GraphORM({
       },
     },
   },
+  UserComment: {
+    requiredColumns: ['votedAt'],
+    fields: {
+      flags: {
+        jsonType: true,
+      },
+    },
+  },
 });
 
 export default obj;
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -24,6 +24,7 @@ import { GQLBookmarkList } from '../schema/bookmarks';
 import { base64 } from '../common';
 import { GQLComment } from '../schema/comments';
 import { GQLUserPost } from '../schema/posts';
+import { UserComment } from '../entity/user/UserComment';
 
 const existsByUserAndPost =
   (entity: string, build?: (queryBuilder: QueryBuilder) => QueryBuilder) =>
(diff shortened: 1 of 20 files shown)


nensidosari (approved):
Looks good

src/graphorm/index.ts
```diff
@@ -390,6 +391,27 @@ const obj = new GraphORM({
...
+              .where(`${childAlias}."userId" = :userId`, { userId: ctx.userId })
+              .andWhere(`${childAlias}."commentId" = "${parentAlias}".id`);
+          },
+        },
+        transform: (value: GQLComment, ctx: Context) => {
+          if (!ctx.userId) {
+            return null;
+          }
+
+          if (!value) {
```
idoshamun, line 408: What's this?

idoshamun (approved):
Looks great!

src/graphorm/index.ts
```diff
@@ -390,6 +391,27 @@ const obj = new GraphORM({
           parentColumn: 'parentId',
         },
       },
+      userState: {
```
rebelchris, line 394: I think you can actually use `transform: nullIfNotLoggedIn,` here just to avoid the manual user exist and extra safety on ensuring it's your user?

src/migration/1711384643544-UserComment.ts
```diff
@@ -0,0 +1,40 @@
...
+              RETURNS TRIGGER
+              LANGUAGE PLPGSQL
+              AS
+            $$
+            BEGIN
+              NEW."votedAt" = now();
+              RETURN NEW;
+            END;
+            $$
+        `)
```
rebelchris, line 25: Can't we just make the column a default date entity so it will just set on insert?

rebelchris: Ah ok you need some logic to see if it's not 0

src/schema/comments.ts
```diff
@@ -211,6 +223,22 @@ export const typeDefs = /* GraphQL */ `
...
+  type UserComment {
+    """
+    The user's vote for the comment
+    """
+    vote: Int!
+
+    """
+    Time when vote for the comment was last updated
+    """
+    votedAt: DateTime
```
rebelchris, line 235: Does GQL actually need this?
We don't send/do anything with it on FE right?

rebelchris (approved):
Nothing breaking, do have some minor questions :)

src/schema/comments.ts
capJavert: It was exposed for user post as well. But yeah, currently not used by FE. I'll check and remove it if not needed.

src/graphorm/index.ts
capJavert: I can't since I also want to return defaults if there is not record. 
```js
if (!value) {
    return ctx.con.getRepository(UserComment).create();
 }
```

If `nullIfNotLoggedIn` would accept function then I would be able to do it.

PR #1780, opened 2024-03-27
feat: lofn integration

capJavert:


src/integrations/feed/configs.ts
```diff
@@ -111,3 +121,43 @@ export class FeedUserStateConfigGenerator implements FeedConfigGenerator {
...
+    this.baseConfig = baseConfig;
+    this.lofnClient = new LofnClient();
+    this.opts = opts;
+  }
+
+  async generate(
+    ctx: Context,
+    opts: DynamicConfig,
+  ): Promise<FeedConfigGeneratorResult> {
+    return runInSpan('FeedLofnConfigGenerator', async () => {
```
idoshamun, line 145: If it ever becomes a bottleneck we can run `fetchConfig` and `FeedPreferencesConfigGenerator` in parallel. Probably will require changing a little bit `FeedPreferencesConfigGenerator`. We can consider it as part of the cleanup

src/integrations/feed/generators.ts
```diff
@@ -151,5 +157,17 @@ export const feedGenerators: Record<FeedVersion, FeedGenerator> = Object.freeze(
...
+  if (version >= 30) {
+    return new FeedGenerator(
+      feedClient,
+      new FeedLofnConfigGenerator(
+        {},
+        {
+          feed_version: version.toString() as FeedVersion,
+        },
+      ),
+    );
```
idoshamun, line 169: During the cleanup, I would rethink it a little bit

src/integrations/feed/types.ts
```diff
@@ -57,8 +58,16 @@ export type FeedConfig = {
 
 export type DynamicConfig = Omit<FeedConfig, 'total_pages'>;
 
+export type FeedConfigGeneratorResult = {
+  config: FeedConfig;
+  tyr_metadata?: TyrMetadata;
```
idoshamun, line 63: Check my comment about making it more generic

src/integrations/feed/clients.ts
```diff
@@ -23,7 +23,12 @@ export class FeedClient implements IFeedClient {
     this.fetchOptions = fetchOptions;
   }
 
-  async fetchFeed(ctx, feedId, config): Promise<FeedResponse> {
+  async fetchFeed(
+    ctx,
+    feedId,
+    config,
+    tyr_metadata = undefined,
```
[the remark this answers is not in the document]
capJavert: Yes, its just a variable name, I will rename to `extraMetadata`, good point.

src/integrations/feed/configs.ts
capJavert: Yes, agreed 👍

src/integrations/feed/generators.ts
capJavert: Yes, basically the issue was that FeedConfig and feed_config_name gets passed around for other generators and I did not want to add `feed_version` to FeedConfig type. Will note it.

src/integrations/feed/types.ts
capJavert: Yes!

src/integrations/feed/configs.ts
capJavert: I'll take a look, it should not be too bad to implement it in this PR.

src/integrations/feed/clients.ts
```diff
@@ -37,10 +43,19 @@ export class FeedClient implements IFeedClient {
...
-      ]),
+      data: res.data.map(({ post_id, metadata }) => {
+        const hasMetadata = !!(metadata || extraMetadata);
+
+        return [
+          post_id,
+          (hasMetadata &&
+            JSON.stringify({
+              ...metadata,
+              mab: extraMetadata,
```
idoshamun, line 54: Originally this is what I meant
```suggestion
              ...extraMetadata,
```

capJavert: Ahh yes, will update.

src/integrations/feed/configs.ts
capJavert: Done 82511f866aaf8b1c58df9a9d1fb977e342e20b09

src/integrations/feed/configs.ts
```diff
@@ -111,3 +121,52 @@ export class FeedUserStateConfigGenerator implements FeedConfigGenerator {
...
+      const [lofnConfig, preferencesConfig] = await Promise.all([
+        this.lofnClient.fetchConfig({
+          user_id: opts.user_id,
+          feed_version: this.opts.feed_version,
+        }),
+        this.feedPreferencesConfigGenerator.generate(ctx, opts),
```
rebelchris, line 153: What would this fetch?
Isn't it just one call to lofn to include everything?

src/integrations/lofn/clients.ts
```diff
@@ -0,0 +1,33 @@
...
+  constructor(
+    url = process.env.LOFN_ORIGIN,
+    fetchOptions: RequestInit = globalFetchOptions,
+  ) {
+    this.url = url;
+    this.fetchOptions = fetchOptions;
+  }
+
+  fetchConfig(payload: LofnFeedConfigPayload): Promise<LofnFeedConfigResponse> {
+    return retryFetchParse(
```
rebelchris, line 23: Optional, since it's new maybe we'd like specific logging for fails?

rebelchris (approved):
Nothing blocking from my side. more questions :)

src/integrations/feed/configs.ts
rebelchris: Oh got you! 🙏

src/integrations/feed/configs.ts
```diff
@@ -129,31 +129,43 @@ type FeedLofnConfigGeneratorOptions = {
...
+      ]);
+
+      const config = getDefaultConfig(
+        {
+          ...preferencesConfig.config,
+          total_pages: lofnConfig.config.total_pages,
```
idoshamun, line 159: We don't really need to take lofn total pages. We can keep it owned by API just like all other pagination params

idoshamun (approved):
Just apply my comment, but looks amazing

src/integrations/lofn/clients.ts
capJavert: Added in 765337e984221288565f15363bbde8457940b2c7

src/integrations/feed/configs.ts
capJavert: `page_size` as well?

capJavert: These should take from API?
- fresh_page_size
- page_size
- total_pages

idoshamun: yes

idoshamun:
@capJavert can you please log the final configuration that are being generated by the lofn config generator? it would be useful while we debug. Include also the feed version and user id

capJavert:
@idoshamun added logging.

PR #1785, opened 2024-03-28
feat: comment downvote

capJavert:
- introduce new `vote` mutation (for all entities)
- add support for comment downvote
  - save to both `user_comment` and `comment_upvote` from old and new mutation
- switch to triggers to increment comment upvote/downvote (works with old and new mutations)
- update tests to confirm old/new use cases work

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -87,5 +87,5 @@ export const parseDate = (date: string | Date): Date | undefined => {
 };
 
 export const toGQLEnum = (value: Record<string, string>, name: string) => {
-  return `enum ${name} { ${Object.keys(value).join(' ')} }`;
+  return `enum ${name} { ${Object.values(value).join(' ')} }`;
 };
--- a/src/common/vote.ts
+++ b/src/common/vote.ts
@@ -1,5 +1,80 @@
+import { ValidationError } from 'apollo-server-errors';
+import { NotFoundError, TypeOrmError } from '../errors';
+import { Post, UserPost } from '../entity';
+import { GQLEmptyResponse } from '../schema/common';
+import { ensureSourcePermissions } from '../schema/sources';
+import { Context } from '../Context';
+
 export enum UserVote {
   Up = 1,
   None = 0,
   Down = -1,
 }
+
+export enum UserVoteEntity {
+  Comment = 'comment',
+  Post = 'post',
+}
+
+export type UserVoteProps = {
+  ctx: Context;
+  id: string;
+  vote: UserVote;
+};
+
+export const votePost = async ({
+  ctx,
+  id,
+  vote,
+}: UserVoteProps): Promise<GQLEmptyResponse> => {
+  try {
+    if (!Object.values(UserVote).includes(vote)) {
+      throw new ValidationError('Unsupported vote type');
+    }
+
+    const post = await ctx.con.getRepository(Post).findOneByOrFail({ id });
+    await ensureSourcePermissions(ctx, post.sourceId);
+    const userPostRepo = ctx.con.getRepository(UserPost);
+
+    switch (vote) {
+      case UserVote.Up:
+        await userPostRepo.save({
+          postId: id,
+          userId: ctx.userId,
+          vote: UserVote.Up,
+          hidden: false,
+        });
+
+        break;
+      case UserVote.Down:
+        await userPostRepo.save({
+          postId: id,
+          userId: ctx.userId,
+          vote: UserVote.Down,
+          hidden: true,
+        });
+
+        break;
+      case UserVote.None:
+        await userPostRepo.save({
+          postId: id,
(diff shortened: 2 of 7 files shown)


src/common/vote.ts
```diff
@@ -1,5 +1,168 @@
...
+
+    const comment = await ctx.con.getRepository(Comment).findOne({
+      where: {
+        id,
+      },
+      relations: ['post'],
+    });
+
+    if (!comment) {
+      throw new NotFoundError('Comment not found');
```
rebelchris, line 105: Why dont we just do findOneOrFail to keep it aligned?

src/common/vote.ts
```diff
@@ -1,5 +1,168 @@
...
+export const voteComment = async ({
+  ctx,
+  id,
+  vote,
+}: UserVoteProps): Promise<GQLEmptyResponse> => {
+  try {
+    validateVoteType({ vote });
+
+    const comment = await ctx.con.getRepository(Comment).findOne({
+      where: {
```
rebelchris, line 98: Maybe add select here to avoid big query?

src/entity/user/UserComment.ts
```diff
@@ -34,8 +34,8 @@ export class UserComment {
   @Column({ default: null, nullable: true })
   votedAt: Date;
 
-  @Column({ type: 'smallint', default: UserVote.None })
-  vote: UserVote = UserVote.None;
+  @Column({ type: 'smallint', default: 0 })
+  vote: UserVote = 0;
```
rebelchris, line 38: Why don't we use the enum here anymore? Was it an issue?

src/schema/comments.ts
```diff
@@ -954,13 +955,15 @@ export const resolvers: IResolvers<any, Context> = {
         const post = await comment.post;
         await ensureSourcePermissions(ctx, post.sourceId);
         await ctx.con.transaction(async (entityManager) => {
+          await entityManager.getRepository(UserComment).save({
+            commentId: id,
+            userId: ctx.userId,
+            vote: UserVote.Up,
+          });
           await entityManager.getRepository(CommentUpvote).insert({
```
rebelchris, line 963: One of the two here can eventually also be cleaned correct? Maybe just add todo as well

src/schema/posts.ts
```diff
@@ -1755,56 +1756,7 @@ export const resolvers: IResolvers<any, Context> = {
...
-        // Foreign key violation
-        if (err?.code === TypeOrmError.FOREIGN_KEY) {
-          throw new NotFoundError('Post or user not found');
-        }
-
-        throw err;
-      }
-
-      return { _: true };
+      return votePost({ ctx, id, vote });
```
rebelchris, line 1759: Would the userVote one not replace this one?

rebelchris (commented):
Nothing blocking just some questions mainly 🙏

src/schema/posts.ts
capJavert: Yes, but when frontend is refactored. I'll add a TODO

src/entity/user/UserComment.ts
capJavert: Dependency cycle, I'll sort it.

capJavert: Should be solved with 255ce7051d825ff95cbe5cb685ea9aad5541335e

src/common/utils.ts
```diff
@@ -87,5 +87,5 @@ export const parseDate = (date: string | Date): Date | undefined => {
 };
 
 export const toGQLEnum = (value: Record<string, string>, name: string) => {
-  return `enum ${name} { ${Object.keys(value).join(' ')} }`;
+  return `enum ${name} { ${Object.values(value).join(' ')} }`;
```
idoshamun, line 90: Make sure it won't break anything 🥵

capJavert: Yeah, actually it should have been like this from beginning.

src/entity/user/UserComment.ts
rebelchris:
```suggestion
  vote: UserVote = UserVote.None;
```

src/schema/comments.ts
```diff
@@ -37,6 +36,8 @@ import {
 import { ensureSourcePermissions, SourcePermissions } from './sources';
 import { generateShortId } from '../ids';
 import { CommentReport } from '../entity/CommentReport';
+import { UserComment } from '../entity/user/UserComment';
+import { UserVote } from '../types';
```
rebelchris, line 40: Could have actually done `import type` i guess?

rebelchris (approved):
Cool, some non blocking comments :)

src/schema/comments.ts
capJavert: We can't because it is used as value in the file.

PR #1786, opened 2024-03-29
feat: migrate from comment_upvote to user_comment

capJavert:
After https://github.com/dailydotdev/daily-api/pull/1785 is merged we can run the migration to bring two tables in parity with data. There is only ~60k rows so it can probably be run in single go.

### DIFF
--- a/bin/copyUserCommentUpvoteData.ts
+++ b/bin/copyUserCommentUpvoteData.ts
@@ -1,0 +1,41 @@
+import '../src/config';
+import createOrGetConnection from '../src/db';
+
+(async (): Promise<void> => {
+  const limitArgument = process.argv[2];
+  const offsetArgument = process.argv[3];
+
+  if (!limitArgument || !offsetArgument) {
+    throw new Error('limit and offset arguments are required');
+  }
+
+  const limit = +limitArgument;
+
+  if (Number.isNaN(limit)) {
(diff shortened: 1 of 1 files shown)


bin/copyUserCommentUpvoteData.ts
```diff
@@ -0,0 +1,41 @@
...
+  const offset = +offsetArgument;
+
+  if (Number.isNaN(offset)) {
+    throw new Error('offset argument is invalid, it should be a number');
+  }
+
+  const con = await createOrGetConnection();
+
+  await con.transaction(async (manager) => {
+    await manager.query(`SET session_replication_role = replica;`);
```
capJavert, line 27: Disable triggers for session, to prevent `comment.upvote` being incremented another time since they were already incremented when rows were added to old `comment_upvote`.

idoshamun: We need to make sure that:
* It actually works 😓 
* Doesn't have any implication on production queries
I think it should impact only the current transaction and work properly but just double check

idoshamun (approved):

bin/copyUserCommentUpvoteData.ts
capJavert: Yep, tested locally.

PR #1787, opened 2024-04-01
feat: reading reminder support for digest system

idoshamun:
As we want to add a daily reminder for users to read a post, I decided to use the existing daily digest system. 💪

This PR extends the digest system and allows us to leverage it for different use cases.

This is a proposal as the RFC wasn't approved yet. So feel free to reject this idea.

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -38,6 +38,8 @@ import {
   UserStreak,
   MarketingCta,
   UserMarketingCta,
+  UserPersonalizedDigestType,
+  UserPersonalizedDigestSendType,
 } from '../src/entity';
 import { sourcesFixture } from './fixture/source';
 import { getTimezonedStartOfISOWeek } from '../src/common';
@@ -2242,8 +2244,8 @@ describe('query personalizedDigest', () => {
 });
 
 describe('mutation subscribePersonalizedDigest', () => {
-  const MUTATION = `mutation SubscribePersonalizedDigest($hour: Int, $day: Int, $timezone: String) {
-    subscribePersonalizedDigest(hour: $hour, day: $day, timezone: $timezone) {
+  const MUTATION = `mutation SubscribePersonalizedDigest($hour: Int, $day: Int, $timezone: String, $type: DigestType) {
+    subscribePersonalizedDigest(hour: $hour, day: $day, timezone: $timezone, type: $type) {
       preferredDay
       preferredHour
       preferredTimezone
@@ -2410,6 +2412,29 @@ describe('mutation subscribePersonalizedDigest', () => {
       preferredDay: DayOfWeek.Friday,
       preferredHour: 22,
       preferredTimezone: 'Europe/Athens',
+    });
+  });
+
+  it('should subscribe to reading reminder', async () => {
+    loggedUser = '1';
+
+    const res = await client.mutate(MUTATION, {
+      variables: {
+        type: UserPersonalizedDigestType.ReadingReminder,
+      },
+    });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.subscribePersonalizedDigest).toMatchObject({
+      preferredDay: DayOfWeek.Monday,
+      preferredHour: 9,
+      preferredTimezone: 'Etc/UTC',
+    });
+    const digest = await con.getRepository(UserPersonalizedDigest).findOneBy({
+      userId: loggedUser,
+      type: UserPersonalizedDigestType.ReadingReminder,
+    });
+    expect(digest.flags).toEqual({
+      sendType: UserPersonalizedDigestSendType.workdays,
     });
   });
 });
--- a/bin/checkScheduledPush.ts
+++ b/bin/checkScheduledPush.ts
@@ -1,9 +1,0 @@
-import '../src/config';
-import { addMinutes } from 'date-fns';
-import { sendReadingReminderPush } from '../src/onesignal';
-
-(async (): Promise<void> => {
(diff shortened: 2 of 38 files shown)


idoshamun:
I forgot to add one change to the subscribe mutation. I'll do it soon, rest is ready to roll

src/workers/personalizedDigestEmail.ts
```diff
@@ -61,15 +66,61 @@ const sendTypeToFeatureMap: Record<
   [UserPersonalizedDigestSendType.workdays]: features.dailyDigest,
 };
 
-const worker: Worker = workerToExperimentWorker({
-  subscription: 'api.personalized-digest-email',
-  handler: async (message, con, logger, pubsub, allocationClient) => {
-    if (process.env.NODE_ENV === 'development') {
-      return;
-    }
+const dedupedSend = async (
```
capJavert, line 69: This should go to `common/personalizedDigest`

capJavert (commented):
I like implementation is low on code changes 👍 

Few places to check:
- route to send digest manually from retool, we need to make sure we pass correct type there
- cron, we need to add `WHERE` so that digest crons ignore reminder type

bin/checkScheduledPush.ts
```diff
@@ -0,0 +1,9 @@
+import '../src/config';
+import { addMinutes } from 'date-fns';
+import { sendReadingReminderPush } from '../src/onesignal';
+
+(async (): Promise<void> => {
```
rebelchris, line 5: This whole file needed?
Feels like a test for your own user?

src/entity/user/UserPersonalizedDigest.ts
```diff
@@ -7,6 +7,11 @@ export enum UserPersonalizedDigestSendType {
   workdays = 'workdays',
 }
 
+export enum UserPersonalizedDigestType {
+  digest = 'digest',
+  reading_reminder = 'reading_reminder',
```
rebelchris, line 12: We always use CamelCase for enum naming (I see the above is actually wrong)
 
```suggestion
  Digest = 'digest',
  ReadingReminder = 'reading_reminder',
```

src/onesignal.ts
```diff
@@ -37,3 +37,25 @@ export async function sendPushNotification(
...
+  push.send_after = at.toISOString();
+  push.contents = { en: 'Build a habit and become more knowledgeable' };
+  push.headings = { en: "Hi, It's reading time" };
+  push.url = addNotificationUtm(
+    process.env.COMMENTS_PREFIX,
+    'push',
+    'reminder',
+  );
+  push.chrome_web_badge =
+    'https://daily-now-res.cloudinary.com/image/upload/v1672745846/public/dailydev.png';
```
rebelchris, line 59: I saw some notes on removing this image, wasn't possible?

bin/checkScheduledPush.ts
idoshamun: Thought it could be nice utility for anyone who wants. But can remove!

src/onesignal.ts
idoshamun: missed that. you're right

src/entity/user/UserPersonalizedDigest.ts
idoshamun: Waiting for this: https://github.com/dailydotdev/daily-api/pull/1785/files#r1545581484
before I can apply the change. Otherwise it would mess the GQL enum. So i prefer to keep it as it is for now and then fix it. Another alternative is to copy this implementation to my PR as well and fix it here already. Wdyt?

idoshamun:
@capJavert 
> cron, we need to add WHERE so that digest crons ignore reminder type

actually i want to use the crons also for this reminder so it makes them generic and the logic is in the worker. does that make sense?

src/entity/user/UserPersonalizedDigest.ts
capJavert: Merged

src/onesignal.ts
idoshamun: notification looks crappy with that, so i'm keeping for now and letting design know

__tests__/users.ts
```diff
@@ -2412,6 +2414,29 @@ describe('mutation subscribePersonalizedDigest', () => {
...
+      variables: {
+        type: UserPersonalizedDigestType.ReadingReminder,
+      },
+    });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.subscribePersonalizedDigest).toMatchObject({
+      preferredDay: DayOfWeek.Monday,
+      preferredHour: 9,
+      preferredTimezone: 'Etc/UTC',
+    });
```
rebelchris, line 2431: Think you need type here now as well :)

rebelchris (approved):
Looks good to me :)
Testing frontend now and most stuff seems to work as intended

__tests__/users.ts
idoshamun: passes locally

capJavert (approved):
Looks good to me 🤘

__tests__/users.ts
capJavert: `toMatchObject` does not test extra values on object, but yeah, we can probably add to be sure it is correct 👍

rebelchris: Yeah this was what I ment :) 
We would have seen it failed as it was not added to object.

Maybe you meant `toStrictEqual`

capJavert: That one would be better agreed 👍

PR #1791, opened 2024-04-02
test: hopefully fix issues with ci

idoshamun:
After the CI optimizations I noticed that all tests are now flaky.

It seems to happen because the database connection already exists. Hopefully this will fix it


src/db.ts
```diff
@@ -1,13 +1,10 @@
 import { AppDataSource } from './data-source';
-import { DataSource } from 'typeorm';
-
-let connection: DataSource;
 
 const createOrGetConnection = async () => {
-  if (!connection) {
-    connection = await AppDataSource.initialize();
+  if (!AppDataSource.isInitialized) {
```
capJavert, line 4: Seems to fail again 😢

idoshamun: Yes I need to keep looking

PR #1794, opened 2024-04-03
feat: use timezone setting from user table

capJavert:
Remove `preferredTimezone` and always use timezone setting from `user` table.


src/cron/dailyDigest.ts
```diff
@@ -35,17 +38,20 @@ const cron: Cron = {
...
       queryBuilder: personalizedDigestQuery,
       logger,
-      handler: async ({ personalizedDigest, emailBatchId }) => {
+      handler: async ({
+        personalizedDigest: personalizedDigestWithTimezome,
+        emailBatchId,
+      }) => {
+        const { timezone = 'Etc/UTC', ...personalizedDigest } =
+          personalizedDigestWithTimezome as UserPersonalizedDigest &
+            Pick<User, 'timezone'>;
```
capJavert, line 47: Just to juggle type because typeorm types for `QueryBuilder` are garbage.

src/cron/dailyDigest.ts
```diff
@@ -18,9 +19,11 @@ const cron: Cron = {
   handler: async (con, logger) => {
     const personalizedDigestQuery = con
       .createQueryBuilder()
+      .select('upd.*, u.timezone')
       .from(UserPersonalizedDigest, 'upd')
+      .leftJoin(User, 'u', 'u.id = upd."userId"')
       .where(
-        'clamp_to_hours("preferredHour" - EXTRACT(HOUR FROM NOW() AT TIME ZONE "preferredTimezone")) = :preferredHourOffset',
+        `clamp_to_hours("preferredHour" - EXTRACT(HOUR FROM NOW() AT TIME ZONE COALESCE(NULLIF(u.timezone, ''), 'Etc/UTC'))) = :preferredHourOffset`,
```
omBratteng, line 26: Intended to change this to use backtick?

capJavert: yes, due to `''` in query.

capJavert: Yes I can also escape but this is more readable in my opinion.

src/cron/dailyDigest.ts
```diff
@@ -35,17 +38,20 @@ const cron: Cron = {
     await schedulePersonalizedDigestSubscriptions({
       queryBuilder: personalizedDigestQuery,
       logger,
-      handler: async ({ personalizedDigest, emailBatchId }) => {
+      handler: async ({
+        personalizedDigest: personalizedDigestWithTimezome,
+        emailBatchId,
+      }) => {
+        const { timezone = 'Etc/UTC', ...personalizedDigest } =
```
idoshamun, line 45: Can we store the default timezone somewhere if it's used multiple times?

idoshamun (approved):

src/cron/dailyDigest.ts
```diff
@@ -18,9 +19,11 @@ const cron: Cron = {
   handler: async (con, logger) => {
     const personalizedDigestQuery = con
       .createQueryBuilder()
+      .select('upd.*, u.timezone')
       .from(UserPersonalizedDigest, 'upd')
+      .leftJoin(User, 'u', 'u.id = upd."userId"')
```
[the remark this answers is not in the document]
capJavert: @idoshamun inner join?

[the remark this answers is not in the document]
idoshamun: yes

src/cron/dailyDigest.ts
capJavert: Yes, added in 0c522af937cc06dec013abfed2363ff1fc9f16b7

src/cron/dailyDigest.ts
omBratteng: Ah, didn't catch the `''`, I was only looking for `${}`

src/cron/dailyDigest.ts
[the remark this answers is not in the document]
capJavert: Since tables have constraint on user id is it not the same?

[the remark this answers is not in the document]
idoshamun: yes but `join` operation by default is inner join. using `leftJoin` means you have a reason to do that and you don't. so it's a cosmetic change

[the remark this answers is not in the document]
capJavert: Makes sense 👍

PR #1796, opened 2024-04-04
chore: update acquisition channels

nensidosari:
Updating acquisition channels based on this ask: https://dailydotdev.slack.com/archives/C02E2C3C13R/p1712140812579529

related PR: https://github.com/dailydotdev/apps/pull/2921


src/entity/user/User.ts
```diff
@@ -412,10 +412,10 @@ export const validateUserUpdate = async (
 
 export enum AcquisitionChannel {
   friend = 'friend',
-  social_media = 'social_media',
+  instagram_facebook = 'instagram_facebook',
+  x = 'x',
+  tiktok = 'tiktok',
   search_engine = 'search_engine',
-  blog = 'blog',
-  extension_store = 'extension_store',
```
capJavert, line 418: I think it would be better to keep the old ones on the API card for back compat and also since we have those values in DB already. On frontend we can remove them yes 👍

capJavert (approved):

src/entity/user/User.ts
idoshamun: I agree. I'd even go more extreme and remove the enum all together from the API and just keep it on the frontend. But it's not a blocker, just easy to do changes in one place only

capJavert: Enum is nice for equality checks. We could fix these kind of double maintenance with commons lib which could be used by apps and api, kinda like `schema` repo but for types and possibly some utils.

nensidosari: Agree on Ante's idea here, leaving this here ATM

idoshamun: API doesn't really do anything with that, it just receives and stores pretty dumb. There's no use in even maintaining that on the API side even as a shared library. It impacts productivity for now real benefit. That would be my two cents. As mentioned, not a blocker just something to think about

PR #1801, opened 2024-04-04
fix: add slug field to post entity

rebelchris:
Added slug field to post db

Uses postgres generated data (the long query basically concats title-id and lowercases it and removes special chars.

FE side coming up soon.

AS-227

### DIFF
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -230,4 +230,7 @@ export class Post {
     lazy: true,
   })
   public relatedPosts: Promise<PostRelation[]>;
+
+  @Column({ type: 'text', nullable: false })
+  slug: string;
 }
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -233,6 +233,8 @@ export class Post {
 
   @Column({
     type: 'text',
+    update: false,
+    insert: false,
     nullable: false,
     unique: true,
     generatedType: 'STORED',
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -232,5 +232,6 @@ export class Post {
   public relatedPosts: Promise<PostRelation[]>;
 
   @Column({ type: 'text', nullable: false, unique: true })
+  @Index('IDX_post_slug', { unique: true })
   slug: string;
 }
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -231,7 +231,13 @@ export class Post {
   })
   public relatedPosts: Promise<PostRelation[]>;
 
-  @Column({ type: 'text', nullable: false, unique: true })
+  @Column({
+    type: 'text',
+    nullable: false,
+    unique: true,
+    generatedType: 'STORED',
+    asExpression: `trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(post.title,'')||'-'||post.id)), '[^a-z0-9-]+', '-', 'gi'))`,
+  })
   @Index('IDX_post_slug', { unique: true })
   slug: string;
 }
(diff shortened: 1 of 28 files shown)


src/migration/1712228361792-PostSlug.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class PostSlug1712228361792 implements MigrationInterface {
+  name = 'PostSlug1712228361792';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "post" ADD "slug" text NOT NULL GENERATED ALWAYS AS (trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(post.title,'empty-title')||'-'||post.id)), '[^a-z0-9\\\\-_]+', '-', 'gi'))) STORED;`,
```
idoshamun, line 8: if title is empty let's fallback to just the id. but i think we should check if title is empty after stripping all the unique characters

rebelchris: The issue is that we can't use concat_ws which does exactly that because it's not immutable.
So we have to do individual fallbacks, can maybe fallback to empty space which will get cleaned?

idoshamun: This field also needs tests to make sure it handles all the special cases. it's high impact change so we must test it

idoshamun: > The issue is that we can't use concat_ws which does exactly that because it's not immutable.
So we have to do individual fallbacks, can maybe fallback to empty space which will get cleaned?
Not sure i understand but I'll see in the test case

rebelchris: @idoshamun Made it so it render empty title rather.
Checked it works and still adds post id as fallback.

Will add some tests as well :)

src/schema/posts.ts
```diff
@@ -1040,10 +1046,12 @@ const getPostById = async (
   id: string,
 ): Promise<GQLPost> => {
   const res = await graphorm.query<GQLPost>(ctx, info, (builder) => ({
-    queryBuilder: builder.queryBuilder.where(
-      `"${builder.alias}"."id" = :id AND "${builder.alias}"."deleted" = false`,
-      { id },
-    ),
+    queryBuilder: builder.queryBuilder
+      .where(
+        `"${builder.alias}"."id" = :id OR "${builder.alias}"."slug" = :id`,
```
capJavert, line 1051: Do we need `INDEX` for slug?

src/migration/1712228361792-PostSlug.ts
rebelchris: @idoshamun Added some tests with the examples from google.
Let me know if that works?

Note we don't do the full UTF-8 as we only rely on PG functions for now.
(I personally think this nuance is good as it allows faster shipping + less overhead/triggers) and our posts generally don't go that wild.

src/schema/posts.ts
rebelchris: We can add one for sure :)

idoshamun (approved):

capJavert (approved):

src/entity/posts/Post.ts
```diff
@@ -230,4 +230,8 @@ export class Post {
     lazy: true,
   })
   public relatedPosts: Promise<PostRelation[]>;
+
+  @Column({ type: 'text', nullable: false, unique: true })
```
capJavert, line 234: I think there is also syntax:

```js
@Column({
    generatedType: 'STORED',
    asExpression: 'sql expression here'
})
```

rebelchris: Oh cool!
Didnt find that as quickly 🫤

rebelchris: @capJavert Modified :) 
Also added the metadata sql part

PR #1803, opened 2024-04-05
feat: add comment downvote reputation events

capJavert:
- reputation workers for comment downvote

### DIFF
--- a/src/common/pubsub.ts
+++ b/src/common/pubsub.ts
@@ -85,6 +85,10 @@ const postCollectionUpdatedTopic = pubsub.topic(
 );
 const userReadmeUpdatedTopic = pubsub.topic('api.v1.user-readme-updated');
 const userReputatioUpdatedTopic = pubsub.topic('user-reputation-updated');
+const commentDownvotedTopic = pubsub.topic('api.v1.comment-downvoted');
+const commentDownvoteCanceledTopic = pubsub.topic(
+  'api.v1.comment-downvote-canceled',
+);
 
 export enum NotificationReason {
   New = 'new',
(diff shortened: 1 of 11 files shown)


src/common/pubsub.ts
```diff
@@ -505,6 +509,26 @@ export const notifyReputationIncrease = async (
     userAfter,
   });
 
+export const notifyCommentDownvoted = async (
```
idoshamun, line 512: I'd suggest to use the type safe version we have now for workers and messages. not a blocker but ideally as we move forward new implementations should utilize that

idoshamun (commented):
Please see my comment and let me know how you'd like to proceed

src/common/pubsub.ts
capJavert: Yep, I'll move, also 👍 for implementing new workers.

capJavert: be0b580ed2de42e5c010dccd863ca5df4e228d08

src/workers/cdc/primary.ts
```diff
@@ -317,6 +317,51 @@ const handleCommentUpvoteChange = async (
...
+          userId: data.payload.after.userId,
+        });
+      }
+      break;
+    }
+    case 'd': {
+      const wasDownvoted = data.payload.before.vote === UserVote.Down;
+
+      if (wasDownvoted) {
+        await triggerTypedEvent(logger, 'api.v1.comment-downvote-canceled', {
```
rebelchris, line 352: Just checking but with the typed ones did we not say we always send the entity as a whole to make it more streamlined for future additions? 
(can't fully recall if that was the decision)

rebelchris (approved):
Looks alright, one non blocking question

src/workers/cdc/primary.ts
capJavert: I think at the end it was more of a make it constant inside the subscriptions of the same topic. The [current typed ones ](https://github.com/dailydotdev/daily-api/blob/0c623e2911702484d5a3b0887071c7f177170142/src/common/typedPubsub.ts#L10)are also defined like that.

PR #1804, opened 2024-04-05
fix: properly handle notify on post downvote delete

capJavert:
While working on https://github.com/dailydotdev/daily-api/pull/1803 I noticed evaluation was wrong in downvote handler. It is an edge case because `user_post` rows never get deleted in the wild but revert to `vote: 0`, it would mostly happen if post row was deleted. Also added tests suite to match what we have with upvote.


idoshamun (approved):
Nice catch!

PR #1805, opened 2024-04-05
fix: sitemaps to use slugs

rebelchris:
Changed sitemaps to use slug instead of URL


idoshamun (approved):

capJavert (approved):
👌 Maybe just wait with merging until we truncate.

PR #1806, opened 2024-04-05
fix: truncate title for slug

rebelchris:
Added truncate of 100 to title part.
Had to add replica to typeorm table (some of the old migration downs would actually fail if we ever needed to run them)

Added specific test case as well to ensure it.

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -161,6 +161,21 @@ describe('slug field', () => {
     expect(res.data.post.slug).toBe('gem-se--p1');
   });
 
+  it('should return the post slug truncated if title is too long', async () => {
+    const repo = con.getRepository(ArticlePost);
+    await repo.update(
+      { id: 'p1' },
+      {
+        title:
+          'Donec vulputate neque a est convallis, at interdum ligula fermentum. Pellentesque euismod semper urna, ac eleifend felis viverra nec. Phasellus sit am',
+      },
+    );
+    const res = await client.query(QUERY);
+    expect(res.data.post.slug).toBe(
+      'donec-vulputate-neque-a-est-convallis-at-interdum-ligula-fermentum-pellentesque-euismod-semper-urn-p1',
+    );
+  });
+
   it('should return the post slug when searching for slug', async () => {
     const SUB_QUERY = `{
     post(id: "p1-p1") {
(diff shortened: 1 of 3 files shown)


__tests__/posts.ts
```diff
@@ -161,6 +161,21 @@ describe('slug field', () => {
     expect(res.data.post.slug).toBe('gem-se--p1');
   });
 
+  it('should return the post slug truncated if title is too long', async () => {
+    const repo = con.getRepository(ArticlePost);
+    await repo.update(
+      { id: 'p1' },
+      {
+        title:
+          'Donec vulputate neque a est convallis, at interdum ligula fermentum. Pellentesque euismod semper urna, ac eleifend felis viverra nec. Phasellus sit am',
```
capJavert, line 170: Love some latin on Fridays, reminds me of school 😆

src/migration/1712315208717-PostSlugMaxLength.ts
```diff
@@ -0,0 +1,72 @@
...
+    await queryRunner.query(`DROP INDEX "public"."IDX_post_slug"`);
+    await queryRunner.query(
+      `DELETE FROM "public"."typeorm_metadata" WHERE "type" = $1 AND "name" = $2 AND "database" = $3 AND "schema" = $4 AND "table" = $5`,
+      ['GENERATED_COLUMN', 'slug', 'api', 'public', 'post'],
+    );
+    await queryRunner.query(
+      `ALTER TABLE "post" DROP CONSTRAINT "UQ_cd1bddce36edc3e766798eab376"`,
+    );
+    await queryRunner.query(`ALTER TABLE "post" DROP COLUMN "slug"`);
+    await queryRunner.query(
```
capJavert, line 51: I know I said generated columns but seeing this it might be better long term to just have it as a regular column to avoid dropping each time we change something and instead do it with migrations? Not blocking for this though.

capJavert (approved):

src/migration/1712315208717-PostSlugMaxLength.ts
capJavert: We can still have it be readonly from API/ORM side.

rebelchris: Hmm hoping we don't have to change often?
It's kind of the same as trigger issues.

For now let's leave it, let's keep in mind if there's another issue, we go for text and just modify the value?

capJavert: Agreed, non blocking for this 👍

PR #1810, opened 2024-04-08
chore: split requests and limits

omBratteng:
Updated pulumi-common so that we can have separate requests and limits for memory.
I had to revert the previous attempt, because the limit and request was set to the same, causing OOMKilled errors.

Work on progress.


idoshamun (approved):

.infra/crons.ts
```diff
@@ -2,9 +2,12 @@ interface Cron {
   name: string;
   schedule: string;
   limits?: {
-    cpu: string;
     memory: string;
   };
+  requests?: {
+    cpu: string;
+    memory: string;
+  }
```
capJavert, line 10: What does `requests` define specifically?

omBratteng: `requests` is the minimum that the deployment needs to run, `limits` is the limit of what the deployment can use.

So now we set that api _**needs**_ 400MiB of memory to be deployed, so if there are two nodes available, but one has 300MiB memory available, and the other has 500MiB, the one with 500MiB is chosen.

Or if the cluster doesn't have enough memory "free" (not requested), it would add a new node to the pool.

capJavert (approved):

PR #1811, opened 2024-04-08
feat: return postId with forbidden error

capJavert:
Adding support for `extensions.postId` to allow clients to query private posts on client side. See https://github.com/dailydotdev/apps/pull/2951

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -1087,7 +1087,25 @@ export const resolvers: IResolvers<any, Context> = {
         partialPost.private ||
         sourceTypesWithMembers.includes(postSource.type)
       ) {
-        await ensureSourcePermissions(ctx, partialPost.sourceId);
+        try {
+          await ensureSourcePermissions(ctx, partialPost.sourceId);
+        } catch (permissionError) {
+          if (permissionError instanceof ForbiddenError) {
+            const forbiddenError = permissionError as ForbiddenError;
+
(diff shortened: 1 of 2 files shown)


src/schema/posts.ts
```diff
@@ -1087,7 +1087,25 @@ export const resolvers: IResolvers<any, Context> = {
...
+          if (permissionError instanceof ForbiddenError) {
+            const forbiddenError = permissionError as ForbiddenError;
+
+            const forbiddenErrorForPost = new ForbiddenError(
+              permissionError.message,
+              {
+                ...forbiddenError.extensions,
+                postId: partialPost.id,
+              },
+            );
```
capJavert, line 1102: This can be bundled in new Error class eg `PostForbiddenError` but I don't think it is needed at this point.

capJavert:
@DragosIorgulescu its hotfix at this point 😆 because private squad post pages upvoting does not work, see https://github.com/dailydotdev/apps/pull/2951

rebelchris (approved):

nensidosari (approved):

DragosIorgulescu:
> @DragosIorgulescu its hotfix at this point 😆 because private squad post pages upvoting does not work, see https://github.com/dailydotdev/apps/pull/2951

No worries 😅 just raising it as non-blocking, although maybe even more reason to add spec coverage as a follow-up ticket?

PR #1812, opened 2024-04-09
fix: vote change handling and move to typed workers

capJavert:
AS-237

Current worker logic that was applied from post never worked when user changes from upvote to downvote or vice verse. This fixes it and adds additional tests to cover the case.

Also refactored rep workers to typed workers because it allowed me to streamline cdc logic for any kind of future vote entities.


rebelchris (approved):
Looks good to me,  nice cleanup!

PR #1815, opened 2024-04-09
fix: prevent rep decrease if you delete own posts

rebelchris:
Prevent decrease of reputation if you delete your own post.
Seeing that we only work based of post table deletion with CDC i introduced a flag for `deletedBy` (might come in handy one day anyway)

I added some logic to say if the scout is the one deleting we don't decrease anyone, as they might spam posts of one user to attack their reputation.
(CC: @idoshamun not sure this is ok, let me know if not)

AS-239 #done

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -1367,7 +1367,10 @@ export const resolvers: IResolvers<any, Context> = {
           { id },
           {
             deleted: true,
-            flags: updateFlagsStatement<Post>({ deleted: true }),
+            flags: updateFlagsStatement<Post>({
+              deleted: true,
+              deletedBy: ctx.userId,
+            }),
           },
         );
         return { _: true };
@@ -1387,7 +1390,10 @@ export const resolvers: IResolvers<any, Context> = {
           { id },
           {
             deleted: true,
-            flags: updateFlagsStatement<Post>({ deleted: true }),
+            flags: updateFlagsStatement<Post>({
+              deleted: true,
+              deletedBy: ctx.userId,
+            }),
           },
         );
       });
--- a/src/workers/postBannedRep.ts
+++ b/src/workers/postBannedRep.ts
@@ -16,7 +16,8 @@ const worker: Worker = {
   subscription: 'post-banned-rep',
   handler: async (message, con, logger): Promise<void> => {
     const data: Data = messageToJson(message);
-    const { id, authorId, scoutId } = data.post;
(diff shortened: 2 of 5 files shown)


src/workers/postBannedRep.ts
```diff
@@ -16,7 +16,8 @@ const worker: Worker = {
   subscription: 'post-banned-rep',
   handler: async (message, con, logger): Promise<void> => {
```
rebelchris, line 17: We actually don't even check if the post is deleted or not here, but guess it's ok for now.

src/schema/posts.ts
```diff
@@ -1367,7 +1367,10 @@ export const resolvers: IResolvers<any, Context> = {
           { id },
           {
             deleted: true,
-            flags: updateFlagsStatement<Post>({ deleted: true }),
+            flags: updateFlagsStatement<Post>({
+              deleted: true,
+              deletedBy: ctx.userId,
+            }),
```
omBratteng, line 1373: Think it's a good thing for historical purposes

rebelchris: Agreed

omBratteng (approved):
LGTM

src/workers/postBannedRep.ts
```diff
@@ -38,15 +39,19 @@ const worker: Worker = {
...
 
-        if (authorId) {
+        /**
+         * Ensure authors can remove their own post
+         * Ensure scouts can't add posts/remove them to decrease author reputation
+         */
+        if (authorId && authorId !== deletedBy && scoutId !== deletedBy) {
           const authorEvent = repo.create({
             ...ownerProps,
             grantToId: authorId,
```
omBratteng, line 49: Maybe we could also set the `grantById` to `deletedBy`?

idoshamun (approved):
Nice! Love this approach

src/workers/postBannedRep.ts
capJavert: We can maybe check in CDC directly and just ignore all deleted posts? That one would be easy to implement even in this PR?

capJavert (approved):
I was expecting another column but baaam flags to the rescue 👌

rebelchris:
@capJavert It triggers on deletion is kind of the case.
We also assign reputation (positive) for those who reported.

So it's hard to distinguish between self deletion or not.
Think for now this will do until we have further cases.

PR #1816, opened 2024-04-10
fix: zombie process for images

rebelchris:
Classic reason why we need the type safe topics.
Didn't implement here though because this topic has way too many subscribers to quickly iterate "please note we should eventually do this as backlog ticket" :hidethepain:

Subscription expects `contentHtml` from the topic.
Topic never publishes this.
```js
export const generateNewImagesHandler =
  <T extends Data<Content> = Data<Content>>(
    key: keyof T,
    type: ContentImageUsedByType,
    contentKey: keyof Data<Content> = 'contentHtml',
  ) =>
  async (message, con): Promise<void> => {
    const data: T = messageToJson(message);
    const obj = data[key];

    if (!obj || !obj?.id || !obj?.[contentKey]) return;

    await updateUsedImagesInContent(con, type, obj.id, obj[contentKey]);
  };
```

Resulting in comment cleaned every 30 days, we can't retro bring these back 😢 
However added a bin script we need to run to fix existing 227 records 

AS-243 #done

### DIFF
--- a/__tests__/workers/cdc/primary.ts
+++ b/__tests__/workers/cdc/primary.ts
@@ -569,6 +569,7 @@ describe('comment', () => {
       'p1',
       '1',
       'c1',
+      '<p>comment</p>',
     ]);
   });
 
@@ -592,6 +593,7 @@ describe('comment', () => {
       '1',
       'c2',
       'c1',
+      '<p>comment</p>',
     ]);
   });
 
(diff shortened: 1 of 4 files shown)


idoshamun (approved):
Nice catch!

capJavert (approved):
Good catch, yeah even typed workers would have screamed about this 👍

PR #1817, opened 2024-04-10
fix: comment don't send as changeObject

rebelchris:
We actually send `post` as ChangeObject<Post>` but comment as raw data basically so we can't just rely on abstracting the type here.

I know it's a little bit messy, but for a quick fix the fastest solution.
Do think we should re-evaluate if we can move this topic to full object rather 🫠

### DIFF
--- a/src/workers/generators/generateNewImagesHandler.ts
+++ b/src/workers/generators/generateNewImagesHandler.ts
@@ -1,41 +1,15 @@
-import { ChangeObject } from '../../types';
 import {
   ContentImageUsedByType,
   updateUsedImagesInContent,
 } from '../../entity';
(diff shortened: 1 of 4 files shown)


src/workers/generators/generateNewImagesHandler.ts
```diff
@@ -17,13 +18,24 @@ export const generateNewImagesHandler =
   <T extends Data<Content> = Data<Content>>(
     key: keyof T,
     type: ContentImageUsedByType,
-    contentKey: keyof Data<Content> = 'contentHtml',
```
rebelchris, line 20: Not sure why we had this, wasn't ever used differently.

capJavert (approved):

PR #1818, opened 2024-04-10
feat: open keyword query and add meta props to flags

capJavert:
AS-241

- add title and description to `flags` 
- open keyword query so we can fetch it on `/tags/:tag` on frontend and get `description`
- added script for CSV import
- also added [Retool app](https://dailydotdev.retool.com/apps/8470354e-f745-11ee-b6ec-8b108f7c86be/%5BWIP%5D%20Update%20keyword%20(tag)) for keyword description update

### DIFF
--- a/__tests__/keywords.ts
+++ b/__tests__/keywords.ts
@@ -363,6 +363,7 @@ describe('flags field', () => {
   const QUERY = `{
     keyword(value: "react") {
       flags {
+        title
         description
       }
     }
@@ -380,12 +381,16 @@ describe('flags field', () => {
     await con.getRepository(Keyword).update(
       { value: 'react' },
       {
-        flags: updateFlagsStatement({ description: 'React is a JS library' }),
+        flags: updateFlagsStatement({
+          title: 'React',
+          description: 'React is a JS library',
+        }),
       },
     );
     const res = await client.query(QUERY);
     expect(res.errors).toBeFalsy();
     expect(res.data.keyword.flags).toEqual({
+      title: 'React',
       description: 'React is a JS library',
     });
   });
@@ -393,6 +398,7 @@ describe('flags field', () => {
   it('should return null values for unset flags', async () => {
     const res = await client.query(QUERY);
     expect(res.data.keyword.flags).toEqual({
+      title: null,
       description: null,
     });
   });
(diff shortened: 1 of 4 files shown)


capJavert:
@rebelchris I see CSV also has titles, should we add that too?

bin/updateKeyword.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      console.log('processing batch', i, 'of', keywords.length);
+
+      for (const keyword of keywordsBatch) {
+        await manager.query(
+          `
+            UPDATE keyword
+            SET flags = flags || $1
+            WHERE value = $2
+          `,
```
omBratteng, line 61: Should this be an upsert statement, in case some tags does not exist in db?

capJavert: Its not supposed to add new ones but good point ill add 'on conflict' just in case.

__tests__/keywords.ts
```diff
@@ -363,3 +358,49 @@ describe('mutation setKeywordAsSynonym', () => {
     expect(postKeywords).toMatchSnapshot();
   });
 });
+
+describe('flags field', () => {
```
rebelchris, line 362: Maybe just add a null test to be safe so keyword not found

rebelchris (commented):
Whats the title?
Didnt really check that, mayve we might as well just for safekeeping or maybe check with Nimrod tomorrow what he envisioned

rebelchris:
@capJavert I see the title in some instances is indeed slightly different then the tag itself.
Yeah let's add it, as we will need it for the overview page soon.

__tests__/keywords.ts
capJavert: There is already `should return null when keyword does not exist` test

bin/updateKeyword.ts
capJavert: Since it is `UPDATE` I do not need `on conflict` since statement will just not do anything because `WHERE` will not match.

__tests__/keywords.ts
rebelchris: Ah ok missed it :)

rebelchris (approved):
Nice!

PR #1819, opened 2024-04-11
chore: update keyword seeds with flags

capJavert:
Updating seeds for new flags that were added to `Keyword`.


rebelchris (approved):

bin/import.ts
```diff
@@ -20,6 +33,7 @@ const start = async (): Promise<void> => {
   await importEntity(con, 'Post');
   await importEntity(con, 'YouTubePost');
   await importEntity(con, 'Category');
+  await importEntity(con, 'Keyword', { conflictPaths: ['value'] });
```
capJavert, line 36: Had to add this option because in https://github.com/dailydotdev/daily-api/blob/main/src/migration/1658242193484-AddNewOneAITags.ts migration `Keyword` are added and then we can't seed because they conflict.

Seeds should still remain once off and clean in the future for other cases.

omBratteng (approved):

PR #1820, opened 2024-04-11
fix: add query to get related tags for source

rebelchris:
Add query to sources to get relatedTags for the source.
Based on:
- last 90 days post
- keyword count
- allowed keywords

Screenshot 2024-04-11 at 16 16 23
(subnote: would actually be nice to have a postman collection shared 😅)

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -29,6 +29,7 @@ import {
   EntityManager,
   EntityNotFoundError,
   FindOptionsWhere,
+  MoreThan,
 } from 'typeorm';
 import { GQLUser } from './users';
 import { Connection } from 'graphql-relay/index';
@@ -52,6 +53,7 @@ import {
 import { validateAndTransformHandle } from '../common/handles';
 import { QueryBuilder } from '../graphorm/graphorm';
 import type { GQLTagResults } from './tags';
+import { subDays } from 'date-fns';
 
 export interface GQLSource {
   id: string;
@@ -1137,6 +1139,7 @@ export const resolvers: IResolvers<any, Context> = {
       return res[0];
     },
     relatedTags: async (_, { sourceId }, ctx): Promise<GQLTagResults> => {
+      const timeThreshold = subDays(new Date(), 90);
       const keywords = await ctx.con
         .createQueryBuilder()
         .from(Post, 'p')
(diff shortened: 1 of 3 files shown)


src/schema/sources.ts
```diff
@@ -1121,6 +1138,30 @@ export const resolvers: IResolvers<any, Context> = {
       }
       return res[0];
     },
+    relatedTags: async (_, { sourceId }, ctx): Promise<GQLTagResults> => {
```
rebelchris, line 1141: @idoshamun Mostly from query point of view, any suggestions here or this will do for now?

idoshamun: that looks ok! let's monitor it and if needed introduce redis in front

idoshamun (approved):

src/schema/sources.ts
rebelchris: We can also consider some graphQl magic on FE with high stale time and combing the queries into one request.

capJavert: For tag recommendation of onboarding we used materialized view which we refresh in cron once a week.

src/schema/sources.ts
```diff
@@ -1121,6 +1138,30 @@ export const resolvers: IResolvers<any, Context> = {
...
+        .createQueryBuilder()
+        .from(Post, 'p')
+        .select('pk.keyword, count(pk.keyword) as count')
+        .innerJoin(
+          PostKeyword,
+          'pk',
+          'pk.postId = p.id and pk.status = :status',
+          { status: 'allow' },
+        )
+        .where({ sourceId })
```
capJavert, line 1153: Do we need to check if source access is allowed?

rebelchris: Nah think its fine, doesnt really expose anything and would cost another query 🥲

src/schema/sources.ts
rebelchris: We can start with this and monitor if we need more if that works for you?

capJavert: Yeah, from my side this looks ok for now.

src/schema/sources.ts
capJavert: Ok, let's leave it as is for now.

capJavert (approved):

PR #1821, opened 2024-04-12
fix: source by tag query

rebelchris:
It was either this query or the reverse double one (query post-keyword then do where IN [sourceId], but felt one query might be easier.

Could potentially be a heavy one due to count, so let's monitor everSQL closely with this :) 
Decided to make it a paginated one in case we want to use it for something else later on.

It can get a list of sources ranked by how many of the keyword occurrences happen on post it has.  (excluding private sources)

Screenshot 2024-04-12 at 10 00 59

On prod run:
Screenshot 2024-04-12 at 10 05 13
Screenshot 2024-04-12 at 10 05 41


capJavert (approved):
Nothing much to add, let's monitor how it performs.

PR #1822, opened 2024-04-12
feat: add source tag materialized view

rebelchris:
Added a materialized view for source-tag-count 
Will modify use-cases in a separate PR to keep things easier to review.

Local example:
Screenshot 2024-04-12 at 16 13 52

### DIFF
--- a/src/entity/SourceTagView.ts
+++ b/src/entity/SourceTagView.ts
@@ -1,0 +1,38 @@
+import { DataSource, ViewColumn, ViewEntity } from 'typeorm';
+import { Post } from './posts';
+import { subDays } from 'date-fns';
+import { PostKeyword } from './PostKeyword';
+import { Source } from './Source';
+
+@ViewEntity({
+  materialized: true,
+  expression: (dataSource: DataSource) =>
+    dataSource
+      .createQueryBuilder()
+      .select(`s.id as sourceId`)
+      .addSelect('pk.keyword AS tag')
+      .addSelect('count(pk.keyword) AS count')
+      .from(Source, 's')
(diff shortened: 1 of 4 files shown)


src/entity/SourceTagView.ts
```diff
@@ -0,0 +1,38 @@
...
+      })
+      .innerJoin(
+        PostKeyword,
+        'pk',
+        'pk.postId = p.id AND pk.status = :status',
+        { status: 'allow' },
+      )
+      .andWhere({ active: true, private: false })
+      .groupBy(`sourceId, tag`)
+      .orderBy('count', 'DESC'),
```
idoshamun, line 27: We don't really need order by but not blocking

src/entity/SourceTagView.ts
```diff
@@ -0,0 +1,38 @@
...
+})
+export class SourceTagView {
```
idoshamun, line 29: We need index for source and index for tag

idoshamun (approved):
Non blocking comments

src/entity/SourceTagView.ts
rebelchris: Thinking was it was cheaper to do here then in other queries using this?

idoshamun: Doing it here doesn't necessarily mean downstream queries will be ordered

rebelchris: Happy to remove it here 👍 will do with the indexes in one go

omBratteng (approved):

capJavert (approved):

PR #1824, opened 2024-04-15
fix: optimized new queries by using MV and added similar source query

rebelchris:
Optimized the recently added queries by leveraging the MV.
Also added the new query for similar sources.

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -6,8 +6,6 @@ import { Context } from '../Context';
 import {
   createSharePost,
   generateMemberToken,
-  Post,
-  PostKeyword,
   Source,
   SourceFeed,
   SourceMember,
@@ -29,7 +27,8 @@ import {
   EntityManager,
   EntityNotFoundError,
   FindOptionsWhere,
-  MoreThan,
+  In,
+  Not,
 } from 'typeorm';
 import { GQLUser } from './users';
 import { Connection } from 'graphql-relay/index';
@@ -53,7 +52,7 @@ import {
 import { validateAndTransformHandle } from '../common/handles';
 import { QueryBuilder } from '../graphorm/graphorm';
 import type { GQLTagResults } from './tags';
-import { subDays } from 'date-fns';
+import { SourceTagView } from '../entity/SourceTagView';
 
 export interface GQLSource {
   id: string;
@@ -298,6 +297,36 @@ export const typeDefs = /* GraphQL */ `
       Tag for which you want to find top sources
       """
       tag: String!
+
+      """
+      Exclude these sources
+      """
+      excludeSources: [String]
+
+      """
+      Paginate after opaque cursor
+      """
+      after: String
+
+      """
+      Paginate first
+      """
+      first: Int
+    ): SourceConnection!
+
+    """
+    Get sources similar to this source
+    """
+    similarSources(
+      """
+      Source to find similar sources for
+      """
+      sourceId: ID!
+
+      """
+      Exclude these sources
+      """
+      excludeSources: [String]
 
       """
       Paginate after opaque cursor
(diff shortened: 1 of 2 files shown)


src/schema/sources.ts
```diff
@@ -604,6 +633,8 @@ export enum SourcePermissions {
   Edit = 'edit',
 }
 
+const alwaysExcludeSources = ['unknown', 'community'];
```
capJavert, line 636: Always excluded for what? Maybe just naming a bit weird, we can also move it to `src/types` as global const.

src/schema/sources.ts
```diff
@@ -1023,24 +1067,68 @@ export const resolvers: IResolvers<any, Context> = {
...
+        .query(`with base as (SELECT * FROM source_tag_view), s as (
+            SELECT *, row_number() over (partition by "sourceId" order by count desc) rn
+            FROM base
+        )
+        SELECT s2."sourceId"
+        FROM s s1
+        JOIN s s2 on s1.tag = s2.tag and s1."sourceId" != s2."sourceId"
+        WHERE s1."sourceId" = '${args.sourceId}' and s1.rn <= 10 and s2.rn <= 10
+        GROUP BY 1
+        ORDER BY count(*) desc
```
capJavert, line 1111: Can you order by `count` column instead?

rebelchris: Oh yeah 100% this was actually just a copy paste thing 🤦

src/schema/sources.ts
rebelchris: I had them per query at first, you think it's better?
As it's only for 2 queries now.

capJavert: I am not sure what it means, for `unknown` is fine because it is a "ghost" source but for example `community` we don't exclude on different places.

capJavert: Someone could take it and use it as "this sources need to be excluded all the time".

rebelchris: Yeah let's just move it to both functions then it's more explicit :)

capJavert: Sounds good, just add it before merging.

rebelchris: Forgot to push, added now :)

src/schema/sources.ts
```diff
@@ -1023,24 +1066,69 @@ export const resolvers: IResolvers<any, Context> = {
...
+
+      const subQuery = await ctx.con
+        .query(`with base as (SELECT * FROM source_tag_view), s as (
+            SELECT *, row_number() over (partition by "sourceId" order by count desc) rn
+            FROM base
+        )
+        SELECT s2."sourceId", s2."count"
+        FROM s s1
+        JOIN s s2 on s1.tag = s2.tag and s1."sourceId" != s2."sourceId"
+        WHERE s1."sourceId" = '${args.sourceId}' and s1.rn <= 10 and s2.rn <= 10
```
capJavert, line 1109: Use params, this is vulnerable to SQL injection. 🙏

src/schema/sources.ts
```diff
@@ -904,6 +933,12 @@ interface SourcesArgs extends ConnectionArguments {
 
 interface SourcesByTag extends ConnectionArguments {
   tag: string;
+  excludeSources?: string[];
```
capJavert, line 936: Btw. what do we need `excludeSources` arg from on the frontend?

capJavert (changes requested):

src/schema/sources.ts
rebelchris: It's more of a additional thing as I build it to function as a paginated approach.
(initially the source-related-sources) would use it, but ended up not using it.

However we might find some case where we want to re-use it and have the functionality.

capJavert: 👍 ok, let's see while you build it, just looked that we might be able to get around without it in the first place if unused and also to limit what can be sent as args on these kind of paginated feed queries.

capJavert (approved):

PR #1829, opened 2024-04-16
fix: clean search suggestions

rebelchris:
Since it's a custom solution, decided to simply mimic the feedResolver to exclude the same things.
Should be alright for this beta release of search.

We are better of looking at adding props to Meleisearch in the future I think.


capJavert (approved):
Looks good, would you mind adding a test case for?
- unknown source
- user has some post hidden

idoshamun (approved):

rebelchris:
@capJavert can do, was more thinking not yet because its a experiment so far.
But happily do so

PR #1831, opened 2024-04-16
feat: node 20.x update

capJavert:
Update to Node 20.x as latest LTS. Will merge after deps update. Checked locally and containers run fine.


omBratteng (approved):

.circleci/config.yml
```diff
@@ -103,7 +103,7 @@ jobs:
   pulumi_up:
     circleci_ip_ranges: true
     docker:
-      - image: cimg/node:18.17
+      - image: cimg/node:20.12
```
rebelchris, line 106: These match node alpine?
Maybe we should set the nvmrc one to match explicitly?

rebelchris (approved):
Non blocking comments

idoshamun (approved):

.circleci/config.yml
capJavert: Yeah, alpine always runs its own version which is updated as the upstream image is updated. Minor version of node won't be breaking but yeah we can probably update .nvmrc to match.

rebelchris: Yeah in apps I set everything to 20.12 just to be more explicit.
(non blocking though)

PR #1833, opened 2024-04-17
fix: similar sources query used the wrong count

idoshamun:


src/schema/sources.ts
```diff
@@ -1099,16 +1099,16 @@ export const resolvers: IResolvers<any, Context> = {
...
         )
-        SELECT s2."sourceId", s2."count"
+        SELECT s2."sourceId"
         FROM s s1
         JOIN s s2 on s1.tag = s2.tag and s1."sourceId" != s2."sourceId"
         WHERE s1."sourceId" = $1 and s1.rn <= 10 and s2.rn <= 10
-        GROUP BY 1, s2."count"
-        ORDER BY s2."count" desc
+        GROUP BY 1
+        ORDER BY count(*) desc
```
rebelchris, line 1111: Ante asked me to use count here, this is why I added it.
Don't think it makes the query different right?

idoshamun: It's not the same value. This is why actually the production query prefers sources with lots of posts.
Basically I'm checking here how many of the top 10 tags are similar to the source in the request.
Using `s2.count` changes completely the logic

PR #1835, opened 2024-04-18
Infra: clickhouse sync deployment

vpol:


.infra/index.ts
```diff
@@ -522,3 +523,25 @@ if (vpcNativeProvider) {
     { provider: vpcNativeProvider.provider },
   );
 }
+
+deployClickhouseSync(
```
capJavert, line 527: This syncs all api tables into clickhouse?

vpol: nope, only specified [here](https://github.com/dailydotdev/daily-api/pull/1835/files#diff-7eb3c0dc4b6e8082a5405da027b746c2ae8567945714b1a8497351207852c6ceR28)

.infra/ch_sync.ts
```diff
@@ -0,0 +1,287 @@
...
+
+export function deployClickhouseSync(
+  name: string,
+  namespace: string | Input<string>,
+  args: ClickhouseSyncArgs,
+  {
+    limits = {
+      cpu: '1',
+      memory: '1024Mi',
+    },
```
omBratteng, line 33: Should split up limits and requests

.infra/index.ts
```diff
@@ -522,3 +523,25 @@ if (vpcNativeProvider) {
...
+      clickhouse_host: config.require('clickhouse_host'),
+      clickhouse_user: config.require('clickhouse_user'),
+      clickhouse_password: config.require('clickhouse_password'),
+      clickhouse_port: config.require('clickhouse_port'),
+      clickhouse_database: config.require('clickhouse_database'),
+    },
+  },
+  {
+    provider: vpcNativeProvider?.provider,
+  }
```
capJavert, line 546:
```suggestion
if (!isAdhocEnv) {
  deployClickhouseSync(
    name,
    namespace,
    {
      propsPath: './clickhouse-sync.yml',
      propsVars: {
        database_pass: config.require('debeziumDbPass'),
        database_user: config.require('debeziumDbUser'),
        database_dbname: name,
        hostname: envVars.typeormHost as string,
        clickhouse_host: config.require('clickhouse_host'),
        clickhouse_user: config.require('clickhouse_user'),
        clickhouse_password: config.require('clickhouse_password'),
        clickhouse_port: config.require('clickhouse_port'),
        clickhouse_database: config.require('clickhouse_database'),
      },
    },
    {
      provider: vpcNativeProvider?.provider,
    }
  );
}
```

.infra/index.ts
capJavert: We probably want to run this only in production since in local development it will fail due to missing clickhouse variables.

vpol: we going to move it to pulumi-common eventually, and it will be configurable the same way we configure debezium.

.infra/ch_sync.ts
vpol: I've used same approach as we use when deploying debezium (from pulumi-common)

https://github.com/dailydotdev/daily-api/pull/1835/files#diff-5cd5e53053e3d73392546492f319c6c184c71023dec170caed7ee3a71f6afc6dR275

no cpu limit for requested, only memory limit.

.infra/index.ts
idoshamun: @omBratteng do you think it will take time to move it to pulumi common? if not maybe we can do it right away to prevent the migration later?

.infra/ch_sync.ts
```diff
@@ -0,0 +1,287 @@
...
+    env = [],
+    image = 'gcr.io/daily-ops/clickhouse-sink-docker:latest',
```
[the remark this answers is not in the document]
vpol: it's a default. we can override it with version. since we have full control over versions, I think it's safe to use latest as default and override in specific cases (like testing)

.infra/ch_sync.ts
```diff
@@ -0,0 +1,287 @@
...
+      },
+      {provider},
+    );
+    volumes.push({
+      name: 'service-account-key',
+      secret: {
+        secretName: chsyncSecretSa.metadata.name,
+      },
+    });
+    volumeMounts.push({
```
idoshamun, line 226: Why do we need google service key here?

vpol: not sure. it was used in the debezium code. I will remove it 👍

.infra/ch_sync.ts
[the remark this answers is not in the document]
idoshamun: every time i see `:latest` i get the chills. Have bad experience with it. I'd rather have it fixed to something by default and if needed we can use latest in specific cases. otherwise we're risking cache issues

.infra/ch_sync.ts
idoshamun: debezium connects to pubsub so it's needed

.infra/ch_sync.ts
omBratteng: I know, but we should start defining a different request that acts as a minimum amount of resources needed. It probably doesn't need/use 1 CPU and 1 GiB as a minimum of resources all of the time

.infra/ch_sync.ts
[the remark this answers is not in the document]
omBratteng: Why do we build our own version of the clickhouse-sink, when they publish their own images?

https://hub.docker.com/layers/altinity/clickhouse-sink-connector/2.0.1-lt/images/sha256-99f38d1b5c34ac8bbcc86a86ca1eadca34e984c77ff1e2b3a54d5ee612029037

.infra/index.ts
omBratteng: Since we'll use this in other services later, we can move it to pulumi-common from the get-go

idoshamun: @omBratteng would be great if you can take care of it please 🙏

omBratteng: Will do it today

omBratteng: Created new PR for it #1858

.infra/ch_sync.ts
omBratteng: In #1858, these are split, and we'll have to circle back later and see what the actual usage is.

.infra/ch_sync.ts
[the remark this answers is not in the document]
omBratteng: In #1858 I used the upstream image

omBratteng:
Merged in #1858

PR #1836, opened 2024-04-18
fix: avoid complex SQL when incrementing streaks

DragosIorgulescu:
- make use of the view record we just created, or the last view record if we did not create one when checking if we need to increment streaks
- avoid timezone adjustments when updating `lastViewAt` -- we simply store the same date which is present on the `view.timestamp`, so when we check against it, we know we are checking against the same timezone and timestamp essentially

ref MI-313

### DIFF
--- a/__tests__/workers/newView.ts
+++ b/__tests__/workers/newView.ts
@@ -159,14 +159,24 @@ describe('reading streaks', () => {
   ) => {
     await prepareTest(currentDate, previousDate, previousStreak);
 
-    const streak = await con.getRepository(UserStreak).findOne({
-      where: {
-        userId: 'u1',
-        lastViewAt: new Date(currentDate),
-      },
+    const user = await con.getRepository(User).findOne({
+      where: { id: 'u1' },
+      relations: ['streak'],
     });
+
+    const streak = await user.streak;
+
+    const lastView = await con.getRepository(View).findOne({
+      where: { userId: 'u1' },
+      order: { timestamp: 'DESC' },
+    });
+
     for (const key in expectedStreak) {
-      expect(streak?.[key]).toEqual(expectedStreak[key]);
+      if (key === 'lastViewAt') {
+        expect(streak?.[key]).toEqual(lastView.timestamp);
+      } else {
+        expect(streak?.[key]).toEqual(expectedStreak[key]);
+      }
     }
   };
 
@@ -313,10 +323,12 @@ describe('reading streaks', () => {
   });
 
   it('should not increment a reading streak if lastViewAt is the same day', async () => {
-    await runTest('2024-01-26T19:17Z', '2024-01-26T17:23Z', defaultStreak, {
-      ...defaultStreak,
-      lastViewAt: new Date('2024-01-26T19:17Z'),
-    });
+    await runTest(
+      '2024-01-26T19:17Z',
+      '2024-01-26T17:23Z',
+      defaultStreak,
+      defaultStreak,
+    );
   });
 
   describe('showMilestone is set correctly', () => {
@@ -333,7 +345,6 @@ describe('reading streaks', () => {
           currentStreak: 6,
           totalStreak: 43,
           maxStreak: 6,
-          lastViewAt: new Date('2024-01-26T19:17Z'),
         },
       );
 
@@ -369,7 +380,6 @@ describe('reading streaks', () => {
           currentStreak: 6,
           totalStreak: 43,
           maxStreak: 10,
-          lastViewAt: new Date('2024-01-26T19:17Z'),
         },
       );
 
@@ -396,7 +406,6 @@ describe('reading streaks', () => {
           currentStreak: 6,
           totalStreak: 43,
           maxStreak: 10,
-          lastViewAt: new Date('2024-01-26T19:17Z'),
         },
       );
 
@@ -417,7 +426,6 @@ describe('reading streaks', () => {
         {
           ...defaultStreak,
           currentStreak: 5,
-          lastViewAt: new Date('2024-01-26T19:17Z'),
         },
       );
 
--- a/src/workers/newView.ts
+++ b/src/workers/newView.ts
@@ -6,7 +6,10 @@ import { isFibonacci } from '../common/fibonacci';
 
 const ONE_WEEK = 604800000;
 
-const addView = async (con: DataSource, entity: View): Promise<boolean> => {
+const addView = async (
+  con: DataSource,
+  entity: View,
+): Promise<{ view: View; isNewView: boolean }> => {
   const repo = con.getRepository(View);
   const existing = await repo.findOne({
     where: {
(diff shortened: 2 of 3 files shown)


idoshamun (commented):
Since we now migrated all users, should we consider this comment? https://github.com/dailydotdev/daily-api/pull/1836/files#diff-9ce2515f0e6544d6bcb887eed3d4a732981b644468f04b7f3985c000458199e7R43

Also what about the migration script, do we need to fix it?

src/workers/newView.ts
```diff
@@ -19,60 +22,24 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
...
-    return true;
+    const viewRecord = await repo.save(entity);
+    return { view: viewRecord, isNewView: true };
   }
-  return false;
+  return { view: existing, isNewView: false };
 };
 
-const DEFAULT_TIMEZONE = 'UTC'; // in case user doesn't have a timezone set
-const INC_STREAK_QUERY = `
```
capJavert, line 29: @kopancek can you take a look at this changes since you were the author of original query. I am all up for removing it since typeorm logic is much easier to parse mentally but I am not 100% if there is some edge case this new implementation does not cover. 

I see some tests fail so something might be off.

DragosIorgulescu: looking into specs now, they may be failing due to how they are set up, but yes, would love @kopancek eyes on this as well 🙏

DragosIorgulescu:
> Since we now migrated all users, should we consider this comment?

@idoshamun thought about that, but did not want to include that in the same PR, I'll push a separate one for this cleanup 🙏

DragosIorgulescu:
updated specs to explicitly validate we are setting `lastViewAt` correctly, and the streaks seed to reflect the fact that we are no longer storing the `lastViewAt` in the user's timezone 

_this does not require a re-run of the seed, because it was correctly converting the `view.timestamp` to the same timezone as the `lastViewAt` value, so it was correctly comparing apples to apples -- this was not happening in the `newView` worker_

idoshamun (dismissed):
Please wait for @capJavert approval please

__tests__/workers/newView.ts
```diff
@@ -159,14 +159,24 @@ describe('reading streaks', () => {
...
+      order: { timestamp: 'DESC' },
     });
+
     for (const key in expectedStreak) {
-      expect(streak?.[key]).toEqual(expectedStreak[key]);
+      if (key === 'lastViewAt') {
+        expect(streak?.[key]).toEqual(lastView.timestamp);
+      } else {
+        expect(streak?.[key]).toEqual(expectedStreak[key]);
+      }
```
kopancek, line 179: This change does not seem correct to me. Previously we were comparing the lastViewAt on the `user_streak` table, now we do it against the `view` table. 

If we want to keep the same test behavior, we should take the `lastViewAt` from the `streak` object you create above.

__tests__/workers/newView.ts
```diff
@@ -313,10 +323,12 @@ describe('reading streaks', () => {
...
   it('should not increment a reading streak if lastViewAt is the same day', async () => {
-    await runTest('2024-01-26T19:17Z', '2024-01-26T17:23Z', defaultStreak, {
-      ...defaultStreak,
-      lastViewAt: new Date('2024-01-26T19:17Z'),
-    });
+    await runTest(
+      '2024-01-26T19:17Z',
+      '2024-01-26T17:23Z',
+      defaultStreak,
+      defaultStreak,
```
kopancek, line 330: `defaultStreak` does not contain the `lastViewAt` timestamp. That's why the test was sending it before. Sending just `defaultStreak` as the last parameter here means, that we do not even try to compare the dates anymore. Which again, seems to defeat the purpose of the test.

__tests__/workers/newView.ts
```diff
@@ -333,7 +345,6 @@ describe('reading streaks', () => {
           currentStreak: 6,
           totalStreak: 43,
           maxStreak: 6,
-          lastViewAt: new Date('2024-01-26T19:17Z'),
```
kopancek, line 336: Same as above

src/workers/newView.ts
```diff
@@ -19,60 +22,24 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
     !existing ||
     entity.timestamp.getTime() - existing.timestamp.getTime() > ONE_WEEK
   ) {
-    await repo.save(entity);
-    return true;
+    const viewRecord = await repo.save(entity);
+    return { view: viewRecord, isNewView: true };
   }
-  return false;
+  return { view: existing, isNewView: false };
```
kopancek, line 28: I think if we extract the `isNewView` into the upper scope, the code would look nicer. E.g.:
```typescript
let view = existing;
const isNewView = !existing ||
    entity.timestamp.getTime() - existing.timestamp.getTime() > ONE_WEEK
if (isNewView) {
    view = await repo.save(entity)
}
return { view, isNewView };
```

src/workers/newView.ts
```diff
@@ -19,60 +22,24 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
...
-  const { userId, timestamp } = data;
+  const { userId, timestamp } = latestView;
 
   if (!userId) {
     return false;
   }
 
-  const viewTime = timestamp ? new Date(timestamp as string) : new Date();
-
   // TODO: This code is temporary here for now because we didn't run the migration yet,
```
kopancek, line 43: Since we are touching the file, we can basically get rid of this logic now. The migration has been run, so all users will have a record in the `user_streak` table AFAIK.

src/workers/newView.ts
```diff
@@ -159,13 +142,13 @@ const worker: Worker = {
       throw err;
     }
 
-    // no need to touch reading streaks if we didn't save a new view event or if we don't have a userId
-    if (!didSave || !data.userId) {
+    // no need to touch reading streaks if we didn't save a new view or if we don't have a userId
+    if (!didSave || !data.userId || !latestView) {
```
kopancek, line 146: Why do we need the `!latestView` condition as well? If `didSave` is `true`, `latestView` will always be defined.

If you want, we can instead base the logic on the `!latestView`. But then we can get rid of the `didSave` variable completely and rely on `latestView` being non null instead.

src/workers/newView.ts
```diff
@@ -88,37 +55,49 @@ const incrementReadingStreak = async (
...
-      await con.getRepository(Alerts).save({
-        userId,
-        showStreakMilestone,
-      });
-    }
+  if (
+    streak &&
+    (!streak.lastViewAt ||
+      streak.currentStreak === 0 ||
+      streak.lastViewAt.getDate() !== timestamp.getDate())
```
kopancek, line 69: Both `lastViewAt` and `timestamp` are in UTC by default. This code does not deal with user timezones at all and will increment the reading streak only with UTC in mind, not the user timezone. I'm not sure that is the intention.

Imagine the following scenarios:
# User in PDT
1. User is in PDT (UTC -08:00)
2. User reads an article at 12:00:00 local time, which is 20:00 UTC. We increment the streak
3. User reads article at 19:00:00 local time, same date, which is 3AM UTC the next day. We end up incrementing the streak, although we should not have.

# User in Sydney
1. User is in Sydney (UTC +10:00)
2. User reads an article at 16:00 local time, which is 2AM UTC. We increment the streak.
3. User reads an article the next day at 7:00 local time, which is 17:00 UTC, the same day as was the 2AM UTC view. We do not increment the streak, even if should have done it.

Does that make sense?

kopancek (changes requested):
I'm not sure the current code takes user timezone in mind at all. The tests also seem to have been crippled by the changes and do not test if the `lastViewAt` column on the streak is what we expected. Was that the intention?

src/workers/newView.ts
DragosIorgulescu: yep, already mentioned will be cleaning this up in a separate PR 🙏

__tests__/workers/newView.ts
DragosIorgulescu: we actually don't want to keep the same behaviour, the goal here is not to compare against the date coming in, but against the timestamp of the last view record. 

essentially what we are doing here is asserting that after the worker runs, our streak `lastViewAt` correctly matches the  `timestamp` of the last view -- we intentionally don't want to compare it against the date being passed in

__tests__/workers/newView.ts
DragosIorgulescu: yep, and it's okay that it does not contain it, as we are no longer comparing against the incoming date (which would require us to do timezone conversion shenanigans) 🙏

src/workers/newView.ts
DragosIorgulescu: sure thing, good point 👍

idoshamun:
> It does not matter here what the user timezone is, as when the View is created, the timestamp is stored at the current time converted to the db timezone, which is UTC. And as long as the streaklastViewAt is in the same timezone, we are doing a valid comparison.

This assumption is wrong isn't it? Because you want to check the two timestamps are on a different day. Day is defined by the time zone. Milan provides good examples that break this assumption

DragosIorgulescu:
My logic was more on making sure we don't have a mismatch between views and streaks, if we store views in UTC, do we end up displaying views in the right timezone?

idoshamun:
@DragosIorgulescu yes we always convert them. User time zone is dynamic so there's no point in storing the timestamp in the right timezones

kopancek:
> My logic was more on making sure we don't have a mismatch between views and streaks, if we store views in UTC, do we end up displaying views in the right timezone?

It depends on how the logic to show the views is implemented. You can do the same shenanigan, e.g. `AT TIMEZONE` of the user, something to check I guess.

The DB always stores all times in UTC. The NodeJS code on the backend uses UTC by default as well. This is not an issue by itself, until you start comparing the dates from the timestamp, which we need to do for streaks.

We can still cut away the complex SQL logic and do the DB queries with ORM, but we need to then do timezone conversion on the backend. E.g. using `moment` timezone functionality. This should be fine to add as a new dependency if we don't already have it.

__tests__/workers/newView.ts
kopancek: If we do not send the `lastViewAt`, then the `expectedStreak` object does not contain the `lastViewAt` key and we do not do this comparison at all:
```typescript
if (key === 'lastViewAt') {
        expect(streak?.[key]).toEqual(lastView.timestamp);
      } else {
        expect(streak?.[key]).toEqual(expectedStreak[key]);
      }
```

DragosIorgulescu:
ok, makes sense, nevermind these changes then 😅  I'll close this PR out without merging. Thanks for the input @kopancek , a very good sanity check on my ramblings 🙏

PR #1843, opened 2024-04-22
fix: clear boot cache on changes to Marketing CTA campaigns

omBratteng:
Removes the boot cache for a user if they are removed from a campaign, or when a campaign is deleted and the user has interacted with it (to prevent clearing cache if the user has a different campaign cached).

Also clears the boot cache for all users that have not interacted with the campaign when the campaign is edited.


capJavert (commented):
Looks good, missing tests coverage.

rebelchris (approved):

omBratteng:
Tests added

src/workers/cdc/primary.ts
```diff
@@ -740,6 +744,49 @@ const onPostRelationChange = async (
...
+  }
+};
+
+const onUserMarketingCtaChange = async (
+  data: ChangeMessage<UserMarketingCta>,
+) => {
+  // Only run on delete and if the user has read the Marketing CTA
+  if (data.payload.op !== 'd' || data.payload.before.readAt !== null) {
+    return;
+  }
```
capJavert, line 779: Can we split this into two separate returns?

__tests__/workers/cdc/primary.ts
```diff
@@ -2412,3 +2425,230 @@ describe('comment downvote', () => {
...
+          ),
+        ),
+      ).not.toBeNull();
+    });
+  });
+
+  describe('on user unassign', () => {
+    // afterEach(async () => {
+    //   await ioRedisPool.execute((client) => client.flushall());
+    // });
```
capJavert, line 2571: cleanup

src/workers/cdc/primary.ts
omBratteng: doen

__tests__/workers/cdc/primary.ts
```diff
@@ -2412,3 +2425,230 @@ describe('comment downvote', () => {
...
+        {
+          userId: usersFixture[0].id,
+        },
+        {
+          readAt: new Date('2024-03-13 13:00:00'),
+        },
+      );
+
+      expect(
+        await getRedisObject(
```
capJavert, line 2485: Can we test that `deleteRedisKey` was called appropriate amount of times. Just to validate it does not change more keys aside those we check. Then we can laso make sure that on `c` it is called 0 times.

omBratteng: Hmm, the tests expect redis storage to be changed, so if I mock it, the keys aren't deleted from redis?

omBratteng: I could instead just call redis instead to get a list of redis keys

omBratteng: I mocked it and just called redis inside the mocked function 😈

src/workers/cdc/primary.ts
```diff
@@ -740,6 +744,48 @@ const onPostRelationChange = async (
   }
 };
 
+const onMarketingCtaChange = async (
+  con: DataSource,
+  data: ChangeMessage<MarketingCta>,
+) => {
+  if (data.payload.op !== 'u') {
+    return;
+  }
```
rebelchris, line 753: What if it's deletion?

src/workers/cdc/primary.ts
```diff
@@ -740,6 +744,48 @@ const onPostRelationChange = async (
...
+    readAt: IsNull(),
+  });
+
+  if (users.length > 0) {
+    await deleteRedisKey(
+      ...users.map((user) =>
+        generateStorageKey(
+          StorageTopic.Boot,
+          StorageKey.MarketingCta,
+          user.userId,
```
rebelchris, line 766: Can't we get away with doing a single up? (so avoid the userId to do it for everyone)

rebelchris (commented):
Minor questions

src/workers/cdc/primary.ts
omBratteng: It cascades and deletes the rows in UserMarketingCta, which triggers the other change and deletes them then.

src/workers/cdc/primary.ts
omBratteng: The cache key is `boot:marketing_cta:XDCZD-PHG`, so if we delete with the `boot:marketing_cta` prefix, we would delete keys for users who might now be in the campaign.

rebelchris: Hmm that might be a interesting nuance, eg what happens if you in 2 campaigns?

src/workers/cdc/primary.ts
rebelchris: Ah got it :)

capJavert: Actually we do need to be careful deleting it could potentially schedule a lot of cdc messages, let's say 100_000 users are in some campaign and it gets deleted, that could clog our cdc worker for quite a lot.

src/workers/cdc/primary.ts
omBratteng: You only see one campaign at a time

rebelchris: Yes see one, but you could technically be spooled for one + another one right?

omBratteng: Yeah. You'd just re-fill the redis cache on your next boot request. It would still be the first in the queue

__tests__/workers/cdc/primary.ts
omBratteng: Ended up just checking for amount of keys in redis

src/workers/cdc/primary.ts
omBratteng: I can remove the deletion button from the retool app, that way they can't trigger a delete

omBratteng: I've disabled the delete button in Retool

src/redis.ts
```diff
@@ -69,3 +69,8 @@ export const setRedisObjectWithExpiry = (
 
 export const getRedisObject = (key) =>
   ioRedisPool.execute((client) => client.get(key));
+
+export const getRedisKeysByPattern = (pattern: string) => {
+  console.log('getRedisKeysByPattern', pattern);
```
capJavert, line 74: cleanup

omBratteng: 🤓

src/redis.ts
```diff
@@ -69,3 +69,8 @@ export const setRedisObjectWithExpiry = (
...
+  return ioRedisPool.execute((client) => client.keys(pattern));
+};
```
omBratteng, line 76:
```suggestion

export const getRedisKeysByPattern = (pattern: string) =>
  ioRedisPool.execute((client) => client.keys(pattern));
```

PR #1844, opened 2024-04-22
feat: enable disabling of Marketing CTA's

omBratteng:
As requested in AS-222, the possibility of disabling a campaign

### DIFF
--- a/src/entity/MarketingCta.ts
+++ b/src/entity/MarketingCta.ts
@@ -23,6 +23,9 @@ export class MarketingCta {
   @Column({ type: 'text' })
   variant: MarketingCtaVariant;
 
+  @Column({ type: 'boolean', default: false })
+  disabled: boolean;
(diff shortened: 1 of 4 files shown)


src/entity/MarketingCta.ts
```diff
@@ -23,6 +23,9 @@ export class MarketingCta {
   @Column({ type: 'text' })
   variant: MarketingCtaVariant;
 
+  @Column({ type: 'boolean', default: false })
+  disabled: boolean;
```
capJavert, line 27: Maybe for future proofing is `status` with enum better? eg. `active | disabled`

omBratteng: Refactored to use enum instead

capJavert (approved):
Nice!

rebelchris (approved):

PR #1845, opened 2024-04-22
fix: search suggestion ordering

rebelchris:
The actual search resolver will order based on position in the idStr.
So we pass the same logic to the suggestion v2 resolver now.

Added to test to notify if this ever changes by accident.


idoshamun (approved):
Perfect!

PR #1846, opened 2024-04-22
fix: remove enum from api

nensidosari:


src/schema/users.ts
```diff
@@ -647,9 +645,7 @@ export const typeDefs = /* GraphQL */ `
     """
     Stores user source tracking information
     """
-    addUserAcquisitionChannel(
-      acquisitionChannel: AcquisitionChannel!
-    ): EmptyResponse @auth
+    addUserAcquisitionChannel(acquisitionChannel: String!): EmptyResponse @auth
```
capJavert, line 648: Can we do `@length(max: 50)` to at least validate partially, someone can basically take this and write really big strings into it. It will still pass our inputs from frontend since they are on the short side.

capJavert (approved):

src/schema/users.ts
nensidosari: It doesn't seem to recognise it - `Unknown directive "@length"`

Maybe I can add a check below as added for other things like: https://github.com/dailydotdev/daily-api/blob/1b7f144e1687e1fc3e90c50f4faa7bc33d05ddf4/src/schema/users.ts#L1359

PR #1848, opened 2024-04-22
fix: get the current date in the user timezone directly from sql

DragosIorgulescu:
- might be more reliable than doing this in javascript


kopancek (approved):

nensidosari (approved):

src/common/users.ts
```diff
@@ -400,15 +400,14 @@ export const checkAndClearUserStreak = async (
...
 
   if (!lastViewAt) {
     return false;
   }
 
-  const today = getTodayTz(timezone);
-  const day = today.getDay();
-  const difference = differenceInDays(today, lastViewAt);
+  const day = currentDate.getDay();
+  const difference = differenceInDays(currentDate, lastViewAt);
```
capJavert, line 410: Tests fail because since you know return date from psql `useFakeTimers` no longer works so you get wrong dates.

Reading slack and this PR I don't think the issue is computing in sql or JS. Basically in both cases you want to get midnight in user timezone?

capJavert (changes requested):

src/common/users.ts
DragosIorgulescu: yeah, I was trying to fix specs, but this is not trivial at all if we get the current time from SQL, which is a pain. 

The reason for this change is due to this part here: https://github.com/dailydotdev/daily-api/blob/264132ab3261b3c78f34a95fa6e64ab9ca28c936/src/common/date.ts#L4 

I am not sure that when we create a new date object based on the one converted to the user timezone, we are not converting it back to UTC again, which then causes issues when we do comparisons.

capJavert: Yeah, dates from database are tricky to mock, I had this issue myself when doing personalized digest. If your onlyy concern is the timezone preservation you can replace `getTodayTz` with following implementation:
```ts
export const getTodayTz = (timeZone: string) => {
  const now = new Date();
  const timeZonedToday = utcToZonedTime(now, timeZone);
  timeZonedToday.setHours(0, 0, 0, 0);

  return timeZonedToday;
};
```

This makes it more clear and I am sure the timezone is preserved.

src/common/users.ts
```diff
@@ -408,7 +408,7 @@ export const checkAndClearUserStreak = async (
 
   const today = getTodayTz(timezone);
   const day = today.getDay();
-  const difference = differenceInDays(today, lastViewAt);
+  const difference = differenceInDays(today.setHours(0, 0, 0, 0), lastViewAt);
```
capJavert, line 411: `setHours` does not return `Date`, it is mutating `today` in place. So better like this?

```suggestion
  today.setHours(0, 0, 0, 0)
  const difference = differenceInDays(today, lastViewAt);
```

src/common/date.ts
```diff
@@ -1,5 +1,6 @@
+import { utcToZonedTime } from 'date-fns-tz';
+
 export const getTodayTz = (timeZone: string) => {
   const now = new Date();
-  const timeZonedToday = now.toLocaleDateString('en', { timeZone });
-  return new Date(timeZonedToday);
+  return utcToZonedTime(now, timeZone);
```
capJavert, line 5: Since this does not really do much now if not setting hours to `0` maybe we can just inline in `src/common/users.ts`?

capJavert (approved):
Just some cleanup, other then that good to go 👍

src/common/users.ts
```diff
@@ -406,7 +406,9 @@ export const checkAndClearUserStreak = async (
     return false;
   }
 
-  const today = getTodayTz(timezone);
+  const today = utcToZonedTime(new Date(), timezone);
```
idoshamun, line 409: Does it really make any difference?

capJavert: I don't think it does, but it does seem more logical to use it instead of our own util.

PR #1849, opened 2024-04-23
fix: add SEO filters to most upvoted and best discussed feed

rebelchris:
@capJavert Please let me know if this works, or you prefer these to be seperate feeds.
I wanted to keep the logic for upvoted/discussed the same just pass either source or tag as variable.

If this works I can add some test as well :)

### DIFF
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -452,6 +452,16 @@ export const typeDefs = /* GraphQL */ `
       Array of supported post types
       """
       supportedTypes: [String!]
+
+      """
+      ID of the source you want to get most upvoted feed for
+      """
+      source: ID
+
+      """
+      Tag to filter by
+      """
+      tag: String
     ): PostConnection!
 
     """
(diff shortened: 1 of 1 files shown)


capJavert:
@rebelchris If we need to reuse some common logic I would rather have it extracted as functions internally but have them separate queries from client perspective. Does that make sense?

rebelchris:
@capJavert Added the test cases :)

__tests__/feeds.ts
```diff
@@ -1196,8 +1196,8 @@ describe('query authorFeed', () => {
 });
 
 describe('query mostUpvotedFeed', () => {
-  const QUERY = (period = 7, first = 10): string => `{
-    mostUpvotedFeed(first: ${first}, period: ${period}) {
+  const QUERY = (period = 7, first = 10, source = '', tag = ''): string => `{
+    mostUpvotedFeed(first: ${first}, period: ${period}, source: "${source}", tag: "${tag}") {
```
capJavert, line 1200: Just to double check existing frontend queries won't break due to missing tag/source in their queries?

capJavert (approved):
Nice!

__tests__/feeds.ts
rebelchris: Screenshot 2024-04-23 at 13 33 42

Yeah works, if it's not set it's just not queried.

PR #1850, opened 2024-04-23
fix: clear boot cache on user assignment to Marketing CTA campaign

omBratteng:
Clears the boot cache when a user is assigned to a campaign.

Important to note, all this logic will most likely be refactored when the customer.io integration is implemented

### DIFF
--- a/src/workers/cdc/primary.ts
+++ b/src/workers/cdc/primary.ts
@@ -773,17 +773,25 @@ const onMarketingCtaChange = async (
 const onUserMarketingCtaChange = async (
   data: ChangeMessage<UserMarketingCta>,
 ) => {
-  // Only run on delete and if the user has read the Marketing CTA
-  if (data.payload.op !== 'd') return;
-  if (data.payload.before.readAt !== null) return;
+  if (data.payload.op === 'c') {
+    await deleteRedisKey(
+      generateStorageKey(
+        StorageTopic.Boot,
+        StorageKey.MarketingCta,
+        data.payload.after.userId,
+      ),
+    );
+  } else if (data.payload.op === 'd') {
+    if (data.payload.before.readAt !== null) return;
 
-  await deleteRedisKey(
-    generateStorageKey(
-      StorageTopic.Boot,
-      StorageKey.MarketingCta,
-      data.payload.before.userId,
-    ),
-  );
+    await deleteRedisKey(
+      generateStorageKey(
+        StorageTopic.Boot,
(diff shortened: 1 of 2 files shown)


src/workers/cdc/primary.ts
```diff
@@ -773,17 +773,25 @@ const onMarketingCtaChange = async (
...
-  );
+  if (data.payload.op === 'c') {
+    await deleteRedisKey(
+      generateStorageKey(
+        StorageTopic.Boot,
+        StorageKey.MarketingCta,
+        data.payload.after.userId,
+      ),
+    );
+  } else if (data.payload.op === 'd') {
```
capJavert, line 784: Boot queries by FIFO so by removing this cache we would not do anything since if there is any other previous user assignment it would just fill with same value?

omBratteng: It would just re-fill with the value it had on the next boot query.

I could've done some checks here to check if the user already had a campaign in cache, but I don't think it's necessary at this moment. Right now we only have the Retool app to assign users, which isn't the end goal.

As I wrote in the body of the PR, lots of this logic will probably be re-factored once customer.io integration is in place, as then we would probably just fill the cache on assignment

capJavert: Why then even invalidate in the first place?

omBratteng: Because right now if you have the SLEEPING value (i.e. not been assigned a campaign), it would not be cleared if you get assigned to a campaign.

I could check if the key value is the `RedisMagicValues.SLEEPING` before I invalidate the cache?

capJavert: It feels more complicated the it should be.

omBratteng: Retool doesn't have access to redis, only the database

capJavert (approved):
To make retool work let's merge but I expect this all to be cleaned up when we do customer.io because now it is too complicated.

src/workers/cdc/primary.ts
omBratteng: Think of this as a temporary fix. Once customer.io integration is implemented, this will be removed on all/most logic be offloaded to webhook call.

capJavert: We can use the agent to expose rpc call https://github.com/dailydotdev/retool-agent, ideally do that when we do customer.io integration as well. https://github.com/dailydotdev/retool-agent

PR #1856, opened 2024-04-25
fix: flags were not parsed as object

rebelchris:
Apparently our flags object is not correctly deep parsed in the messageToJSON.
Decided to manually re-parse it for the worker.

(See example)
Screenshot 2024-04-25 at 15 47 17

### DIFF
--- a/src/workers/postBannedRep.ts
+++ b/src/workers/postBannedRep.ts
@@ -45,18 +45,7 @@ const worker: Worker = {
          * Ensure authors can remove their own post
          * Ensure scouts can't add posts/remove them to decrease author reputation
          */
-        console.log(
-          'authorId',
-          authorId,
-          'scoutId',
-          scoutId,
-          'deletedBy',
-          deletedBy,
-        );
-        if (
-          (authorId && authorId !== deletedBy) ||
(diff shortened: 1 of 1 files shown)


omBratteng (approved):

capJavert (approved):

nensidosari (approved):

src/workers/postBannedRep.ts
```diff
@@ -17,7 +17,10 @@ const worker: Worker = {
   handler: async (message, con, logger): Promise<void> => {
     const data: Data = messageToJson(message);
     const { id, authorId, scoutId, flags } = data.post;
-    const { deletedBy } = flags;
+    const parsedFlags =
+      typeof flags === 'string' ? JSON.parse(flags as string) : flags;
+    const { deletedBy } = parsedFlags;
```
omBratteng, line 22: Oh, I bet because in the tests, we send and actual object to the mock handler, the tests was successful. But in the real word, it gets sent as a string that gets parsed 😬

rebelchris: Yep, wasn't sure if there was some case where a topic could send it as parsed, so this felt safer 😅

PR #1857, opened 2024-04-26
refactor: source entity to add flags property

sshanzel:
Introduce a new `flags` column for `source` entity. Workers and such will be worked on a different PR.

MI-321 #done


src/entity/Source.ts
```diff
@@ -22,6 +22,19 @@ export enum SourceType {
...
+  totalViews: number;
+  totalPosts: number;
+  totalUpvotes: number;
+}
+
+const defaultFlags: Omit<SourceFlags, 'featured'> = {
+  totalPosts: 0,
+  totalViews: 0,
+  totalUpvotes: 0,
+};
```
capJavert, line 36: Flags should be empty by default, so any defaulting to zero should be handled in other API logic or on graphorm side. 

For example having defaultValues and then updating them after some time would require migration to update all previous rows that had different default value. Which is complicated to do.

src/entity/Source.ts
```diff
@@ -91,6 +104,9 @@ export class Source {
   @Index('IDX_source_handle', { unique: true })
   handle: string;
 
+  @Column({ type: 'jsonb', default: defaultFlags })
+  flags: SourceFlags;
```
capJavert, line 108: @sshanzel since this is such a small PR can you just add "real logic" like workers in it as well? Just to make sure we did not miss something here and then require additional migrations in follow up PRs. If you want to chunk it still let's make an epic branch for this series of tasks?

sshanzel: I agree making it be merged all at once would allow us to have a better view on this whole series of changes. Though I think part of the reason why the tasks were chunked into smaller ones is to allow working in parallel. From the JIRA ticket, this blocks 3 items, merging this small PR early would allow us to work on those 3 tickets at the same time.

Yeah, maybe we can create an epic branch here in the API.

src/entity/Source.ts
sshanzel: I actually did that at first but hesitated so I went to the RFC and checked if we specified a default value, found they mentioned 0 so I went ahead and applied it. @DragosIorgulescu thoughts?

src/entity/Source.ts
capJavert: @sshanzel yeah got it, if you can at least hook up graphorm in the PR and at tests for the field and I am fine with merging this as well. You can see example on `flags field` tests in `posts`.

src/entity/Source.ts
```diff
@@ -22,6 +22,19 @@ export enum SourceType {
   Squad = 'squad',
 }
 
+interface SourceFlags {
+  featured: boolean;
+  totalViews: number;
+  totalPosts: number;
+  totalUpvotes: number;
+}
```
capJavert, line 30: Also please follow `SourceFlagsPublic` nomenclature, you can see example for  `UserPostEntity`. Basically public flags are those that can be requested/updated from graphql.

src/entity/Source.ts
sshanzel: Yeah, was thinking that would allow us to unblock FE work too. Sure, let me update those items.

src/entity/Source.ts
rebelchris: It can be solved on graphORM side not db side basically.

src/migration/1714116154573-SourceFlags.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceFlags1714116154573 implements MigrationInterface {
+  name = 'SourceFlags1714116154573';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "source" ADD "flags" jsonb NOT NULL DEFAULT '{"totalPosts":0,"totalViews":0,"totalUpvotes":0}'`,
```
rebelchris, line 8: We should remove the defaults before merging :)

src/entity/Source.ts
sshanzel: Yeah, I understand, it is just on the RFC, the table was dedicated to the entity, not the gql schema. In any case, I will revert to empty curly brackets similar to our other implementations of flags.

rebelchris (approved):
Looks good to me, just wait for ante 👍

src/schema/sources.ts
```diff
@@ -1622,5 +1645,11 @@ export const resolvers: IResolvers<any, Context> = {
 
       return referralUrl;
     },
+    flags: ({ flags }: GQLSource): SourceFlagsPublic => ({
+      totalPosts: flags?.totalPosts ?? 0,
+      totalViews: flags?.totalViews ?? 0,
+      totalUpvotes: flags?.totalUpvotes ?? 0,
+      featured: isNullOrUndefined(flags?.featured) ? false : flags.featured,
```
capJavert, line 1652: Also I think better to do it on graphorm level because it will be available directly in queries vs only after query resolves. Basically under `jsonType: true` with `transform`.

src/graphorm/index.ts
```diff
@@ -224,6 +224,9 @@ const obj = new GraphORM({
   Source: {
     requiredColumns: ['id', 'private', 'handle', 'type'],
     fields: {
+      flags: {
+        jsonType: true,
```
capJavert, line 228:
```suggestion
        jsonType: true,
        transform: (flags: Partial<SourceFlagsPublic>) => {
          return {
            totalPosts: flags?.totalPosts ?? 0,
            totalViews: flags?.totalViews ?? 0,
            totalUpvotes: flags?.totalUpvotes ?? 0,
            featured: flags?.featured ?? false,
          };
        },
```

capJavert: Also if you can add tests, check here for example 🙏 https://github.com/dailydotdev/daily-api/blob/main/__tests__/sources.ts#L2795-L2834

capJavert (approved):
Looks good, just some minor adjustments.

src/schema/sources.ts
capJavert: https://github.com/dailydotdev/daily-api/pull/1857#discussion_r1581177349

sshanzel: Yeah, that makes sense, we will have more control and avoid the null value on certain situations.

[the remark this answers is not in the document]
sshanzel: Introduced a function to determine the value so it will be consistent.

src/graphorm/index.ts
sshanzel: I introduced a new function as I realized the properties can grow and the scenario can easily happen in the future. That is to make the process consistent.

__tests__/commons/object.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect(result).toEqual(defaultPublicSourceFlags);
+  });
+
+  it('should return the first param if default value is null or undefined', () => {
+    const result = setDefaultsIfNull(defaultPublicSourceFlags, null);
+
+    expect(result).toEqual(defaultPublicSourceFlags);
+  });
+
+  it('should set the value if property is null or undefined', () => {
```
nensidosari, line 17:
```suggestion
  it('should set the default value if property is null or undefined', () => {
```

__tests__/commons/object.ts
```diff
@@ -0,0 +1,44 @@
...
+    const result = setDefaultsIfNull(
+      { ...defaultPublicSourceFlags, totalPosts: null },
+      defaultPublicSourceFlags,
+    );
+
+    expect(result.totalPosts).toEqual(0);
+  });
+
+  it('should NOT set the value if property is NOT null or undefined', () => {
```
nensidosari, line 26:
```suggestion
  it('should NOT set the default value if property is NOT null or undefined', () => {
```

nensidosari (approved):
Looks good on my side as well

DragosIorgulescu (approved):
nothing to add 🙌

src/common/object.ts
```diff
@@ -40,3 +40,25 @@ export const nameRegex = new RegExp(/^(.){1,60}$/);
...
+    const value = obj[key];
+
+    if (isNullOrUndefined(value)) {
+      return result;
+    }
+
+    result[key] = value;
+
+    return result;
+  }, defaults);
```
capJavert, line 63: While I get what you are trying to do this seems too complicated in my opinion. 
- it is doing iteration (over keys) which just adds more computation on field resolver using it
- what if someone introduces nested object, this would not be able to resolve it? 

I personally much more prefer just doing it inline in gql as you did before https://github.com/dailydotdev/daily-api/pull/1857/commits/e31fd547bf5bc65c2e4e4b152a8e6295860d2e5c#diff-95123b05e21b34d3705f1b3748fe295fa17db5caeeb22238a294368729175a07R1649-R1652

capJavert: @rebelchris @DragosIorgulescu tagging for your opinion as well.

sshanzel: Yeah, I just realized if we add another property, we might miss another `obj.key ?? value` to put since everything is inline. For now, yeah, this won't cover a nested object, but that can definitely add that processing once we get the need for it.

capJavert: > Yeah, I just realized if we add another property, we might miss another obj.key ?? value 

Yeah, figured, did you look into option to take care of this with Typescript instead by defining return type of field resolver, which would be caught in lint/test before any deployment. So it would make sure additional fields do not miss handling without introducing runtime cost.

sshanzel: Yeah, I checked, and it could be possible, just needs more work on the GraphORM side since it doesn't automatically understand the type it needs to return.

capJavert: I tested with:
```ts
transform: (value: SourceFlagsPublic): SourceFlagsPublic => {
  return {
    totalPosts: value?.totalPosts ?? 0,
    totalViews: value?.totalViews ?? 0,
    totalUpvotes: value?.totalUpvotes ?? 0,
    featured: value?.featured ?? false,
  };
}
```

and then adding new field `test` to `SourceFlagsPublic`

I automatically get the error:
image

Which would block any pipeline and force developer to add handling in graphorm.

sshanzel: Right. Yeah, we can probably do that instead of adding generic types to GraphORM.

sshanzel: Ideally, we can do generic types for better devex and avoid forgetting to add that type on the function since our lint does not ask for it.

capJavert: Yeah, problem partially is that we don't have [strict option in typescript](https://dailydotdev.atlassian.net/browse/WT-1532), it would light up red all over the place even more during dev.

PR #1858, opened 2024-04-29
infra: deploy ClickhouseSync

omBratteng:
Superseeds #1835

Related PR in https://github.com/dailydotdev/pulumi-common/pull/308


.infra/clickhouse-sync.yml
```diff
@@ -0,0 +1,131 @@
...
+#binary.handling.mode: "base64"
+
+# ignore_delete: If set to true, the connector will ignore delete events. The default is false.
+#ignore_delete: "true"
+
+#disable.ddl: If set to true, the connector will ignore DDL events. The default is false.
+#disable.ddl: "false"
+
+#disable.drop.truncate: If set to true, the connector will ignore drop and truncate events. The default is false.
+#disable.drop.truncate: "false"
```
rebelchris, line 131: These you still need here?

.infra/index.ts
```diff
@@ -522,3 +524,48 @@ if (vpcNativeProvider) {
...
+            'clickhouseSync',
+          ).vars,
+        },
+      },
+      image: {
+        repository: 'altinity/clickhouse-sink-connector',
+        tag: '2.0.2-lt',
+      },
+      resources: {
+        // TODO: adjust resources based on the actual usage
```
rebelchris, line 558: This still needed?

rebelchris (approved):

.infra/index.ts
```diff
@@ -522,3 +524,48 @@ if (vpcNativeProvider) {
...
+        },
+        vars: {
+          ...config.requireObject<{ vars: Record<string, string> }>(
+            'clickhouseSync',
+          ).vars,
+        },
+      },
+      image: {
+        repository: 'altinity/clickhouse-sink-connector',
+        tag: '2.0.2-lt',
```
omBratteng, line 555:
```suggestion
        digest:
          'sha256:1c0db9877331a0aaceb2e0f2838db3a5156a3cfbfdb83a92235b6287b576c5d0',
```
We can also use the digest if we want to be sure that the tag isn't force updated

.infra/index.ts
omBratteng: Yeah, will need to deploy and see actual real usage first

.infra/clickhouse-sync.yml
omBratteng: We could come back once deployed and all good and cleanup commented out config lines?

.infra/index.ts
omBratteng: What do you think @vpol? You used a forked version (with no changes?) in your PR #1835

vpol: I use custom build pipeline to build our own images from git repo. so IF and WHEN I need to deploy patched version, I can do it without pain. I do it for other projects as well. we can use our gcr image that I've built and tag.

.infra/index.ts
vpol: it's basically debezium under the hood with some steroids, but with only one tiny table to monitor, so we can allocate roughly 1/3 of the resources we give debezium currently.

.infra/clickhouse-sync.yml
vpol: can we 🙏 keep config readable? just to avoid pain of finding it again if needed.

.infra/index.ts
omBratteng: Yup. My plan is to come back to this once deployed for a few days/weeks to see real usage

.infra/index.ts
omBratteng: I changed it to use our gcr image. Seeing as we might need to do some patching

.infra/clickhouse-sync.yml
omBratteng: I don't mind either way

PR #1862, opened 2024-04-30
chore: reduce redis instance size

omBratteng:
Last week I deleted ~4 GiB worth of old and untouched Redis keys. At the time we were using 5 out of 10 GiB of memory, which had been stable for months. Now we use 1.3 out of 10 GiB, so we can reduce it to 2 GiB and disable read-replica as we're not using it.

image

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -63,7 +63,7 @@ if (isAdhocEnv) {
 const redis = new Redis(`${name}-redis`, {
   isAdhocEnv,
   name: `${name}-redis`,
-  tier: 'STANDARD_HA',
+  tier: 'BASIC',
   memorySizeGb: 2,
   region: location,
   authEnabled: true,
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -64,7 +64,7 @@ const redis = new Redis(`${name}-redis`, {
   isAdhocEnv,
   name: `${name}-redis`,
   tier: 'STANDARD_HA',
-  memorySizeGb: 10,
+  memorySizeGb: 2,
   region: location,
   authEnabled: true,
   redisVersion: 'REDIS_6_X',
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -64,7 +64,7 @@ const redis = new Redis(`${name}-redis`, {
   isAdhocEnv,
   name: `${name}-redis`,
   tier: 'BASIC',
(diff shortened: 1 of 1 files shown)


.infra/index.ts
```diff
@@ -63,8 +63,8 @@ if (isAdhocEnv) {
 const redis = new Redis(`${name}-redis`, {
   isAdhocEnv,
   name: `${name}-redis`,
-  tier: 'STANDARD_HA',
-  memorySizeGb: 10,
+  tier: 'BASIC',
+  memorySizeGb: 2,
```
capJavert, line 67: What happens if we go over?

omBratteng: 🤷‍♂️

It'll probably not let us write any more

capJavert: Maybe put a bit more breathing space, 3Gb?

idoshamun: It starts evicting keys depending on our policy

capJavert: Looking at the code I think marketing CTA is the only one that could bump the usage drastically if we start using it due to a lot of potential redis keys.

.infra/index.ts
```diff
@@ -63,8 +63,8 @@ if (isAdhocEnv) {
 const redis = new Redis(`${name}-redis`, {
   isAdhocEnv,
   name: `${name}-redis`,
-  tier: 'STANDARD_HA',
-  memorySizeGb: 10,
+  tier: 'BASIC',
```
idoshamun, line 66: It may fail to downgrade so let's keep an eye

idoshamun (approved):

.infra/index.ts
omBratteng: Yeah, it might. It could probably go through via click-ops if pulumi fails.
I'd also like to do it on saturday/sunday morning to reduce any chances of it evicting all keys.

capJavert (approved):

.infra/index.ts
omBratteng: We can easily scale up if we see marketing CTA starts to eat more memory.

rebelchris (approved):

.infra/index.ts
omBratteng, line 67:
```suggestion
  memorySizeGb: 3,
```
We settled at 3, cc: @idoshamun

PR #1865, opened 2024-04-30
feat: set watercooler to be opt-in

idoshamun:
Make watercooler as opt-in even though it's a public feed. Watercooler mostly has memes which not all the users want, surfacing it as default to all users seems to be controversial.

I also fixed a bug that even though "show on my feed" was set to false there could be an edge case where we show the posts of the squad. Unrelated to the original task but I noticed it as I go.

### DIFF
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -155,19 +155,16 @@ export const feedToFilters = async (
 
   // Split memberships by hide flag
   const membershipsByHide = memberships.reduce(
-    (
-      acc: { true: string[]; false: string[] },
-      value: { sourceId: SourceMember['sourceId']; hide: boolean },
-    ) => {
-      acc[value.hide ? 'true' : 'false'].push(value.sourceId);
+    (acc, value: { sourceId: SourceMember['sourceId']; hide: boolean }) => {
+      acc[value.hide ? 'hide' : 'show'].push(value.sourceId);
       return acc;
     },
-    { true: [], false: [] },
+    { hide: [], show: [] },
   );
 
   // If the user is not a member of the watercooler, hide it
-  if (!membershipsByHide.false.includes(WATERCOOLER_ID)) {
-    membershipsByHide.true.push(WATERCOOLER_ID);
+  if (!membershipsByHide.show.includes(WATERCOOLER_ID)) {
+    membershipsByHide.hide.push(WATERCOOLER_ID);
   }
 
   return {
@@ -175,8 +172,8 @@ export const feedToFilters = async (
     excludeTypes,
     excludeSources: excludeSources
       .map((sources: Source) => sources.id)
-      .concat(membershipsByHide.true),
-    sourceIds: membershipsByHide.false,
+      .concat(membershipsByHide.hide),
+    sourceIds: membershipsByHide.show,
   };
 };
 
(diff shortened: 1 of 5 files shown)


src/common/feedGenerator.ts
```diff
@@ -150,13 +153,30 @@ export const feedToFilters = async (
...
+      value: { sourceId: SourceMember['sourceId']; hide: boolean },
+    ) => {
+      acc[value.hide ? 'true' : 'false'].push(value.sourceId);
+      return acc;
+    },
+    { true: [], false: [] },
+  );
+
+  // If the user is not a member of the watercooler, hide it
+  if (!membershipsByHide.false.includes(WATERCOOLER_ID)) {
```
rebelchris, line 169: This had me reading this section 4 times 🤣

rebelchris (approved):
Looks alright to me

src/common/feedGenerator.ts
```diff
@@ -150,13 +153,30 @@ export const feedToFilters = async (
...
+  // Split memberships by hide flag
+  const membershipsByHide = memberships.reduce(
+    (
+      acc: { true: string[]; false: string[] },
+      value: { sourceId: SourceMember['sourceId']; hide: boolean },
+    ) => {
+      acc[value.hide ? 'true' : 'false'].push(value.sourceId);
+      return acc;
+    },
+    { true: [], false: [] },
```
capJavert, line 165: Can we use something that is not `false | true` since they are boolean keyword tokens in JavaScript so it will look wonky. Maybe `hide | show`

src/common/feedGenerator.ts
```diff
@@ -125,10 +127,11 @@ export const feedToFilters = async (
           .getRepository(SourceMember)
           .createQueryBuilder('sm')
           .select('sm."sourceId"')
-          .where('sm."userId" = :userId', { userId })
-          .andWhere(
-            `COALESCE((flags->'hideFeedPosts')::boolean, FALSE) = FALSE`,
```
capJavert, line 130: So now we pull all source member rows for each user and then manually hide depending on the flag in below iteration?

src/common/feedGenerator.ts
idoshamun: lol this is why i added the comment, show i change true/false to show/hide. will make it more explicit

capJavert (approved):

src/common/feedGenerator.ts
idoshamun: yes. we need to block them, not only remove them from the list

idoshamun:
> What was exact fix for this? Just interested where the mishap happened?

Feed service can "randomly" surface public squad posts even if you are no subscribed. So if API doesn't provide a squad id in the squad_ids list, the feed service may still surface some posts. Providing it as blocked will force feed service to completely ignore the squad

src/common/feedGenerator.ts
idoshamun: Done 👌

src/common/feedGenerator.ts
capJavert: Yeah, figured from here https://github.com/dailydotdev/daily-api/pull/1865#issuecomment-2085038766, tnx for explanation!

PR #1866, opened 2024-04-30
refactor: squad posting to increment squad total posts

sshanzel:
As the RFC has mentioned, we should update the mutation for creating new posts related to Squad.

There was no mention of deleting posts, but I assumed we should reduce the count by 1. Feel free to mention if we should not make any changes to it.

Notable changes here:
- Created a generic function for saving changes to posts, this is to allow us to streamline the updating of flags (maybe we could have done the increment on the new post worker, but I might have missed the discussion around it).
- Created a generic function for deleting a post, to do the same above.

MI-322 #done

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -3630,8 +3630,7 @@ describe('mutation votePost', () => {
     );
   });
 
-  it('should upvote', async () => {
-    loggedUser = '1';
+  const testUpvote = async () => {
     await con.getRepository(Post).save({
       id: 'p1',
       upvotes: 3,
@@ -3652,6 +3651,32 @@ describe('mutation votePost', () => {
     });
     const post = await con.getRepository(Post).findOneBy({ id: 'p1' });
     expect(post?.upvotes).toEqual(4);
+  };
+
+  it('should upvote', async () => {
+    loggedUser = '1';
+    const source = await con.getRepository(Source).findOneByOrFail({ id: 'a' });
+    expect(source.flags.totalUpvotes).toEqual(undefined);
+
+    await testUpvote();
+
+    const updatedSource = await con
+      .getRepository(Source)
+      .findOneByOrFail({ id: 'a' });
+    expect(updatedSource.flags.totalUpvotes).toEqual(undefined);
+  });
+
+  it('should upvote and increment squad total upvotes', async () => {
+    loggedUser = '1';
+    const repo = con.getRepository(Source);
+    await repo.update({ id: 'a' }, { type: SourceType.Squad });
+    const source = await repo.findOneByOrFail({ id: 'a' });
+    expect(source.flags.totalUpvotes).toEqual(undefined);
+
+    await testUpvote();
+
+    const updatedSource = await repo.findOneByOrFail({ id: 'a' });
+    expect(updatedSource.flags.totalUpvotes).toEqual(1);
   });
 
   it('should downvote', async () => {
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -2036,7 +2036,6 @@ describe('mutation viewPost', () => {
   });
 
   it('should submit view event', async () => {
-    loggedUser = '1';
     await con
       .getRepository(Post)
       .update({ id: 'p1' }, { type: PostType.Share });
@@ -2050,6 +2049,31 @@ describe('mutation viewPost', () => {
     const res = await client.mutate(MUTATION, { variables });
     expect(res.errors).toBeFalsy();
     expect(notifyView).toBeCalledTimes(0);
+  });
+});
+
+describe('trigger increment_squad_views_count', () => {
+  it('should NOT update source total views', async () => {
+    const repo = con.getRepository(Source);
+    const source = await repo.findOneByOrFail({ id: 'a' });
+    expect(source.flags.totalViews).toEqual(undefined);
+
+    await con.getRepository(Post).update({ id: 'p1' }, { views: 1 });
+
+    const updatedSource = await repo.findOneByOrFail({ id: 'a' });
+    expect(updatedSource.flags.totalViews).toEqual(undefined);
+  });
+
+  it('should update squad total views', async () => {
+    const repo = con.getRepository(Source);
+    await repo.update({ id: 'a' }, { type: SourceType.Squad });
+    const source = await repo.findOneByOrFail({ id: 'a' });
+    expect(source.flags.totalViews).toEqual(undefined);
+
+    await con.getRepository(Post).update({ id: 'p1' }, { views: 1 });
+
+    const updatedSource = await repo.findOneByOrFail({ id: 'a' });
+    expect(updatedSource.flags.totalViews).toEqual(1);
   });
 });
 
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -3797,7 +3797,7 @@ describe('mutation votePost', () => {
 
     await testDownvote();
 
-    // should not be affected since this is not a squad
+    // should not be affected since the existing vote was a downvote
     const updatedSource = await con
       .getRepository(Source)
       .findOneByOrFail({ id: 'a' });
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -1225,6 +1225,9 @@ describe('mutation deletePost', () => {
       sharedPostId: 'p1',
       authorId,
     });
+    await con
+      .getRepository(Source)
+      .update({ id: 'a' }, { flags: { totalPosts: 1 } });
   };
 
   it('should not authorize when not logged in', () =>
@@ -1289,6 +1292,19 @@ describe('mutation deletePost', () => {
     const id = 'sp1';
     await createSharedPost(id);
     await verifyPostDeleted(id, loggedUser);
+  });
+
+  it('should allow member to delete their own shared post and reduce squads flags total posts', async () => {
+    loggedUser = '2';
+    const id = 'sp1';
+    await createSharedPost(id);
+    const source = await con.getRepository(Source).findOneBy({ id: 'a' });
+    expect(source.flags.totalPosts).toEqual(1);
+    await verifyPostDeleted(id, loggedUser);
+    const updatedSource = await con
+      .getRepository(Source)
+      .findOneBy({ id: 'a' });
+    expect(updatedSource.flags.totalPosts).toEqual(0);
   });
 
   const verifyPostDeleted = async (id: string, user: string) => {
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -715,6 +715,7 @@ describe('welcomePost type', () => {
 
   it('should add welcome post and increment squad total posts', async () => {
     const repo = con.getRepository(Source);
+    await repo.update({ id: 'a' }, { type: SourceType.Squad });
     const source = await repo.findOneBy({ id: 'a' });
     const post = await createSquadWelcomePost(con, source, '1');
     expect(post.showOnFeed).toEqual(false);
@@ -1297,8 +1298,10 @@ describe('mutation deletePost', () => {
   it('should allow member to delete their own shared post and reduce squads flags total posts', async () => {
     loggedUser = '2';
     const id = 'sp1';
+    const repo = con.getRepository(Source);
+    await repo.update({ id: 'a' }, { type: SourceType.Squad });
     await createSharedPost(id);
-    const source = await con.getRepository(Source).findOneBy({ id: 'a' });
+    const source = await repo.findOneBy({ id: 'a' });
     expect(source.flags.totalPosts).toEqual(1);
     await verifyPostDeleted(id, loggedUser);
     const updatedSource = await con
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -2078,6 +2078,52 @@ describe('trigger increment_squad_views_count', () => {
   });
 });
 
+describe('trigger deduct_squad_stats', () => {
+  beforeEach(async () => {
+    const repo = con.getRepository(Source);
+    await repo.update({ id: 'a' }, { type: SourceType.Squad });
+    const source = await repo.findOneByOrFail({ id: 'a' });
+    await createSquadWelcomePost(con, source, '1');
+  });
+
+  it('should decrement total posts', async () => {
+    const repo = con.getRepository(Source);
+    const source1 = await repo.findOneByOrFail({ id: 'a' });
+    expect(source1.flags.totalPosts).toEqual(1);
+    await con.getRepository(Post).update({ id: 'p1' }, { deleted: true });
+    const source2 = await repo.findOneByOrFail({ id: 'a' });
+    expect(source2.flags.totalPosts).toEqual(0);
+  });
+
+  it('should deduce deleted posts views to squads total views', async () => {
+    const repo = con.getRepository(Source);
+    const source1 = await repo.findOneByOrFail({ id: 'a' });
+    expect(source1.flags.totalViews).toEqual(undefined);
+    await con.getRepository(Post).update({ id: 'p1' }, { views: 1 });
+    await con.getRepository(Post).update({ id: 'p1' }, { views: 2 });
+    await con.getRepository(Post).update({ id: 'p4' }, { views: 1 });
+    const source2 = await repo.findOneByOrFail({ id: 'a' });
+    expect(source2.flags.totalViews).toEqual(3);
+
+    await con.getRepository(Post).update({ id: 'p1' }, { deleted: true });
+
+    const source3 = await repo.findOneByOrFail({ id: 'a' });
+    expect(source3.flags.totalViews).toEqual(1);
+  });
+
+  it('should update squad total views', async () => {
+    const repo = con.getRepository(Source);
+    await repo.update({ id: 'a' }, { type: SourceType.Squad });
+    const source = await repo.findOneByOrFail({ id: 'a' });
+    expect(source.flags.totalViews).toEqual(undefined);
+
+    await con.getRepository(Post).update({ id: 'p1' }, { views: 1 });
+
+    const updatedSource = await repo.findOneByOrFail({ id: 'a' });
+    expect(updatedSource.flags.totalViews).toEqual(1);
+  });
+});
+
 describe('mutation submitExternalLink', () => {
   const MUTATION = `
   mutation SubmitExternalLink($sourceId: ID!, $url: String!, $commentary: String, $title: String, $image: String) {
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -3685,9 +3685,7 @@ describe('mutation votePost', () => {
     await testUpvote();
 
     const updatedSource = await repo.findOneByOrFail({ id: 'a' });
-    // the reason this will become 2 is we touched the post.upvotes twice.
-    // first was through the manual update, second was through the mutation
-    expect(updatedSource.flags.totalUpvotes).toEqual(2);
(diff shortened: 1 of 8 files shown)


src/entity/posts/utils.ts
```diff
@@ -300,24 +372,29 @@ export const createSharePost = async (
...
-      private: privacy,
-      origin: PostOrigin.UserGenerated,
-      visible,
-      visibleAt: visible ? new Date() : null,
-      flags: {
+    const post = await savePost<SharePost>({
+      con,
+      type: SharePost,
+      shouldUpdateFlags: true,
+      post: {
```
sshanzel, line 379: Nothing was changed with the object we passed as `post`, just indentation due to being inside an object.

src/common/post.ts
```diff
@@ -126,33 +126,37 @@ export const createSquadWelcomePost = async (
...
-    content,
-    contentHtml: markdown.render(content),
-    image: defaultImage.welcomePost,
-    banned: true,
-    flags: {
+  return savePost({
+    con,
+    type: WelcomePost,
+    shouldUpdateFlags: true,
+    post: {
```
sshanzel, line 135: Nothing was changed with the object we passed as `post`, just indentation due to being inside an object.

src/entity/posts/utils.ts
```diff
@@ -36,6 +39,75 @@ type StringPostStats = {
...
+
+  if (shouldUpdateFlags) {
+    const source = await con
+      .getRepository(Source)
+      .findOneBy({ id: res.sourceId });
+
+    const currentTotalPosts = source.flags?.totalPosts || 0;
+    const flags = updateFlagsStatement({
+      totalPosts: currentTotalPosts + 1,
+    });
```
capJavert, line 65: Might be out of loop but did we not say we are going to do this with triggers? This way counts will be updated no matter from where the post is deleted (API, script, psql etc.) we do similar for `upvotes` and `downvotes` for posts.

FYI @DragosIorgulescu

sshanzel: Yeah, DB triggers are a good option as well. Looking forward to the reply as the ticket mentioned the mutation 👀

sshanzel: I just saw the note about DB triggers being an option. I can maybe try and see how we can incorporate it. However, it only mentions the `totalPosts` to be part of it and the others would be within a cron job.

sshanzel: The RFC highlights updating the mutations, we can maybe reopen a ticket if we ever decide to finalize using DB triggers. @capJavert feel free to make the final call, I am cool either way and willing to make the changes if necessary.

sshanzel:
@capJavert I pushed some changes towards using triggers. Would love to hear your initial thoughts. This is only for total posts count. I will be adding the triggers for the other 2 properties. Do you maybe want them in a separate migration file or I can keep them all in the same one?

src/migration/1714633817175-SquadTotalPostsTrigger.ts
```diff
@@ -0,0 +1,67 @@
...
+    queryRunner.query(`
+        CREATE OR REPLACE FUNCTION increment_squad_posts_count()
+          RETURNS TRIGGER
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        DECLARE
+          s public.source;
+        BEGIN
+          SELECT * INTO s FROM public.source WHERE id = NEW."sourceId";
```
capJavert, line 16: I think better to just do `AND s.type = 'squad'` in UPDATE statement, that will avoid select query in each trigger execution if I see correctly?

src/migration/1714633817175-SquadTotalPostsTrigger.ts
```diff
@@ -0,0 +1,67 @@
...
+          IF s.type = 'squad' THEN
+            UPDATE source
+            SET flags = jsonb_set(flags, array['totalPosts'], to_jsonb(COALESCE(CAST(flags->>'totalPosts' AS INTEGER), 0) + 1))
```
capJavert, line 19: Never seen this `array['totalPosts']` syntax. It actually returns `{totalPosts}` which is what `jsonb_set` requires. I would prefer to just use string, seems more simple if there is not anything else I am missing?

src/migration/1714633817175-SquadTotalPostsTrigger.ts
capJavert: UPDATE that does not modify any rows should be faster then SELECT + OPTIONALLY UPDATE.

capJavert: Actually we can also use `WHEN (NEW.type = 'squad'` in trigger definition, that should be the best.

capJavert:
@sshanzel 

>  I will be adding the triggers for the other 2 properties. Do you maybe want them in a separate migration file or I can keep them all in the same one?

Whatever you prefer. 👍

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: But the type here is the post, so it will never be a `squad`.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: The `array['totalPosts']` just looks more object-oriented and modern, I have no hard preference, so either way works for me.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
capJavert: Yeah but you are doing UPDATE query on `source` where you can do a `WHERE` filter on its type.

sshanzel: Oh, right. Forgot about that. Yes, we can do that.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
```diff
@@ -0,0 +1,59 @@
...
+    queryRunner.query(`
+        CREATE OR REPLACE FUNCTION decrement_squad_posts_count()
+          RETURNS TRIGGER
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          IF OLD.deleted = false AND NEW.deleted = true THEN
+            UPDATE  source
+            SET     flags = jsonb_set(flags, '{totalPosts}', to_jsonb(CAST(flags->>'totalPosts' AS INTEGER) - 1))
```
capJavert, line 35:
```suggestion
            SET     flags = jsonb_set(flags, '{totalPosts}', to_jsonb(GREATEST(0, COALESCE(CAST(flags->>'totalPosts' AS INTEGER), 0) - 1)))
```

Just to make sure it handles no value and not go into negative with `GREATEST`.

capJavert (approved):
One edge case, but good to go after it, nice work!

kopancek (approved):
I like the trigger implementation, agree with Ante on safeguarding the delete though.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: Yep. Makes sense to catch the edge-case.

sshanzel:
@capJavert @kopancek @nensidosari all triggers to update the stats (totalViews, totalPosts, totalUpvotes) should now be available.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
```diff
@@ -0,0 +1,65 @@
...
+        CREATE OR REPLACE FUNCTION increment_squad_posts_count()
+          RETURNS TRIGGER
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          UPDATE  source
+          SET     flags = jsonb_set(flags, '{totalPosts}', to_jsonb(COALESCE(CAST(flags->>'totalPosts' AS INTEGER) , 0) + 1))
+          WHERE   id = NEW."sourceId"
+          AND     type = 'squad';
```
rebelchris, line 17: We decided on this just for performance? (Just meant maybe not a bad thing to have for all sources, but also not needed now)

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
```diff
@@ -0,0 +1,47 @@
...
+                            )
+            WHERE   id = NEW."sourceId"
+            AND     type = 'squad';
+          END IF;
+          RETURN NEW;
+        END;
+        $$
+      `);
+    queryRunner.query(
+      `CREATE TRIGGER update_squad_upvotes_count AFTER UPDATE ON "post" FOR EACH ROW EXECUTE PROCEDURE update_squad_upvotes_count()`,
```
rebelchris, line 37: More of a question @capJavert is there no way we can only attach trigger to set column?

rebelchris: Alternative, we listen to the same thing here, wondering if it's "cheaper" to simply just do all with 1 trigger?

rebelchris (approved):
Looks pretty good, just some questions none blocking.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
```diff
@@ -0,0 +1,65 @@
...
+        CREATE OR REPLACE FUNCTION deduct_squad_stats()
+          RETURNS TRIGGER
+          LANGUAGE PLPGSQL
+          AS
+        $$
+        BEGIN
+          IF OLD.deleted = false AND NEW.deleted = true THEN
+            UPDATE  source
+            SET     flags = jsonb_set(
+                              jsonb_set(
```
capJavert, line 36: Why two `jsonb_set`? Should it not the second one be separate trigger? That is how the rest were done.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
```diff
@@ -0,0 +1,65 @@
...
+                                flags,
+                                '{totalPosts}',
+                                to_jsonb(GREATEST(0, CAST(flags->>'totalPosts' AS INTEGER) - 1))
+                              ),
+                              '{totalViews}',
+                              to_jsonb(GREATEST(0, CAST(flags->>'totalViews' AS INTEGER) - OLD.views))
```
capJavert, line 42: Also not sure if we need to decrement views when post is deleted, views are still there and happened, no matter if post is deleted.

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
capJavert: Hmm I think each set of triggers should listen on the source table:
- totalUpvotes triggers to `user_post` table
- totalPosts triggers `post` table (as it is now)
- totalView triggers on `view` table

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
```diff
@@ -0,0 +1,47 @@
...
+          IF NEW.upvotes <> OLD.upvotes THEN
+            UPDATE  source
+            SET     flags = jsonb_set(
+                              flags,
+                              '{totalUpvotes}',
+                              to_jsonb(
+                                GREATEST(
+                                  0,
+                                  COALESCE(CAST(flags->>'totalUpvotes' AS INTEGER), 0) +
+                                  (CASE WHEN NEW.upvotes > OLD.upvotes THEN 1 ELSE -1 END)
```
capJavert, line 25: But you don't know if upvotes for how much the upvotes changed. For example upvotes could change from 8 to 10. You will increment by 1, when in reality it should add two upvotes.

capJavert: See my comment https://github.com/dailydotdev/daily-api/pull/1866#discussion_r1588928823

src/migration/1714633817175-SquadTotalPostsTrigger.ts
capJavert: Sorry, asked in discussion, did not see it before https://dailydotdev.slack.com/archives/C069XNYMDGV/p1714669514654279

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
capJavert: As to listening to specific table that might not be possible due to https://dailydotdev.slack.com/archives/C069XNYMDGV/p1714669514654279

rebelchris: Disregard my comment on combining and column listening :)

sshanzel: Yeah, it would have been ideal to listen to each those tables. The only downside is that we have to fetch the source id from the parent entity which is the post (for view and user_post).

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: This is to set two properties of the same flags property.

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
sshanzel: Would it? Do we not send an upvote one at a time per user and they will run on their own even if two users hit the upvote at the same time?

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
sshanzel: If we are cool doing for that, I'd be glad to attach the triggers closest to where the action is being recorded.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
capJavert: Yeah sure, I just think move it to separate trigger I think so only field is updated in each trigger, as you did for others.

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
capJavert: Yeah, lets close this discussion and do it on post as you did, and in regards to slack thread.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
capJavert: Resolved in https://github.com/dailydotdev/daily-api/pull/1866#discussion_r1589028400

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
capJavert: Most of the time yes, and if you listened to `user_post` that would be the case since each new row means upvote count changed by at most 1 (+ or -). But since you are looking at accumulator on post you have to take it into account.

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
capJavert, line 37: I am missing the trigger that on post soft deletion you decrement upvotes from squad, right?

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
capJavert: Between I think in this trigger you can actually use `WHEN NEW.upvotes <> OLD.upvotes` to trigger only when upvotes count change.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
capJavert: Also @sshanzel what happens if post gets un-deleted. I am fine in not covering that case because it was not mentioned in RFC as well but we have to make a note of it somewhere.

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
sshanzel: Yep. Since we plan on separating the triggers for each properties, I will also add a separate one for this.

src/migration/1714662136871-SquadTotalUpvotesTrigger.ts
sshanzel: Oh, nice. Didn't know you can actually specify that.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: Yeah, we can probably do that. That would just mean we will recalculate the stats for all sources on launch to have its merits.

capJavert: Does RFC say anything about this? If not let's keep it as is and we can see next week when Drgaos is back where do we want to go. I don't have strong opinions if we want to have it on all sources. I would expect this kind of aggregations could be done by clickhouse in the future so that is the reason against calculating now for every source.

rebelchris: Yeah would indeed roll out like this for now.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: Yeah, sounds good to me. That was also the reason I asked it, I wanted the deduction to not happen. Let's do it.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: We are not doing any deduction on the stats when a post gets deleted. Should make our lives easier on that edge case.

src/migration/1714633817175-SquadTotalPostsTrigger.ts
sshanzel: Awesome. Thank you both.

__tests__/posts.ts
```diff
@@ -2003,6 +2038,31 @@ describe('mutation viewPost', () => {
   });
 });
 
+describe('trigger increment_squad_views_count', () => {
```
capJavert, line 2041: We have triggers in separate test file `__tests__/triggers/post.ts`, sorry for missing this in earlier review.

__tests__/posts.ts
```diff
@@ -3554,20 +3709,44 @@ describe('mutation votePost', () => {
...
+    await testDownvote();
+
+    // should not be affected since this is not a squad
+    const updatedSource = await con
+      .getRepository(Source)
+      .findOneByOrFail({ id: 'a' });
+    expect(updatedSource.flags.totalUpvotes).toEqual(undefined);
   });
 
-  it('should cancel vote', async () => {
```
capJavert, line 3559: Can we keep this test as is and then just add new tests after. Old tests should pass.

capJavert (commented):
Lee thanks for work on this 💪 , just a minor comments on tests, other then that looks ok to me.

__tests__/posts.ts
```diff
@@ -3508,8 +3639,7 @@ describe('mutation votePost', () => {
     );
   });
 
-  it('should upvote', async () => {
```
capJavert, line 3511: Same as https://github.com/dailydotdev/daily-api/pull/1866#discussion_r1589596497

__tests__/posts.ts
sshanzel: Ah, sure, we can move it.

__tests__/posts.ts
sshanzel: Yeah, it is probably better to keep old ones untouched.

capJavert (approved):

PR #1870, opened 2024-04-30
feat: add search tags suggestions query

capJavert:
AS-281

Add query to use in search panel. https://github.com/dailydotdev/apps/pull/3067

Similar query is in `schema/keywords` but it is only for moderators and I wanted to expose this one from`schema/search` to match specific format we return for all of our search providers.

### DIFF
--- a/src/common/search.ts
+++ b/src/common/search.ts
@@ -10,13 +10,5 @@ export const maxSearchLimit = 100;
 export const getSearchLimit = ({
   limit,
 }: Pick<SearchSuggestionArgs, 'limit'>) => {
-  if (limit > maxSearchLimit) {
-    return maxSearchLimit;
-  }
-
-  if (limit < 1) {
-    return 1;
-  }
-
-  return limit ?? defaultSearchLimit;
+  return Math.max(Math.min(limit ?? defaultSearchLimit, maxSearchLimit), 1);
 };
--- a/src/common/search.ts
+++ b/src/common/search.ts
@@ -1,0 +1,22 @@
+export type SearchSuggestionArgs = {
+  query: string;
+  version: number;
+  limit?: number;
+};
+
+export const defaultSearchLimit = 3;
+export const maxSearchLimit = 100;
+
+export const getSearchLimit = ({
+  limit,
+}: Pick<SearchSuggestionArgs, 'limit'>) => {
+  if (limit > maxSearchLimit) {
+    return maxSearchLimit;
+  }
+
+  if (limit < 1) {
+    return 1;
+  }
+
+  return limit ?? defaultSearchLimit;
+};
--- a/src/schema/search.ts
+++ b/src/schema/search.ts
@@ -22,18 +22,23 @@ import { ConnectionArguments } from 'graphql-relay/index';
 import { FeedArgs, feedResolver, fixedIdsFeedBuilder } from '../common';
 import { GQLPost } from './posts';
 import { searchMeili } from '../integrations/meilisearch';
-import { Post, Source, UserPost } from '../entity';
+import { Keyword, Post, Source, UserPost } from '../entity';
+import {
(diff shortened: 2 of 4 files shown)


src/schema/search.ts
```diff
@@ -312,6 +337,28 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
         query: args.query,
       };
     },
+    searchTagSuggestions: async (
```
rebelchris, line 340: I noticed this the other day why do we use keyword and tag differently 👀

src/schema/search.ts
```diff
@@ -312,6 +337,28 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
...
+      source,
+      { query, limit }: SearchSuggestionArgs,
+      ctx,
+    ): Promise<GQLSearchSuggestionsResults> => {
+      const searchQuery = ctx.con
+        .getRepository(Keyword)
+        .createQueryBuilder()
+        .select(`value as id, COALESCE(flags->>'title', value) as title`)
+        .where(`status = 'allow'`)
+        .andWhere(`value ILIKE :query`, {
```
rebelchris, line 350: Well need to monitor performance on this one

rebelchris (approved):

src/schema/search.ts
idoshamun: it's a legacy reason. At first we had tags and then i wanted to change the way i manage and discover the tags without affecting everything so I had to come up with a new name, keyword. Everything else was already using the tags term so I kept it. 🤷 
Bad decisions

src/common/search.ts
```diff
@@ -0,0 +1,22 @@
+export type SearchSuggestionArgs = {
+  query: string;
+  version: number;
+  limit?: number;
+};
+
+export const defaultSearchLimit = 3;
+export const maxSearchLimit = 100;
+
+export const getSearchLimit = ({
```
idoshamun, line 10: You can simply do `Math.max(Math.min(limit, maxSearchLimit), 1)`

src/schema/search.ts
```diff
@@ -312,6 +337,28 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
...
     },
+    searchTagSuggestions: async (
+      source,
+      { query, limit }: SearchSuggestionArgs,
+      ctx,
+    ): Promise<GQLSearchSuggestionsResults> => {
+      const searchQuery = ctx.con
+        .getRepository(Keyword)
+        .createQueryBuilder()
+        .select(`value as id, COALESCE(flags->>'title', value) as title`)
```
idoshamun, line 348: What is this title thingy?

idoshamun (approved):

src/schema/search.ts
rebelchris: Ah so moving forward can we just start using keyword everywhere 👍

src/schema/search.ts
rebelchris: I think its the flag that nimrod can set for sep purpose.
I think we should just search based on value (i think its called)

idoshamun: thanks for the explanation

nensidosari (approved):

src/common/search.ts
capJavert: Yep, nice!

src/schema/search.ts
capJavert: I'll refactor so it is called keyword here, on frontend though we still use Tag almost everywhere.

idoshamun: I don't mind keeping it tags everywhere

capJavert: @rebelchris gonna merge this as is, let's sync on it tomorrow when you are back.

PR #1871, opened 2024-04-30
fix: set correct ch port

dailydotdevbot:


omBratteng (approved):
Looks good, why did the order of them change?
Also, why PR opened by bot?

idoshamun (approved):

PR #1872, opened 2024-05-01
feat: export data to customerio

idoshamun:
Update user profiles in customerio and delete them when needed.

Sidenote:  I refactored relevant workers to typed pubsub 💪

### DIFF
--- a/__tests__/workers/userCreatedCio.ts
+++ b/__tests__/workers/userCreatedCio.ts
@@ -1,0 +1,60 @@
+import nock from 'nock';
+import worker from '../../src/workers/userCreatedCio';
+import { ChangeObject } from '../../src/types';
+import { expectSuccessfulTypedBackground } from '../helpers';
+import { User } from '../../src/entity';
+import { PubSubSchema } from '../../src/common';
+import { cio } from '../../src/cio';
+
+jest.mock('../../src/cio', () => ({
+  ...(jest.requireActual('../../src/cio') as Record<string, unknown>),
+  cio: { identify: jest.fn() },
+}));
+
+beforeEach(async () => {
+  jest.clearAllMocks();
+  nock.cleanAll();
+});
+
+describe('userCreatedCio', () => {
+  type ObjectType = Partial<User>;
+  const base: ChangeObject<ObjectType> = {
+    id: '1',
+    username: 'cio',
+    name: 'Customer IO',
+    infoConfirmed: true,
+    createdAt: 1714577744717,
+    updatedAt: 1714577744717,
+    bio: 'bio',
+    readme: 'readme',
+  };
+
+  it('should update customer.io', async () => {
+    await expectSuccessfulTypedBackground(worker, {
+      user: base,
+    } as unknown as PubSubSchema['api.v1.user-created']);
+    expect(cio.identify).toHaveBeenCalledWith('1', {
+      created_at: 1714577744,
+      first_name: 'Customer',
+      name: 'Customer IO',
+      updated_at: 1714577744,
+      username: 'cio',
+    });
+  });
+
+  it('should not update customer.io if user is ghost user', async () => {
+    const before: ChangeObject<ObjectType> = { ...base, id: '404' };
+    await expectSuccessfulTypedBackground(worker, {
+      user: before,
+    } as unknown as PubSubSchema['api.v1.user-created']);
+    expect(cio.identify).toHaveBeenCalledTimes(0);
+  });
+
+  it('should not update customer.io if user is not confirmed', async () => {
+    const before: ChangeObject<ObjectType> = { ...base, infoConfirmed: false };
+    await expectSuccessfulTypedBackground(worker, {
+      user: before,
+    } as unknown as PubSubSchema['api.v1.user-created']);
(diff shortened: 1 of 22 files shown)


src/workers/userUpdatedCio.ts
```diff
@@ -0,0 +1,54 @@
...
+function camelCaseToSnakeCase(
+  obj: Record<string, unknown>,
+): Record<string, unknown> {
+  const snakeObj: Record<string, unknown> = {};
+  for (const key in obj) {
+    snakeObj[key.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`)] =
+      obj[key];
+  }
+  return snakeObj;
+}
```
omBratteng, line 15: make this a global util?

src/workers/userUpdatedCio.ts
```diff
@@ -0,0 +1,54 @@
...
+  'bio',
+  'devcardEligible',
+  'readme',
+  'readmeHtml',
+  'infoConfirmed',
+  'profileConfirmed',
+  'notificationEmail',
+];
+
+const worker: TypedWorker<'user-updated'> = {
```
omBratteng, line 28: Just for updates? Or does it also run on user created?

src/workers/userUpdatedCio.ts
idoshamun: yes!

src/workers/userUpdatedCio.ts
idoshamun: originally i thought that on creation a user can't be confirmed but i think with email they can. i'll double check

src/cio.ts
```diff
@@ -0,0 +1,10 @@
+import { TrackClient } from 'customerio-node';
+
+export const cio = new TrackClient(
+  process.env.CIO_SITE_ID,
+  process.env.CIO_API_KEY,
+);
```
capJavert, line 6: Just to make sure, will this work locally where we don't have credentials in env?

src/workers/userUpdatedCio.ts
capJavert: One edge case with implementation:
```js
camelCaseToSnakeCase({
    SomeProp: 'test'
})
```

outputs:
```
{
    _some_prop: 'test'
}
```

capJavert: Since we already have lodash maybe use `_.snakeCase`?

src/workers/userUpdatedCio.ts
```diff
@@ -0,0 +1,54 @@
...
+const OMIT_FIELDS: (keyof User)[] = [
+  'id',
+  'bio',
+  'devcardEligible',
+  'readme',
+  'readmeHtml',
+  'infoConfirmed',
+  'profileConfirmed',
+  'notificationEmail',
+];
```
capJavert, line 26: Does it make sense to do it other way around, specify fields we want to send? I am thinking we add some field to `User` eg. `flags` and then forget to update it we accidentally send it to customerio.

capJavert (approved):
Looks good, some minor comments.

src/workers/userUpdatedCio.ts
idoshamun: I rather have them sent mistakenly than not send. So imagine we add experience level (like we did) and totally forget about it. Then we have to import historical data and fix the code, more like risk management

src/cio.ts
idoshamun: i'll double check but it shouldn't break, probably will just throw an error if you try to actually run it, but not on init

src/workers/userUpdatedCio.ts
idoshamun: Good idea!

src/cio.ts
capJavert: It is gonna fire when we create user locally right, and then worker will try to retry indefinitely. I think we had similar thing with sendgrid and then we just added NODE_ENV check to avoid running worker code where it is required in development.

__tests__/workers/userCreatedCio.ts
```diff
@@ -0,0 +1,60 @@
...
+  ...(jest.requireActual('../../src/cio') as Record<string, unknown>),
+  cio: { identify: jest.fn() },
+}));
+
+beforeEach(async () => {
+  jest.clearAllMocks();
+  nock.cleanAll();
+});
+
+describe('userCreatedCio', () => {
```
capJavert, line 19: If you can add this kind of test https://github.com/dailydotdev/daily-api/blob/45364867794974b43f4713e3fcf362da2eb27f65/__tests__/workers/commentDownvotedRep.ts#L53-L59 for new CIO workers, just to make sure they are actually registered in `workers/index`.

capJavert (approved):

src/workers/userUpdatedCio.ts
capJavert: 🤝

__tests__/workers/userCreatedCio.ts
idoshamun: i missed both of your comments. i'll fix that in a follow up PR

PR #1874, opened 2024-05-03
fix: actually send post_id as prop

rebelchris:
After tedious debugging I realized our `feedResolverCursor` doesn't allow post_id by default, thus a manual generator setting is needed to actually add the `post_id`.

### DIFF
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -46,6 +46,7 @@ import { ConnectionArguments } from 'graphql-relay';
 import graphorm from '../graphorm';
 import {
   feedClient,
+  FeedConfigName,
   FeedGenerator,
   feedGenerators,
   SimpleFeedConfigGenerator,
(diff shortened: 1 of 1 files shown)


idoshamun (approved):

src/schema/feeds.ts
```diff
@@ -1250,11 +1251,18 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       ctx,
       info,
     ) => {
+      const generator = new FeedGenerator(
+        feedClient,
+        new SimpleFeedConfigGenerator({
+          feed_config_name: FeedConfigName.PostSimilarity,
+          ...args,
```
capJavert, line 1258: Just watch out `args` is few other things like `before`, `offset`... that do not belong in generator. So maybe just pass `args.post_id`?

capJavert (approved):

src/schema/feeds.ts
capJavert: It is not doing anything here but just don't like when args are spread since it could do something accidentally. Again, TS `strict` would scream here.

rebelchris: I'll adjust if the call now returns and I do the test cleanup :)

PR #1880, opened 2024-05-07
Revert "AS-293 content quality feed filter (#1877)"

denisb0:
This reverts commit 23b7b48a12e161c314858d305e70560a80cf27a0.


capJavert (commented):
Why are we reverting?

PR #1882, opened 2024-05-07
feat: add missing fk to user relations

capJavert:
In addition to https://github.com/dailydotdev/daily-api/pull/1878, fixing other cases where FK was missing on `userId` column.

- adding 1:1 and M:1 relations depending on PK
- migration file each has `DELETE` statement to delete leftover entries from relation tables for deleted users, eg. leftover alerts, bookmarks, settings from those deleted users
- tests updates are mostly due to users fixture missing now that we added FK

this will delete:
- user_action - 6,596 rows
- alerts - 33,347 rows
- bookmark - 4,278 rows
- settings - 74,233 rows
- user_state - 9 rows

### DIFF
--- a/__tests__/actions.ts
+++ b/__tests__/actions.ts
@@ -1,5 +1,5 @@
 import { DataSource } from 'typeorm';
-import { UserAction, UserActionType } from '../src/entity';
+import { User, UserAction, UserActionType } from '../src/entity';
 import createOrGetConnection from '../src/db';
 import {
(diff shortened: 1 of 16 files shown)


rebelchris (approved):
Could you maybe include the warn->error issue here as well?
Secondary should we consider cleaning the deleteUser function?

(Although it can't harm like this, so maybe safer to leave as is)

idoshamun (approved):

PR #1884, opened 2024-05-08
chore: bump dependencies

capJavert:
Bump dependencies.

Some major upgrades but no breaking changes since packages just incremented major because of old node@14,16 support.


capJavert:
I think we can also remove `node-fetch` as dependency since node 20.x supports fetch natively. I'll do it in separate PR.

package.json
```diff
@@ -121,38 +121,38 @@
...
-    "@swc/core": "^1.4.14",
+    "@fastify/static": "^7.0.4",
+    "@swc/core": "^1.5.5",
     "@types/express": "^4.17.21",
     "@types/jest": "^29.5.12",
     "@types/jsonwebtoken": "^9.0.6",
-    "@types/lodash": "^4.17.0",
-    "@types/node": "^18.16.17",
+    "@types/lodash": "^4.17.1",
+    "@types/node": "^20.12.10",
```
omBratteng, line 130: I usually just do `20.x`, then it's taken care of for this version

```suggestion
    "@types/node": "20.x",
```

omBratteng (approved):

package.json
capJavert: I don't like wildcards 😆 they can bite you really easily

rebelchris (approved):
Nice nice!

package.json
omBratteng: Same, but I usually find it alright for `@types/*`, as it's just types, and shouldn't cause issues™️

omBratteng: But non-blocker

PR #1889, opened 2024-05-09
feat: handle lofn extra field

idoshamun:
Some feed experiments are not part of the inner config but the request itself. Something we will probably have to fix in the future but for now we need it for the AIGC experiment.


rebelchris (approved):
Lets go for now to unblock. 🙏

capJavert (approved):

PR #1898, opened 2024-05-10
feat: add custom feed fields

capJavert:
AS-300 - new fields needed for custom feed

Also implemented AS-303 - `feedList` query


src/entity/Feed.ts
```diff
@@ -8,4 +15,28 @@ export class Feed {
...
+    asExpression: `trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(LEFT(feed.flags->>'name',100),'')||'-'||feed.id)), '[^a-z0-9-]+', '-', 'gi'))`,
+  })
+  @Index('IDX_feed_slug', { unique: true })
+  slug: string;
+
+  @ManyToOne(() => User, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  user: Promise<User>;
```
capJavert, line 41: We have `userId` already so we don't need `createdBy` from RFC I think.

omBratteng: Is it possible that two users can share the same feed ID?
In that case, it would probably be useful to have createdBy, as that's the person who can delete a feed?

<table>
<tr>
 <td>Feed ID
 <td>User ID
<tr>
 <td>feed123
 <td>user1
<tr>
 <td>feed123
 <td>user2
</table>

capJavert: No, `id` is PK and unique in `feed` table.

__tests__/feeds.ts
```diff
@@ -2040,3 +2040,217 @@ describe('query userUpvotedFeed', () => {
...
+    });
+  });
+
+  it('should contain all default values in db query', async () => {
+    const feed = await con.getRepository(Feed).findOneBy({ id: 'cf1' });
+    expect(feed?.flags).toEqual({});
+  });
+});
+
+describe('slug filed', () => {
```
idoshamun, line 2103:
```suggestion
describe('slug field', () => {
```

idoshamun (approved):
Minor typo, but overall LGTM

src/entity/Feed.ts
```diff
@@ -1,11 +1,43 @@
...
+
+  @Column({
+    type: 'text',
+    update: false,
+    insert: false,
+    nullable: false,
+    unique: true,
+    generatedType: 'STORED',
+    asExpression: `trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(LEFT(feed.flags->>'name',100),'')||'-'||feed.id)), '[^a-z0-9-]+', '-', 'gi'))`,
+  })
```
rebelchris, line 34: Nice, love to see more of these 🤩

rebelchris (approved):
Nice work!

PR #1900, opened 2024-05-10
feat: disable dependabot for dependencies

capJavert:
Disabling dependabot for `npm`, we update at our own pace monthly so his PRs just spam. Security alerts are still enabled per docs.


omBratteng (commented):
Removing it entirely should also work

idoshamun:
yap let's delete the file

PR #1902, opened 2024-05-13
refactor: sources query to filter featured sources

sshanzel:
Allow the sources query to filter by flags.featured.

This is backwards compatible as seen by the still passing old tests. We don't have to wait to manually update existing sources.

MI-348 #done


src/schema/sources.ts
```diff
@@ -308,6 +309,11 @@ export const typeDefs = /* GraphQL */ `
       Fetch public Squads
       """
       filterOpenSquads: Boolean
+
+      """
+      Add filter for featured sources
+      """
+      featured: Boolean
```
capJavert, line 316: We should add index for this.

__tests__/sources.ts
```diff
@@ -102,32 +103,73 @@ beforeEach(async () => {
...
+      { flags: updateFlagsStatement({ featured: true }) },
+    );
+    await repo.update(
+      { id: 'b' },
+      { flags: updateFlagsStatement({ featured: false }) },
+    );
+  };
+
+  it('should return only featured sources', async () => {
+    await prepareFeaturedTests();
```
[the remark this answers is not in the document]
sshanzel: Just a bit time-consuming whenever those changes happen as we will have to update all the copy-pasted code blocks. Yeah, maybe we can enclose this in a separate `describe` - something like `describe('query sources with featured parameter')`. Not sure if that would be ideal, but we can probably test some other options as well.

src/schema/sources.ts
sshanzel: Should be added now.

src/migration/1715592299433-SourceFlagFeaturedIndex.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceFlagFeaturedIndex1715592299433
+  implements MigrationInterface
+{
+  name = 'SourceFlagFeaturedIndex1715592299433';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_source_flags_featured" ON source USING HASH (((flags->'featured')::boolean))`,
```
rebelchris, line 10: We might have to check since we updated postgres not sure we need the hash and casting @capJavert ?

rebelchris (approved):
Non blocking comments

__tests__/sources.ts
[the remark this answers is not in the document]
capJavert: Yep, most of the time tests are connected either way so no need to nest, but if needed that is an option.

src/schema/sources.ts
```diff
@@ -1058,6 +1065,11 @@ export const resolvers: IResolvers<any, Context> = {
         filter.type = SourceType.Squad;
         filter.private = false;
       }
+
+      if (!isNullOrUndefined(args.featured)) {
+        filter.flags = { featured: args.featured };
+      }
```
capJavert, line 1071: You are gonna have to do nested `andWhere` and specifiy `flags->'featured' = :featured` because like this typeorm creates a query that just compares the full jsonb field:
`AND WHERE "source_1fm"."flags" = :orm_param_1`, which would fail if you had any other jsonb properties in flags.

src/migration/1715592299433-SourceFlagFeaturedIndex.ts
capJavert: HASH is generally more efficient (storage wise) for equality operations, thats why we use it for this bool flags. 

As for casting we can check with `EXPLAIN ANALYZE` to see if it is hit without cast, although it might also be depending on type of equality check we do, look at my comment here https://github.com/dailydotdev/daily-api/pull/1902#discussion_r1598250440, since we are gonna cast to boolean either way.

src/schema/sources.ts
capJavert: and yes, typeorm jsonb support is garbage :hidethepain:

src/migration/1715592299433-SourceFlagFeaturedIndex.ts
rebelchris: Yeah nothing major, just remembered we only did certain things because the previous version didn't support everything right? (but yeah if we already cast it's fine)

kopancek (commented):
Great work, looks good to me, except issues pointed out by @capJavert already.

__tests__/sources.ts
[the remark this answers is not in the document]
sshanzel: Cool. Yeah, I agree, nested describe is the way to go.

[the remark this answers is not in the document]
sshanzel: Updated the tests to do nested describe.

src/schema/sources.ts
sshanzel: Should now be converted to this. All tests are still passing 🚀

nensidosari:
> This is backwards compatible as seen by the still passing old tests.

Since script is already run yesterday to add featured to all current public squads, do we still need backwards compatibility?

sshanzel:
> Since script is already run yesterday to add featured to all current public squads, do we still need backwards compatibility?

Still probably handy if we want to filter non-featured, etc.

capJavert (approved):
Great work 🚀

PR #1905, opened 2024-05-14
chore: update pulumi-common to 1.34.1 to remove ch liveness check 

vpol:


omBratteng (approved):

omBratteng:
We might need to go in an force apply it as you tweaked the yaml manually. We'll see if the deploy fails after merging

vpol:
> We might need to go in an force apply it as you tweaked the yaml manually. We'll see if the deploy fails after merging

yes, it will fail, but I remove deployment before pulumi_up step, it's safe.

PR #1907, opened 2024-05-14
fix: remove source advanced settings

rebelchris:
Legacy system, cleanup basically.

AS-318 #done


sshanzel (approved):
This is nice. What did we replace this with? Or are we just removing this on Source entity? Because if not, we should probably delete the whole entity?

rebelchris:
@sshanzel Basically this wasn't really used only in the old API feed resolver.
But new system leverages the actually `content_curation` types per post, instead of per source. 🚀

capJavert (approved):

PR #1910, opened 2024-05-15
feat: custom feed query

capJavert:
AS-302

### DIFF
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -215,7 +215,7 @@ export interface FeedOptions {
   refresh?: boolean;
 }
 
-export type FeedArgs = ConnectionArguments & FeedOptions;
+export type FeedArgs = ConnectionArguments & FeedOptions & { feedId?: string };
 
 export const applyFeedWhere = (
   ctx: Context,
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -47,10 +47,12 @@ import { GQLPost } from './posts';
 import { Connection, ConnectionArguments } from 'graphql-relay';
 import graphorm from '../graphorm';
 import {
+  FeedClient,
   feedClient,
   FeedConfigName,
   FeedGenerator,
   feedGenerators,
+  FeedPreferencesConfigGenerator,
   SimpleFeedConfigGenerator,
   versionToFeedGenerator,
 } from '../integrations/feed';
@@ -690,6 +692,32 @@ export const typeDefs = /* GraphQL */ `
       """
       feedIdOrSlug: ID!
     ): Feed @auth
+
+    """
+    Get custom feed
+    """
+    customFeed(
+      """
+      Feed id
+      """
+      feedId: ID!
+      """
+      Paginate after opaque cursor
+      """
+      after: String
+      """
+      Paginate first
+      """
+      first: Int
+      """
+      Ranking criteria for the feed
+      """
+      ranking: Ranking = POPULARITY
+      """
+      Array of supported post types
+      """
+      supportedTypes: [String!]
+    ): PostConnection! @auth
   }
 
   extend type Mutation {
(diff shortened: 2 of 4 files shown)


src/common/feedGenerator.ts
```diff
@@ -215,7 +215,7 @@ export interface FeedOptions {
   refresh?: boolean;
 }
 
-export type FeedArgs = ConnectionArguments & FeedOptions;
+export type FeedArgs = ConnectionArguments & FeedOptions & { feedId?: string };
```
omBratteng, line 218: Moved `{ feedId?: string }` into `FeedOptions`?

src/schema/feeds.ts
```diff
@@ -1024,24 +1052,33 @@ const anonymousFeedResolverV1: IFieldResolver<
...
-    applyFeedPaging,
-    {
-      fetchQueryParams: async (ctx) =>
-        feedToFilters(ctx.con, ctx.userId, ctx.userId),
-    },
-  );
+const feedResolverV1: IFieldResolver<
+  unknown,
+  Context,
+  ConfiguredFeedArgs & { feedId?: string }
```
rebelchris, line 1058: This could probably also be merged?

src/schema/feeds.ts
```diff
@@ -1024,24 +1052,33 @@ const anonymousFeedResolverV1: IFieldResolver<
...
+      unreadOnly,
+      builder,
+      alias,
+      queryParams,
+    ),
+  feedPageGenerator,
+  applyFeedPaging,
+  {
+    fetchQueryParams: async (ctx, args) =>
+      feedToFilters(ctx.con, args.feedId || ctx.userId, ctx.userId),
```
rebelchris, line 1079: Nitpicking but see we do like this here and above as:
```
const feedId = opts.feedId || userId;
      const filters = await feedToFilters(ctx.con, feedId, userId);
```

src/schema/feeds.ts
```diff
@@ -1171,6 +1208,61 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       }
       return feedResolverV1(source, args, ctx, info);
     },
+    customFeed: async (
+      source,
+      args: ConfiguredFeedArgs,
+      ctx: Context,
+      info,
+    ) => {
+      const feedIdOrSlug = args.feedId;
```
rebelchris, line 1217: Oh configuredFeedArgs already has it, so we can remove it in above query as extra?

src/schema/feeds.ts
```diff
@@ -1171,6 +1208,61 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+      ) {
+        const feedGenerator = new FeedGenerator(
+          new FeedClient(process.env.POPULAR_FEED),
+          new FeedPreferencesConfigGenerator(
+            {},
+            {
+              includePostTypes: true,
+              includeBlockedSources: true,
+              includeBlockedTags: true,
+              feedId: feedId,
```
rebelchris, line 1239: I think we'll also need the new one `includeBlockedContentCuration` for this feed?

src/schema/feeds.ts
```diff
@@ -1171,6 +1208,61 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+            },
+          ),
+          'popular',
```
rebelchris, line 1242: Isn't this a enum somewhere?

rebelchris (approved):
Some minor nitpick things, nothing really blocking :)

src/schema/feeds.ts
capJavert: Nope, its just union type.

src/schema/feeds.ts
capJavert: For now it should be the same as popular as mentioned in RFC. We can fine tune during testing.

rebelchris: Didn't I add it in popular? 😅 (think it's one that goes for all feeds)

src/schema/feeds.ts
capJavert: Its just different places, how would you simplify it?

rebelchris: Personally just do this in both 😅
feedToFilters(ctx.con, args.feedId || ctx.userId, ctx.userId)

src/schema/feeds.ts
capJavert: Yes!

src/common/feedGenerator.ts
capJavert: yep!

src/schema/feeds.ts
capJavert: You are right, missed it in merge.

src/schema/feeds.ts
capJavert: Ah you mean to extract it as `feedId` variable?

rebelchris (approved):
Nice!

omBratteng (approved):

idoshamun (approved):

PR #1911, opened 2024-05-15
feat: add filters support for custom feeds

capJavert:
AS-323

- add `feedId` to both filters mutations so filters can be added to custom feeds
- `feedId` is optional to preserve BC with my feed


src/schema/feeds.ts
```diff
@@ -1021,14 +1041,19 @@ const applyFeedPagingWithPin = (
...
+}: {
+  ctx: Context;
+  info: GraphQLResolveInfo;
+  feedId: string;
+}): Promise<GQLFeedSettings> => {
   const res = await graphorm.query<GQLFeedSettings>(ctx, info, (builder) => {
     builder.queryBuilder = builder.queryBuilder.andWhere(
-      `"${builder.alias}".id = :userId AND "${builder.alias}"."userId" = :userId`,
-      { userId: ctx.userId },
+      `"${builder.alias}".id = :feedId AND "${builder.alias}"."userId" = :userId`,
```
rebelchris, line 1055: Don't you need to do the `feedId || userId` here as well?

rebelchris (approved):
Non blocking comment, just making sure :)

src/schema/feeds.ts
capJavert: Yeah, but I did it by modfying the `getFeedSettings` to accept `feedId` as well so it is passed correctly from the outside.

nensidosari (approved):
Looks good

omBratteng (approved):

PR #1912, opened 2024-05-15
feat: customer.io webhook for marketing cta

omBratteng:
Adds webhook endpoint for use with customer.io and assigning users to marketing CTA.
The POST requests have to have a valid HMAC signature

AS-328


src/routes/webhooks.ts
```diff
@@ -109,4 +141,36 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
+      const con = await createOrGetConnection();
+      await con.getRepository(UserMarketingCta).upsert(
+        {
+          userId: userId,
+          marketingCtaId: marketingCtaId,
+        },
+        ['userId', 'marketingCtaId'],
+      );
+
+      await getMarketingCta(con, logger, userId);
```
omBratteng, line 171: It'll throw a 500 error of either of userId or marketingCtaId doesn't exist.
Should probably wrap it in a try...catch so that we can return a different status

src/schema/users.ts
```diff
@@ -773,31 +800,8 @@ export const getMarketingCta = async (
...
-    const redisValue = userMarketingCta
-      ? JSON.stringify(marketingCta)
-      : RedisMagicValues.SLEEPING;
-
-    await setRedisObjectWithExpiry(redisKey, redisValue, ONE_WEEK_IN_SECONDS);
-  }
-
-  return marketingCta;
+  const marketingCta: MarketingCta | null = JSON.parse(rawRedisValue || null);
+  return marketingCta || cachePrefillMarketingCta(con, userId, redisKey);
```
omBratteng, line 804: Didn't need this change, should I keep it or revert it?

src/routes/webhooks.ts
capJavert: Yeah, webhoook providers often disable webhooks that return 2xx status code repediately so I would prefer we catch, log error and return 2xx always.

src/schema/users.ts
capJavert: If we don't need it lets remove. Less code is better.

__tests__/webhooks.ts
```diff
@@ -0,0 +1,127 @@
...
+        .post('/webhooks/customerio/marketing_cta')
+        .set('x-cio-timestamp', timestamp.toString())
+        .set('x-cio-signature', hash)
+        .send(payload)
+        .expect(200);
+
+      expect(body.success).toEqual(true);
+    });
+
+    it('should return 200 when signature is valid (dynamic)', async () => {
```
capJavert, line 104: what does `dynamic` mean in test name?

src/routes/webhooks.ts
```diff
@@ -45,6 +51,32 @@ const verifyRequest = (publicKey: string, req: FastifyRequest): boolean => {
...
+  hmac.update(req.rawBody);
+
+  const hash = hmac.digest();
+
+  if (!timingSafeEqual(hash, Buffer.from(signature, 'hex'))) {
+    logger.debug("Signature didn't match");
+    return false;
+  }
+
+  logger.debug('Signature matched!');
```
capJavert, line 76: I know it is just debug logging, but probably something like "CIO signature matched"? Same above.

src/workers/cdc/primary.ts
```diff
@@ -878,9 +854,6 @@ const worker: Worker = {
         case getTableName(con, MarketingCta):
           await onMarketingCtaChange(con, data);
           break;
-        case getTableName(con, UserMarketingCta):
-          await onUserMarketingCtaChange(data);
-          break;
```
capJavert, line 883: Nice!

src/routes/webhooks.ts
capJavert: We can return `success: false` in that case as well.

capJavert (commented):
Looks good, looking forward to implementation. Few comments.

__tests__/webhooks.ts
omBratteng: The other one (static) has the values hardcoded. And this one uses real timestamp for hmac signature

capJavert: So its the same test?

omBratteng: Basically yeah

capJavert: Never enough tests but interested what is the benefit of having both? Any edge case that you found?

omBratteng: No edge cases, but my reasoning is if we modify the signature checker we'll also modify the signature generator in the test, so we might introduce the same error in both places. Then the static test will stand as a safeguard

capJavert: Makes sense 👍

src/routes/webhooks.ts
omBratteng: Added

src/schema/users.ts
omBratteng: Reverted it

rebelchris (approved):
Nice work, besides Ante's comment nothing to add.

src/routes/webhooks.ts
omBratteng: 👍 I will put it in a try...catch and return false and logs if it fails.

omBratteng: I return 400 with `success:false`

omBratteng: > We have a 4 second timeout and if we get a 408, 409, 429, or 5xx response code in that period, Customer.io will retry sending the notification several times with exponential backoff. We will retry your request up to 10 times over a period of approximately 3 hours.

src/workers/cdc/primary.ts
```diff
@@ -771,30 +771,6 @@ const onMarketingCtaChange = async (
...
-    if (data.payload.before.readAt !== null) return;
-
-    await deleteRedisKey(
-      generateStorageKey(
-        StorageTopic.Boot,
-        StorageKey.MarketingCta,
-        data.payload.before.userId,
-      ),
-    );
-  }
```
omBratteng, line 795: I will create a another PR for the deletion webhook

.infra/Pulumi.prod.yaml
```diff
@@ -46,6 +46,8 @@ config:
       secure: AAABAKsjQ79BKL/Msk8ERCFq7TeKGQ6NnzRcQrOoUzQ7hFEDNdq4uxOcG3w5M5aODolHDw==
     cioSiteId:
       secure: AAABAACN2IeoykRGHprr1nYvk1snUvT733+1dp8QI7+WchlaErfJJ3X6LLPYe8BthHnpqw==
+    cioWebhookSecret:
+      secure: AAABAHkXsqVvvtA1VZ3lLrNQoV0Kh3FM31l7QxQVVwQvR4CHbLyqbtg2N+MJ5URkaRwYsMMPv+8RNnRcCiC/v33GjNBc4e57x5MpVpQC2VPUyjMornJI53n7rmeQBtsQ
```
omBratteng, line 50: This is the testing workspace secret, will revert to production/daily.dev workspace once we know it actually works 🙈

src/routes/webhooks.ts
```diff
@@ -109,4 +141,41 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
+            marketingCtaId: marketingCtaId,
+          },
+          ['userId', 'marketingCtaId'],
+        );
+
+        await getMarketingCta(con, logger, userId);
+
+        return res.send({ success: true });
+      } catch (error) {
+        logger.error(error, 'Error processing CIO webhook');
```
capJavert, line 176: We usually send `logger.error({ err: error }, 'Error processing CIO webhook'');`, I think because GC then extracts from `err`?

src/routes/webhooks.ts
capJavert: Nice 👍

capJavert (approved):
🚀

PR #1915, opened 2024-05-16
fix: new post created should not be visible

capJavert:


omBratteng (approved):

denisb0 (approved):

src/routes/privateRpc.ts
```diff
@@ -79,6 +79,9 @@ export default function (router: ConnectRouter) {
         ...req,
         id: postId,
         shortId: postId,
+        visible: false,
+        sentAnalyticsReport: false,
```
idoshamun, line 83: This maybe needs to be true by default. I'm not sure just try to see if there are any side effects. We use it in cdc and cron job i think

capJavert: reverted to db default which is `true`

idoshamun (approved):

PR #1918, opened 2024-05-17
feat: add filters support for feed preview

capJavert:
- add option to pass filters to feed preview directly and pass it to popular feed used for custom feed
- added additional `FeedLocalConfigGenerator` that creates a config based on inline filters

### DIFF
--- a/src/integrations/feed/configs.ts
+++ b/src/integrations/feed/configs.ts
@@ -5,7 +5,7 @@ import {
   FeedConfigGeneratorResult,
   FeedVersion,
 } from './types';
-import { feedToFilters } from '../../common';
+import { AnonymousFeedFilters, feedToFilters } from '../../common';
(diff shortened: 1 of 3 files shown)


src/integrations/feed/configs.ts
```diff
@@ -55,6 +55,44 @@ export class SimpleFeedConfigGenerator implements FeedConfigGenerator {
...
+const addFiltersToConfig = ({
+  config,
+  filters,
+  opts,
+}: {
+  config: FeedConfig;
+  filters: AnonymousFeedFilters;
+  opts: Options;
+}): FeedConfigGeneratorResult['config'] => {
+  const baseConfig = { ...config };
```
capJavert, line 67: Just extracted to util function to avoid repeat in new generator.

rebelchris (approved):

PR #1921, opened 2024-05-17
feat: set scout during post update

capJavert:
After introduction of #1915 all post get created on API first to reserve post id. 

Our scout assignment logic runs only on `createPost` so it would miss attributing scouts for these new posts.

This PR adds the logic in `updatePost` as well and `scoutId` is set if it was not set before.

For more info see [thread](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1715872489501549)

### DIFF
--- a/__tests__/workers/postUpdated.ts
+++ b/__tests__/workers/postUpdated.ts
@@ -1078,6 +1078,107 @@ describe('on post update', () => {
       is_ai_probability: 0.52,
     });
   });
+
+  it('should update post with the relevant scout id and update submission', async () => {
+    const uuid = randomUUID();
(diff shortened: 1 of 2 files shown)


capJavert:
@denisb0 after this is merged you can hopefully enable the create post for community picks again.

denisb0 (commented):
Looks ok for me, but I am not so familiar with logic and data structures :)

omBratteng (approved):

rebelchris (approved):
We can definetly do with some cleanup.
Also the fact it's called `postUpdated` but handles also creation is confusion 😂

PR #1923, opened 2024-05-20
chore: script to update flags for squads

sshanzel:
Script to update all of Squads' stats.

After this, engineering should be able to do some internal testing.

### DIFF
--- a/bin/syncSquadFlags.ts
+++ b/bin/syncSquadFlags.ts
@@ -1,0 +1,62 @@
+import fastq from 'fastq';
+import '../src/config';
+import createOrGetConnection from '../src/db';
+import { Post, Source, SquadSource } from '../src/entity';
+import { updateFlagsStatement } from '../src/common';
+
+interface Result {
+  sourceId: string;
+  count: number;
+  views: number;
+  upvotes: number;
+}
+
+const QUEUE_CONCURRENCY = 1;
+
+(async () => {
+  const con = await createOrGetConnection();
+  const stream = await con
+    .getRepository(Post)
+    .createQueryBuilder('p')
+    .select('COUNT(p.*)', 'count')
+    .addSelect('SUM(p.views)', 'views')
+    .addSelect('SUM(p.upvotes)', 'upvotes')
+    .addSelect('s.id', 'sourceId')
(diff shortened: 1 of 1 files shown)


bin/syncSquadFlags.ts
```diff
@@ -0,0 +1,62 @@
...
+        { id: sourceId },
+        {
+          flags: updateFlagsStatement<Source>({
+            totalViews: views,
+            totalUpvotes: upvotes,
+            totalPosts: count,
+          }),
+        },
+      );
+      console.log('stats updated for: ', sourceId);
```
capJavert, line 44: Probably want to remove this line it is gonna spam a lot when you run the script. But up to you!

capJavert: Also above log `processing stats..`

bin/syncSquadFlags.ts
```diff
@@ -0,0 +1,62 @@
...
+    .innerJoin(SquadSource, 's', 's.id = p."sourceId"')
+    .where(`s.type = 'squad'`)
+    .groupBy('s.id')
+    .stream();
+
+  let insertCount = 0;
+  const insertQueue = fastq.promise(
+    async ({ sourceId, views, upvotes, count }: Result) => {
+      console.log('processing stats for: ', sourceId);
+      await con.getRepository(SquadSource).update(
```
capJavert, line 34: We can probably do all of this in a single transaction to avoid multiple `UPDATE` calls.

capJavert (approved):
Looks good, some optional optimizations.

nensidosari (approved):

bin/syncSquadFlags.ts
sshanzel: Yeah, makes sense since we can easily query the stats that need to be updated. Let me check quickly.

bin/syncSquadFlags.ts
sshanzel: Agree with that. Replaced it with just a log when initializing, just so we know it is actually starting.

capJavert (approved):
Awesome!

rebelchris (approved):

PR #1924, opened 2024-05-20
chore: reduce snapshot.fetch.size 

vpol:


omBratteng (commented):
Put it in the js file, so that it triggers the actual update. I haven’t fixed the issue in pulumi-commen yet

vpol:
ugh, it will fail anyway, because this key is not in `ClickHouseSyncConfig`

omBratteng (approved):
I see in preview it will re-create k8s secret, so good to go. Still working on fixing just detecting changes in yaml file

PR #1931, opened 2024-05-23
feat: sync accepted marketing to cio

idoshamun:
(Un)subscribe users from the relevant topic based on the `acceptedMarketing` property

### DIFF
--- a/__tests__/workers/userUpdatedCio.ts
+++ b/__tests__/workers/userUpdatedCio.ts
@@ -69,6 +69,21 @@ describe('userUpdatedCio', () => {
     });
   });
 
+  it('should support accepted marketing false', async () => {
+    mocked(getShortGenericInviteLink).mockImplementation(async () => '');
+    await expectSuccessfulTypedBackground(worker, {
+      newProfile: { ...base, acceptedMarketing: false },
+      user: base,
+    } as unknown as PubSubSchema['user-updated']);
+    expect(
+      mocked(cio.identify).mock.calls[0][1].cio_subscription_preferences,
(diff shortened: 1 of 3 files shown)


src/cio.ts
```diff
@@ -47,5 +47,10 @@ export async function identifyUser(
     created_at: dateToCioTimestamp(debeziumTimeToDate(dup.createdAt)),
     updated_at: dateToCioTimestamp(debeziumTimeToDate(dup.updatedAt)),
     referral_link: genericInviteURL,
+    cio_subscription_preferences: {
+      topics: {
+        topic_4: user.acceptedMarketing,
```
omBratteng, line 52: What is `topic_4`?

__tests__/workers/userUpdatedCio.ts
```diff
@@ -59,6 +60,12 @@ describe('userUpdatedCio', () => {
       updated_at: 1714577744,
       username: 'cio',
       referral_link: referral,
+      accepted_marketing: true,
+      cio_subscription_preferences: {
+        topics: {
+          topic_4: true,
+        },
+      },
```
rebelchris, line 68: Maybe we should add a test for false as well, just to verify down the line if breaking stuff comes?

rebelchris (approved):
Suggestion, but non blocking

omBratteng (approved):

__tests__/workers/userUpdatedCio.ts
idoshamun: Done

PR #1932, opened 2024-05-23
fix: digest options as array

rebelchris:
We can review it, but can only merge once FE changes are done :) 
The idea is that we now have 2 already and third option coming so we need to get all for set user.


src/schema/users.ts
```diff
@@ -1104,16 +1117,14 @@ export const resolvers: IResolvers<any, Context> = {
...
       ctx: Context,
-    ): Promise<GQLPersonalizedDigest> => {
+    ): Promise<GQLPersonalizedDigest[]> => {
       const personalizedDigest = await ctx
         .getRepository(UserPersonalizedDigest)
-        .findOneBy({ userId: ctx.userId, type });
+        .findBy({ userId: ctx.userId });
 
-      if (!personalizedDigest) {
+      if (!personalizedDigest || personalizedDigest.length === 0) {
```
capJavert, line 1127: its always gonna be array so latter is enough

capJavert (approved):

src/schema/users.ts
omBratteng:
```suggestion
      if (personalizedDigest?.length === 0) {
```
maybe?

capJavert: it will never be undefined, `findBy` returns array

rebelchris:
@capJavert Also had to add sendType flag to mutation as we now allow it to be passed from FE.

capJavert (approved):

PR #1933, opened 2024-05-23
feat: add feeds to logged boot

capJavert:
Our sidebar is prone to layout shifts so feeds have to go to boot. 😞 

It will be preloaded into `useFeeds` query on frontend during boot fetch see https://github.com/dailydotdev/apps/pull/3106/commits/4ca6cdf6ac7a935b0ce718b4e6be303b66a895c1

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -1,6 +1,6 @@
 import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';
 import createOrGetConnection from '../db';
-import { DataSource, EntityManager } from 'typeorm';
+import { DataSource, EntityManager, Not } from 'typeorm';
 import { clearAuthentication, dispatchWhoami } from '../kratos';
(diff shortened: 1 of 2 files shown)


src/routes/boot.ts
```diff
@@ -177,6 +179,24 @@ const getSquads = async (
...
+  const feeds = con.getRepository(Feed).find({
+    where: {
+      id: Not(userId),
+      userId,
+    },
+    take: maxFeedsPerUser,
+  });
+
+  return feeds;
+};
```
omBratteng, line 198: Should we store a cache in redis?

capJavert: nah, it should be fast enough, if needed we can add down the line

capJavert: I don't want to complicate with invalidation at this point

omBratteng: 👍

omBratteng (approved):

PR #1934, opened 2024-05-23
fix: switching back to private mode

sshanzel:
Allow admins to make their public Squad go private (vice-versa) once approval has been received.

Something we missed on the RFC: https://dailydotdev.slack.com/archives/C069XNYMDGV/p1715851833290429?thread_ts=1715851745.713089&cid=C069XNYMDGV

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -1495,6 +1495,14 @@ export const resolvers: IResolvers<any, Context> = {
         inputHandle !== current.handle,
       );
 
+      const updates: Partial<SquadSource> = {
+        name,
+        handle,
+        description,
+        memberPostingRank: sourceRoleRank[memberPostingRole],
+        memberInviteRank: sourceRoleRank[memberInviteRole],
+      };
+
       if (!isNullOrUndefined(isPrivate) && current.private !== isPrivate) {
         const approved = await ctx.con
           .getRepository(SquadPublicRequest)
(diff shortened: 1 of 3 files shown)


nensidosari (approved):

src/schema/sources.ts
```diff
@@ -1483,6 +1491,20 @@ export const resolvers: IResolvers<any, Context> = {
...
+      if (!isNullOrUndefined(isPrivate) && current.private !== isPrivate) {
+        const approved = await ctx.con
+          .getRepository(SquadPublicRequest)
+          .findOne({
+            where: { sourceId, status: SquadPublicRequestStatus.Approved },
+          });
+
+        if (!approved) {
+          throw new ValidationError(
+            'Squad has not been approved yet of becoming public',
```
capJavert, line 1503: Add the error message to `errors.ts` file

src/schema/sources.ts
```diff
@@ -1496,6 +1518,7 @@ export const resolvers: IResolvers<any, Context> = {
                 description,
                 memberPostingRank: sourceRoleRank[memberPostingRole],
                 memberInviteRank: sourceRoleRank[memberInviteRole],
+                private: isPrivate,
```
capJavert, line 1521: If someone sends `private: null` this part will throw `UNEXPECTED` because `private` is non null column. Maybe extract the update payload at the top and then only set `private` on it if above `if` statement that you added passes.

sshanzel: Yeah, thought at first it might become a bit of a change, but yes, that should be the right way to go. Will update quickly.

src/schema/sources.ts
sshanzel: Oh cool. Let me move this. Thank you!

src/schema/sources.ts
sshanzel: Updated and included a test for it 🚀

src/schema/sources.ts
```diff
@@ -1483,21 +1495,36 @@ export const resolvers: IResolvers<any, Context> = {
...
+              name,
+              handle,
+              description,
+              memberPostingRank: sourceRoleRank[memberPostingRole],
+              memberInviteRank: sourceRoleRank[memberInviteRole],
+            };
+
+            if (!isNullOrUndefined(isPrivate)) {
+              updates.private = isPrivate;
+            }
```
capJavert, line 1524: Maybe streamline even more so we set it only if private changed and in the same place as approved check? https://github.com/dailydotdev/daily-api/commit/e997d8567da8c205d1dd31d2ed1d538dd6f0b73f

capJavert (approved):

src/schema/sources.ts
sshanzel: Right. Yeah, we can do it within the previous if statement. We might have to bring the whole `updates` object way above, but shouldn't hurt.

capJavert (approved):
🚀

PR #1935, opened 2024-05-23
feat: add CIO endpoint to unassign user from marketing CTA

omBratteng:


src/routes/webhooks/customerio.ts
```diff
@@ -0,0 +1,98 @@
...
+        const con = await createOrGetConnection();
+        await con.getRepository(UserMarketingCta).upsert(
+          {
+            userId: userId,
+            marketingCtaId: marketingCtaId,
+          },
+          ['userId', 'marketingCtaId'],
+        );
+
+        await cachePrefillMarketingCta(con, userId);
```
rebelchris, line 65: This will give a edge-case right?
EG user gets added to two campaigns without viewing one he'll get auto assigned the later in redis, not the first?

rebelchris (commented):
One small remark

src/routes/webhooks/customerio.ts
omBratteng: `cachePrefillMarketingCta` is the same internal function as the boot request uses. I.e. it selects from database the first in queue (there's a test for it)

omBratteng: I.e. you've been assigned to campaign A, first which loads A into cache, then get assigned to campaign B, which then loads A into cache because A was first in the queue (based in createdAt)

omBratteng: https://github.com/dailydotdev/daily-api/blob/b067f758fc865b0c92cbc4c4fdea437bd9d9f869/src/schema/users.ts#L764-L774

rebelchris (approved):

src/schema/users.ts
```diff
@@ -751,6 +751,38 @@ const timestampAtTimezone = `"timestamp"::timestamptz ${userTimezone}`;
 
 const MAX_README_LENGTH = 10_000;
 
+export const cachePrefillMarketingCta = async (
```
capJavert, line 754: Maybe move it to common/marketingCta just to avoid routes importing schema files, trying to get the imports in order so we don't have so many circular dependencies.

capJavert (approved):

src/schema/users.ts
omBratteng: I'll see if I find a good spot for it. I think I tried to move the `getMarketingCta` at one point, which really messed up tests.

capJavert: Yeah, I remember that one, let's see 🙏

omBratteng: I was able to move it

capJavert (approved):
Awesome!

PR #1937, opened 2024-05-24
refactor: remove deprecated `meterProvider.addMetricReader`

omBratteng:
Noticed that `addMetricReader` was deprecated
Also adds prometheus exporter to production, so we can test out the prometheus scraping in gcloud.


src/telemetry/metrics.ts
```diff
@@ -10,26 +10,30 @@ import { GcpDetectorSync } from '@google-cloud/opentelemetry-resource-util';
 import { isProd } from '../common';
 
 export const startMetrics = (serviceName: string): void => {
-  const metricReader = isProd
-    ? new metrics.PeriodicExportingMetricReader({
+  const readers: metrics.MetricReader[] = [
+    new PrometheusExporter({}, () => {
+      const { endpoint, port } = PrometheusExporter.DEFAULT_OPTIONS;
+      console.log(`metrics endpoint: http://localhost:${port}${endpoint}`);
```
capJavert, line 16: Use the logger? Or do we need it even?

src/telemetry/metrics.ts
```diff
@@ -10,26 +10,30 @@ import { GcpDetectorSync } from '@google-cloud/opentelemetry-resource-util';
 import { isProd } from '../common';
 
 export const startMetrics = (serviceName: string): void => {
-  const metricReader = isProd
-    ? new metrics.PeriodicExportingMetricReader({
```
capJavert, line 14: Why didn't we use prometheus exporter before in production?

omBratteng: Because we didn't have prometheus in production before. Added it a few weeks ago, and as it totes that it's cheaper metrics than gcloud, it's worth a look.

capJavert: Ok, if it is not super urgent we can also merge it after the weekend to avoid any issues with current metrics.

src/telemetry/metrics.ts
omBratteng: I removed the console.log, and added logger for errors instead.

src/telemetry/metrics.ts
omBratteng: It won't affect regular metrics, as we can have multiple metric exporters.
So it will both do the old one we used, and add prometheus metrics

capJavert: Yes I see, but I am not 100% on what can go wrong so that is why I am saying.

capJavert (approved):

src/telemetry/metrics.ts
omBratteng: I'll keep an eye out, and revert it if I see it causes any issues. I was planning on playing around with the prometheus this weekend 😬

PR #1940, opened 2024-05-26
feat: migrate transactional emails to cio

idoshamun:
Still WIP but most of the work is done. I placed some todos that I need to fix, I'm waiting for a response from their support.

It actually required more changes than I expected as some things are not supported:
* Sending emails in batches with personalization parameters
* Generating batch id for the digest


__tests__/notifications/index.ts
```diff
@@ -1049,24 +991,6 @@ describe('storeNotificationBundle', () => {
     expect(actual.attachments.length).toEqual(0);
   });
 
-  it('should generate squad_access notification', () => {
```
idoshamun, line 1052: This notification is no longer used

src/growthbook.ts
```diff
@@ -127,7 +119,6 @@ export const features = {
   }),
   dailyDigest: new Feature('daily_personalized_digest', {
     ...digestFeatureBaseConfig,
-    templateId: 'd-925d2759ddd641f99220b3c7c6836458',
```
capJavert, line 130: Digests now have the same template ID by default?

src/routes/index.ts
```diff
@@ -131,4 +136,32 @@ Disallow: /`);
       message: 'ok',
     });
   });
+
+  // https://customer.io/docs/journeys/custom-unsubscribe-links/#step-1-configure-your-system-to-receive-the-post-when-recipients-click-unsubscribe
+  fastify.post<{ Querystring: { token: string } }>(
+    '/unsubscribe',
```
capJavert, line 142: Maybe move under `/notifications`, separate file or just add to path.

capJavert (commented):
Minor stuff so far, actually surprised because I expected more major changes. Nice work 🚀

src/growthbook.ts
```diff
@@ -104,15 +104,7 @@ export class Feature<T extends JSONValue> {
...
-  templateId: 'd-328d1104d2e04fa1ab91e410e02751cb',
-  meta: {
-    from: {
-      email: 'informer@daily.dev',
-      name: 'daily.dev',
-    },
-    category: 'Digests',
-    asmGroupId: 23809,
-  },
+  templateId: '46',
```
rebelchris, line 107: Don't we need to update more templateId's?
Or we'll do the main ones seperate later?

rebelchris (approved):
Wow really not that bad of a change!

src/growthbook.ts
idoshamun: yes. nimrod asked to merge the templates together

PR #1941, opened 2024-05-27
fix: expose prometheus ports internally in cluster

omBratteng:
For the prometheus scraper to find the metrics, we need to expose the metrics endpoint.

NOTE: This is a temporary change, will move it to pulumi-common if prometheus works out great

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -258,7 +258,6 @@ if (isAdhocEnv) {
 } else {
   appsArgs = [
     {
-      port: 3000,
       env: [nodeOptions(memory), ...jwtEnv],
       minReplicas: 3,
       maxReplicas: 25,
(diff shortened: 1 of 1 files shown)


.infra/index.ts
```diff
@@ -258,7 +258,6 @@ if (isAdhocEnv) {
 } else {
   appsArgs = [
     {
-      port: 3000,
```
capJavert, line 261: This is replace with `ports` prop below?

capJavert (approved):

.infra/index.ts
omBratteng: Yeah. `port` is "deprecated", and internally in pulumi-common, `port` translates to `ports`
https://github.com/dailydotdev/pulumi-common/blob/6845b5d4e62d30e0c9b6efb9a8beb7d62a5404cd/src/suite/index.ts#L477-L479

PR #1945, opened 2024-05-28
fix: add missing jwt keys to adhoc deployments

omBratteng:
Latest commits broke api-bg in adhoc, as it is unable to load JWT keys.


idoshamun (approved):
Sorry!!! 🙈

capJavert (approved):

PR #1946, opened 2024-05-28
feat: add post content updated topic

capJavert:
- add new content updated topic for API being source of truth for post data
- utilized typed workers since they already accept types which we can load from schema package 👌 
- ~~have an issue with JSON encoding, opened an issue~~


src/workers/cdc/common.ts
```diff
@@ -8,3 +19,76 @@ export const getTableName = <Entity>(
...
+    updatedAt: post.metadataChangedAt,
+    source: {
+      ...source,
+      createdAt: +source.createdAt,
+      // TODO issue with null can't be parsed as JSON
+      headerImage: source.headerImage || undefined,
+      color: source.color || undefined,
+      twitter: machineSource.twitter || undefined,
+      website: machineSource.website || undefined,
+      description: source.description || undefined,
```
capJavert, line 62: https://github.com/bufbuild/protobuf-es/issues/860, I might have to fallback `null` values, I am not even sure why this happens

capJavert:
Opening up for review, but as said I might have to fallback some values due to JSON encode issues.

src/workers/cdc/common.ts
```diff
@@ -8,3 +19,76 @@ export const getTableName = <Entity>(
   con: DataSource,
   target: EntityTarget<Entity>,
 ): string => con.getRepository(target).metadata.tableName;
+
+export const notifyPostContentUpdated = async ({
+  con,
+  logger,
+  post,
+}: {
```
omBratteng, line 27: Not related to this PR

We could start by not passing the logger down in everything, if we need the logger, we can just import it where we actually need it. Less hoisting.
What do you think @capJavert?

capJavert: yeah, i forgot we had it, I'll replace

capJavert: 422e119738d28d1a7f65558e2f5c13585fac17d0

src/workers/cdc/common.ts
capJavert: Looks like they are gonna port the fix so our issue will be fixed.

capJavert: I also adjusted schema so it supports it when they fix it on their end https://github.com/dailydotdev/schema/pull/27

capJavert: Fixed in 55f7487fde1a9e6b4695cc8a98e10f31a54351ed

src/workers/cdc/common.ts
```diff
@@ -8,3 +24,90 @@ export const getTableName = <Entity>(
...
+    origin: post.origin,
+    url: articlePost.url,
+    canonicalUrl: articlePost.canonicalUrl,
+    image: articlePost.image,
+    description: articlePost.description,
+    readTime: articlePost.readTime,
+    summary: articlePost.summary,
+    content: freeformPost.content,
+    language: post.language,
+    contentMeta: post.contentMeta
```
capJavert, line 91: @denisb0 is https://github.com/dailydotdev/schema/blob/main/daily-api/posts.proto#L33 `int64` for some specific reason or?

capJavert: @vpol do you maybe know if we need that big of a numbers, we usually use int32 in other schemas and on JS side it is a bit tricky because int64 has to be [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) which is more complex to serialize to JSON.

vpol: If basically defines a vector length, event int16 will do the job.

src/workers/cdc/common.ts
```diff
@@ -8,3 +24,86 @@ export const getTableName = <Entity>(
...
+      ],
+      order: {
+        createdAt: 'ASC',
+      },
+    }),
+  ]);
+  const articlePost = post as ChangeObject<ArticlePost>;
+  const freeformPost = post as ChangeObject<FreeformPost>;
+
+  const contentUpdatedMessage = new ContentUpdatedMessage({
```
rebelchris, line 64: Don't we also need the `steps` here?
(not sure what we called them)

rebelchris (approved):
Nothing blocking just one question.

src/workers/cdc/common.ts
capJavert: what steps, for scraping?

capJavert: It should be in `content_meta`, @vpol do you know if one of the fields in [schema](https://github.com/dailydotdev/schema/blob/main/daily-api/pubsub/content-updated.proto) represent it?

capJavert:
Merging to see how it works in production. No subscribers yet so it won't really be used instantly.

PR #1947, opened 2024-05-29
feat: feed feedback reminder reset

sshanzel:
We are planning to show some surveys to our users, and in order to understand that they've seen the survey already, we need to store the timestamp of when they filled or dismissed it.

For new users, we should let them see the survey after the first cycle so they will have ample time to see how their feed works. To do that, we set the last time they saw the card to when we created the user itself.

MI-354 #done


src/migration/1716976350324-AlertFeedSettingsFeedback.ts
```diff
@@ -0,0 +1,27 @@
...
+      `ALTER TABLE "alerts" ADD "lastFeedSettingsFeedback" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now()`,
+    );
+    await queryRunner.query(
+      `
+        UPDATE  public.alerts AS a
+        SET     "lastFeedSettingsFeedback" = u."createdAt"
+        FROM    public.user AS u
+        WHERE   u.id = a."userId"
+      `,
+    );
```
sshanzel, line 19: @capJavert let me know if you want me to put this in a bin script instead.

capJavert: It is gonna update all rows in the table so it might be heavy so I would prefer it in bin so we can also potentially re-run if we have to.

capJavert (approved):

src/migration/1716976350324-AlertFeedSettingsFeedback.ts
idoshamun: Since it's a transaction are we sure it's gonna trigger CDC? (which is my primary concern)

idoshamun (commented):
I think in boot we should add the logic to check for the 30 days, don't we?

sshanzel:
> I think in boot we should add the logic to check for the 30 days, don't we?

Since the boot will include the `alerts` object, we can do it in the FE. This would also allow us to make the experiment flag value to be a number.

idoshamun:
> Since the boot will include the alerts object, we can do it in the FE

it's just that we don't usually do it on the api side, so maybe it's better to stick to the existing practice?
anyway we don't need to experiment on the days for now

src/routes/boot.ts
```diff
@@ -124,12 +132,17 @@ const excludeProperties = <T, K extends keyof T>(
...
   }
-  return obj;
+
+  const clone = structuredClone(obj);
+
+  properties.forEach((prop) => {
+    delete clone[prop];
+  });
+
+  return clone;
```
sshanzel, line 145: Had to update since JS in nature would use the same object referenced from the parameters, hence affecting the original object where it came from. And on the other side, we still need that property for some calculations. With that, we you run a function to modify an object, you normally expect it to not affect the original object and return a new one instead.

src/migration/1716976350324-AlertFeedSettingsFeedback.ts
sshanzel: We should be good since we don't track CDC changes for the `alerts` table (not anymore since sunsetting the gateway repo).

src/routes/boot.ts
```diff
@@ -477,7 +493,10 @@ const anonymousBoot = async (
       shouldVerify,
     },
     visit,
-    alerts: { ...ALERTS_DEFAULT, changelog: false },
+    alerts: {
+      ...excludeProperties(ALERTS_DEFAULT, ['lastFeedSettingsFeedback']),
```
rebelchris, line 503: Wait did you want to exclude all here?

rebelchris (approved):
One non blocking question.

src/routes/boot.ts
sshanzel: The function would only exclude the second parameter, which is the new property. This is just to make it consistent on what is being returned.

idoshamun:
@sshanzel don't wait for me, seems like you got enough approvals :)

PR #1948, opened 2024-05-29
feat: expose list of all tags

rebelchris:
Expose a list of all tags (had to use keywords internally as they have the right data)

### DIFF
--- a/src/schema/tags.ts
+++ b/src/schema/tags.ts
@@ -6,6 +6,8 @@ import { TagRecommendation } from '../entity/TagRecommendation';
 import { In, Not } from 'typeorm';
 import { ValidationError } from 'apollo-server-errors';
 import { SubmissionFailErrorMessage } from '../errors';
+import graphorm from '../graphorm';
+import { GQLKeyword } from './keywords';
 
 interface GQLTag {
   name: string;
@@ -58,6 +60,11 @@ export const typeDefs = /* GraphQL */ `
 
   extend type Query {
     """
+    Get all tags
+    """
+    tags: [Keyword]
+
+    """
     Get the most popular tags
     """
     popularTags: [Tag] @cacheControl(maxAge: 600)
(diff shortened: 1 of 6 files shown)


src/schema/tags.ts
```diff
@@ -57,6 +59,11 @@ export const typeDefs = /* GraphQL */ `
   }
 
   extend type Query {
+    """
+    Get all tags
+    """
+    tags: [Keyword]
```
rebelchris, line 65: Little bit awkward, but with the upcoming ones I want to keep them all in this file 😅

src/schema/tags.ts
```diff
@@ -89,6 +96,14 @@ export const typeDefs = /* GraphQL */ `
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const resolvers: IResolvers<any, Context> = traceResolvers({
   Query: {
+    tags: (_, __, ctx, info): Promise<GQLKeyword[]> =>
+      graphorm.query<GQLKeyword>(ctx, info, (builder) => {
+        builder.queryBuilder = builder.queryBuilder.where({
+          status: 'allow',
+        });
```
capJavert, line 103: limit?

capJavert (commented):
Also if you can add a quick test for the query 🙏

src/schema/tags.ts
rebelchris: Since it should render all in one visual I decided not to add limit to it 😅

capJavert: it will list 827 of them as currently stands, I think we should add upper limit just in case. Maybe 1000?

rebelchris: Not sure what it would add to be honest?
We don't get many new ones, and even if we do get 1001 we want to render them all right?

rebelchris: I did add it :)

capJavert (approved):

PR #1949, opened 2024-05-29
fix: add new tag endpoints for tag directory

rebelchris:
Needed the endpoints, we need to adjust the actual query based on MV we adding.
Added test as well to ensure they keep working.

### DIFF
--- a/src/schema/tags.ts
+++ b/src/schema/tags.ts
@@ -65,9 +65,24 @@ export const typeDefs = /* GraphQL */ `
     tags: [Keyword]
 
     """
+    Get the most trending tags
+    """
+    trendingTags(
+      """
+      Limit the number of tags returned
+      """
+      limit: Int
+    ): [Tag] @cacheControl(maxAge: 600)
+
+    """
     Get the most popular tags
     """
-    popularTags: [Tag] @cacheControl(maxAge: 600)
+    popularTags(
+      """
+      Limit the number of tags returned
+      """
+      limit: Int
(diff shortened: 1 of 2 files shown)


src/schema/tags.ts
```diff
@@ -106,11 +121,25 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+    trendingTags: async (source, args, ctx): Promise<GQLTag[]> => {
+      // TODO: Implement query based on new MV
+      const { limit = 10 } = args;
+      const hits = await ctx.getRepository(Keyword).find({
+        select: ['value'],
+        order: { value: 'ASC' },
+        where: { status: 'allow' },
+        take: limit,
+      });
+      return hits.map((x) => ({ name: x.value }));
```
omBratteng, line 133:
```suggestion
      const keywords = await ctx.getRepository(Keyword).find({
        select: ['value'],
        order: { value: 'ASC' },
        where: { status: 'allow' },
        take: limit,
      });
      return keywords.map((keyword) => ({ name: keyword.value }));
```

Just a personal preference, but I prefer naming the variables to be similar to what we're fetching/looping over

omBratteng (approved):

src/schema/tags.ts
rebelchris: Ah this will be completely removed anyway :) 
It's just the old placeholder for local dev until we have the MV

nensidosari (approved):

PR #1959, opened 2024-05-30
chore: more cpu for ch-sync

vpol:


.infra/index.ts
```diff
@@ -594,7 +594,7 @@ if (!isAdhocEnv) {
       resources: {
         // TODO: adjust resources based on the actual usage
         requests: {
-          cpu: '1',
```
omBratteng, line 597: Just as an FYI, this is just the requested CPU. There are no limits set, so it wouldn't be throttled

vpol: Screenshot 2024-05-30 at 18 56 52

vpol: looks like it is throttled. on my local linux machine it easily eats 4-6 cores.

omBratteng: That's interesting
image
image

PR #1962, opened 2024-05-31
feat: add new setting - optOutReadingStreak

nensidosari:
Add a new user setting to opt out from reading streak. We can now remove also the weekly goal setting just didn't want to do in the same PR

MI-364 #done


omBratteng (approved):

src/schema/settings.ts
```diff
@@ -118,6 +118,11 @@ export const typeDefs = /* GraphQL */ `
     """
     optOutWeeklyGoal: Boolean!
 
+    """
+    Whether the user opted out from reading streak
+    """
+    optOutReadingStreak: Boolean!
```
capJavert, line 124: Update the `GQLSettings` typescript definition as well

src/schema/settings.ts
```diff
@@ -211,6 +216,11 @@ export const typeDefs = /* GraphQL */ `
     """
     optOutWeeklyGoal: Boolean
 
+    """
+    Whether the user opted out from reading streak
+    """
+    optOutReadingStreak: Boolean
```
capJavert, line 222: also update `GQLUpdateSettingsInput` typescript definition

src/entity/Settings.ts
```diff
@@ -52,6 +52,9 @@ export class Settings {
   @Column({ default: false })
   optOutWeeklyGoal: boolean;
 
+  @Column({ default: false })
+  optOutReadingStreak: boolean;
```
capJavert, line 56: Not blocking but did we decide anywhere that it is going to be a setting, maybe user `flags` is better place?

capJavert (commented):
Looks good 👍 some questions for sanity check

src/entity/Settings.ts
nensidosari: Well I don't think there's been any decision on this, just since weekly goal was using settings went to use the same for this. 
U think it's best we use flags?

capJavert: Just a thought, but it is fine like this as well, especially since frontend is also done 👍

capJavert: Let's leave as is.

capJavert (approved):
Good to go when comments/tests are resolved.

src/schema/settings.ts
nensidosari: updated

src/schema/settings.ts
nensidosari: updated

PR #1965, opened 2024-06-03
refactor: get rid of sendgrid

idoshamun:
Remove all SendGrid related code (except the analytics part which we keep to monitor everything)


.infra/workers.ts
```diff
@@ -23,18 +23,10 @@ export const workers: Worker[] = [
     topic: 'views',
     subscription: 'add-views-v2',
   },
-  {
```
rebelchris, line 26: Did you check if these topics can also be removed now?
(not sure if they have other subscriptions)

rebelchris (approved):
Gotta love a good cleanup :)

src/common/personalizedDigest.ts
```diff
@@ -268,13 +264,8 @@ export const schedulePersonalizedDigestSubscriptions = async ({
...
 }) => {
-  // TODO: check if there's an alternative in CIO
-  const emailBatchId = await createEmailBatchId();
-
-  if (!emailBatchId) {
-    throw new Error('failed to create email batch id');
-  }
-
+  // Keep email batch id around just in case
+  const emailBatchId = null;
```
capJavert, line 268: If we are keeping it maybe generate with `generateUUID`? It will be useful since each message log has it so if any errors we know from which cron it comes from.

.infra/workers.ts
```diff
@@ -249,10 +237,6 @@ export const workers: Worker[] = [
     topic: 'api.v1.source-created',
     subscription: 'api.source-created-squad-user-action',
   },
-  {
-    topic: 'api.v1.source-created',
-    subscription: 'api.source-created-squad-owner-mailing',
-  },
```
capJavert, line 255: We don't send this anymore or send it through CIO somehow?

idoshamun: CIO has everything needed 😬

.infra/workers.ts
rebelchris: Ah felt like it could be a optimization cleanup for pushing to topics that have no subscriber.
But def not as important.

src/common/personalizedDigest.ts
idoshamun: yes, fixed

PR #1967, opened 2024-06-04
chore: explicitly exclude alerts table for cdc

sshanzel:


sshanzel:
As per docs, adding both `include` and `exclude` list won't really matter as we should only do one or the other.

PR #1968, opened 2024-06-04
feat: reading streak reminder

omBratteng:


src/onesignal.ts
```diff
@@ -87,3 +91,35 @@ export async function sendReadingReminderPush(
...
+    // @TODO: check what this should be
+    'streak_reminder',
+  );
+
+  // @TODO: check what these should be
+  push.chrome_web_badge =
+    'https://daily-now-res.cloudinary.com/image/upload/v1672745846/public/dailydev.png';
+  // @TODO: check what these should be
+  push.chrome_web_icon =
+    'https://daily-now-res.cloudinary.com/image/upload/s--9vc188bS--/f_auto/v1712221649/1_smcxpz';
```
omBratteng, line 122: @rebelchris could you check what these TODOs should be?

src/workers/personalizedDigestEmail.ts
```diff
@@ -141,6 +144,24 @@ const digestTypeToFunctionMap: Record<
       },
     );
   },
+  [UserPersonalizedDigestType.ReadingStreakReminder]: async (data, con) => {
```
rebelchris, line 147: Lol we should probably rename this worker 😂

src/workers/personalizedDigestEmail.ts
```diff
@@ -141,6 +144,24 @@ const digestTypeToFunctionMap: Record<
...
+    const { personalizedDigest, emailSendTimestamp, deduplicate = true } = data;
+    const emailSendDate = new Date(emailSendTimestamp);
```
rebelchris, line 149: Naming wise it might become confusing as it's not email (I know the above also has it)

src/onesignal.ts
rebelchris: @idoshamun I think you have more context on these, who do I check this with?

rebelchris (changes requested):
Just blocking for the TODO's will find out for you

src/workers/personalizedDigestEmail.ts
omBratteng: Haha, yeah. I thought about doing some renaming of variables that say email in them 🤣🙈

rebelchris: I guess also the cron runs hourly? I think so probably ok to not remove on read for now?

src/onesignal.ts
```diff
@@ -87,3 +91,35 @@ export async function sendReadingReminderPush(
...
+  };
+  push.url = addNotificationUtm(
+    process.env.COMMENTS_PREFIX,
+    'push',
+    // @TODO: check what this should be
+    'streak_reminder',
+  );
+
+  // @TODO: check what these should be
+  push.chrome_web_badge =
```
idoshamun, line 118: This one should be kept as is

src/onesignal.ts
idoshamun: we should ask design what image we should use for this. you can send them this link as reference but probably they'll come up with something related to streaks

idoshamun (commented):
We should consider renaming the notification to streak reminder instead of reading streak reminder to avoid confusion with the reading reminder

idoshamun (commented):
I don't see the logic for excluding users who already hit their streak today. i'm assuming it's because the PR is still draft

omBratteng:
> I don't see the logic for excluding users who already hit their streak today. i'm assuming it's because the PR is still draft

Yeah

src/workers/personalizedDigestEmail.ts
omBratteng: Removed email from the variable name to make it less confusin

src/onesignal.ts
```diff
@@ -81,9 +86,38 @@ export async function sendReadingReminderPush(
...
+  };
+  push.headings = {
+    en: streakReminderHeading,
+  };
+  push.url = addNotificationUtm(
+    process.env.COMMENTS_PREFIX,
+    'push',
+    // @TODO: check what this should be
+    'streak_reminder',
+  );
```
omBratteng, line 115: @idoshamun @rebelchris what should this value be? Is `streak_reminder` ok?

Example URL `http://webapp.local.com:5002/?utm_source=notification&utm_medium=push&utm_campaign=streak_reminder`

src/onesignal.ts
```diff
@@ -81,9 +86,38 @@ export async function sendReadingReminderPush(
...
+
+  push.chrome_web_badge = chromeWebBadge;
+  // @TODO: check what these should be
+  push.chrome_web_icon =
+    'https://daily-now-res.cloudinary.com/image/upload/s--9vc188bS--/f_auto/v1712221649/1_smcxpz';
```
omBratteng, line 120: [Asked design](https://dailydotdev.slack.com/archives/C02AA18RD1D/p1717668424412599)

src/cron/hourlyNotifications.ts
```diff
@@ -0,0 +1,79 @@
...
+  UserPersonalizedDigest,
+  UserPersonalizedDigestSendType,
+  UserPersonalizedDigestType,
+} from '../entity';
+import { Cron } from './cron';
+import { isWeekend, addHours, startOfHour, subDays } from 'date-fns';
+import { DEFAULT_TIMEZONE } from '../types';
+
+const sendType = UserPersonalizedDigestSendType.workdays;
+const digestTypes = [UserPersonalizedDigestType.StreakReminder];
```
omBratteng, line 17: Will move the reading reminder into this in a different PR, as I want to deliver this by EOD

src/workers/personalizedDigestEmail.ts
omBratteng: Yeah, probably 😬
Having the streak and reading reminders in personalized digest doesn't make sense. We could probably refactor it into to tables too

src/onesignal.ts
rebelchris: From my side it's ok :)

src/cron/hourlyNotifications.ts
```diff
@@ -0,0 +1,79 @@
...
+      handler: async ({
+        personalizedDigest: personalizedDigestWithTimezome,
+        emailBatchId,
+      }) => {
+        const { timezone = DEFAULT_TIMEZONE, ...personalizedDigest } =
+          personalizedDigestWithTimezome as UserPersonalizedDigest &
+            Pick<User, 'timezone'>;
+
+        // We run the script just before the hour,so we add 1 hour to the
+        // timestamp to send the email at the beginning of the hour
```
rebelchris, line 54:
```suggestion
        // timestamp to send the reminder at the beginning of the hour
```

src/cron/dailyDigest.ts
```diff
@@ -8,12 +8,17 @@ import {
...
 } from '../entity';
 import { Cron } from './cron';
 import { isWeekend, addHours, startOfHour, subDays } from 'date-fns';
 import { DEFAULT_TIMEZONE } from '../types';
 
 const sendType = UserPersonalizedDigestSendType.workdays;
+const digestTypes = [
+  UserPersonalizedDigestType.Digest,
+  UserPersonalizedDigestType.ReadingReminder,
+];
```
rebelchris, line 21: This is pretty neat and flexible moving forward to move between crons.
Isn't moving the reading reminder just moving this to the other one?

rebelchris (approved):
Let's see if we can narrow the open TODOS today

src/cron/dailyDigest.ts
omBratteng: Technically it is 😬
But I want to do it in a different PR nonetheless just to test it properly and ensure nothing is borked by it 😄

src/onesignal.ts
omBratteng:
```suggestion
  push.url = addNotificationUtm(
    process.env.COMMENTS_PREFIX,
    'push',
    'streak_reminder',
  );
```

src/workers/personalizedDigestEmail.ts
```diff
@@ -129,10 +132,59 @@ const digestTypeToFunctionMap: Record<
...
+    // but for some reason still have a streak reminder scheduled
+    if (userSettings?.optOutReadingStreak) {
+      return;
+    }
+
+    const userStreak = await con
+      .getRepository(UserStreak)
+      .findOneBy({ userId });
+
+    if (!userStreak) {
```
sshanzel, line 171: Maybe we should not send too when the streak count is 0.

src/workers/personalizedDigestEmail.ts
rebelchris: I guess the one table works, but the naming is a bit confusing maybe 🤔
(not urgent/blocking)

But more generic would probably be nice.

src/cron/hourlyNotifications.ts
```diff
@@ -0,0 +1,79 @@
...
+  handler: async (con, logger) => {
+    const personalizedDigestQuery = con
+      .createQueryBuilder()
+      .select('upd.*, u.timezone')
+      .from(UserPersonalizedDigest, 'upd')
+      .innerJoin(User, 'u', 'u.id = upd."userId"')
+      .where(
+        `clamp_to_hours(upd."preferredHour" - EXTRACT(HOUR FROM NOW() ${atTimeZone})) = 1`,
+        {
+          defaultTimezone: DEFAULT_TIMEZONE,
```
sshanzel, line 33: Do we still need this parameter?

sshanzel (approved):
Looks great overall! Just had some minor non-blocking comments/questions.

Also, I'm a bit confused why personalized digest is tightly coupled with this one. Is it due to the nature of working with timezones?

src/cron/hourlyNotifications.ts
omBratteng: Yeah, it's used in
```ts
const atTimeZone =
  "AT TIME ZONE COALESCE(NULLIF(u.timezone, ''), :defaultTimezone)";
```

src/workers/personalizedDigestEmail.ts
omBratteng: Added a check for it :)

omBratteng:
> Also, I'm a bit confused why personalized digest is tightly coupled with this one. Is it due to the nature of working with timezones?

@sshanzel Its not really coupled with it, but it's re-using the same crons and workers that personalized digest has. But yeah, it is confusing, and we probably should do a bit of refactoring here to make the names a bit more generic

src/workers/personalizedDigestEmail.ts
rebelchris: @omBratteng Do we check here if you already read today?

omBratteng: @rebelchris yes. We check if the day of current time and lastViewAt is the same here
https://github.com/dailydotdev/daily-api/blob/afacdd327e2127f649e1e4e06ceea395f2952cbc/src/workers/personalizedDigestEmail.ts#L179

src/cron/hourlyNotifications.ts
```diff
@@ -0,0 +1,79 @@
...
+          personalizedDigestWithTimezome as UserPersonalizedDigest &
+            Pick<User, 'timezone'>;
+
+        // We run the script just before the hour,so we add 1 hour to the
+        // timestamp to send the reminder at the beginning of the hour
+        const notificationSendTimestamp = addHours(timestamp, 1).getTime();
+        const sendDateInTimezone = utcToZonedTime(
+          notificationSendTimestamp,
+          timezone,
+        );
```
capJavert, line 59: Be aware that depending on number of items that need to be sent it could take more then 2 minutes to process all messages, so it might just be better to generate timestamp at this point and send immediately?

src/workers/personalizedDigestEmail.ts
```diff
@@ -129,10 +132,62 @@ const digestTypeToFunctionMap: Record<
...
+      logger.error(
+        `User streak not found for user ${personalizedDigest.userId} when sending streak reminder.`,
+      );
```
capJavert, line 174: This is not really an error, maybe info, but once this runs in prod for few days I would just move it to debug level.

capJavert (commented):
Looks good, minor comments.

src/workers/personalizedDigestEmail.ts
omBratteng: Just changed it to debug

src/onesignal.ts
idoshamun: yap sounds good

src/cron/hourlyNotifications.ts
omBratteng: Don't think it matters that much for notifications, as the timestamp is used in `send_after` param of onesignal. So even if the timestamp is 20:00, and the worker is finished at 20:03, it gets sent immediately.

capJavert: Check these [errors please](https://console.cloud.google.com/logs/query;query=labels.%22k8s-pod%2Fapp%22%3D%22api-personalized-digest%22%0Aseverity%3DERROR%0ASEARCH%2528%22Schedule%20Notifications%20may%20not%20be%20scheduled%20in%20the%20past%22%2529;summaryFields=:false:32:beginning;cursorTimestamp=2024-06-04T07:54:42.962Z;duration=P7D?project=devkit-prod&authuser=2), I think they might not accept it if it is the same one

"Schedule Notifications may not be scheduled in the past"

capJavert: If that is using the same thing as other reminder.

rebelchris: Maybe Ante's suggestion is safer right?
It won't hurt to change to direct sends anyway?

src/cron/hourlyNotifications.ts
sshanzel: Oh, cool. Then yeah, let's ignore this.

sshanzel:
> > Also, I'm a bit confused why personalized digest is tightly coupled with this one. Is it due to the nature of working with timezones?
> 
> @sshanzel Its not really coupled with it, but it's re-using the same crons and workers that personalized digest has. But yeah, it is confusing, and we probably should do a bit of refactoring here to make the names a bit more generic

Cool. Makes sense, yeah, maybe we can do some refactoring in the future. Looks great!

src/cron/hourlyNotifications.ts
omBratteng: I tested it up to 10 minutes in the past, above that and it gives the same error.
I'll change it to be immediately

src/cron/hourlyNotifications.ts
```diff
@@ -0,0 +1,72 @@
...
+        `clamp_to_hours(upd."preferredHour" - EXTRACT(HOUR FROM NOW() ${atTimeZone})) = 1`,
+        {
+          defaultTimezone: DEFAULT_TIMEZONE,
+        },
+      )
+      .andWhere(`flags->>'sendType' = :sendType`, { sendType })
+      .andWhere(`upd.type in (:...digestTypes)`, { digestTypes });
+
+    // Make sure digest is sent at the beginning of the hour
+    const timestamp = new Date().getTime();
```
capJavert, line 40: You need to move this inside `handler` right to make the time be exact at when the cron schedules?

capJavert: Like this the timestamp will still be XX:58

omBratteng: I just used the one inside the worker that the cron triggers
https://github.com/dailydotdev/daily-api/blob/ce43ed767acbc6dfd260fc36e5c5298adb9f65ab/src/workers/personalizedDigestEmail.ts#L155
https://github.com/dailydotdev/daily-api/blob/ce43ed767acbc6dfd260fc36e5c5298adb9f65ab/src/workers/personalizedDigestEmail.ts#L185

omBratteng: I moved it into the handler, though it doesn't use it for anything except checking if it is weekend. The `previousSendTimestamp`, `emailSendTimestamp` and `emailBatchId` have no effect on the worker

capJavert: It would be good to have it standardized from outside, at least `emailSendTimestamp` so it controls when worker sends for hourly. I get why you don't need `previousSendTimestamp`.

capJavert: That way the worker is not coupled with specific send schedule, it just takes the `emailSendTimestamp` and sends.

src/onesignal.ts
```diff
@@ -81,9 +88,35 @@ export async function sendReadingReminderPush(
...
+
+export async function sendStreakReminderPush(
+  userIds: string[],
+  at: Date,
+): Promise<null | OneSignal.CreateNotificationSuccessResponse> {
+  if (!appId || !apiKey) return null;
+  const push = new OneSignal.Notification();
+  push.app_id = appId;
+  push.include_external_user_ids = userIds;
+  push.send_after = at.toISOString();
```
rebelchris, line 104: Can't we just omit this one actually to send it directly?

src/cron/hourlyNotifications.ts
omBratteng: Could also do as @rebelchris suggested, omit `send_after` in the onesignal function. So we don't need the timestamp for this worker?

src/onesignal.ts
omBratteng: Take a look now, I removed it

src/cron/hourlyNotifications.ts
omBratteng: I removed the send_after config for streak reminder function, so it gets sent immediately after invocation.
Does that work?

rebelchris: Just omit it yeah.
It will send directly which I think is fine right?

capJavert: 👍

capJavert (approved):
Let's go!

PR #1970, opened 2024-06-05
refactor: source flags trigger to include machine type

sshanzel:
Should be good to review. The only difference here between each function is the removal of the `AND type = 'squad'`.

Note: it is a good sign that some tests failed, it means it worked. Will update those cases in a few seconds.


src/migration/1717589112159-SourceStatsTrigger.ts
```diff
@@ -0,0 +1,128 @@
...
+        $$
+        BEGIN
+          UPDATE  source
+          SET     flags = jsonb_set(flags, '{totalPosts}', to_jsonb(COALESCE(CAST(flags->>'totalPosts' AS INTEGER) , 0) + 1))
+          WHERE   id = NEW."sourceId"
+          AND     type = 'squad';
+          RETURN NEW;
+        END;
+        $$
+      `);
```
sshanzel, line 81: For the rollback, we just need to copy the initial function:
https://github.com/dailydotdev/daily-api/blob/722a3016a868efeda40cd1b0629a825483ff3823/src/migration/1714633817175-SquadTotalPostsTrigger.ts#L7-L21

src/migration/1717589112159-SourceStatsTrigger.ts
```diff
@@ -0,0 +1,128 @@
...
+                                COALESCE(CAST(flags->>'totalUpvotes' AS INTEGER), 0) + (NEW.upvotes - OLD.upvotes)
+                              )
+                            )
+                          )
+          WHERE   id = NEW."sourceId"
+          AND     type = 'squad';
+          RETURN NEW;
+        END;
+        $$
+      `);
```
sshanzel, line 106: For the rollback, we just need to copy the initial function:

https://github.com/dailydotdev/daily-api/blob/722a3016a868efeda40cd1b0629a825483ff3823/src/migration/1714662136871-SquadTotalUpvotesTrigger.ts#L9-L32

src/migration/1717589112159-SourceStatsTrigger.ts
```diff
@@ -0,0 +1,128 @@
...
+                            flags,
+                            '{totalViews}',
+                            to_jsonb(COALESCE(CAST(flags->>'totalViews' AS INTEGER), 0) + (NEW.views - OLD.views))
+                          )
+          WHERE   id = NEW."sourceId"
+          AND     type = 'squad';
+          RETURN NEW;
+        END;
+        $$
+      `);
```
sshanzel, line 126: For the rollback, we just need to copy the initial function:

https://github.com/dailydotdev/daily-api/blob/722a3016a868efeda40cd1b0629a825483ff3823/src/migration/1714667947050-SquadTotalViewsTrigger.ts#L7-L25

src/migration/1717589112159-SourceStatsTrigger.ts
```diff
@@ -0,0 +1,128 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceStatsTrigger1717589112159 implements MigrationInterface {
+  name = 'SourceStatsTrigger1717589112159';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    queryRunner.query(`
+        CREATE OR REPLACE FUNCTION increment_squad_posts_count()
```
rebelchris, line 8: Not sure if maybe we should delete and re-add with better naming?
`increment_source_posts_count` (but only if it doesn't add much overhead)

rebelchris (approved):
One minor suggestion.
So we just need to be careful to quickly run the bin script once we merge right?
Else the trigger will already fire before hand

sshanzel:
> So we just need to be careful to quickly run the bin script once we merge right?

Just as long as we run the script after merging, we should be good whichever time we pick since they will just be back to what they should really count.

rebelchris:
@sshanzel Oh right I guess it will accumulate whatever on that time 👍

rebelchris (approved):

nensidosari (approved):

sshanzel:
Ran through the migrations locally again, and applied to the local DB. After refresh, there are no more triggers specific to squads as shown below.

Screenshot 2024-06-06 at 4 37 04 PM

PR #1971, opened 2024-06-06
feat: move user streak out of auth, add query by id

nensidosari:


__tests__/users.ts
```diff
@@ -320,19 +320,41 @@ describe('query userStreaks', () => {
...
   });
 
-  it('should not allow unauthenticated users', () =>
-    testQueryErrorCode(client, { query: QUERY }, 'UNAUTHENTICATED'));
+  it('should throw error when no user id provided and no user is authenticated', async () => {
+    await testQueryErrorCode(
+      client,
+      { query: QUERY },
+      'GRAPHQL_VALIDATION_FAILED',
+      'User is missing!',
```
rebelchris, line 341: Not sure we need this additions as it won't ever persist to the FE

__tests__/users.ts
```diff
@@ -320,19 +320,41 @@ describe('query userStreaks', () => {
...
     loggedUser = '1';
     const res = await client.query(QUERY);
     expect(res.errors).toBeFalsy();
   });
 
+  it('should return the user streaks based on user id', async () => {
+    const resById = await client.query(QUERY_BY_USER_ID, {
+      variables: { id: '2' },
+    });
+    expect(resById.errors).toBeFalsy();
```
rebelchris, line 355: You probably also want to ensure the actual data here on top of no errors.

rebelchris (approved):
Looks pretty solid!
Minor comment on tests, but non blocking.

Just wait for Ante's review as well 🙏

__tests__/users.ts
nensidosari: The message addition u mean? can remove

rebelchris: Yes, don't think it will add much if we just use it for test cases basically.

__tests__/users.ts
nensidosari: That will always return the default data, added that test. There is also a test doing just that on https://github.com/dailydotdev/daily-api/blob/cca8714abe73b7efcaf653684b9225e6095f4a74/__tests__/users.ts#L428.

sshanzel (approved):
Great work! 🥳

__tests__/users.ts
rebelchris: Perfect, just extra failsafe if someone changes defaults by accident :)

src/schema/users.ts
```diff
@@ -941,7 +954,9 @@ export const resolvers: IResolvers<any, Context> = {
             )
             .addSelect('u.timezone', 'timezone')
             .innerJoin(User, 'u', `"${builder.alias}"."userId" = u.id`)
-            .where(`"${builder.alias}"."userId" = :id`, { id: ctx.userId }),
+            .where(`"${builder.alias}"."userId" = :id`, {
+              id: id ?? ctx.userId,
+            }),
```
capJavert, line 959: Mix and matching between auth and non auth seems complicated, can we maybe just add another userStreakProfile query which will return only stuff needed for profile? It will also avoid below `hasClearedStreak` check.

capJavert: Because even now in tests we are testing for cases where the above can break 😄

src/schema/users.ts
```diff
@@ -963,6 +973,24 @@ export const resolvers: IResolvers<any, Context> = {
...
+      ctx: Context,
+      info,
+    ): Promise<GQLUserStreak> => {
+      const streak = await getUserStreakQuery(id, ctx, info);
+
+      if (!streak) {
+        return {
+          max: 0,
+          total: 0,
+          userId: ctx.userId,
```
capJavert, line 988: this should be `id` right?

capJavert (approved):
Love it, just one check! 🚀

omBratteng (approved):

src/schema/users.ts
nensidosari: 🤦‍♀️ yess

PR #1972, opened 2024-06-06
fix: check if member exists before running joined logic

capJavert:
[Thread](https://dailydotdev.slack.com/archives/C014VT1J59V/p1717652642933269)


rebelchris (approved):
Thanks for the fix and test case

PR #1975, opened 2024-06-07
chore: temp fixes for FF issues

rebelchris:
Temp send FF users to webapp as extension doesn't work
And add message on old auth system to inform people.


omBratteng (approved):
AAAAAAAAAAAAA

PR #1976, opened 2024-06-07
fix: add roadmap flag to keywords

rebelchris:
Ability to public expose roadmap flags

### DIFF
--- a/src/entity/Keyword.ts
+++ b/src/entity/Keyword.ts
@@ -12,9 +12,13 @@ export type KeywordFlags = Partial<{
   onboarding: boolean;
   title: string;
   description: string;
+  roadmap: string;
 }>;
 
-export type KeywordFlagsPublic = Pick<KeywordFlags, 'title' | 'description'>;
+export type KeywordFlagsPublic = Pick<
+  KeywordFlags,
+  'title' | 'description' | 'roadmap'
+>;
 
 @Entity()
 @Index('IDX_status_value', ['status', 'value'])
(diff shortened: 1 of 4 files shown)


src/schema/keywords.ts
```diff
@@ -46,6 +46,11 @@ export const typeDefs = /* GraphQL */ `
     Description of the keyword
     """
     description: String
+
+    """
+    Roadmap.sh link for the keyword
+    """
+    roadmap: String
```
capJavert, line 53: Its a string but maybe we want to link the slug and then add domain through field resolver just so we don't hardcode it inside the db?

src/entity/Keyword.ts
```diff
@@ -12,9 +12,13 @@ export type KeywordFlags = Partial<{
   onboarding: boolean;
   title: string;
   description: string;
+  roadmap: string;
 }>;
 
-export type KeywordFlagsPublic = Pick<KeywordFlags, 'title' | 'description'>;
+export type KeywordFlagsPublic = Pick<
+  KeywordFlags,
+  'title' | 'description' | 'roadmap'
```
capJavert, line 20: Just update the tests as well 🙏

capJavert (approved):

src/entity/Keyword.ts
capJavert: https://github.com/dailydotdev/daily-api/blob/10896baa0d508cd4e2fe52064d916bb0a0dc3e43/__tests__/keywords.ts#L362

nensidosari (approved):

PR #1977, opened 2024-06-10
fix: handle properly duplicate emails

idoshamun:
Found several edge cases which may lead to duplicate emails


vpol (approved):

rebelchris (approved):
Good find, guess we only handled it on the checkEmail call before (and handles)

PR #1978, opened 2024-06-10
feat(cio): webhook move to delete path instead of method

capJavert:
AS-369

cio does not support `DELETE` method so opted for `/delete` path. Alternative to use custom header can be limited by header size (because we will be sending JSON) so I think this is easier and in the future we can easily switch back the method if cio supports it.


idoshamun (approved):

rebelchris (approved):
Thanks!

PR #1983, opened 2024-06-11
feat: add onboarding checklist setting

capJavert:
Add setting where we can save onboarding checklist state

Frontend PR: https://github.com/dailydotdev/apps/pull/3203


nensidosari (approved):
Looks good!

capJavert:
> With db values how will we override it on the FE side for experiment users?

@rebelchris I will update settings from FE during registration (as we did for closed sidebar experiment).

rebelchris:
@capJavert Cool that works for now.
We'll just have to keep in mind what we'll do when it wins/if we need to remove it as it won't be as easy as turning experiment off I guess.

PR #1986, opened 2024-06-13
feat: add longest streak to devcard

nensidosari:
- adding longest streak to the devcard, until now we're using only getDevCardData, not getDevCardDataV1
- related PR https://github.com/dailydotdev/apps/pull/3216 

Can't remember if we'd do anything else about image generation with this version


idoshamun:
Should it be longest or active streak?

nensidosari:
@idoshamun longest says the specs https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/919470082/Reading+streaks+follow+up

rebelchris (approved):

src/common/devcard.ts
```diff
@@ -82,18 +84,23 @@ export async function getDevCardDataV1(
...
+      select: ['maxStreak'],
+    }),
   ]);
   return {
     user,
     articlesRead,
     tags,
     sourcesLogos: sources.map((source) => source.image),
     rank,
+    longestStreak: streak?.maxStreak ?? 0,
```
capJavert, line 103: Can we just make it `maxStreak` as in streak column to avoid confusion in naming?

src/common/devcard.ts
```diff
@@ -128,13 +136,18 @@ export async function getDevCardData(
       })
     ).map(({ value }) => value),
     getFavoriteSources(con, userId),
+    await con.getRepository(UserStreak).findOneOrFail({
```
capJavert, line 139: you don't need `await` here

__tests__/devcard.ts
```diff
@@ -331,6 +333,7 @@ describe('query devCard(id)', () => {
       isProfileCover: true,
       showBorder: false,
       articlesRead: 0,
+      longestStreak: 0,
```
capJavert, line 336: Can we also add a test where strak is non 0

src/common/devcard.ts
nensidosari: Sure

src/common/devcard.ts
nensidosari: removing

sshanzel (approved):
Don't have anything to add on top of Ante's comments 🥳

src/common/devcard.ts
sshanzel: On that note, we can actually avoid making any changes here, we can just add the existing `userStreak` query to the same call.

capJavert: Yeah, good idea @sshanzel, I agree 👍

nensidosari:
Closing in favour of directly userStreak

PR #1988, opened 2024-06-13
feat: users leaderboard page

omBratteng:
AS-358

### DIFF
--- a/src/schema/leaderboard.ts
+++ b/src/schema/leaderboard.ts
@@ -8,7 +8,7 @@ import { User, UserStreak } from '../entity';
 
 export type GQLUserLeaderboard = {
   score: number;
-  user: GQLUser;
+  user: GQLUser | Promise<GQLUser>;
 };
 
 export const typeDefs = /* GraphQL */ `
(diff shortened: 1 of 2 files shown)


src/schema/leaderboard.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  extend type Query {
+    """
+    Get the users with the highest reputation
+    """
+    highestReputation(
+      """
+      Limit the number of users returned
+      """
+      limit: Int
```
capJavert, line 29: Add range for `limit` so someone can't say 100k or something big.

src/entity/user/UserStats.ts
```diff
@@ -0,0 +1,60 @@
...
+          FROM "comment" c
+          WHERE c."userId" = u."id"
+        )`,
+        'commentUpvotes',
+      )
+      .from('user', 'u')
+      .andWhere('u.infoConfirmed = TRUE')
+      .andWhere("u.id != '404'"),
+})
+export class UserStats {
```
capJavert, line 45: Added MV based on query that Ole-Martin started working on, double check the data and that it returns what we expect.

rebelchris (approved):
Should we consider some select criteria?
Also wondering if we should just combine comment and post upvotes in one in the mv? (But open to this as well)

capJavert:
@rebelchris it is combined in one MV but uses SELECT, I might move it to `WITH` so I can reuse nested queries.

rebelchris:
Meant more for the other queries that ole-martin amreay added for select minimizing 👍

capJavert:
I think for the current ones there is no point since they are light but we can test how it performs.

src/entity/user/UserStats.ts
capJavert: Looks good, will double check in production.

PR #1993, opened 2024-06-17
feat: sync subscriptions with cio and back

idoshamun:
As part of our effort to solve to unsubscribe issues we need to sync updates to and from cio.

Note:  I wanted to introduce new hook to deal with the subscription thing to not have one big hook for everything but I didn't manage to get it work properly

### DIFF
--- a/src/common/mailing.ts
+++ b/src/common/mailing.ts
@@ -1,7 +1,13 @@
-import { APIClient, SendEmailRequest } from 'customerio-node';
+import { APIClient, SendEmailRequest, TrackClient } from 'customerio-node';
 import { SendEmailRequestOptionalOptions } from 'customerio-node/lib/api/requests';
 import { SendEmailRequestWithTemplate } from 'customerio-node/dist/lib/api/requests';
 import { signJwt } from '../auth';
+import { DataSource } from 'typeorm';
+import {
+  User,
+  UserPersonalizedDigest,
+  UserPersonalizedDigestType,
+} from '../entity';
 
 export enum UnsubscribeGroup {
   Notifications = 'notifications',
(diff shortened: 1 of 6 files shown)


src/common/mailing.ts
```diff
@@ -46,6 +52,52 @@ export const baseNotificationEmailData: SendEmailRequestOptionalOptions = {
...
+    return;
+  }
+
+  const atts = await cioApi.getAttributes(userId);
+  const subs = JSON.parse(
+    atts?.customer?.attributes?.cio_subscription_preferences || '{}',
+  );
+  const marketing = isSubscribed(subs, '4');
+  const notifications = isSubscribed(subs, '7');
+  const digest = isSubscribed(subs, '8');
```
rebelchris, line 85: Not blocking but would be great if we can ENUM these values somewere?
I envision we'll do other stuff with it in the future and magic numbers are not super clear.

rebelchris (approved):
Minor suggestion, but non blocking

src/schema/users.ts
```diff
@@ -1300,6 +1302,8 @@ export const resolvers: IResolvers<any, Context> = {
         flags,
       });
 
+      await resubscribeUser(cio, ctx.userId);
```
capJavert, line 1305: We subscribe to everything or?

src/routes/webhooks/customerio.ts
```diff
@@ -141,22 +162,14 @@ export const customerio = async (fastify: FastifyInstance): Promise<void> => {
...
-        visit_id: payload.event_id,
-        user_id: userId,
-        event_name: `${payload.object_type} ${payload.metric}`,
-        app_platform: 'customerio',
-        extra: JSON.stringify(payload.data),
-      };
-      await sendAnalyticsEvent([event]);
+
+      // There are multiple events that should trigger the subscription sync
+      if (payload.metric.includes('subscri')) {
```
capJavert, line 167: Feels error prone, do we control the metric values?

capJavert (approved):

src/routes/webhooks/customerio.ts
idoshamun: nope, do you rather see hard coded values? i'm just afraid to miss some :triggered:

src/schema/users.ts
idoshamun: cio remembers the last state, so i'm just removing the user from the global unsubscribe because they clearly want to receive email at this point

src/routes/webhooks/customerio.ts
capJavert: I would, but I also don't know how many of those is there.

src/common/mailing.ts
idoshamun: done

src/routes/webhooks/customerio.ts
idoshamun: done

PR #1994, opened 2024-06-17
refactor: remove custom unsubscribe

idoshamun:
As we're going to manage subscription via cio, we no longer need this.

I'll merge once marketing changed all the emails accordingly.


capJavert (approved):
Nice!

PR #1998, opened 2024-06-19
fix: replace profile fallback image

omBratteng:
<table>
 <tr>
  <td>Old
  <td>New
 <tr>
  <td>
  <td>
</table>

The migration will affect 87756 rows

AS-389


idoshamun (commented):
That's gonna trigger hell lot of cdc messages so we have to deal with care. Should we consider not to update existing users?

omBratteng:
> That's gonna trigger hell lot of cdc messages so we have to deal with care. Should we consider not to update existing users?

YOLO? 😈

idoshamun:
it's better to merge in low time if you want to proceed with that

rebelchris (approved):
We should really consider just setting it null and having the app itself do fallback 😂
That makes it way easier to switch.

Also happy to migrate this as a single query (think in the past we used to be able to disable CDC temporary somehow right @idoshamun ?)

idoshamun:
Having profile image as null has more implications than this. We will need to make sure we support it everywhere, otherwise things will start break. If we want to fix it as part of this scope we can do it, but it's more work

omBratteng:
I could remove the migration and run it directly, but in batches to a few thousand to not overwhelm CDC?

idoshamun:
sounds ok!

idoshamun (approved):

rebelchris (approved):

PR #2001, opened 2024-06-20
fix: ensure v2 search only searches on title

rebelchris:
Didn't explicitly test it locally due to long replication setup, but follows documentation implementation.
@vpol is there any quick way to validate this only search on title compared to before?

AS-391 #done


idoshamun (approved):

omBratteng (approved):

vpol:
@rebelchris you can search for this

Screenshot 2024-06-20 at 14 45 01

PR #2002, opened 2024-06-21
fix: validate timezone input

omBratteng:
Adds validation for timezones. Users are now able to send in their own value, with would break the digest cron jobs as postgres expects valid timezones.

AS-394

### DIFF
--- a/src/entity/user/User.ts
+++ b/src/entity/user/User.ts
@@ -32,6 +32,7 @@ import {
 import { generateTrackingId } from '../../ids';
 import { UserStreak } from './UserStreak';
 import { DEFAULT_TIMEZONE } from '../../types';
+import { validateValidTimeZone } from '../../common';
 
 @Entity()
 @Index('IDX_user_lowerusername_username', { synchronize: false })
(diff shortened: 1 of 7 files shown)


src/entity/user/User.ts
```diff
@@ -398,6 +399,15 @@ export const validateUserUpdate = async (
     );
   }
 
+  if ('timezone' in data) {
+    const isValidTimeZone = validateValidTimeZone(data.timezone);
+    if (!isValidTimeZone) {
+      throw new ValidationError(
+        JSON.stringify({ timezone: 'invalid timezone' }),
+      );
```
rebelchris, line 407: We just need to check that on FE we capture this (although very unlikely to happen with a natural flow?)

rebelchris (approved):

src/entity/user/User.ts
omBratteng: We don't at the moment. But yeah, it wouldn't happen in a natural flow.
I was just looking at it, I can add a error toast to it in FE.

rebelchris: Ah since it's already a edge-case don't bother for now I think :)

PR #2003, opened 2024-06-21
feat: new CTA variant

omBratteng:
I forgot, but I think this is all that's needed on API side to add new CTA variant 😅 

AS-396


rebelchris (approved):
Looks about right yes.
Nothing in the endpoint that validates based on graphQL? (we probably use toGraphEnum so it's fine)

omBratteng:
> Nothing in the endpoint that validates based on graphQL? (we probably use toGraphEnum so it's fine)

We don't use it there at all, there's just a mutation and the mutation just takes the campaignId.

rebelchris:
@omBratteng But what about the one that adds them to the DB?

omBratteng:
> @omBratteng But what about the one that adds them to the DB?

@rebelchris I've added the new CTA to the retool app. And from CIO it just uses campaignId and userId. Nothing about the variant. So we should be good to go.

PR #2005, opened 2024-06-21
fix: trying to fix some flaky tests

rebelchris:
It doesn't fix the main one we have :( (can't find what could cause the source id to change, maybe the parralism? although this shouldn't be affected from what I can read per suite)

Does fix some other flaky tests we had (same named tests basically is never a good idea)

Tried some random flaky test package and it didn't even find the main one. (So makes me wonder it's really due to parralism :))

Others found though:
Screenshot 2024-06-21 at 15 33 19

### DIFF
--- a/__tests__/keywords.ts
+++ b/__tests__/keywords.ts
@@ -377,7 +377,7 @@ describe('mutation setKeywordAsSynonym', () => {
   });
 });
 
-describe('flags field', () => {
+describe('keywords flags field', () => {
   const QUERY = `{
     keyword(value: "react") {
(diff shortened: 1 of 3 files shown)


nensidosari:
Wait, u say it fixes some flaky test? but the changes is just in the test descriptions?

rebelchris:
@nensidosari Yeah the had the same name over multiple suites basically.
Now the describe makes them unique.

Not a 100% sure, but think in some instances this will cause flakiness on these (as most of the test inside these describes where flaky)

PR #2006, opened 2024-06-23
fix: fetch sources using our best practices

idoshamun:
When we introduced the source directory, we did a workaround to overcome graphorm issue. But now I fixed the queries to follow our best practices

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -1091,7 +1091,6 @@ const getFormattedSources = async <Entity>(
         'ts',
         `ts."sourceId" = ${builder.alias}.id`,
       )
-      .where({ active: true, type: SourceType.Machine })
       .orderBy('r', 'DESC')
       .limit(limit);
     return builder;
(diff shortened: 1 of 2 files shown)


src/schema/sources.ts
```diff
@@ -1075,20 +1076,26 @@ const togglePinnedPosts = async (
...
-    .getRawMany();
+
+  return await graphorm.query(ctx, info, (builder) => {
+    builder.queryBuilder
+      .innerJoin(
+        ctx.con.getMetadata(entity).tableName,
+        'ts',
+        `ts."sourceId" = ${builder.alias}.id`,
+      )
+      .where({ active: true, type: SourceType.Machine })
```
rebelchris, line 1094: I thought we didn't really care to showcase public squads here as well?

rebelchris (approved):
Oh, wait it was already possible 🤔

src/schema/sources.ts
idoshamun: It was by mistake 🙈

rebelchris (approved):

PR #2009, opened 2024-06-26
feat: add reading streak to cio

rebelchris:
Added new topic to update when reading streak changes.
Also added subscription to update custom io based on streak data.

Data in action:
Screenshot 2024-06-26 at 10 46 21

Stream needed:
https://github.com/dailydotdev/streams/pull/32

AS-407 #done


src/cron/updateCurrentStreak.ts
```diff
@@ -0,0 +1,41 @@
+import { Cron } from './cron';
+import { User, UserStreak } from '../entity';
+import { checkAndClearUserStreak } from '../common';
+
+const cron: Cron = {
+  name: 'update-current-streak',
+  handler: async (con, logger) => {
+    console.log('opened cron');
```
omBratteng, line 8:
```suggestion
```

src/cron/updateCurrentStreak.ts
```diff
@@ -0,0 +1,40 @@
+import { Cron } from './cron';
+import { User, UserStreak } from '../entity';
+import { checkAndClearUserStreak } from '../common';
+
+const cron: Cron = {
+  name: 'update-current-streak',
```
omBratteng, line 6: Forgot to add the cron to infra so it gets triggered?

src/common/typedPubsub.ts
```diff
@@ -66,6 +66,9 @@ export type PubSubSchema = {
     userId: string;
     validUntil: string; // ISO 8601 str
   };
+  'api.v1.user-streak-updated': {
+    request: ChangeObject<UserStreak>;
```
idoshamun, line 70: Should we call this property streak?

rebelchris: Done

src/cron/updateCurrentStreak.ts
```diff
@@ -0,0 +1,40 @@
...
+          .innerJoin(User, 'u', 'u.id = us."userId"')
+          .where(`us."currentStreak" != 0`)
+          .andWhere(
+            `(date_trunc('day', us. "lastViewAt" at time zone COALESCE(u.timezone, 'utc'))::date) < (date_trunc('day', now() at time zone COALESCE(u.timezone, 'utc'))::date) - interval '1 day' `,
+          )
+          .getRawMany();
+
+        await Promise.all(
+          usersPastStreakTime.map(async (userStreak) => {
+            return await checkAndClearUserStreak(
```
idoshamun, line 25: This function runs already the query, which means if we have 10k users to reset the streak, we will have 10k queries. We can instead keep an array of user IDs to reset and then do it in a single query or batch

src/cron/updateCurrentStreak.ts
```diff
@@ -0,0 +1,40 @@
+import { Cron } from './cron';
+import { User, UserStreak } from '../entity';
+import { checkAndClearUserStreak } from '../common';
+
+const cron: Cron = {
+  name: 'update-current-streak',
+  handler: async (con, logger) => {
+    try {
+      await con.transaction(async (entityManager): Promise<void> => {
+        const usersPastStreakTime = await entityManager
```
idoshamun, line 10: Should we consider streaming the results?

idoshamun: Maybe at the current scale it's fine not to. So your call

src/cron/updateCurrentStreak.ts
rebelchris: Hmm yeah we can extract to a `checkUserStreak` and seperate query for clearing here?

src/cron/updateCurrentStreak.ts
rebelchris: I actually thought about it, but noticed we never really do for crons only bin scripts.
Happy to change it though.

src/cron/updateCurrentStreak.ts
idoshamun: Yes, would be better to separate concerns

src/cron/updateCurrentStreak.ts
idoshamun: Let's add a metric maybe instead of many users are being returned from this query and then also how many we reset from. Since we run it every 30mins the batch should be relatively small I believe which is great

rebelchris: That sounds good, it should be minimal past the first I think

rebelchris: Added a counter let me know if it works. (CC: @omBratteng )
And rewrote the cron logic :)

src/cron/updateCurrentStreak.ts
```diff
@@ -0,0 +1,59 @@
...
+          .from(UserStreak, 'us')
+          .innerJoin(User, 'u', 'u.id = us."userId"')
+          .where(`us."currentStreak" != 0`)
+          .andWhere(
+            `(date_trunc('day', us. "lastViewAt" at time zone COALESCE(u.timezone, 'utc'))::date) < (date_trunc('day', now() at time zone COALESCE(u.timezone, 'utc'))::date) - interval '1 day' `,
+          )
+          .getRawMany();
+
+        const userIdsToReset = [];
+        usersPastStreakTime.map(async (userStreak) => {
```
idoshamun, line 30: Not an async anymore, and forEach is a better alternative in this case

```suggestion
        usersPastStreakTime.forEach((userStreak) => {
```

src/cron/updateCurrentStreak.ts
```diff
@@ -0,0 +1,59 @@
...
+            userIdsToReset.push(userStreak.userId);
+          }
+        });
+
+        if (!userIdsToReset.length) {
+          logger.info('no user streaks to reset');
+          return;
+        }
+
+        const updateResult = await con
```
idoshamun, line 41: Shouldn't we use the clear streak function here to keep the logic in one place?

rebelchris: Ah missed this, I can refactor that one to take multiple yes :)

idoshamun (changes requested):
I totally missed that we didn't add the streak table to debezium!

idoshamun (commented):
we also need to add the migration for the replica identity?

idoshamun (approved):

PR #2016, opened 2024-06-28
fix: update to queries to avoid writes

rebelchris:
We detected that in the current streak system we always update each row in `user_streak` with a new `lastViewAt` and `updatedAt` this has heavy implications on the writes on this table.

In general it's not that bad, but when we add debezium to this table/cdc it overloads that system.

We decided to do two separate queries
- Query if streak should be updated
- If yes: Run update

This results in only 1 write per user a day.

Once approved I'll adjust the CDC flow again to bring it back.

For reference the old PR has notes around what's implemented already:
https://github.com/dailydotdev/daily-api/pull/1653

### DIFF
--- a/__tests__/workers/newView.ts
+++ b/__tests__/workers/newView.ts
@@ -162,8 +162,8 @@ describe('reading streaks', () => {
     const streak = await con.getRepository(UserStreak).findOne({
       where: {
         userId: 'u1',
-        lastViewAt: new Date(currentDate),
       },
+      order: { lastViewAt: 'DESC' },
     });
     for (const key in expectedStreak) {
       expect(streak?.[key]).toEqual(expectedStreak[key]);
@@ -315,7 +315,7 @@ describe('reading streaks', () => {
   it('should not increment a reading streak if lastViewAt is the same day', async () => {
     await runTest(
       '2024-01-26T17:23Z',
-      '2024-01-26T17:23Z',
+      '2024-01-26T15:23Z',
       defaultStreak,
       defaultStreak,
     );
--- a/__tests__/workers/newView.ts
+++ b/__tests__/workers/newView.ts
@@ -313,10 +313,12 @@ describe('reading streaks', () => {
   });
 
   it('should not increment a reading streak if lastViewAt is the same day', async () => {
-    await runTest('2024-01-26T19:17Z', '2024-01-26T17:23Z', defaultStreak, {
-      ...defaultStreak,
-      lastViewAt: new Date('2024-01-26T19:17Z'),
-    });
+    await runTest(
+      '2024-01-26T17:23Z',
+      '2024-01-26T17:23Z',
+      defaultStreak,
+      defaultStreak,
+    );
   });
 
   describe('showMilestone is set correctly', () => {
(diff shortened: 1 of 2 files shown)


__tests__/workers/newView.ts
```diff
@@ -408,7 +410,7 @@ describe('reading streaks', () => {
 
     it('should not set showStreakMilestone if lastViewAt is the same day', async () => {
       await runTest(
-        '2024-01-26T19:17Z',
+        '2024-01-26T17:23Z',
```
rebelchris, line 413: Needed since the new system doesn't update the lastView if it's same day

src/workers/newView.ts
```diff
@@ -117,76 +106,78 @@ const worker: Worker = {
...
-        con.getRepository(View).create({
-          postId: data.postId,
-          userId: data.userId,
-          referer: data.referer,
-          timestamp: data.timestamp && new Date(data.timestamp as string),
-        }),
-      );
-      if (!didSave) {
-        logger.debug(
+    await con.transaction(async (manager: EntityManager) => {
```
rebelchris, line 109: Wrapped everything in transaction for safety

src/workers/newView.ts
```diff
@@ -25,46 +33,35 @@ const addView = async (con: DataSource, entity: View): Promise<boolean> => {
   return false;
 };
 
-const DEFAULT_TIMEZONE = 'UTC'; // in case user doesn't have a timezone set
-const INC_STREAK_QUERY = `
```
rebelchris, line 29: This is the one that got removed :)

nensidosari (approved):
Looks good, can't find anything that might break the existing implementation.

src/workers/newView.ts
```diff
@@ -73,42 +70,34 @@ const incrementReadingStreak = async (
...
-        maxStreak: 1,
+  const shouldIncrementResult = await shouldIncrementStreak(
+    manager,
+    userId,
+    viewTime,
+  );
+  const { currentStreak, totalStreak, maxStreak, shouldIncrement } =
+    shouldIncrementResult ?? {};
+  if (shouldIncrement) {
+    const newCurrentStreak = currentStreak + 1;
```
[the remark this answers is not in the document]
rebelchris: Ah this was the only one that was re-used so wanted it to be one source of truth if that makes sense?

__tests__/workers/newView.ts
```diff
@@ -313,10 +313,12 @@ describe('reading streaks', () => {
   });
 
   it('should not increment a reading streak if lastViewAt is the same day', async () => {
-    await runTest('2024-01-26T19:17Z', '2024-01-26T17:23Z', defaultStreak, {
-      ...defaultStreak,
-      lastViewAt: new Date('2024-01-26T19:17Z'),
-    });
+    await runTest(
+      '2024-01-26T17:23Z',
+      '2024-01-26T17:23Z',
```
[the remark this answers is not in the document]
rebelchris: I'll add a failing test as well then :)

sshanzel (approved):
Looks amazing. Don't have anything to add on top of what Nensi has already mentioned 🥳

omBratteng (approved):

PR #2018, opened 2024-06-28
fix: add analytics to track streak resets

rebelchris:
@idoshamun Let me know if this approach works.
Noticed we don't have API only tracking (only the parsed ones)

Also wasn't sure if more fields where needed here.


idoshamun (approved):

src/common/users.ts
```diff
@@ -377,6 +378,13 @@ export const clearUserStreak = async (
   con: DataSource | EntityManager,
   userIds: string[],
 ): Promise<number> => {
+  const events = userIds.map((userId) => ({
```
idoshamun, line 381: Are we sure all these users are being affected before we send the event?
Should we consider moving it to CDC?

idoshamun (changes requested):

src/common/users.ts
rebelchris: I mean here the query will always set.
We do checks before to see it needs to be cleared or is not 0.

idoshamun (approved):

PR #2020, opened 2024-06-28
fix: we emptied to many people

rebelchris:
The evaluation was missing the `current` !== 0 meaning we still spooled empty users


src/common/users.ts
```diff
@@ -419,7 +419,7 @@ export const checkAndClearUserStreak = async (
   info: GraphQLResolveInfo,
   streak: GQLUserStreakTz,
 ): Promise<boolean> => {
-  if (checkUserStreak(streak)) {
+  if (strak.checkUserStreak(streak)) {
```
omBratteng, line 422: `strak`.

omBratteng (approved):

PR #2021, opened 2024-07-01
refactor: sync email subscriptions via cron

omBratteng:
The fastest and easiest solution to do it in batches and not exceed CIOs rate limit.


src/common/mailing.ts
```diff
@@ -69,36 +75,47 @@ export const resubscribeUser = async (
...
-  const atts = await cioApi.getAttributes(userId);
-  const subs = JSON.parse(
-    atts?.customer?.attributes?.cio_subscription_preferences || '{}',
+  // This is fetched from APIClient source code
+  // TODO: Remove this once APIClient supports fetching attributes in bulk
+  const request = new CIORequest(process.env.CIO_APP_KEY, {});
+  const userAttributes = await request.post(
+    `${RegionUS.apiUrl}/customers/attributes`,
+    { ids: userIds },
   );
```
omBratteng, line 91: The function didn't exist in their APIClient, so I extracted the same way they do it internally.
If it is alright with @rebelchris, I'll open a PR with CIO to add it. Probably after I'm done with the notifications endpoint, just to get that out of the way first.

src/common/mailing.ts
```diff
@@ -69,36 +75,47 @@ export const resubscribeUser = async (
...
-  const digest =
-    isSubscribed(subs, CioUnsubscribeTopic.Digest) && !unsubscribed;
+
   await con.transaction(async (manager) => {
-    await manager
-      .getRepository(User)
-      .update(
-        { id: userId },
-        { notificationEmail: notifications, acceptedMarketing: marketing },
+    userAttributes?.customers.forEach(async (customer) => {
```
omBratteng, line 94: Was a bit unsure what best practice would be here. I couldn't find any way of bulk updating users via TypeORM, and would probably have to use raw SQL for it.
So I went with the easiest solution, forEach loop inside the transaction.

src/cron/syncSubscription.ts
```diff
@@ -0,0 +1,29 @@
...
+  handler: async (con) => {
+    logger.info('syncing subscriptions');
+
+    while ((await getRedisListLength(redisKey)) > 0) {
+      // Store it in a set to remove duplicates
+      const userIds = new Set(await popFromRedisList(redisKey, 100));
+
+      await syncSubscription([...userIds], con);
+
+      logger.info(`synced subscriptions for ${userIds.size} users`);
```
omBratteng, line 24:
```suggestion
      await syncSubscription([...userIds], con);
      await new Promise((resolve) => setTimeout(resolve, 1000));

      logger.info(`synced subscriptions for ${userIds.size} users`);
```

Probably could add this too, just to ensure we don't exceed the rate limit of 10 requests per second. In case the the list has >1000 entries.

.infra/crons.ts
```diff
@@ -100,4 +100,8 @@ export const crons: Cron[] = [
     name: 'update-current-streak',
     schedule: '30 * * * *',
   },
+  {
+    name: 'sync-subscription',
+    schedule: '*/5 * * * *',
```
capJavert, line 105: Maybe something more specific to what it does, subscriptions is a bit broad term. Also does it manage to execute in 5min, we don't want them to overlap?

src/common/mailing.ts
capJavert: Thats fine, just see if we need to batch (as we do in scripts) so we don't have a large transaction that is gonna clog other updates.

src/cron/syncSubscription.ts
```diff
@@ -0,0 +1,29 @@
...
+  StorageKey.Reporting,
+  'global',
+);
+
+const cron: Cron = {
+  name: 'sync-subscription',
+  handler: async (con) => {
+    logger.info('syncing subscriptions');
+
+    while ((await getRedisListLength(redisKey)) > 0) {
```
capJavert, line 18: Can we make this finite somehow, `while` loops are always scary.

capJavert (commented):
Some general notes, not 100% what exact issue we are solving?

src/common/mailing.ts
omBratteng: At most, it'll only do 100 updates, as that's the max length of the userIds list (and response from customer.io)

omBratteng:
> not 100% what exact issue we are solving?

Yesterday morning, we got a bunch of calls to the reporting CIO webhook, which in turn does a call back to CIO. And we exceeded their rate limit of 10 req/s, resulting in lots of 500 errors in our API and alerts triggering.

So this takes care of it, by deferring the updating of subscription status to a cron job where it can do it in batches.

.infra/crons.ts
omBratteng: I renamed it. And I think 5 minutes is enough, can adjust it once in production and see how long an execution takes.

rebelchris (commented):
In general I'm not really sure if this is the only blocker.
I think we also need to batch maybe the request we SEND to them? (@idoshamun wdyt?)

omBratteng:
> I think we also need to batch maybe the request we SEND to them? (@idoshamun wdyt?)

What do you mean? I've replaced the call to CIO to use the endpoint that accepts up to 100 users, i.e. doing batches of up to 100 at a time?

rebelchris:
@omBratteng It's only the incoming webhook right?
We also send data to them, so just wondering if we also need to batch those up? (not sure if those affect it)

It includes:
- identify
- identifyStreak
- tracking via reporting endpoint

omBratteng:
We haven't encountered any rate limiting on those calls just yet, but if we deem it necessary, I can take it as a another task?

omBratteng:
Also, just for information, the identify endpoint isn't rate limited by CIO.

src/cron/syncSubscription.ts
omBratteng: Added a bit more robustness to break out of the loop. Is that alright?

src/common/mailing.ts
capJavert: Ok, note it in the comment above code and link docs just so next person is aware 👍

.infra/crons.ts
capJavert: 👌

src/cron/syncSubscription.ts
capJavert: What did you add, I don't see anything? (might be github lagging)

src/common/mailing.ts
omBratteng: Done.

src/cron/syncSubscription.ts
omBratteng: Forgot to push 🙈

omBratteng: Now they're pushed

rebelchris (approved):

omBratteng:
Added tests for the cron

omBratteng:
And a test for syncSubscription

src/cron/syncSubscriptionWithCIO.ts
```diff
@@ -0,0 +1,38 @@
...
+        // Store it in a set to remove duplicates
+        const userIds = new Set(userIdsRaw);
+
+        await syncSubscription([...userIds], con);
+
+        logger.info(`synced subscriptions for ${userIds.size} users`);
+      } catch (err) {
+        logger.error({ err }, 'error syncing subscriptions');
+        break;
+      }
```
capJavert, line 33: The error thrown would break the loop either way so it more fixes the logging part but if for any reason redis keys don't get popped properly in `popFromRedisList` we would still get infinite loop. 

Is there a way to get the keys before hand and then execute in a for loop?

Or just limit amount of keys we process in single cron (since it can only take 5 minutes), for example 1000 and solve it like that.

src/routes/webhooks/customerio.ts
```diff
@@ -208,8 +210,10 @@ export const customerio = async (fastify: FastifyInstance): Promise<void> => {
       const payload = req.body;
 
       if (subscriptionMetrics.includes(payload.metric)) {
-        const con = await createOrGetConnection();
-        await syncSubscription(payload.data.identifiers.id, con);
+        pushToRedisList(
```
capJavert, line 213: Additionally, redis can always get cleared at any point so if that happens we lose some subscriptions to sync? Giant edge case but just so we are aware of it?

rebelchris: I also raised this one before we started. It needs to probably be validated by @idoshamun we ok with it. 
(As it "never" happened so far "knock on wood")

omBratteng: I'd say it's a quick for to stop the alerts and 500 errors we are currently getting lots of.
@rebelchris you also mentioned that @sshanzel might implement temporal? Once that's done, we could probably hook into that.

Till then I think using redis is fine, it hadn't been cleared for ages (years even) until we downscaled it. And we're talking about loss of possibly 5 minutes

src/cron/syncSubscriptionWithCIO.ts
omBratteng: Cron jobs have a max runtime of 5 minutes, but I've set it do 4 minutes for this. So it will not run forever.

omBratteng: I don't really want to get the entries beforehand, as there is a possibility that the job gets killed (it's k8s after all) before it has executed all of the pre-fetched, which will then make the entries lost

omBratteng: Another solution could be to make it into a function, and at the end, run `llen` to get length of list. And if it is >0, call the same function again. But that's just recursion loop

capJavert: Agree the redis logic is good as is, I would just add (at the end of `while` block:
```js
iterations += 1

if (iterations > X) {
    break
}
```

Like that we would parse amount of X and then the cron would pickup in next execution.

capJavert: Better to limit like that vs with execution limit since it can break mid execution eg. 4 minutes expire mid parsing a batch.

omBratteng: Added limit of 100, is that enough. I think if it were to fail to pop, it would happen rather quick. So 100 iterations would be over quite quick?

omBratteng: 100 is also something we could easily adjust.

Also not sure how many possible users would be queued up. Right now we just get lots of 500 errors, but some of them are retries from CIO, so unsure of the actual amount.

capJavert: It's 100 * 100 users, so 10000 users, which seems ok to process in 5 min?

capJavert (approved):
Only thing left is https://github.com/dailydotdev/daily-api/pull/2021/files#r1662316502 but I am fine with it as temp solution if everyone is onboard that we can lose some data 👍

src/cron/syncSubscriptionWithCIO.ts
omBratteng: In 5 minutes, yeah.

Could also add this between each execution, just to be sure we don't hit the 10req/s limit of CIO
https://github.com/dailydotdev/daily-api/pull/2021#discussion_r1661145136

src/cron/syncSubscription.ts
capJavert: Yes, I imagine this can resolve quite fast, how many requests we send per `syncSubscription` call? @omBratteng

src/cron/syncSubscriptionWithCIO.ts
capJavert: Yeah, agreed, left a comment on that comment 😆

src/cron/syncSubscription.ts
omBratteng: Per `syncSubscription`, we do one call to them. so it is very much possible that we manage to do >10req/s.
But, that all depends on how many users are queued up. Could be 50, then we'd suffice with just one call, as it does up to 100 users.

omBratteng: But then, better safe than sorry. Doesn't hurt to have it there. It's also a cron job, so doesn't affect execution time of any endpoints.

And if the while loop were to go haywire, it would run 100 times, taking 100 seconds, but that's fine I guess?

capJavert: Yes, I agree

capJavert: We can wait `100ms` , even if we executed support fast we would be at 10 req/s

omBratteng: `200ms` just to be on the safe side?

omBratteng: Added 200ms

capJavert: measure

src/routes/webhooks/customerio.ts
idoshamun: I agree with the compromise @omBratteng suggests. We should reconsider once we have a queue system until then let's go with redis. I appreciate everyone's concern to keep the data in tact though 🙏

src/cron/syncSubscriptionWithCIO.ts
```diff
@@ -28,6 +28,9 @@ const cron: Cron = {
 
         await syncSubscription([...userIds], con);
 
+        // Wait for a bit to avoid rate limiting
+        await new Promise((resolve) => setTimeout(resolve, 200));
```
capJavert, line 32: @omBratteng you can use `import { setTimeout } from 'node:timers/promises'` which returns a promise `await setTimeout(resolve, 200)`

omBratteng: done.

src/common/mailing.ts
rebelchris: Happy with that :) 
Nice work on fixing it for them, keep us posted 🙏

PR #2022, opened 2024-07-02
chore: cleanup search_version

rebelchris:
Remove the old v1 search support as we now only support v2 on these.
We still have the logic for bookmark and reading history search


src/schema/search.ts
```diff
@@ -185,11 +170,6 @@ export const typeDefs = /* GraphQL */ `
       """
       query: String!
 
-      """
-      Version of the search algorithm
-      """
-      version: Int = 1
```
capJavert, line 191: Do we want to remove it? In the future we could have multiple search versions again and then we have to return all this code.

capJavert: Instead we could just adjust what default is (if not sent), eg. now to be `2`.

rebelchris: Also open to that :) 
It might just be confusing what 2 means if there is no 1?

capJavert: I look at it as feed. There the old versions also phase out. So anything below current version will just get `2`.

src/schema/search.ts
```diff
@@ -274,62 +252,43 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
       getSession(ctx.userId, id),
     searchPostSuggestions: async (
       source,
-      { query, version }: { query: string; version: number },
+      { query }: { query: string; version: number },
```
rebelchris, line 255: Don't use version here to avoid backward issues

__tests__/__snapshots__/feeds.ts.snap
```diff
@@ -7,8 +7,8 @@ Object {
     "golang",
   ],
   "excludeSources": Array [
-    "b",
     "c",
+    "b",
```
capJavert, line 11: Flaky test?

src/schema/search.ts
capJavert: Additionally, what does this mean for local, search no longer works because we don't have meili locally?

rebelchris:
@capJavert Yep won't work locally indeed.
But can't recall last time we did anything that needed search output (posts wise)

Maybe something we can slowly add as mock?

__tests__/__snapshots__/feeds.ts.snap
rebelchris: I guess so, my global snapshot update updated this one

capJavert (approved):
Approving, we can supply API key to the team when we get it.

PR #2024, opened 2024-07-02
fix: use `userId`

omBratteng:


capJavert (approved):

src/common/mailing.ts
```diff
@@ -114,7 +114,7 @@ export const syncSubscription = async function (
         );
       if (!digest) {
         await manager.getRepository(UserPersonalizedDigest).delete({
-          id: customer.id,
+          userId: customer.id,
```
capJavert, line 117: damn did we delete more than we should have?

rebelchris: id doesn't exist :)  (so should be ok)

omBratteng: Nah, just got error. That row doesn't exist anyways. Though, strange that linter didn't catch that the table doesn't have `id` column
 
> Property "id" was not found in "UserPersonalizedDigest". Make sure your query is correct.

omBratteng: That would've been bad stupid from their side :triggered:

capJavert: I have to speed up that `strict` mode update.. it would have caught this

omBratteng: Not sure... this is the type definition
image

capJavert: [strict](https://www.typescriptlang.org/tsconfig/#strict) mode in typescript yes

capJavert: But even then it would not since their internal `ObjectId` always has `id` property 🤦

omBratteng: Yeah 😿

PR #2027, opened 2024-07-02
fix: add canonical check on update aswell

rebelchris:
The update hook didn't check if canonical/url already exists on the change causing catch errors.
Explicitly check now and uniform the check function

### DIFF
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -200,7 +200,18 @@ type CheckExistingPostProps = {
   errorMsg: string;
   id?: string;
 };
-const checkExistingPost = async ({
+
+/**
+ * Function to check wheter an url/canonical url is already in use
+ *
+ * @param entityManager
+ * @param data
+ * @param counter
+ * @param logger
+ * @param errorMsg
+ * @param id - By passing the id we only search for other posts containing these URLs
+ */
+const checkExistingUrl = async ({
   entityManager,
   data,
   counter,
@@ -241,7 +252,7 @@ const createPost = async ({
   questions,
 }: CreatePostProps): Promise<Post | null> => {
   if (
-    await checkExistingPost({
+    await checkExistingUrl({
       entityManager,
       data,
       counter,
@@ -370,7 +381,7 @@ const updatePost = async ({
   }
 
   if (
-    await checkExistingPost({
+    await checkExistingUrl({
       entityManager,
       data,
       counter,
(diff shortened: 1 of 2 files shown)


src/workers/postUpdated.ts
```diff
@@ -192,29 +192,63 @@ const assignScoutToPost = async ({
...
     .where(
-      'url = :url or url = :canonicalUrl or "canonicalUrl" = :url or "canonicalUrl" = :canonicalUrl',
+      '(url = :url or url = :canonicalUrl or "canonicalUrl" = :url or "canonicalUrl" = :canonicalUrl)',
       { url: data.url, canonicalUrl: data.canonicalUrl },
-    )
-    .getRawOne();
+    );
+  if (id) {
+    builder = builder.andWhere('id != :id', { id });
+  }
```
capJavert, line 221: Is `id` not the sole identifier of the post we are trying to update?

capJavert: Ah, I see, you are trying to query if any other post has the canonical/url that we are now updating for this post?

capJavert: I would maybe move this clause in update specifically since when you provide `id` it does not really look for post matching that `id` but looks if any other post matches the url, which to me is not intuitive and opposite to how it works when `id` is not provided for `checkExistingPost` function.

capJavert: Maybe `checkExistingPost` and `checkExistsUrl`?

rebelchris: @capJavert Good feedback, the naming was a bit off actually.
I revamped and added tsdoc to better explain the ID.

Let me know if this works?

capJavert: I would still prefer two functions personally but if going this route can we maybe name `id` as `excludeId` so it is immediately obvious that it will exclude?

rebelchris: Hmm but it needs to be one query I think?
You want to check all posts that are not the current one to see if it exists right? (maybe i'm misunderstanding you here?)

If so the excludeId is maybe best?

capJavert: > You want to check all posts that are not the current one to see if it exists right

Function does different things, when you don't provide `id` it will tell you if another post exists, but if you provide `id` you are checking if some other post exists aside from the one with `id`.

in short, yeah, I am fine with `excludeId` if two functions complicate it.

rebelchris: Ah got you :) 
Yeah felt too much was the same, for now just went with excludeId.

I do think we need to start abstracting both into seperate ones soon anyway.

capJavert (approved):

PR #2028, opened 2024-07-02
feat: add team member output

rebelchris:
Ole-Martin's favorite feature enablement.
This way we can simply add it to each user fragment wherever we want to know it :)

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -77,12 +77,15 @@ const obj = new GraphORM({
         transform: nullIfNotSameUser,
       },
       isTeamMember: {
-        select: (_, alias, qb) =>
-          qb
-            .select('count(*)')
+        select: (_, alias, qb) => {
+          const query = qb
+            .select('1')
             .from(Feature, 'f')
             .where(`f."userId" = ${alias}.id`)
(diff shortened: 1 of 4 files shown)


src/graphorm/index.ts
```diff
@@ -74,6 +76,15 @@ const obj = new GraphORM({
       notificationEmail: {
         transform: nullIfNotSameUser,
       },
+      isTeamMember: {
+        select: (_, alias, qb) =>
+          qb
+            .select('count(*)')
+            .from(Feature, 'f')
+            .where(`f."userId" = ${alias}.id`)
+            .andWhere(`f."feature" = :feature`, { feature: FeatureType.Team }),
```
capJavert, line 85: We should be able to use `EXISTS` it will be more efficient then count. 

```suggestion
        select: (_, alias, qb) => {
          const query = qb
            .select('1')
            .from(Feature, 'f')
            .where(`f."userId" = ${alias}.id`)
            .andWhere(`f."feature" = :feature`, { feature: FeatureType.Team });

          return `EXISTS${query.getQuery()}`;
        },
```

capJavert (approved):
I think we can use exists for better performance, approving either way.

PR #2029, opened 2024-07-03
refactor: change the log output of sync-subscription-with-cio cron

omBratteng:
Easier to see how many was synced, instead of having to sum all of the logs 🙈

### DIFF
--- a/src/cron/syncSubscriptionWithCIO.ts
+++ b/src/cron/syncSubscriptionWithCIO.ts
@@ -17,6 +17,7 @@ const cron: Cron = {
     logger.info('syncing subscriptions with customer.io');
 
     let iterations = 0;
+    let syncedSubscriptions = 0;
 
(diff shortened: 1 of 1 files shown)


rebelchris (approved):

src/cron/syncSubscriptionWithCIO.ts
```diff
@@ -44,6 +46,7 @@ const cron: Cron = {
         break;
       }
     }
+    logger.info(`synced subscriptions for ${syncedSubscriptions} users`);
```
capJavert, line 49: You don't have to do it retro but better to pass as JSON object eg. `logger.info({ count: syncedSubscriptions }, 'synced subscriptions')`

omBratteng: True, that's better. See #2030

PR #2030, opened 2024-07-03
refactor: use json output on sync-subscription-with-cio

omBratteng:


rebelchris (approved):

src/cron/syncSubscriptionWithCIO.ts
```diff
@@ -46,7 +46,7 @@ const cron: Cron = {
         break;
       }
     }
-    logger.info(`synced subscriptions for ${syncedSubscriptions} users`);
+    logger.info({ count: syncedSubscriptions }, 'synced subscriptions');
```
capJavert, line 49: We can also change the above `debug` call to the same pattern then.

capJavert (approved):
Tnx for retro updating 🚀

src/cron/syncSubscriptionWithCIO.ts
omBratteng: Done

PR #2034, opened 2024-07-04
fix: experience level

nensidosari:
Just one small check for experience level, this does not cover all of our use cases like SSO but as a first step to not allowing registration with no experience level.

issue: https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1719910800166109?thread_ts=1719910800.166109&cid=C01L0QXQJNB

I'll check all other signup options we have in preview to make sure none of those is broken.

### DIFF
--- a/__tests__/webhooks.ts
+++ b/__tests__/webhooks.ts
@@ -17,9 +17,23 @@ import {
 import { StorageKey, StorageTopic, generateStorageKey } from '../src/config';
 import nock from 'nock';
 import { triggerTypedEvent } from '../src/common';
+import { NotificationPayload } from '../src/routes/webhooks/customerio';
+import { sendGenericPush } from '../src/onesignal';
 
 let app: FastifyInstance;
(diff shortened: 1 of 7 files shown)


src/entity/user/User.ts
```diff
@@ -183,6 +183,9 @@ type AddNewUserResult =
   | { status: 'failed'; reason: UserFailErrorKeys; error?: Error };
 
 const checkRequiredFields = (data: AddUserData): boolean => {
+  if (data && data.username && !data.experienceLevel) {
```
capJavert, line 186: `data?.username` use optional chaining on both?

capJavert (approved):

src/entity/user/User.ts
nensidosari: changed to ante's suggestion, should be good as data?.username would never be true if there was no data

rebelchris (approved):
Yeah this works :)

PR #2036, opened 2024-07-04
chore: new copy

rebelchris:
New copy requested by product:
https://dailydotdev.slack.com/archives/C07A93GC63H/p1720081953149869?thread_ts=1719925744.424499&cid=C07A93GC63H


capJavert (approved):
We need a better way of storing these 😆

PR #2037, opened 2024-07-04
fix: add new private feed by ids

rebelchris:
Private feed for platform guys to play with.
I know it's not 100% secure on this side, but don't think it's exposing much realisticly.

The FE part is actually limited to team members (hence the boot addition)

### DIFF
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -335,6 +335,31 @@ export const typeDefs = /* GraphQL */ `
     ): PostConnection! @auth
 
     """
+    Get feed by providing post ids
+    """
+    feedByIds(
+      """
+      Paginate after opaque cursor
+      """
+      after: String
+
+      """
+      Paginate first
+      """
+      first: Int
+
+      """
+      Array of post ids
+      """
+      postIds: [String!]!
+
+      """
+      Array of supported post types
+      """
+      supportedTypes: [String!]
+    ): PostConnection! @auth
+
+    """
     Get an adhoc feed using a provided config
     """
     feedByConfig(
@@ -1343,6 +1368,13 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
         info,
       );
     },
+    feedByIds: feedResolver(
+      (ctx, { postIds }: FeedArgs & { postIds: string[] }, builder, alias) =>
+        fixedIdsFeedBuilder(ctx, postIds, builder, alias),
+      feedPageGenerator,
+      applyFeedPaging,
+      {},
+    ),
     feedByConfig: (
       source,
       args: ConnectionArguments & { config: string },
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -1,5 +1,7 @@
 import {
   BookmarkList,
+  Feature,
+  FeatureType,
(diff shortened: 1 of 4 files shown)


src/schema/feeds.ts
```diff
@@ -1343,6 +1368,13 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
         info,
       );
     },
+    feedByIds: feedResolver(
```
nensidosari, line 1371: Looks goos, though not sure how it works 🤔 what does it mean? what id?

So do you get a feed with only posts, ids of which u've provided?

nensidosari (approved):

src/schema/feeds.ts
rebelchris: Yeah a feed with post from whom you provider the post id in URL.

capJavert: Can we add check here if accessing user is a team member?

rebelchris: We can, but wanted to avoid it as it would need either a custom parsing in the JWT (which adds complexity)
Or a separate query.

And tbh didn't care too much if someone wanted to exploit the endpoint manually for now.
Let me know if you think query is the way to go (as the JWT seems a bit excessive for now)

capJavert: We would need one feature query for `ctx.userId`, I think thats easy?

capJavert: Exploiting this endpoint could be pretty bad, imagine someone sends 10k post ids in the array. Nothing is stopping from doing that.

rebelchris: Yeah it's indeed one query to Feature :) 
We do it in boot already (so technically CTX could directly access it if parsed in JWT) (but let's do it if a "actual" endpoint needs it.)

Ok will add it np.

capJavert: Yeah, no need to do it in JWT now, tnx! 👌

rebelchris: @capJavert added functionality to block non team members

src/schema/feeds.ts
```diff
@@ -1371,9 +1373,21 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
...
+
       feedPageGenerator,
       applyFeedPaging,
-      {},
+      {
+        fetchQueryParams: async (ctx): Promise<void> => {
+          const isTeamMember = await ctx.con
+            .getRepository(Feature)
+            .findOneByOrFail({ feature: FeatureType.Team, userId: ctx.userId });
+          if (!isTeamMember || isTeamMember.value !== 1) {
```
capJavert, line 1384: Nice! Can you add a test as well? 🙏

capJavert (approved):

src/schema/feeds.ts
rebelchris: Wanted to avoid it as it's only internal nonsense but yeah 🤣  (just laziness)

rebelchris: @capJavert Put my laziness aside and added test cases let me know if we good to go?

capJavert (approved):
Awesome!

PR #2038, opened 2024-07-05
feat: slack integration

capJavert:
Slack integration 
- [x] UserIntegration entity
- [x] UserSourceIntegration entity
- [x] oAuth2 callback for slack app installation 
- [x] slack events subscription to remove expired/removed integrations 
- [x] userIntegrations query
- [x] slack related queries for integration
- [x] removed outdated Integration entity and replace schema with new one
- [x] connecting source is currently limited only for squad admins, additionally only single source integration is allowed per source

AS-472


src/routes/index.ts
```diff
@@ -46,6 +47,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
   fastify.register(redirects);
   fastify.register(automations, { prefix: '/auto' });
   fastify.register(sitemaps, { prefix: '/sitemaps' });
+  fastify.register(slack, { prefix: '/slack' });
```
rebelchris, line 50: Wondering if we want it under `integrations/slack` or `webhooks/slack`?

src/routes/slack.ts
```diff
@@ -0,0 +1,157 @@
...
+        );
+
+        throw error;
+      }
+
+      const existingIntegration = await con
+        .getRepository(UserIntegrationSlack)
+        .createQueryBuilder()
+        .where('"userId" = :userId', { userId: req.userId })
+        .andWhere(`meta->>'teamId' = :teamId`, { teamId: result.team.id })
```
rebelchris, line 118: I was wondering about meta having all objects (I'm find with it) considering mostly used in bg scripts (and gives high flexibility with other integrations)

Just have to keep monitoring we don't ever need heavy filtering/queries on meta fields?

rebelchris (approved):
Nice no real blockers from my side.
Only thing popped in my head is JSON field and having to query on it. (but don't think it matters much)

src/routes/index.ts
capJavert: Its not only webhook, but could be `/integrations/`, will adjust to that 👌

src/routes/slack.ts
capJavert: We can add indexes on those fields as well, especially now with psql 15.

.infra/Pulumi.adhoc.yaml
```diff
@@ -78,5 +78,6 @@ config:
     typeormUser: postgres
     tz: UTC
     urlPrefix: http://api.local.com
+  api:image: api-image:tilt-f53bc87b4aa5a6bc
```
idoshamun, line 81: This one shouldn't be committed

src/entity/UserIntegration.ts
```diff
@@ -0,0 +1,63 @@
...
+  teamId: string;
+  teamName: string;
+};
+
+@Entity()
+@TableInheritance({
+  column: { type: 'text', name: 'type' },
+})
+@Index('IDX_user_id_type', ['userId', 'type'])
+export class UserIntegration {
```
idoshamun, line 33: Why do we assign it per user and not per squad?

idoshamun: nvm i understand now :)

src/routes/index.ts
idoshamun: I second the `/integrations/` directory

src/routes/slack.ts
```diff
@@ -0,0 +1,157 @@
...
+        .getOne();
+
+      const integrationMeta: IntegrationMetaSlack = {
+        appId: result.app_id,
+        slackUserId: result.authed_user.id,
+        scope: result.scope,
+        tokenType: result.token_type,
+        accessToken: result.access_token,
```
idoshamun, line 126: We should encrypt the access token. Storing it as plain text is not a best practice

src/schema/slack.ts
```diff
@@ -0,0 +1,153 @@
...
+import { ValidationError } from 'apollo-server-errors';
+import { UserSourceIntegrationSlack } from '../entity/UserSourceIntegration';
+import { Source } from '../entity';
+
+export type GQLSlackChannels = {
+  id?: string;
+  name?: string;
+};
+
+export const typeDefs = /* GraphQL */ `
```
idoshamun, line 17: Should we make the terms here more generic to support future integrations?
Should we consider changing the file name as well?

src/schema/slack.ts
```diff
@@ -0,0 +1,153 @@
...
+  type Channel {
+    id: String
+    name: String
+  }
+
+  extend type Query {
+    """
+    Get slack channels
+    """
+    slackChannels(
```
idoshamun, line 27: Pagination support?

src/schema/users.ts
```diff
@@ -1185,6 +1223,25 @@ export const resolvers: IResolvers<any, Context> = {
         },
       );
     },
+    userIntegrations: async (
+      _,
+      __,
+      ctx: Context,
+    ): Promise<GQLUserIntegration[]> => {
+      const userIntegrations = await ctx.con
```
idoshamun, line 1231: Best practice would be to use graphorm to ensure graphql tree is resolved properly

src/schema/slack.ts
capJavert: I am not sure there is gonna be the same calls for other integrations but maybe `integrations` schema makes more sense yes.

src/routes/slack.ts
capJavert: Done in a0f06700409992e4bc6078e4fca88aba9e24b188 with tests

src/routes/integrations/slack.ts
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            const con = await createOrGetConnection();
+
+            await con
+              .getRepository(UserIntegrationSlack)
+              .createQueryBuilder()
+              .delete()
+              .where(`meta->>'teamId' = :teamId`, { teamId })
+              .execute();
```
capJavert, line 233: Basically, because we use bot token revoking it or removing app from workspace invalidates tokens for all users. So basically there will be some duplication of each `UserIntegration` row having same encrypted token but I think its ok since in the future this kind of table organization will serve us better if we have to obtain user token (which is specific for each user).

src/routes/integrations/slack.ts
```diff
@@ -0,0 +1,257 @@
...
+        tokenType: result.token_type,
+        accessToken: await encrypt(
+          result.access_token,
+          process.env.SLACK_DB_KEY,
+        ),
+        teamId: result.team.id,
+        teamName: result.team.name,
+      };
+
+      if (existingIntegration) {
```
idoshamun, line 151: What happens if two users try to add the same slack org?

src/schema/integrations.ts
```diff
@@ -1,38 +1,191 @@
...
+        ['userIntegrationId', 'sourceId'],
+      );
+
+      await client.conversations.join({
+        channel: args.channelId,
+      });
+
+      await client.chat.postMessage({
+        channel: args.channelId,
+        text: `Connected source "${source.name}" to this channel 🙌`,
```
idoshamun, line 185: Let's get the copy from product for this. Also I suggest adding an invitation link to the message

idoshamun (approved):
Everything looks great just validate the copy as suggested

src/schema/integrations.ts
capJavert: I asked for copy but I'll merge this and update after since I'll do another PR for some stuff later.

idoshamun: please remember to add also link in the next PR :)

capJavert: Updated the copy now. I'll see about the link in that next one yes when I sort out joining.

PR #2039, opened 2024-07-05
feat: apply strict typescript rules to scripts

capJavert:

This one focuses on:
- scripts
  - some global types

### DIFF
--- a/bin/addWelcomePosts.ts
+++ b/bin/addWelcomePosts.ts
@@ -14,7 +14,7 @@ import { createSquadWelcomePost } from '../src/common';
 
   resStream.on('data', async (squadSource: SquadSource) => {
     // Fetch first admin based on created data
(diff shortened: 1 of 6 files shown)


bin/addWelcomePosts.ts
```diff
@@ -14,7 +14,7 @@ import { createSquadWelcomePost } from '../src/common';
 
   resStream.on('data', async (squadSource: SquadSource) => {
     // Fetch first admin based on created data
-    const { userId } = await con.getRepository(SourceMember).findOne({
```
rebelchris, line 17: Unrelated but we can probably actually just get rid of some bin scripts 😅

rebelchris (approved):

bin/addWelcomePosts.ts
capJavert: Agreed

PR #2042, opened 2024-07-08
fix: meili pagination

rebelchris:
Added pagination to meilisearch.
Noticed one minor issue we'll need to fix in the `getOffsetWithDefault`.

In some cases at one stage it will return 8 instead of 9 offset which can cause duplicates.
(Actually wondering if this could cause some of the duplication in feed we see in general?) (CC: @idoshamun )

Example:
Screenshot 2024-07-08 at 20 02 34

However think it's better then current max 9 limit on search results 🙏


docker-compose.yml
```diff
@@ -32,3 +33,13 @@ services:
...
+
+  search:
+    image: getmeili/meilisearch:v1.9.0
+    restart: always
+    environment:
+        MEILI_MASTER_KEY: topsecret
+    volumes:
+     - search:/data.ms
+    ports:
+      - '7700:7700'
```
capJavert, line 45: @omBratteng I added this to temp have local meilisearch instance, can this be moved to local adhoc somehow? Do we have support for it, it works as is just want to get a feeling since it will allow us to play with search locally? cc: @rebelchris

rebelchris: Would be a super addition to our local instance, really helps with any search debugging.
(And nice pro to have the local meili interface to play with)

.infra/Pulumi.adhoc.yaml
```diff
@@ -78,5 +81,6 @@ config:
     typeormUser: postgres
     tz: UTC
     urlPrefix: http://api.local.com
+  api:image: api-image:tilt-0bb0f170100909ab
```
capJavert, line 84: I commited this, we don't need this change 🙏

src/integrations/meilisearch.ts
```diff
@@ -28,5 +43,13 @@ export const searchMeili = async (params: string): Promise<Meili[]> => {
       ...requestOptions,
     },
   );
-  return res.hits;
+  return {
+    pagination: {
+      current: res.hits.length + res.offset + res.limit,
```
capJavert, line 48: is not current offset + limit?

src/schema/common.ts
```diff
@@ -148,6 +149,25 @@ export const getSearchQuery = (param: string): string =>
...
+  connArgsToPage: ({ pagination }): OffsetPage => {
+    return {
+      limit: pagination?.total || 10,
+      offset: pagination?.offset + 1 || 0,
+      current: pagination?.current,
+    };
+  },
+  nodeToCursor: (page, args, node, i): string =>
+    offsetToCursor(page.offset + i),
+  hasNextPage: (page): boolean => page.current <= page.limit,
```
capJavert, line 167: next page is if total is lower then current right?

capJavert: eg. `estimatedTotalHits`

capJavert (commented):
Some sanity check comments, awesome work!

src/schema/common.ts
rebelchris: yeah limit is `estimatedTotalHits` and current = current fetched total.
Hope that makes sense?

src/integrations/meilisearch.ts
capJavert: From my logic current hits are <= limit (depending on if there is a next page), so if you add limit it becomes `next`?

rebelchris: Then probably current = offset + hits.length right?

What's already offset + what actually got returned in current query.

capJavert: yeah, that sounds right to me

rebelchris:
@capJavert Thanks for the support, managed to make the offset work now :)

src/schema/common.ts
```diff
@@ -148,6 +150,27 @@ export const getSearchQuery = (param: string): string =>
...
+  unknown
+> => ({
+  connArgsToPage: ({ pagination }): OffsetPage => {
+    return {
+      limit: pagination?.total || 10,
+      offset: pagination?.offset || 0,
+      current: pagination?.current,
+    };
+  },
+  nodeToCursor: (page, { ids }, node): string =>
```
rebelchris, line 168: ids, is just the current ones, we need to determine which one is actually there (exclude deleted ones)

src/schema/common.ts
```diff
@@ -156,6 +179,7 @@ export const offsetPageGenerator = <TReturn>(
   connArgsToPage: (args: ConnectionArguments): OffsetPage => {
     const limit = Math.min(args.first || defaultLimit, maxLimit);
     const offset = getOffsetWithDefault(args.after, -1) + 1;
+    console.log('my feed', limit, offset);
```
rebelchris, line 182:
```suggestion

```

.infra/Pulumi.adhoc.yaml
```diff
@@ -78,5 +81,6 @@ config:
     typeormUser: postgres
     tz: UTC
     urlPrefix: http://api.local.com
+  api:image: api-image:tilt-71438a7697ad6ee4
```
rebelchris, line 84:
```suggestion
```

capJavert: Literally pulumi changed as I was doing `git add`

capJavert (approved):
Thank you for added work!

PR #2043, opened 2024-07-09
refactor: bookmark entity to include remind at column

sshanzel:
Introduce a new property to store the date when to remind the user about their bookmark. Felt like the PR would be too small with just the entity changes, hence, added the mutation.

This would unblock working with the FE. Also, regardless of how we want to approach the queueing system, there won't be any difference with the mutation introduced here.

Added TODOs on places we need to once the Queueing system is up.

MI-426 #done
MI-429 #done


rebelchris (approved):

__tests__/bookmarks.ts
```diff
@@ -603,3 +603,61 @@ describe('query searchBookmarks', () => {
...
+    expect(bookmark.remindAt).toEqual(date);
+    const res = await client.mutate(mutation, {
+      variables: { postId: 'p1', remindAt: null },
+    });
+    expect(res.errors).toBeFalsy();
+
+    const result = await repo.findOneBy({ postId: 'p1', userId: loggedUser });
+    expect(result.remindAt).toBeNull();
+
+    // we should add another check whether the task from the queue was cleared
```
capJavert, line 661: if you have the task up in Jira you can add TODO MI-XXX so we can track in the codebase.

capJavert: same with below TODOs

capJavert (approved):

__tests__/bookmarks.ts
sshanzel: Right. Let me create a ticket for this.

sshanzel: Updated with the ticket number as below 🥳 

https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31?selectedIssue=MI-425

capJavert: Tnx!

PR #2044, opened 2024-07-09
refactor: separate metrics and tracing and define counters in one place

omBratteng:
This separates metrics and tracing into separate parts. Additionally, we now define the counters in one place so that they are initialised with 0 at startup. This is because prometheus does not fully like distributed metrics, where one pod not might have the counter and the other does, so it'll skew metrics.

Though, will need to actually deploy it to verify that the metrics show up in GCP.

At the moment, no metrics from cron jobs are actually counted, as prometheus is a scraper, and the job might be over before prometheus is able to scrape it. Will introduce an intermediate push proxy at a later point.


src/cron/updateCurrentStreak.ts
```diff
@@ -1,17 +1,13 @@
...
 const cron: Cron = {
   name: 'update-current-streak',
   handler: async (con, logger) => {
     try {
-      const streakCounter = opentelemetry.metrics
-        .getMeter('api-bg')
-        .createCounter('streak_update', {
-          description: 'How many streaks get updated',
-        });
+      const streakCounter = counters.cron.streakUpdate;
```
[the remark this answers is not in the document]
omBratteng: That's correct

src/directive/rateLimit.ts
```diff
@@ -95,7 +96,7 @@ export const onLimit: RateLimitOnLimit<Context> = (
     case 'createFreeformPost':
     case 'submitExternalLink':
     case 'sharePost':
-      context.rateLimitCouner.add(1, { type: 'createPost' });
+      context.meter && counters.api.rateLimit.add(1, { type: 'createPost' });
```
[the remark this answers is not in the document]
omBratteng: Probably not, but just a nice safety measure to ensure that metrics is loaded into context

src/cron/updateCurrentStreak.ts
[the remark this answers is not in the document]
omBratteng: It might work for a while as long as we still also push metrics via GCP. But if all works with prometheus, we'll remove that metric export method.

And it hasn't worked till now anyways, as cron never had trace/metrics enabled.

src/directive/rateLimit.ts
```diff
@@ -95,7 +96,7 @@ export const onLimit: RateLimitOnLimit<Context> = (
     case 'createFreeformPost':
     case 'submitExternalLink':
     case 'sharePost':
-      context.rateLimitCouner.add(1, { type: 'createPost' });
+      context.meter && counters?.api?.rateLimit?.add(1, { type: 'createPost' });
```
capJavert, line 99: can this be in some utility method which under the hood does `meter &&`, seems clunky to call it each time and also everyones understanding when to call it?

src/workers/newNotificationV2Mail.ts
```diff
@@ -659,7 +658,7 @@ const worker: Worker = {
         },
         'failed to send notification email',
       );
-      counter.add(1, { channel: 'email' });
+      counters?.background?.notificationFailed?.add(1, { channel: 'email' });
```
capJavert, line 661: no meter here?

src/cron/updateCurrentStreak.ts
```diff
@@ -1,17 +1,13 @@
 import { Cron } from './cron';
 import { User, UserStreak } from '../entity';
 import { checkUserStreak, clearUserStreak } from '../common';
-import { opentelemetry } from '../telemetry/opentelemetry';
+import { counters } from '../telemetry';
```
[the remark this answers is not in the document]
capJavert: I think its just an object to store all the counters.

src/directive/rateLimit.ts
omBratteng: Might not need `context.meter &&` when we're optionally chaining it

src/cron/updateCurrentStreak.ts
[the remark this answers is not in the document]
omBratteng: Yeah, it's just an object that stores all counters, shouldn't add much load

src/workers/newNotificationV2Mail.ts
omBratteng: What do you mean?

src/directive/rateLimit.ts
```diff
@@ -104,12 +105,13 @@ export const onLimit: RateLimitOnLimit<Context> = (
       });
     case 'commentOnPost':
     case 'commentOnComment':
-      context.rateLimitCouner.add(1, { type: 'createComment' });
+      context.meter &&
+        counters?.api?.rateLimit?.add(1, { type: 'createComment' });
```
rebelchris, line 109: We don't need it here anymore right?
```suggestion
      counters?.api?.rateLimit?.add(1, { type: 'createComment' });
```

rebelchris (approved):
Only one left I think :)

src/directive/rateLimit.ts
omBratteng: 🙈

omBratteng: Removed

src/workers/newNotificationV2Mail.ts
capJavert: Ah I see you removed it from everywhere now?

omBratteng: Meter is what "holds" the metrics. From a meter you create counters, and I've moved the creation of all counters into `metrics.ts`. So now we just import `counters` and select the specific counter you want.

omBratteng: > Ah I see you removed it from everywhere now?

Yeah

capJavert: Makes sense 👍

src/Context.ts
```diff
@@ -59,6 +46,10 @@ export class Context {
     return this.req.span;
   }
 
+  get meter(): opentelemetry.Meter {
+    return this.req.meter;
+  }
```
capJavert, line 51: Do we need this one now?

src/telemetry/metrics.ts
```diff
@@ -38,4 +118,47 @@ export const startMetrics = (serviceName: string): void => {
 
   const meterProvider = new metrics.MeterProvider({ resource, readers });
   api.metrics.setGlobalMeterProvider(meterProvider);
+
+  channel.subscribe(({ fastify }: { fastify: FastifyInstance }) => {
```
capJavert, line 122: Should this and below be executed if `METRICS_ENABLED` is not `true`?

capJavert (approved):
Just two more sanity checks questions, otherwise looks good to me.

src/telemetry/metrics.ts
omBratteng: Added early return instead

src/Context.ts
omBratteng: Might not need it. I'll test locally if I can remove it completely

omBratteng: Remove that and other meters

PR #2045, opened 2024-07-09
feat: introduce bookmark type on queries

sshanzel:
Add the `type Bookmark` to make it easier to fetch the bookmarked property.

Also introduced a common function for Date. Along with it, I found a very interesting issue related to the `user.createdAt` on our fragments.

MI-428 #done


__tests__/posts.ts
```diff
@@ -339,6 +339,41 @@ describe('bookmarked field', () => {
...
+    loggedUser = '1';
+    const repo = con.getRepository(Bookmark);
+    await repo.save({
+      postId: 'p1',
+      userId: loggedUser,
+      remindAt: new Date(),
+    });
+    const res = await client.query(QUERY);
+    expect(res.data.post.bookmark.createdAt).toBeTruthy();
+    expect(res.data.post.bookmark.remindAt).toBeTruthy();
```
sshanzel, line 373: I was bashing my head to the keyboard wondering why the `remindAt` is always `null`. It seems that there is something wrong with our DateTime parser.

src/graphorm/index.ts
```diff
@@ -360,16 +380,17 @@ const obj = new GraphORM({
       },
     },
   },
+  Bookmark: {
+    from: 'Bookmark',
+    fields: {
+      remindAt: { transform: transformDate },
```
sshanzel, line 386: After further investigation, I saw a lot of fields here using a similar function doing `() => new Date(value)`. Before adding this here, the `remindAt` was always `null` while the `createdAt` was always correct.

src/graphorm/index.ts
```diff
@@ -74,6 +77,9 @@ const obj = new GraphORM({
       notificationEmail: {
         transform: nullIfNotSameUser,
       },
+      createdAt: {
+        transform: transformDate,
+      },
```
sshanzel, line 82: The parent entity of this field is the `type User`. I have a feeling that after adding this, the issue with having `createdAt` on our FE will be resolved.

src/graphorm/index.ts
```diff
@@ -360,16 +380,17 @@ const obj = new GraphORM({
...
     fields: {
-      createdAt: {
-        transform: (value: string | Date): Date => new Date(value),
-      },
-      lastUpdatedAt: {
-        transform: (value?: string | Date): Date =>
-          value ? new Date(value) : undefined,
-      },
+      createdAt: { transform: transformDate },
+      lastUpdatedAt: { transform: transformDate },
```
sshanzel, line 393: Just updated these ones to use the same function.

capJavert (approved):

src/graphorm/index.ts
sshanzel: Good point. It was working when I tested it, but probably best if we include it to use the same transform function.

src/graphorm/index.ts
rebelchris: Isnt it the generic typeorm date issue we have?

rebelchris (commented):
Nice

PR #2046, opened 2024-07-09
chore: add metric to see which table spools how many req

rebelchris:
Would be nice to know...
I see sometimes the CDC still gets overworked and kind of want to see which tables cause these spikes.

Do let me know if anyone thinks it's too much 😅

### DIFF
--- a/src/telemetry/metrics.ts
+++ b/src/telemetry/metrics.ts
@@ -60,6 +60,10 @@ const counterMap = {
       name: 'cio_events',
       description: 'How many customerio events were sent to analytics',
     },
+    cdcTrigger: {
+      name: 'cdc_trigger',
+      description: 'How many times the cdc trigger was called',
+    },
   },
   background: {
     postError: {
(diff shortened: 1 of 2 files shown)


src/telemetry/metrics.ts
```diff
@@ -60,6 +60,10 @@ const counterMap = {
       name: 'cio_events',
       description: 'How many customerio events were sent to analytics',
     },
+    cdcTrigger: {
+      name: 'cdc_trigger',
+      description: 'How many times the cdc trigger was called',
+    },
```
omBratteng, line 66: Wrong place, it's a worker, so it goes into `background`

omBratteng (changes requested):

src/telemetry/metrics.ts
rebelchris: My bad, changed it now.

omBratteng (approved):

capJavert (approved):

PR #2047, opened 2024-07-09
fix: remove gcp metrics exporter

omBratteng:
This removes the GCP metric exporter, now we only use prometheus metrics

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "@fastify/helmet": "^11.1.1",
     "@fastify/rate-limit": "^9.1.0",
     "@fastify/websocket": "^10.0.1",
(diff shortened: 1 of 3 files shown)


rebelchris (approved):

package.json
```diff
@@ -38,7 +38,6 @@
     "@fastify/helmet": "^11.1.1",
     "@fastify/rate-limit": "^9.1.0",
     "@fastify/websocket": "^10.0.1",
-    "@google-cloud/opentelemetry-cloud-monitoring-exporter": "^0.18.0",
```
capJavert, line 41: everything worked fine then?

capJavert (approved):

package.json
omBratteng: Yeah, this is just the Google specific one. We still have prometheus exporter

PR #2048, opened 2024-07-09
feat: cache cors requests

capJavert:


omBratteng:
Same goes for `/e` and `/e/*`, which is sent to analytics-api service

capJavert:
> In production, `/v1/a` and `/v1/a/*` is sent to `monetization` service, and not handled by API

Yes, thats my TODO for tomorrow.

src/index.ts
```diff
@@ -85,8 +85,9 @@ export default async function app(
 
   app.register(helmet);
   app.register(cors, {
-    origin: isProd ? /daily\.dev$/ : true,
+    origin: isProd ? /^(?:https:\/\/)?(?:[\w-]+\.)*daily\.dev$/ : true,
```
capJavert, line 88: before it would allow `adaily.dev`

omBratteng: Good catch, otherwise we'd have to buy *daily.dev domains there are 💸

rebelchris (approved):

omBratteng (approved):

PR #2049, opened 2024-07-09
feat(strict-ts): apply to common utils

capJavert:
daily.dev strict mode PR template

This one focuses on:
- common
- top level utils

### DIFF
--- a/src/common/object.ts
+++ b/src/common/object.ts
@@ -18,7 +18,7 @@ export const isNullOrUndefined = (value: unknown) =>
   typeof value === 'undefined' || value === null;
 
 type Key = string;
-type Value = string;
+type Value = string | undefined;
 type IsRequired = boolean;
 export type ValidateRegex = [Key, Value, RegExp, IsRequired?];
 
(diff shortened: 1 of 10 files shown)


src/routes/rss.ts
```diff
@@ -90,7 +91,9 @@ const generateRSS =
         }),
       );
       return res.type('application/rss+xml').send(feed.xml());
-    } catch (err) {
+    } catch (originalError) {
+      const err = originalError as TypeORMQueryFailedError;
+
```
omBratteng, line 96: I really wish we could set type inside `catch` statement :triggered:

omBratteng (approved):

src/routes/rss.ts
rebelchris: Yep really sad :(

src/common/object.ts
```diff
@@ -18,7 +18,7 @@ export const isNullOrUndefined = (value: unknown) =>
   typeof value === 'undefined' || value === null;
 
 type Key = string;
-type Value = string;
+type Value = string | undefined;
```
rebelchris, line 21: Wouldnt this do?
```suggestion
type Value = string?;
```

rebelchris (approved):
For the undefined ones can't we just make things optional?
I think it's slightly cleaner as you don't have to parse `as string` everywhere.

src/common/object.ts
rebelchris: Oh sigh 😮‍💨

PR #2051, opened 2024-07-10
fix: flaky worker test

capJavert:
~~From looking at the tests it looks like some other test inserts `PostRelation` (there are few in notification and cdc tests) and then post used in `collectionUpdated` gets different sources. I changed the post ids to be unique in this test suite which will hopefully avoid collision and flakiness.~~

it failed..

actually, looking at it again, it could be the ordering issue since in `getAllSourcesBaseQuery` inside worker we order based on `createdAt` so basically `community` source is added to the post but db sometimes spills them other way around depending on the insert.

If I force it I get exact test fail which is good because now I can repro locally:
image

- [x] TODO order while querying or adjust test condition

### DIFF
--- a/__tests__/workers/notifications/collectionUpdated.ts
+++ b/__tests__/workers/notifications/collectionUpdated.ts
@@ -44,9 +44,9 @@ beforeEach(async () => {
   await saveFixtures(con, User, usersFixture);
   await saveFixtures(con, ArticlePost, [
     {
-      id: 'cp1',
-      shortId: 'cp1',
-      url: 'http://cp1.com',
+      id: 'cupp1',
+      shortId: 'cupp1',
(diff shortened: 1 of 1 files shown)


omBratteng (approved):

capJavert:
I am running it few more times to see if I can make it fail.

capJavert:
Hmm nope, it failed

rebelchris (approved):
I would have sworn it was something with a update on comm picks from another script now that they can run concurrently.
Ideally tests should be a bit more isolated i think (like saving fixtures with test prefix or something?)

Happy with this though 👍

capJavert:
Yeah, fixtures, while good can bite you back if you reuse, especially since we share a single table for all tests running.

PR #2053, opened 2024-07-11
feat(strict-ts): apply to boot and user utils

capJavert:

This one focuses on:
- boot handler
- user utils
  - streak utils


src/common/users.ts
```diff
@@ -327,7 +329,7 @@ export const getUserReadingRank = async (
   const now = new Date();
   const getReadingTags = () => {
     if (version === 1) {
-      return Promise.resolve(null);
+      return Promise.resolve([]);
```
capJavert, line 332: This is why snapshots were updated, the type even expected `tags` to always be array so I think more logical. No real impact on frontend.

src/common/users.ts
```diff
@@ -343,10 +345,16 @@ export const getUserReadingRank = async (
...
 
-  const [{ thisWeek, lastWeek, lastReadTime }, tags] = await Promise.all([
+  const [readingStreakResult, tags] = await Promise.all([
     req.getRawOne<ReadingRankQueryResult>(),
     getReadingTags(),
   ]);
+
+  if (!readingStreakResult) {
+    throw new Error('failed to get user reading streak');
+  }
```
capJavert, line 355: This is new, basically it would break if for any reason streak would not exist so at least here now we have dedicated error.

src/routes/boot.ts
```diff
@@ -317,10 +317,13 @@ export function getReferralFromCookie({
...
-  userId: string,
-  con: DataSource | EntityManager,
-): Promise<Experimentation> => {
+const getExperimentation = async ({
+  userId,
+  con,
+}: {
+  userId?: string;
+  con: DataSource | EntityManager;
+}): Promise<Experimentation> => {
```
capJavert, line 326: Refactored to object syntax so I can use optional props

src/common/users.ts
```diff
@@ -20,7 +20,7 @@ export interface User {
   username?: string;
   timezone?: string;
   acceptedMarketing?: boolean;
-  experienceLevel?: string;
+  experienceLevel: string | null;
```
rebelchris, line 23: Why the variation here?
Do we want explicit `null` value (*vs undefined)

src/entity/Settings.ts
```diff
@@ -77,7 +77,7 @@ export class Settings {
   onboardingChecklistView: ChecklistViewState;
 
   @UpdateDateColumn()
-  updatedAt: Date;
+  updatedAt: Date | null;
```
rebelchris, line 80: This can technically never be null right?

src/entity/Settings.ts
```diff
@@ -65,7 +65,7 @@ export class Settings {
   optOutCompanion: boolean;
 
   @Column({ type: 'text', array: true, default: null })
-  customLinks: string[];
+  customLinks: string[] | null;
```
rebelchris, line 68: Why not use ? here?

src/routes/boot.ts
```diff
@@ -95,8 +95,8 @@ export type BootUserReferral = Partial<{
 }>;
 
 interface AnonymousUser extends BootUserReferral {
-  id: string;
-  firstVisit: string;
+  id?: string;
+  firstVisit: string | null;
```
rebelchris, line 99: Why not?
```suggestion
  firstVisit?: string;
```

rebelchris (commented):
Mainly some validation questions on why we choose `| null` over optional types in some cases?

src/routes/boot.ts
capJavert: mostly because underlying typeorm and redis methods return null instead of undefined and i did not want to make more type errors while I change those. Usually yes, we want undefined (optional).

capJavert: this applies to all comments above

capJavert: also null and undefined are different things so TS complains about it, you cant use it interchangeably

rebelchris: Ah ok, yeah figured it was related to something like that.
Ok I'll resolve all above :)

rebelchris (approved):

PR #2056, opened 2024-07-12
feat: typed notification worker

sshanzel:
- Introduce a typed notification worker.
- Removes the necessity of having to use `messageToJson` by introducing a worker generator.
- Typed data based on the topic.
- See the error below if I try to deconstruct an unknown property to what was defined.

Screenshot 2024-07-12 at 1 45 45 PM

### DIFF
--- a/src/workers/notifications/postBookmarkReminder.ts
+++ b/src/workers/notifications/postBookmarkReminder.ts
@@ -1,18 +1,11 @@
-import { messageToJson } from '../worker';
 import { Bookmark } from '../../entity';
 import { NotificationType } from '../../notifications/common';
-import { NotificationWorker } from './worker';
+import { generateTypedNotificationWorker } from './worker';
 import { buildPostContext } from './utils';
 
-interface Data {
-  postId: string;
-  userId: string;
-}
-
-const worker: NotificationWorker = {
-  subscription: 'api.post-bookmark-reminder-notification',
-  handler: async (message, con) => {
-    const { postId, userId }: Data = messageToJson(message);
+const worker = generateTypedNotificationWorker({
+  subscription: 'api.v1.post-bookmark-reminder',
+  handler: async ({ postId, userId }, con) => {
     const postCtx = await buildPostContext(con, postId);
 
     if (!postCtx) {
@@ -34,6 +27,6 @@ const worker: NotificationWorker = {
       },
     ];
   },
-};
+});
 
 export default worker;
--- a/src/workers/notifications/worker.ts
+++ b/src/workers/notifications/worker.ts
@@ -1,8 +1,9 @@
 import { DataSource } from 'typeorm';
 import { FastifyBaseLogger } from 'fastify';
-import { Message } from '../worker';
+import { Message, messageToJson } from '../worker';
 import { NotificationBaseContext } from '../../notifications';
 import { NotificationType } from '../../notifications/common';
+import { PubSubSchema } from '../../common';
 
 export type NotificationHandlerReturn =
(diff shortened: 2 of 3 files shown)


src/workers/notifications/worker.ts
```diff
@@ -11,12 +12,48 @@ export type NotificationHandlerReturn =
...
+  logger: FastifyBaseLogger,
+) => Promise<NotificationHandlerReturn>;
+
 export interface NotificationWorker {
   subscription: string;
+  handler: NotificationWorkerHandler;
+  maxMessages?: number;
+}
+
+interface TypedNotificationWorkerProps<
```
capJavert, line 27: Would like to have `TypedWorker` used here, we can extend with return type something like `interface TypedWorker<T extends keyof PubSubSchema, R = void>
  extends BaseTypedWorker<PubSubSchema[T], R> {}`

but I am ok with this as interim soluion.

src/workers/notifications/worker.ts
```diff
@@ -11,12 +12,48 @@ export type NotificationHandlerReturn =
...
   ) => Promise<NotificationHandlerReturn>;
-  maxMessages?: number;
 }
+
+interface GenerateTypeWorkerResult<T extends keyof PubSubSchema> {
+  subscription: T;
+  handler: NotificationWorkerHandler;
+}
+
+export const generateTypedNotificationWorker = <
```
capJavert, line 44: Works nice to not break a lot of other code.

capJavert (approved):

src/workers/notifications/worker.ts
rebelchris: Maybe just note it down somewhere in web board so we don't forget

src/workers/notifications/postBookmarkReminder.ts
```diff
@@ -1,18 +1,11 @@
...
-interface Data {
-  postId: string;
-  userId: string;
-}
-
-const worker: NotificationWorker = {
-  subscription: 'api.post-bookmark-reminder-notification',
-  handler: async (message, con) => {
-    const { postId, userId }: Data = messageToJson(message);
+const worker = generateTypedNotificationWorker({
```
rebelchris, line 6: Isn't it missing the type parsing here somewhere?
EG:
```
const worker: TypedWorker<'comment-upvoted'> = {
```

rebelchris (approved):
Non blocking, but maybe action items :) ?

src/workers/notifications/postBookmarkReminder.ts
sshanzel: We don't have to anymore. The types will automatically be inferred through the `subscription` property.

rebelchris: Ah perfect 👌

src/workers/notifications/worker.ts
sshanzel: Yeah, I think it would be better to ultimately originate from the same type which is the `TypedWorker`. Just did this for a quick setup we can follow through. Let me write it down on the board.

src/workers/notifications/worker.ts
```diff
@@ -11,12 +12,48 @@ export type NotificationHandlerReturn =
...
+  T extends keyof PubSubSchema,
+  D extends PubSubSchema[T],
+>({
+  subscription,
```
capJavert, line 48: @sshanzel basically here subscription and topic names differ so you have to at least have two params topic and subscription or refactor to use `TypedWorker` type

PR #2059, opened 2024-07-12
fix: topic used as subscription

sshanzel:
The issue of it all was the wrong subscription name. To avoid it ever happening again, I've added some tests to validate the subscriptions.

### DIFF
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -30,5 +30,5 @@ export const personalizedDigestWorkers = commonDigestWorkers.map((worker) => {
     return worker;
   }
 
-  return { ...worker, ...args };
+  return { ...worker, args };
 });
--- a/.infra/workers.ts
+++ b/.infra/workers.ts
@@ -1,295 +1,34 @@
 import * as gcp from '@pulumi/gcp';
+import {
+  workers,
+  personalizedDigestWorkers as commonDigestWorkers,
+  digestDeadLetter,
+  WorkerArgs,
+} from './common';
 
-interface Worker {
-  topic: string;
-  subscription: string;
-  args?: {
-    enableMessageOrdering?: boolean;
-    ackDeadlineSeconds?: number;
-    expirationPolicy?: {
-      ttl: string;
-    };
-    deadLetterPolicy?: {
-      deadLetterTopic: string;
-      maxDeliveryAttempts: number;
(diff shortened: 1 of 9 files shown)


__tests__/workers/workers.ts
```diff
@@ -0,0 +1,39 @@
+import {
+  personalizedDigestWorkers,
+  workers as infraWorkers,
+} from '../../.infra/workers';
```
capJavert, line 4: This won't work currently since jest will start importing files and dependencies from `.infra` and we don't install those in tests. 😢

sshanzel: Ah, right. The deps will then be requested to be installed. Let me think of another way.

sshanzel: I found a workaround to make the importing from the test work, though not sure if it is ideal. Though at least it tests all subscriptions.

.infra/workers.ts
```diff
@@ -1,295 +1,34 @@
...
+
+export const personalizedDigestWorkers = commonDigestWorkers.map((worker) => {
+  const args: WorkerArgs = digestWorkersArgsMap[worker.subscription];
+
+  if (!args) {
+    return worker;
+  }
+
+  return { ...worker, ...args };
+});
```
capJavert, line 34: Can you please double check sure this merges correctly with what we had before? 🙏 Its pretty critical.

__tests__/workers/workers.ts
```diff
@@ -0,0 +1,45 @@
...
+  (acc, { subscription }) => ({ ...acc, [subscription]: true }),
+  {},
+);
+
+const infraDigestWorkersMap = infraDigestWorkers.reduce(
+  (acc, { subscription }) => ({ ...acc, [subscription]: true }),
+  {},
+);
+
+describe('pubsub workers', () => {
```
capJavert, line 21: love this!

capJavert (approved):

.infra/workers.ts
sshanzel: Ah, I just test ran this, hence the merge issue, just so we can try this path out. I can clear it out now.

sshanzel: Yes, I will double check to ensure things are still working as intended.

PR #2064, opened 2024-07-15
test: update snapshots

sshanzel:
@idoshamun @capJavert just wanted to make sure it was intended to update the id of the email. If so, I've updated the snapshots to reflect the new value.


capJavert (approved):
Yes, its the new default since it won! Tnx!

PR #2065, opened 2024-07-16
feat: add support for v3 search

rebelchris:
Basically v2 was a title based only search
v3 is an experiment to try out full body search again

Keep in mind default now is v2.

AS-448

### DIFF
--- a/src/schema/search.ts
+++ b/src/schema/search.ts
@@ -257,11 +257,11 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
       getSession(ctx.userId, id),
     searchPostSuggestions: async (
       source,
-      { query }: { query: string; version: number },
+      { query, version }: { query: string; version: number },
       ctx,
     ): Promise<GQLSearchSuggestionsResults> => {
       const { hits } = await searchMeili(
-        `q=${query}&attributesToRetrieve=post_id,title&attributesToSearchOn=title`,
+        `q=${query}&attributesToRetrieve=post_id,title${version === 2 && '&attributesToSearchOn=title'}`,
       );
       // In case ids is empty make sure the query does not fail
       const idsStr = hits.length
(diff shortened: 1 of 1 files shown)


src/schema/search.ts
```diff
@@ -310,7 +310,7 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
       const limit = Math.min(args.first || 10);
       const offset = getOffsetWithDefault(args.after, -1) + 1;
       const meilieSearchRes = await searchMeili(
-        `q=${args.query}&attributesToRetrieve=post_id&attributesToSearchOn=title&limit=${limit}&offset=${offset}`,
+        `q=${args.query}&attributesToRetrieve=post_id${args.version === 2 && '&attributesToSearchOn=title'}&limit=${limit}&offset=${offset}`,
```
capJavert, line 313: I think you need `args.version === 2 ? '&attributesToSearchOn=title' : ''`, because like this you will write `'false'`

Maybe good time to move to `URLSearchParams`, non blocking.

src/schema/search.ts
```diff
@@ -257,11 +257,11 @@ export const resolvers: IResolvers<unknown, Context> = traceResolvers({
       getSession(ctx.userId, id),
     searchPostSuggestions: async (
       source,
-      { query }: { query: string; version: number },
+      { query, version }: { query: string; version: number },
       ctx,
     ): Promise<GQLSearchSuggestionsResults> => {
       const { hits } = await searchMeili(
-        `q=${query}&attributesToRetrieve=post_id,title&attributesToSearchOn=title`,
+        `q=${query}&attributesToRetrieve=post_id,title${version === 2 && '&attributesToSearchOn=title'}`,
```
capJavert, line 264: Same here as above.

nensidosari (commented):
nthn to add apart Ante's comments

rebelchris:
@capJavert Good point, added now :)

capJavert (approved):

PR #2068, opened 2024-07-18
feat: bookmark reminder with queueing system

sshanzel:
- Introduced Temporal as our Queueing system.
  - Had to switch distros (need opinion from @omBratteng).
  - Discussed with Ido about local env: we can keep it simple for now as we don't need it on many things, however, we should eventually have this as part of our Tilt ecosystem. At the moment, to run it locally, you need to:
    - Install temporal: `brew install temporal`.
    - Run local dev server `temporal server start-dev --ip=0.0.0.0`.
  - We will discuss more about deployment once we finish reviewing this PR. Dennis has all the details for the Temporal Cloud server setup (we will connect with him). We might have to introduce a different `namespace`.
  - This system will have its own instance running similar to `digest`, etc.
- Bookmark reminder, send or remove reminder.
- Tested locally. Follow the instructions above if you want to do so on your local. To easily test it, make your FE send the value of `remindAt` to be T+10.


src/index.ts
```diff
@@ -279,3 +283,13 @@ export default async function app(
 
   return app;
 }
+
+run()
```
sshanzel, line 287: Temporary, just to run it locally with local dev temporal server. Should be part of our Tilt ecosystem.

Dockerfile
```diff
@@ -1,5 +1,15 @@
...
+    && sed -i -- 's/# deb-src/deb-src/g' /etc/apt/sources.list \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+               build-essential \
+               cdbs \
+               devscripts \
+               equivs \
+               fakeroot \
+    && apt-get clean \
+    && rm -rf /tmp/* /var/tmp/*
```
sshanzel, line 12: Just grabbed the snippet somewhere, pretty much clueless about Docker scripts. @omBratteng let me know if we should change something about this.

bin/runWorkflow.ts
```diff
@@ -0,0 +1,21 @@
+import { runReminderWorkflow } from '../src/queue/bookmark/utils';
```
sshanzel, line 1: Just a script to help with local testing. Nothing much.

src/queue/client.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Connection as TemporalConnection } from '@temporalio/client/lib/connection';
+
+const TEMPORAL_NAMESPACE = process.env.TEMPORAL_NAMESPACE || 'default';
+
+export const getTemporalClient = async (): Promise<Client> => {
+  const connection = await TemporalConnection.connect({
+    address: 'host.docker.internal:7233',
+  });
+
+  return new Client({ connection, namespace: TEMPORAL_NAMESPACE });
```
sshanzel, line 11: This will change based on how we connect to Temporal Cloud server, similar to `gatekeeper`.

Dockerfile
omBratteng: I'll hopefully be back tomorrow, so I'll take a look

sshanzel: No rush, just tagging you for more visibility 🙏

.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     resource_class: large
     docker:
       - image: cimg/node:20.12
-      - image: postgres:15-alpine
+      - image: postgres:15-bullseye
```
idoshamun, line 32: Do we really need to change pg image as well?

docker-compose.yml
```diff
@@ -12,7 +12,7 @@ volumes:
 services:
 
   db:
-    image: postgres:15.4-alpine
+    image: postgres:15.4-bullseye
```
idoshamun, line 15: Same question

src/index.ts
```diff
@@ -279,3 +283,15 @@ export default async function app(
 
   return app;
 }
+
+if (process.env.NODE_ENV === 'development') {
```
idoshamun, line 287: That should be removed I guess?

src/queue/bookmark/activities.ts
```diff
@@ -0,0 +1,32 @@
+import { InjectedProps } from '../common';
```
idoshamun, line 1: Instead of `queue` let's name it temporal to be explicit. The directory I mean

src/queue/bookmark/index.ts
```diff
@@ -0,0 +1,20 @@
+import { NativeConnection, Worker } from '@temporalio/worker';
+import { createActivities } from './activities';
+import createOrGetConnection from '../../db';
+import { WorkflowQueue } from '../common';
+import { logger } from '../../logger';
+
+export async function run() {
```
idoshamun, line 7: Let's make a generic executable for temporal that one of its use cases is the bookmark reminder, but we should think about the future here. Similarly we don't create a new index per subscription or notification

src/queue/bookmark/activities.ts
```diff
@@ -0,0 +1,32 @@
...
+import { triggerTypedEvent } from '../../common';
+import { Bookmark } from '../../entity';
+
+interface CommonBookmarkReminderParams {
+  userId: string;
+  postId: string;
+}
+
+export const createActivities = ({ con, logger }: InjectedProps) => ({
```
capJavert, line 10: You can import global logger instance, no need to pass it through arguments.

src/queue/bookmark/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  remindAt,
+}: BookmarkReminderParams) =>
+  generateWorkflowId(WorkflowTopic.Bookmark, WorkflowTopicScope.Reminder, [
+    userId,
+    postId,
+    remindAt.toString(),
+  ]);
+
+export const runReminderWorkflow = async (params: BookmarkReminderParams) => {
+  const workflowId = getReminderWorkflowId(params);
```
idoshamun, line 22: One thing to keep in mind is that the IDs are unique, so we can't trigger the workflow with the same params multiple times. Is that a concern you think?

src/queue/bookmark/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  const connection = await NativeConnection.connect({
+    address: 'host.docker.internal:7233',
+  });
+  const dbCon = await createOrGetConnection();
+  const worker = await Worker.create({
+    connection,
+    workflowsPath: require.resolve('./workflows'),
```
capJavert, line 14: You can't do this with normal path eg. `__dirname` or `path.resolve`? `require` will bug us if we ever move to ES modules.

src/queue/bookmark/workflows.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  const isValid = await validateBookmark({ userId, postId });
+
+  if (!isValid) {
+    return;
+  }
+
+  const diff = remindAt - Date.now();
+
+  await sleep(diff);
```
idoshamun, line 26: I'm not sure but potentially there's a function that accepts date and triggers automatically. Can you check please instead of using sleep?

src/queue/client.ts
capJavert: Do you maybe want only one instance so that only one connection and client are opened per app:
```ts
let temporalClient: Client;

export const getTemporalClient = async (): Promise<Client> => {
  if (!temporalClient) {
    const connection = await TemporalConnection.connect({
      address: 'host.docker.internal:7233',
    });
  
    temporalClient = new Client({ connection, namespace: TEMPORAL_NAMESPACE }) 
  }

  return temporalClient
}
```

src/schema/bookmarks.ts
```diff
@@ -21,6 +21,10 @@ import {
 import { SelectQueryBuilder } from 'typeorm';
 import { GQLPost } from './posts';
 import { Connection } from 'graphql-relay';
+import {
```
idoshamun, line 24: It's better to manage the workflows on the CDC side of things. It's much more fault tolerant and you also get the before and after values for easily canceling existing workflows. Should make everything a little bit simpler

idoshamun (changes requested):

capJavert (commented):
Excited for this, some initial comments

src/index.ts
sshanzel: Yes, just for testing purposes. Let me add a comment.

src/queue/bookmark/activities.ts
sshanzel: Cool. Let's do that.

src/queue/bookmark/utils.ts
sshanzel: We should be safe here, I included the date (turned to int) as params which will provide the uniqueness of the workflow.

src/queue/bookmark/workflows.ts
sshanzel: Yeah, sure I can check for alternatives. I will get back to this.

src/queue/bookmark/index.ts
sshanzel: I can try, let's see how it goes.

src/queue/client.ts
sshanzel: Yeah, we can do that too.

.circleci/config.yml
sshanzel: Didn't really dig into it, most likely it is not necessary, but just thought there could be compat issues. I can check with Ole-Martin if he sees any potential issue.

src/schema/bookmarks.ts
sshanzel: Yeah, thought of that too at first but did not push through. Let's do it.

src/queue/bookmark/index.ts
sshanzel: Yeah, just to note, that would mean we will use a single worker to host all workflows, having this in a separate folder/index, we can register the worker that will solely focus on bookmarks-related tasks. Then register another worker for other scope if we need to. That would allow us to scale per scope (keep bookmarks low, and high amount of workers on heavy computing). I guess I shouldn't be too worried about it for now and handle scalability issues when we get there.

sshanzel: Yes, that's actually the dilemma that I was having. Whether to scope this under `bookmark` or `notifications`. I feel `bookmark` scope is a bit small. Maybe we should aim for the bigger one which is `notification`, so every notification workflows are going to be stored here.

src/queue/bookmark/index.ts
sshanzel: Did not work 😢 

We could probably use `import.meta` but it requires `es2020`.

src/queue/bookmark/activities.ts
sshanzel: Cool. Let's do that.

__tests__/bookmarks.ts
```diff
@@ -636,8 +636,6 @@ describe('mutation setBookmarkReminder', () => {
 
     const result = await repo.findOneBy({ postId: 'p1', userId: loggedUser });
     expect(result.remindAt.toISOString()).toEqual(date);
-
-    // TODO MI-436: we should add another check whether the task was sent to the queue
```
sshanzel, line 640: There should be a relevant test in our CDC.

src/queue/bookmark/workflows.ts
sshanzel: The `sleep` method is told to be the [way to go](https://community.temporal.io/t/delayed-execution-of-workflow/3390/2), though the answer from the moderator was 3 years ago. Then a year after that there is an introduction of `delay` as part of the `WorkflowOptions`. I can maybe ask them for an updated answer.

src/queue/bookmark/index.ts
idoshamun: I didn't mean notification scope, I just meant broader scope for the run function. I don't mind having a bookmarks scope within the temporal (or queue) directory. But the runner needs to be generic. It's not a major issue we can always fix it later

src/migration/1721316842386-BookmarkReplica.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class BookmarkReplica1721316842386 implements MigrationInterface {
```
idoshamun, line 3: Remember to add the bookmark table to debezium include list

src/queue/bookmark/index.ts
sshanzel: Agreed. We can re-organize things when we get there.

sshanzel:
@omBratteng would love to get your opinion when you get the chance 🙏

.circleci/config.yml
omBratteng: Nah, it can stay alpine. Though changing it doesn't really matter.

Dockerfile
omBratteng: On line 21-23 there's a `RUN` that must be removed

omBratteng (changes requested):

Dockerfile
```diff
@@ -1,5 +1,15 @@
-FROM node:20-alpine
-RUN apk add g++ make python3 dumb-init
+FROM node:20-bullseye
```
omBratteng, line 1:
```suggestion
FROM node:20-bookworm
```
`bookworm` is latest stable version of Debian

omBratteng: Might even get away with slim version
```suggestion
FROM node:20-bookworm-slim
```

.circleci/config.yml
sshanzel: Awesome, let's keep unchanged in that case. Thank you!

src/queue/bookmark/workflows.ts
sshanzel: [They replied](https://community.temporal.io/t/sleep-vs-start-delay/12877/2?u=sshanzel) and it seems a start delay is being recommended. I shall update the process.

Dockerfile
sshanzel: Yep. Slim worked. I also used the RUN command provided by Temporal for slim images.

__tests__/workers/cdc/primary.ts
```diff
@@ -3263,3 +3276,161 @@ describe('user streak change', () => {
...
+      expect(runReminderWorkflow).toHaveBeenCalledWith({
+        postId: 'p1',
+        remindAt: date.getTime(),
+        userId: '1',
+      });
+    });
+  });
+
+  describe('on update', () => {
+    it('should NOT cancel reminder workflow if remind at is NOT present before', async () => {
```
nensidosari, line 3329: What does before mean here?

__tests__/workers/cdc/primary.ts
```diff
@@ -3263,3 +3276,161 @@ describe('user streak change', () => {
...
+        }),
+      );
+      expect(cancelReminderWorkflow).toHaveBeenCalledWith({
+        postId: 'p1',
+        remindAt: date.getTime(),
+        userId: '1',
+      });
+    });
+
+    it('should NOT run reminder workflow if remind at is NOT present after', async () => {
```
nensidosari, line 3362: what does after mean here?

bin/runWorkflow.ts
```diff
@@ -0,0 +1,21 @@
...
+
+runReminderWorkflow(params)
+  .then(() => {
+    console.log('Workflow started');
+  })
+  .catch((err) => {
+    console.log('Workflow failed:', err);
+  })
+  .finally(() => {
+    process.exit(1);
```
nensidosari, line 20: Do we want a log for workflow successful?

__tests__/workers/cdc/primary.ts
nensidosari: I did see before.remindAt and after.remindAt, so just more to understand how we separate those

nensidosari (commented):
Nothing much to add, looks good from what I can see.

package.json
```diff
@@ -23,7 +23,9 @@
     "start:background": "npm run cli -- background",
     "pretest": "npm run lint && cross-env NODE_ENV=test npm run db:migrate:reset",
     "test": "jest --testEnvironment=node --runInBand",
-    "typeorm": "typeorm-ts-node-commonjs"
+    "typeorm": "typeorm-ts-node-commonjs",
+    "temporal:dev": "cross-env TEMPORAL_ADDRESS=0.0.0.0:7233 nodemon ./src/temporal/local.ts",
+    "temporal:server": "temporal server start-dev --ip=0.0.0.0"
```
idoshamun, line 28: does it require anything to be installed locally?

package.json
```diff
@@ -23,7 +23,9 @@
     "start:background": "npm run cli -- background",
     "pretest": "npm run lint && cross-env NODE_ENV=test npm run db:migrate:reset",
     "test": "jest --testEnvironment=node --runInBand",
-    "typeorm": "typeorm-ts-node-commonjs"
+    "typeorm": "typeorm-ts-node-commonjs",
+    "temporal:dev": "cross-env TEMPORAL_ADDRESS=0.0.0.0:7233 nodemon ./src/temporal/local.ts",
```
idoshamun, line 27: we have our convention of `dev:background` and `start:background` can we stick to it please?

src/temporal/client.ts
```diff
@@ -0,0 +1,19 @@
...
+
+let client: Client;
+
+export const getTemporalClient = async (): Promise<Client> => {
+  if (client) {
+    return client;
+  }
+
+  const connection = await TemporalConnection.connect({
+    address: TEMPORAL_ADDRESS,
```
idoshamun, line 13: This is still not production ready right?

src/temporal/notifications/index.ts
```diff
@@ -0,0 +1,20 @@
+import { NativeConnection, Worker } from '@temporalio/worker';
+import { createActivities } from './activities';
+import createOrGetConnection from '../../db';
+import { WorkflowQueue } from '../common';
+import { TEMPORAL_ADDRESS } from '../config';
+
+export async function run() {
+  const connection = await NativeConnection.connect({
```
idoshamun, line 8: Why don't we use the client function here?

idoshamun:
I think we're almost there :)

package.json
sshanzel: Right. Let me update this quickly.

package.json
sshanzel: Yes, you will have to install `temporal` while we don't have it within our Tilt ecosystem. I might update the README, once we get merge.

bin/runWorkflow.ts
sshanzel: Yes, we did have console log for then and catch, but this is just a bin script anyway.

__tests__/workers/cdc/primary.ts
sshanzel: In CDC, there is an object for `before` and object for `after`. That's what we mean here.

src/temporal/client.ts
sshanzel: Yep, this will completely be different based on what I saw from `gatekeeper`.

src/temporal/notifications/index.ts
sshanzel: This one is for the worker, they use different libraries/APIs for connection.

capJavert: maybe also abstract this one so only one connection is used? Same as we did for client.

sshanzel: Yeah, sure, we can do that.

src/temporal/client.ts
sshanzel: Connected with Denis, and as discussed, the setup for client on how they do it now is the same as the one in our `gatekeeper` repo. I've now updated how we fetch the options for connecting to either locally or to prod.

src/temporal/notifications/utils.ts
```diff
@@ -23,27 +24,35 @@ export const getReminderWorkflowId = ({
...
 };
 
 export const cancelReminderWorkflow = async (
   params: ReminderWorkflowParams,
 ) => {
-  const client = await getTemporalClient();
-  const workflowId = getReminderWorkflowId(params);
-  const handle = client.workflow.getHandle(workflowId);
-  const description = await handle.describe();
+  try {
```
idoshamun, line 46: Please remove try/catch

src/temporal/notifications/utils.ts
```diff
@@ -23,27 +24,35 @@ export const getReminderWorkflowId = ({
   ]);
 
 export const runReminderWorkflow = async (params: ReminderWorkflowParams) => {
-  const workflowId = getReminderWorkflowId(params);
-  const client = await getTemporalClient();
-  const delay = params.remindAt - Date.now();
+  try {
```
idoshamun, line 27: Please remove try/catch

src/temporal/notifications/index.ts
```diff
@@ -1,20 +1,22 @@
...
 
 export async function run() {
-  const connection = await NativeConnection.connect({
-    address: TEMPORAL_ADDRESS,
-  });
+  const connection = await getTemporalWorkerConnection();
   const dbCon = await createOrGetConnection();
   const worker = await Worker.create({
     connection,
+    namespace: TEMPORAL_NAMESPACE,
```
idoshamun, line 13: No need for that

src/temporal/worker.ts
```diff
@@ -0,0 +1,21 @@
...
+import { logger } from '../logger';
+
+let connection: NativeConnection;
+
+export const getTemporalWorkerConnection = async () => {
+  if (connection) {
+    return connection;
+  }
+
+  try {
```
idoshamun, line 12: Please remove try/catch

src/temporal/client.ts
```diff
@@ -9,11 +10,14 @@ export const getTemporalClient = async (): Promise<Client> => {
     return client;
   }
 
-  const connection = await TemporalConnection.connect({
-    address: TEMPORAL_ADDRESS,
-  });
+  try {
```
idoshamun, line 13: Please remove try/catch

idoshamun (approved):

PR #2069, opened 2024-07-18
feat: post stats updated and triggers

capJavert:
- move `comments` property increment/decrement on `comment` table
- move `comments` property increment/decrement on `post` table
- add `statsUpdatedAt` column to `post` table and add trigger to update it
- added tests for everything I think (do tell if any cases miss)
  - also did manual test locally

notes:
  - gonna move cdc `metadataChangedAt` to trigger in separate PR (cc: Ido)


src/migration/1721296457196-CommentTriggers.ts
```diff
@@ -0,0 +1,42 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class CommentTriggers1721296457196 implements MigrationInterface {
```
capJavert, line 3: Updates `comments` count on `comment` table when sub comment is created/deleted

src/migration/1721302074280-PostCommentTriggers.ts
```diff
@@ -0,0 +1,42 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class PostCommentTriggers1721300643032 implements MigrationInterface {
```
capJavert, line 3: Updates `comments` count on `post` table when comment is created/deleted

src/migration/1721310640980-PostStatsTriggers.ts
```diff
@@ -0,0 +1,69 @@
...
+export class PostStatsTriggers1721310640980 implements MigrationInterface {
+    name = 'PostStatsTriggers1721310640980'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+          CREATE OR REPLACE FUNCTION user_post_vote_insert_trigger_function()
+          RETURNS TRIGGER AS $$
+          BEGIN
+              IF NEW.vote = 1 THEN
+                  UPDATE post SET upvotes = upvotes + 1, "statsUpdatedAt" = NOW() WHERE id = NEW."postId";
```
capJavert, line 13: Same trigger functions only adding `statsUpdatedAt` update.

src/migration/1721310640980-PostStatsTriggers.ts
```diff
@@ -0,0 +1,69 @@
...
+              RETURN OLD;
+          END;
+          $$ LANGUAGE plpgsql;
+        `)
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        const previousMigration = new VoteTriggers1692265909018()
+        await previousMigration.up(queryRunner)
+    }
```
capJavert, line 67: Basically run `up` from last migration touching these triggers, looks more readable then copying over all the SQL again.

idoshamun (commented):
I'd imagine the `statsUpdatedAt` as an independent trigger that checks if upvotes, comments, or downvotes have changed, if so it updated the timestamp.
Just to keep every trigger very focused. Wdyt?

src/migration/1721310640980-PostStatsTriggers.ts
capJavert: > I'd imagine the statsUpdatedAt as an independent trigger that checks if upvotes, comments, or downvotes have changed, if so it updated the timestamp.
Just to keep every trigger very focused. Wdyt?

@idoshamun commenting here so we can talk in a thread.

I thought about it but it would introduce another UPDATE call to `post` table which would again trigger CDC so opted for this approach.

capJavert:
Answering in thread https://github.com/dailydotdev/daily-api/pull/2069#discussion_r1682988980

src/migration/1721310640980-PostStatsTriggers.ts
idoshamun: Are you sure? I think we can manipulate the payload within the trigger so we can get away with a single statement without introducing an overhead. With a before update trigger i'm pretty certain it's possible

capJavert: You mean `NEW` and `OLD` inside triggers?

idoshamun: Yes the `NEW` specifically on the before update trigger. I think we have some examples around here

capJavert: But in this one `NEW` is a comment entity, since trigger is on comment table, and we need to update it on post entity

capJavert: Triggers would have to be on post table so I can do `NEW.statsUpdatedAt = NOW()`?

capJavert: Ignore my comments, yeah you are right, we are doing it for `votedAt`, I'll do it like that.

src/migration/1721301769783-PostStatsUpdated.ts
```diff
@@ -0,0 +1,31 @@
...
+        await queryRunner.query(`ALTER TABLE "post" ADD "statsUpdatedAt" TIMESTAMP NOT NULL DEFAULT now()`);
+
+        await queryRunner.query(`
+          CREATE OR REPLACE FUNCTION post_stats_updated_at_time()
+            RETURNS TRIGGER
+            LANGUAGE PLPGSQL
+            AS
+          $$
+          BEGIN
+            NEW."statsUpdatedAt" = now();
```
rebelchris, line 16: Ah for some reason I assumed we'd move the whole logic of plussing the actual numbers via trigger. (but maybe misunderstood that part)

rebelchris (commented):
Mainly more of a question if we didn't want to move the existing logic of plussing all the stats to a trigger?

src/migration/1721301769783-PostStatsUpdated.ts
capJavert: I did, comments were only ones that were not incremented via triggers.

rebelchris: Ahhh got it :)

rebelchris (approved):

idoshamun (approved):
Neat!

PR #2070, opened 2024-07-18
feat: add `includeSources` to feed settings

ilasw:
## Changes
- Added `includeSources ` field in feed settings;
- Added mutation logic;
- Updated feed tests for new feature;


src/entity/FeedSource.ts
```diff
@@ -12,6 +12,9 @@ export class FeedSource {
   @Index()
   sourceId: string;
 
+  @Column({ default: true })
+  blocked: boolean;
```
capJavert, line 16: maybe better to make it enum with `included` and  `excluded` or whatever naming. you want, so we can expand more easily?

rebelchris: This follows the tag approach and we thought about it, but you can actually only "block" or "unblock" (neutral meaning removal) so would keep this for now atleast. (as it's only a temporary solution till we get experiment results and we know wheter to go full blown with this)

capJavert: Ok, not really in the loop what it should done.

rebelchris (approved):
Looks good to me, but wait for Ante's approval as well please 🙏

src/entity/FeedSource.ts
capJavert: @ilasw do you have frontend PR which uses this, just would like to see how it works as well before merging?

ilasw: Hi @capJavert , yes! You can use is pulling from this branch: https://github.com/dailydotdev/apps/pull/3360 both in source page and in post article.

capJavert: tnx, gonna check it out!

sshanzel (approved):
Looks amazing! One question, would we get another PR for updating user preference on whether to receive new Squad post notifications if they were previously subscribed?

src/schema/feeds.ts
```diff
@@ -1461,7 +1464,6 @@ export const resolvers: IResolvers<any, Context> = traceResolvers({
       info,
     ): Promise<GQLFeedSettings> => {
       const feedId = args.feedId || ctx.userId;
-
```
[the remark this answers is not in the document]
ilasw: @capJavert tried but cannot revert, IDE auto format this way even if copy-paste from old source or rebase from main. There is some config that I have to enable/change?

ilasw:
> would we get another PR for updating user preference on whether to receive new Squad post notifications if they were previously subscribed?
Currently there is no reference for evolutions on the API side, not sure to understand this behavior!

sshanzel:
> > would we get another PR for updating user preference on whether to receive new Squad post notifications if they were previously subscribed?
> 
> Currently there is no reference for evolutions on the API side, not sure to understand this behavior!

When users follow and subscribe to a source, we store some information in the database. As a user, if I end up blocking a source that I subscribed to before, I expect to not receive notifications from them, and that information was the one stored in the database.

src/schema/feeds.ts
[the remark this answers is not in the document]
capJavert: do you have some other formatter enabled in your IDE, because eslint/prettied that is setup in project does not format that part.

ilasw:
> When users follow and subscribe to a source, we store some information in the database. As a user, if I end up blocking a source that I subscribed to before, I expect to not receive notifications from them, and that information was the one stored in the database.

@sshanzel  I've tested this behavior, and when you block a source, you will no longer receive notifications. It seems to have been like this even before this PR!

sshanzel:
> > When users follow and subscribe to a source, we store some information in the database. As a user, if I end up blocking a source that I subscribed to before, I expect to not receive notifications from them, and that information was the one stored in the database.
> 
> @sshanzel I've tested this behavior, and when you block a source, you will no longer receive notifications. It seems to have been like this even before this PR!

Nice. Thank you for confirming 🚀

PR #2072, opened 2024-07-19
fix: add a temporary log on source CDC

rebelchris:
Let's see what causes these 🥶

### DIFF
--- a/src/workers/cdc/primary.ts
+++ b/src/workers/cdc/primary.ts
@@ -624,7 +624,7 @@ const onSourceChange = async (
   logger: FastifyBaseLogger,
   data: ChangeMessage<Source>,
 ) => {
-  console.log('source operation: ', data.payload.op);
+  logger.info('source operation: ', data.payload.op);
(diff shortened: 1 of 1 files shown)


src/workers/cdc/primary.ts
```diff
@@ -624,6 +624,7 @@ const onSourceChange = async (
   logger: FastifyBaseLogger,
   data: ChangeMessage<Source>,
 ) => {
+  console.log('source operation: ', data.payload.op);
```
capJavert, line 627: probably wont to use `logger`?

capJavert:
```suggestion
logger.info(
{ op, before, after },
'source change',
);
```

do everything in single log?

rebelchris: was mainly afraid it could spool to many on creation, but we can try for a short term anyway.

omBratteng (approved):

capJavert (approved):

PR #2073, opened 2024-07-19
Hotfix debezium filter

rebelchris:
Exclude flag updates from filters

### DIFF
--- a/.infra/application.properties
+++ b/.infra/application.properties
@@ -9,12 +9,12 @@ debezium.source.database.password=%database_pass%
 debezium.source.database.dbname=%database_dbname%
 debezium.source.database.server.name=api
 debezium.source.table.include.list=public.comment,public.user_comment,public.comment_mention,public.source_request,public.post,public.user,public.post_report,public.source_feed,public.settings,public.reputation_event,public.submission,public.user_state,public.notification_v2,public.source_member,public.feature,public.source,public.post_mention,public.content_image,public.comment_report,public.user_post,public.banner,public.post_relation,public.marketing_cta,public.squad_public_request,public.user_streak
-debezium.source.column.exclude.list=public.post.tsv,public.post.placeholder,public.source.flags
+debezium.source.column.exclude.list=public.post.tsv,public.post.placeholder
 debezium.source.plugin.name=pgoutput
 debezium.source.heartbeat.interval.ms=60000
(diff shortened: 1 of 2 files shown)


.infra/application.properties
```diff
@@ -27,4 +27,7 @@ debezium.transforms.ReadOperationFilter.condition=!(valueSchema.field('op') && v
 debezium.transforms.PostsFilter.type=io.debezium.transforms.Filter
 debezium.transforms.PostsFilter.language=jsr223.groovy
 debezium.transforms.PostsFilter.condition=!(valueSchema.field('op') && value.op == 'u' && value.source.table == 'post' && value.before.views != value.after.views)
+debezium.transforms.SourceFilter.type=io.debezium.transforms.Filter
+debezium.transforms.SourceFilter.language=jsr223.groovy
+debezium.transforms.SourceFilter.condition=!(valueSchema.field('op') && value.op == 'u' && value.source.table == 'source' && value.before.flags != value.after.flags)
```
capJavert, line 32: Just hopping `flags` here does not become JSON object, can be check locally I guess or just YOLO. 🚀

src/workers/cdc/primary.ts
```diff
@@ -624,14 +624,6 @@ const onSourceChange = async (
...
   data: ChangeMessage<Source>,
 ) => {
-  logger.info(
-    {
-      op: data.payload.op,
-      before: data.payload.before,
-      after: data.payload.after,
-    },
-    'source change',
-  );
```
omBratteng, line 634: Can keep this here till after debezium change, just to be sure.

capJavert (approved):

src/workers/cdc/primary.ts
capJavert: +1

.infra/application.properties
omBratteng, line 32: Wondering if we could replace `value.before.flags != value.after.flags` with `valueSchema.field('after.flags')`? Since we want to exclude all flags?

rebelchris: I'm 99% sure debezium works based on regex flow anyway.
But let's see as logs still on now?

capJavert: Actually it missed events where flags is changed but something else also changed

rebelchris:
Closing for now.

PR #2075, opened 2024-07-19
feat(strict-ts): apply to users schema

capJavert:

This one focuses on:
- users schema
- context (splitting to BaseContext & AuthContext)
  - schemas that are not updated yet default to Context

This one squashed about 100 errors :success:

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -1,5 +1,11 @@
 import { opentelemetry } from './telemetry';
-import { DataSource, EntitySchema, ObjectType, Repository } from 'typeorm';
+import {
+  DataSource,
+  EntitySchema,
+  ObjectLiteral,
+  ObjectType,
+  Repository,
+} from 'typeorm';
 import { FastifyRequest, FastifyBaseLogger } from 'fastify';
 import { GraphQLDatabaseLoader } from '@mando75/typeorm-graphql-loader';
 import { Roles } from './roles';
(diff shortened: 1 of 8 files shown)


src/Context.ts
```diff
@@ -59,3 +65,10 @@ export type SubscriptionContext = {
   log: FastifyBaseLogger;
   userId?: string;
 };
+
+export type BaseContext = Omit<Context, 'userId' | 'trackingId'>;
+
+export type AuthContext = BaseContext & {
+  userId: string;
+  trackingId: string;
+};
```
capJavert, line 74: Two new contexts:
- `BaseContext` - used as type base in schema resolver
- `AuthContext` used in query & mutations with `@auth` directive to make `userId` being defined which is guaranteed by the directive

src/schema/comments.ts
```diff
@@ -550,7 +550,7 @@ export const reportCommentReasons = new Map([
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const resolvers: IResolvers<any, Context> = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  Query: traceResolverObject<any, any>({
+  Query: traceResolverObject<any, any, Context>({
```
capJavert, line 553: Added generic for `Context` to resolver types so for schemas I did not yet convert to strict TS I am just giving `Context` for backward compatibility.

src/schema/users.ts
```diff
@@ -963,16 +980,25 @@ export const resolvers: IResolvers<any, Context> = {
...
+      getUserReadHistory({
+        con: ctx.con,
+        userId: id,
+        after: new Date(after),
+        before: new Date(before),
+      }),
+    userStreak: async (
+      _,
+      __,
+      ctx: AuthContext,
```
capJavert, line 992: Query `userStreak` has `@auth` directive in gql definition so `ctx.userId` is required to be present, this is why it is using `AuthContext` type.

capJavert: Other queries that do not have `@auth` can just use `Context` as before.

src/Context.ts
rebelchris, line 74: Wouldn't this do?
```suggestion
export type AuthContext = Context;
```

rebelchris (approved):
I see what you mean, quite a mission

src/Context.ts
capJavert: No, since Context has `userId` and `trackingId` optional and then that breaks a lot of queries/mutations where it expects for `ctx.userId` to be defined.

PR #2076, opened 2024-07-19
feat: add new social handles and validation

capJavert:
- added validation for different social handles
- expanded `validateRegex` func to be able to extract matched value eg. user submits `https://github.com/capJavert` but in the DB we save only handle part `capJavert`
- take a look at regex tests to see accepted/not accepted values for each social handle 

Front PR https://github.com/dailydotdev/apps/pull/3378

AS-444


sshanzel (approved):
Looks great! 🚀

src/common/object.ts
```diff
@@ -22,20 +22,35 @@
...
+  const mutatedData: ObjectLiteral = {
+    ...data,
+  };
   const result = params.reduce((result, [key, value, regex, isRequired]) => {
     if (isNullOrUndefined(value)) {
       return isRequired ? { ...result, [key]: `${key} is required!` } : result;
     }
 
-    const isValid = regex.test(value as string);
+    const matchResult = value.match(regex);
```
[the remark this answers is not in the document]
capJavert: @idoshamun do you have time to take a look, basically I think its false positive since `value`  is string which I match by regex and it is not constructed from user input its a static `handleRegex` value in `object.ts`?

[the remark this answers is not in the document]
idoshamun: I see. We can safely ignore

__tests__/users.ts
```diff
@@ -2047,6 +2143,264 @@ describe('mutation updateUserProfile', () => {
...
+  });
+
+  it('should validate reddit handle', () => {
+    const valid = [
+      'lee',
+      'reddit.com/u/lee',
+      'reddit.com/user/lee',
+      'https://reddit.com/u/lee',
+      'https://reddit.com/user/lee',
+      'https://reddit.com/user/lee/',
```
nensidosari, line 2287: Would `https://reddit.com/u/lee/` be valid as well?

nensidosari: not sure if we intend to test all valid use cases, if yes examples like this are missing in some tests

nensidosari (approved):
Looks great, just one minor comment on tests

__tests__/users.ts
capJavert: I could have missed ones I will check, I did not really test all the valid ones (too many) but the most common ones users could paste into the box. 

> Would https://reddit.com/u/lee/ be valid as well?

yes, I'll add!

capJavert: Added in 617ed0d749301ffd7bbc6d573d885da618e34455

__tests__/users.ts
```diff
@@ -2047,6 +2143,280 @@ describe('mutation updateUserProfile', () => {
...
+  it('should validate github handle', () => {
+    const valid = [
+      'lee',
+      'github.com/lee',
+      'https://github.com/lee',
+      'https://github.com/lee/',
+    ];
+    const invalid = [
+      'lee#',
+      'http://github.com/lee',
```
rebelchris, line 2156: Considering this one or the one with queryparams maybe lets log some fails if need be we can just abstract these ourselves (but doubt they happen much)

src/entity/user/User.ts
```diff
@@ -88,6 +96,38 @@ export class User {
   @Index('users_github_unique', { unique: true })
   github?: string;
 
+  @Column({ length: 39, nullable: true })
```
rebelchris, line 99: Are these set to 39 a max?

rebelchris (approved):
Looks great nothing blocking

src/common/users.ts
```diff
@@ -449,3 +449,36 @@ export enum LogoutReason {
...
+  /^(?:(?:https:\/\/)?(?:www\.)?youtube\.com\/)?@?(?<value>[\w-]{2,})\/?$/;
+
+export const linkedinSocialUrlMatch =
+  /^(?:(?:https:\/\/)?(?:www\.)?linkedin\.com\/in\/)?(?<value>[\w-]{2,})\/?$/;
+
+export const mastodonSocialUrlMatch =
+  /^(?<value>https:\/\/(?:[a-z0-9-]+\.)*[a-z0-9-]+\.[a-z]{2,}\/@[\w-]{2,}\/?)$/;
+
+export const socialUrlMatch =
+  /^(?<value>https:\/\/(?:[a-z0-9-]{1,50}\.){0,5}[a-z0-9-]{1,50}\.[a-z]{2,24}\/?[\w\-?&=]{0,100})$/;
```
omBratteng, line 484: Since these don't really restrict to any domains, perhaps we should add a blacklist of (common) domains we don't want (i.e. porn sites etc)

src/entity/user/User.ts
capJavert: yes, existing handles are the same so seems like reasonable limit

src/common/users.ts
capJavert: Since we allowed anything up until this point this is already much better. But I agree this could be cool as well. I'll look into it.

__tests__/users.ts
capJavert: We should see increase in validation errors, I added `warn` log.

src/common/users.ts
capJavert: Added task for now https://dailydotdev.atlassian.net/browse/AS-466

omBratteng (approved):

PR #2077, opened 2024-07-23
feat: use cursor resolver in anon feed

capJavert:
For anonymous feed we want to use the cursor resolver but due to no user being authenticated this before lead to memory leaks due to `userId` not being provided to queries leading to a lot of rows being selected.

This reverts anonymous feed to use cursor resolver while also avoiding queries with optional `feedId` and `userId`.

Original fix https://github.com/dailydotdev/daily-api/pull/1692

Will also monitor pods after deploy to check OOM not happening again.


src/common/feedGenerator.ts
```diff
@@ -103,23 +107,27 @@ export const feedToFilters = async (
...
+              .select('fs."sourceId"')
+              .from(FeedSource, 'fs')
+              .where('fs."feedId" = :feedId', { feedId })
+              .getQuery();
+
+            return `s.id IN (${subQuery})`;
+          })
+          .execute()
+      : [],
+    feedId && userId
```
capJavert, line 130: Avoid doing sub queries if no feedId/userId since they should be empty.

src/common/feedGenerator.ts
```diff
@@ -192,7 +200,7 @@ export const selectRead = (
   const query = builder
     .select('1')
     .from(View, 'view')
-    .where(`view.userId = :userId`, { userId })
+    .where(`view.userId = :userId`, { userId: userId || '' })
```
capJavert, line 203: Since query is used in a lot of different ones as `EXISTS${subquery}` I added this to avoid querying with empty values.

idoshamun (approved):

PR #2079, opened 2024-07-23
fix: temporarily disable pino telemetry instrumentation

omBratteng:
Temporarily disables the pino telemetry instrumentation, as I noticed the logs appears twice, once as a single line, the once output on multiple lines.
Will come back later to this to look at why it happens, but in the meantime, not much value is lost.


src/telemetry/opentelemetry.ts
```diff
@@ -77,13 +77,13 @@ const instrumentations = [
...
-      traceFlags: 'traceSampled',
-    },
-  }),
+  // new PinoInstrumentation({
+  //   logKeys: {
+  //     traceId: 'trace',
+  //     spanId: 'spanId',
+  //     traceFlags: 'traceSampled',
+  //   },
+  // }),
```
capJavert, line 86: just remove the lines, we have it in history, also maybe add a backlog task for it?

omBratteng: Lines removed

omBratteng: Will create backlog task for it

capJavert (approved):

PR #2080, opened 2024-07-23
fix: check if IP is in subnet

omBratteng:
I've checked that it finds the IP in the array of IPs


vpol (approved):

idoshamun (approved):

.infra/Pulumi.adhoc.yaml
```diff
@@ -82,6 +82,5 @@ config:
     typeormUser: postgres
     tz: UTC
     urlPrefix: http://api.local.com
-    vordrIps: ""
```
capJavert, line 85: why removed?

omBratteng: `process.env.VORDR_IPS?.split(',')` returns `['']`, which makes the `isInSubnet` subnet fail. So if I remove it, it falls back to `[]`

capJavert: add `filter(Boolean)`?

omBratteng: Yeah, just did that now

idoshamun (approved):

PR #2082, opened 2024-07-24
fix: bookmark remind at precision

sshanzel:
After further investigation, the workflows are not ending, and have found out that the `remindAt` value being sent is a bit large than expected.

The thing is, postgresql timestamp's precision is 6, and JavaScript uses only 3 fractional seconds.

Verified locally that workflows are now receiving the correct value.


nensidosari (approved):

capJavert (commented):
Maybe we can reparse it when sending to temporal, that seems more bulletproof (to not depend on implementation or language)?

sshanzel:
> Maybe we can reparse it when sending to temporal, that seems more bulletproof (to not depend on implementation or language)?

Yeah, was thinking of that too, but how can we also tell that the precision is always 6 that we need to parse it? Do we base it off the length of the numbers?

capJavert:
@sshanzel what is the exact value you get in JS? Can you add a test that fails or provide example?

sshanzel:
Good point. Let me add a test, but to answer, the values we are getting from postgres to JS is `1721825010547000`. We have an extra `000`.

sshanzel:
If I push the test it would success now, but the result is along the lines of the presented data below (just imagine removing the decimal place). We can sync on it tomorrow if you get some time.

Screenshot 2024-07-24 at 11 38 38 PM

src/common/utils.ts
```diff
@@ -102,3 +102,13 @@ export function camelCaseToSnakeCase(
 export function debeziumTimeToDate(time: number): Date {
   return new Date(Math.floor(time / 1000));
 }
+
+export const ensureMsFormat = (value: number): number => {
```
sshanzel, line 106: To support the tests that are not able to replicate the same value that debezium provides.

capJavert (approved):

PR #2087, opened 2024-07-25
fix: post author can not receive notifications from replies in post;

ilasw:
## Changes:
Removed from excluded list post author and post scout. We are only looking for thread replier and those users are stakeholders that may want to be notified on replies. This fix MI-455 but need an extensive test.


capJavert (commented):
We should add tests for the cases we want to cover and verify what happens currently.

src/workers/notifications/commentReply.ts
```diff
@@ -49,12 +49,7 @@ const worker: NotificationWorker = {
       .distinct(true)
       .where('"parentId" = :parentId', { parentId: parent.id })
       .andWhere('"userId" not in (:...exclude)', {
-        exclude: [
-          parent.userId,
-          comment.userId,
-          postCtx.post.authorId ?? '',
-          postCtx.post.scoutId ?? '',
-        ],
+        exclude: [parent.userId, comment.userId],
```
idoshamun, line 52: The reason we had them in the list is to prevent multiple notifications for the same reason.
I think authors or scouts receive notifications for any new comment anyway, so please double check that

ilasw: I've tested but you still receive only one notification because on `articleNewCommentHandler` (from `src/workers/notifications/utils.ts`) you unsubscribe onece you comment. There is this piece of code:

```ts
  if (isReply && (post.authorId || post.scoutId)) {
    const ids = [...new Set([post.authorId, post.scoutId])];
    const threadFollower = await repo
      .createQueryBuilder()
      .select('"userId"')
      .where(`(id = :id OR "parentId" = :id)`, { id: comment.parentId })
      .andWhere({ userId: In(ids) })
      .groupBy('"userId"')
      .getRawMany();

    if (threadFollower.length) {
      threadFollower.forEach(({ userId }) => excludedUsers.push(userId));
    }
  }
```

I recently tried this locally with 3 users and noticed that, as an author, I stopped receiving notifications only once I (as the author) commented on the post. I don't know if the fix below is the best solution, but something is conflicting between these two handlers.

capJavert: I'll take a look next week, tnx for checking

capJavert: Basically above logic @ilasw mentioned is to avoid including any users following a thread from getting new comment notification since they would get reply notification. 

Looks to me its safe to remove scout and author from array in commentReply as was done here because if they commented in the thread they probably want a notification for reply cc: @idoshamun ?

Locally it does not look there are any double notifications from what I can see.

__tests__/workers/notifications.ts
```diff
@@ -1734,4 +1733,36 @@ it('should add squad reply notification', async () => {
...
+  await con.getRepository(Comment).save([
+    {
+      id: 'c6',
+      postId: 'p1',
+      userId: '3',
+      content: 'sub comment5',
+      createdAt: new Date(2020, 1, 6, 0, 0),
+      parentId: 'c1',
+    },
+  ]);
```
capJavert, line 1749: If you can add additional test case for this and not modify the current test. So something in line `author should get a reply notification if they commented`, same for scout (it can be two test cases).

ilasw: @capJavert I've updated with a more generic one `users should get a reply notification if they commented in the same thread`, is this ok? Because the fix is involving the thread users more than the author/scout. What do you think?

capJavert: yeah, looks good, let's keep are eyes also on any reports with this after merge if we missed some edge case 🙏

capJavert (approved):

src/workers/notifications/commentReply.ts
ilasw: @idoshamun is ok to merge?

idoshamun: Ok! Go ahead :)

idoshamun (approved):

PR #2088, opened 2024-07-26
feat(strict-ts): apply to gql schemas 2

capJavert:

This one focuses on: 
- actions schema
- alerts schema
- bookmarks schema
- comments schema

### DIFF
--- a/src/schema/alerts.ts
+++ b/src/schema/alerts.ts
@@ -1,7 +1,7 @@
 import { Alerts, ALERTS_DEFAULT, UserActionType } from '../entity';
 import { IResolvers } from '@graphql-tools/utils';
 import { traceResolvers } from './trace';
-import { Context } from '../Context';
+import { AuthContext, BaseContext } from '../Context';
 import { DataSource } from 'typeorm';
 import { insertOrIgnoreAction } from './actions';
 import { GQLEmptyResponse } from './common';
@@ -154,7 +154,7 @@ export const typeDefs = /* GraphQL */ `
     """
     Get the alerts for user
     """
-    userAlerts: Alerts!
+    userAlerts: Alerts! @auth
   }
 `;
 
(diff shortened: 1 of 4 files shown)


src/schema/bookmarks.ts
```diff
@@ -161,7 +162,7 @@ export const typeDefs = /* GraphQL */ `
       The query to search for
       """
       query: String!
-    ): SearchBookmarksSuggestionsResults!
+    ): SearchBookmarksSuggestionsResults! @auth
```
capJavert, line 165: bookmark suggestions should be behind auth if bookmarks search is, userId was required either way

src/schema/alerts.ts
```diff
@@ -207,29 +207,35 @@ export const updateAlerts = async (
...
+  userId?: string,
 ): Promise<Omit<Alerts, 'flags'>> => {
-  const alerts = await con.getRepository(Alerts).findOneBy({
-    userId,
-  });
+  const alerts = userId
+    ? await con.getRepository(Alerts).findOneBy({
+        userId,
+      })
+    : undefined;
```
omBratteng, line 216: Instead of a ternary here, could we do an early return if there is no userId?

capJavert: Usually yes, but since we return `ALERTS_DEFAULT` on the bottom I just left it like this so the rest of the code is the same. I would have to return `ALERTS_DEFAULT` twice.

omBratteng: I see. Shouldn't the findOneBy return null/undefined if it doesn't find anything?

omBratteng (approved):

src/schema/alerts.ts
capJavert: yeah, but its a type issue where it can do one or the order, and also we don't want to even run a query if no userId

PR #2091, opened 2024-07-30
fix: temporal workflow has already started

sshanzel:
- Create a function to catch errors that results to NotFound. As per documentation, running the `handle. describe` would result in a not found error rather than returning null.
- When the same workflow id is running, we should just ignore executing the same workflow.
- When workflow is not found and trying to cancel the id, we should just ignore it.
- Setup a test environment for Temporal. This involves worker registration in the test environment for integration testing.

~~TODO:~~
- [x] ~~Write tests for the whole Temporal integration.~~

### DIFF
--- a/src/temporal/common.ts
+++ b/src/temporal/common.ts
@@ -37,9 +37,11 @@ export const getWorkflowHandle = async (
   return client.workflow.getHandle(workflowId);
 };
 
-export const getDescribeOrError = (handle: WorkflowHandle) => {
+export const getDescribeOrError = async (
+  handle: WorkflowHandle,
+): Promise<WorkflowExecutionDescription | undefined> => {
   try {
-    return handle.describe();
+    return await handle.describe();
   } catch (error) {
     if (error.name === TemporalError.NotFound) {
       return;
@@ -51,7 +53,7 @@ export const getDescribeOrError = (handle: WorkflowHandle) => {
 
 export const getWorkflowDescription = async (
   workflowId: string,
-): Promise<WorkflowExecutionDescription> => {
+): Promise<WorkflowExecutionDescription | undefined> => {
   const handle = await getWorkflowHandle(workflowId);
 
   return getDescribeOrError(handle);
--- a/src/temporal/common.ts
+++ b/src/temporal/common.ts
@@ -1,4 +1,7 @@
 import { DataSource } from 'typeorm';
+import { WorkflowHandle } from '@temporalio/client';
+import { WorkflowExecutionDescription } from '@temporalio/client/src/types';
+import { getTemporalClient } from './client';
 
 export enum WorkflowTopic {
   Notification = 'notification',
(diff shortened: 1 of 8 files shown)


src/temporal/common.ts
```diff
@@ -21,3 +24,35 @@ export const generateWorkflowId = (
...
+      return;
+    }
+
+    throw error;
+  }
+};
+
+export const getWorkflowDescription = async (
+  workflowId: string,
+): Promise<WorkflowExecutionDescription> => {
```
capJavert, line 54: This then returns workflow description or undefined (in case of NotFound) right?

src/temporal/notifications/utils.ts
```diff
@@ -38,12 +46,16 @@ export const runReminderWorkflow = async (params: ReminderWorkflowParams) => {
...
 ) => {
-  const client = await getTemporalClient();
   const workflowId = getReminderWorkflowId(params);
-  const handle = client.workflow.getHandle(workflowId);
-  const description = await handle.describe();
+  const handle = await getWorkflowHandle(workflowId);
+
+  if (!handle) {
+    return;
+  }
```
capJavert, line 54: Maybe will be more visible when tests are added, but in what case the handle is missing?

src/temporal/common.ts
sshanzel: Yes. Should we explicitly say `| undefined` as part of the TS strict mode?

src/temporal/notifications/utils.ts
sshanzel: It is a third-party package, so I was just assuming the worst that it could be null or undefined.

src/temporal/common.ts
capJavert: I think so yes

src/temporal/notifications/utils.ts
capJavert: Assuming worst is fine but returning early does not fix it, I think its important to understand if and when it can be nullish and if we need to handle it.

capJavert: But lets see the test cases

sshanzel: In this case, it is canceling a workflow, so if it does not exist, we can just ignore the process because we will have nothing to cancel.

src/temporal/common.ts
```diff
@@ -21,3 +24,37 @@ export const generateWorkflowId = (
...
   con: DataSource;
 }
+
+export enum TemporalError {
+  NotFound = 'WorkflowNotFoundError',
+}
+
+export const getWorkflowHandle = async (
+  workflowId: string,
+): Promise<WorkflowHandle> => {
```
capJavert, line 34: This can also be `WorkflowHandle | undefined` then, if `getHandle` can be `undefined`?

src/temporal/config.ts
```diff
@@ -12,7 +12,7 @@ interface ServerOptions {
 }
 
 export const TEMPORAL_ADDRESS =
-  process.env.TEMPORAL_ADDRESS || 'host.docker.internal:7233';
+  process.env.TEMPORAL_ADDRESS || 'localhost:7233';
```
capJavert, line 15: is this accidental change?

capJavert (approved):

src/temporal/config.ts
sshanzel: No, this should be good until we have the workers spooled within Tilt.

src/temporal/common.ts
sshanzel: Agreed. Let me update this one.

src/temporal/config.ts
capJavert: 👌

PR #2092, opened 2024-08-01
feat: add streak config

omBratteng:
Related apps PR https://github.com/dailydotdev/apps/pull/3386

AS-437

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,18 @@ module.exports = {
   rules: {
     'prettier/prettier': 'error',
     'unused-imports/no-unused-imports': 'error',
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
+            name: 'date-fns',
+            importNames: ['isWeekend'],
+            message: 'Please import from src/common instead.',
+          },
+        ],
+      },
+    ],
   },
   overrides: [
     {
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -30,12 +30,12 @@ export interface GQLUserStreak {
   current?: number;
   lastViewAt?: Date;
   userId: string;
+  weekStart: DayOfWeek;
 }
 
 export interface GQLUserStreakTz extends GQLUserStreak {
   timezone: string;
   lastViewAtTz: Date;
-  weekStart: DayOfWeek;
 }
 
 export const fetchUser = async (
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -7,6 +7,7 @@ import { getTimezonedStartOfISOWeek, getTimezonedEndOfISOWeek } from './utils';
 import { GraphQLResolveInfo } from 'graphql';
 import { utcToZonedTime } from 'date-fns-tz';
 import { sendAnalyticsEvent } from '../integrations/analytics';
+import { DayOfWeek, DEFAULT_WEEK_START } from './date';
 
 export interface User {
   id: string;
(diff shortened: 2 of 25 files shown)


.eslintrc.js
```diff
@@ -9,6 +9,18 @@ module.exports = {
   rules: {
     'prettier/prettier': 'error',
     'unused-imports/no-unused-imports': 'error',
+    'no-restricted-imports': [
```
rebelchris, line 12: Nice thanks for adding this failsafe :D

rebelchris (approved):
Looks good from my side 🙏

src/routes/boot.ts
```diff
@@ -383,6 +383,7 @@ const getUser = (con: DataSource, userId: string): Promise<User | null> =>
       'createdAt',
       'cover',
       'experienceLevel',
+      'weekStart',
```
capJavert, line 386: do we need it in boot?

src/cron/hourlyNotifications.ts
```diff
@@ -24,7 +26,7 @@ const cron: Cron = {
   handler: async (con, logger) => {
     const personalizedDigestQuery = con
       .createQueryBuilder()
-      .select('upd.*, u.timezone')
+      .select('upd.*, u.timezone, u.weekStart')
```
capJavert, line 29: think it needs `u."weekStarts"` due to casing

src/common/users.ts
```diff
@@ -406,7 +408,20 @@ export const clearUserStreak = async (
...
+  startOfWeek: DayOfWeek = DEFAULT_WEEK_START,
+) => {
+  if (startOfWeek === DayOfWeek.Sunday) {
+    return (
+      (day === Day.Saturday && difference > FREEZE_DAYS_IN_A_WEEK) ||
+      (day === Day.Sunday &&
+        difference > FREEZE_DAYS_IN_A_WEEK + MISSED_LIMIT) ||
+      (day > Day.Sunday && difference > MISSED_LIMIT)
+    );
+  }
```
capJavert, line 423: :fml: 😆 a bit hard to parse, but it is what iti s

src/cron/dailyDigest.ts
```diff
@@ -25,7 +27,7 @@ const cron: Cron = {
   handler: async (con, logger) => {
     const personalizedDigestQuery = con
       .createQueryBuilder()
-      .select('upd.*, u.timezone')
+      .select('upd.*, u.timezone, u.weekStart')
```
capJavert, line 30: same as https://github.com/dailydotdev/daily-api/pull/2092#discussion_r1701642228

src/common/users.ts
omBratteng: I know 😢
Tried to think of a better way to solve it, but alas to no avail. Maybe worth to take a look another time.

src/routes/boot.ts
omBratteng: Technically not needed in boot, but we need it when we render the streak modal.
Same on notifications settings page

src/cron/hourlyNotifications.ts
omBratteng: Not sure, but I added just to be safe

.eslintrc.js
omBratteng: Yeah, thought it was a good addition just to be safe that we don't accidentally import `isWeekend` from `date-fns` and expect it to work differently.
Actually surprise that `isWeekend` from `date-fns` didn't have this feature...

capJavert: Even worse they have it on `startOfWeek` util 🙈

src/routes/boot.ts
capJavert: If its not used anywhere from boot I'd remove it, because streak we pull from separate query I think?

capJavert (approved):

src/routes/boot.ts
omBratteng: I'll see if I can add it to `userStreak` query, and not in boot.

omBratteng:
@capJavert refactored it out of boot and into streak query. Working on refactoring in apps PR.

src/common/users.ts
omBratteng: @capJavert what do you think about this? A little bit more readable hopefully

```ts
export const shouldResetStreak = (
  day: number,
  difference: number,
  startOfWeek: DayOfWeek = DEFAULT_WEEK_START,
) => {
  const firstDayOfWeek =
    startOfWeek === DayOfWeek.Monday ? Day.Monday : Day.Sunday;

  const lastDayOfWeek =
    startOfWeek === DayOfWeek.Monday ? Day.Sunday : Day.Saturday;

  return (
    (day === lastDayOfWeek && difference > FREEZE_DAYS_IN_A_WEEK) ||
    (day === firstDayOfWeek &&
      difference > FREEZE_DAYS_IN_A_WEEK + MISSED_LIMIT) ||
    (day > firstDayOfWeek && difference > MISSED_LIMIT)
  );
};
```

rebelchris (approved):
Approved.
Personally find this one better indeed: https://github.com/dailydotdev/daily-api/pull/2092/files#r1703623882

src/common/users.ts
omBratteng: Changed the function to this ☝️

capJavert: Looks better yes, thanks for the effort. I would personally maybe split or expressions in separate if's if we add more down the line, but good for this stage.

nensidosari (approved):
Was trying to find names for the return conditions but too hard 😅

src/common/users.ts
omBratteng: I wanted to extract them into separate constants, but I couldn't come up with any good names for them :triggered:

PR #2094, opened 2024-08-01
feat(strict-ts): apply to posts schema

capJavert:

This one focuses on: 
- posts schema
- minor fixes from past few PRs that were added to `main`
- almost half way there 💦

### DIFF
--- a/src/common/pubsub.ts
+++ b/src/common/pubsub.ts
@@ -202,7 +202,7 @@ export const notifyView = (
   log: EventLogger,
   postId: string,
   userId: string,
-  referer: string,
+  referer: string | undefined,
   timestamp: Date,
   tags?: string[],
 ): Promise<void> =>
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -38,7 +38,7 @@ export type PostFlags = Partial<{
   private: boolean;
   visible: boolean;
   showOnFeed: boolean;
-  promoteToPublic: number;
+  promoteToPublic: number | null;
   deletedBy: string;
 }>;
 
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -11,8 +11,8 @@ import {
   SourcePermissions,
   sourceTypesWithMembers,
 } from './sources';
-import { Context } from '../Context';
-import { traceResolverObject } from './trace';
+import { AuthContext, BaseContext, Context } from '../Context';
+import { traceResolvers } from './trace';
 import {
   CreatePost,
   CreatePostArgs,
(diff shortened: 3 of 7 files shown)


src/entity/posts/Post.ts
```diff
@@ -38,7 +38,7 @@ export type PostFlags = Partial<{
   private: boolean;
   visible: boolean;
   showOnFeed: boolean;
-  promoteToPublic: number;
+  promoteToPublic: number | null;
```
capJavert, line 41: You can't unsed jsonb value with `undefined` so we set null (its already happening this is just the type update)

src/workers/postCommentedRedis.ts
```diff
@@ -14,7 +14,10 @@ const worker: Worker = {
     const data: Data = messageToJson(message);
     try {
       const postNotificatiion = await getPostNotification(con, data.postId);
-      await redisPubSub.publish('events.posts.commented', postNotificatiion);
+
+      if (postNotificatiion) {
+        await redisPubSub.publish('events.posts.commented', postNotificatiion);
+      }
```
capJavert, line 20: It would trigger `publish` with `undefined` which has no point

src/common/pubsub.ts
```diff
@@ -202,7 +202,7 @@ export const notifyView = (
   log: EventLogger,
   postId: string,
   userId: string,
-  referer: string,
+  referer: string | undefined,
```
capJavert, line 205: can't make it `referer?` because it is not object argument and don't want to refactor all over the place

src/schema/posts.ts
```diff
@@ -173,12 +177,12 @@ export interface SubmitExternalLinkArgs extends ExternalLink {
...
   con: DataSource,
   postId: string,
-): Promise<GQLPostNotification> => {
+): Promise<GQLPostNotification | undefined> => {
   const post = await con
     .getRepository(Post)
     .findOne({ where: { id: postId }, select: ['id', 'upvotes', 'comments'] });
   if (!post) {
-    return null;
+    return undefined;
```
rebelchris, line 185: Probably fine, but keep a close eye on prod if it doesn't break anything accidental (by expecting nulls or something like that.

src/workers/postCommentedRedis.ts
rebelchris: Nice, we need more of these 😂

rebelchris (approved):
Overall no big issues it seems :)

src/schema/posts.ts
capJavert: its used for https://github.com/dailydotdev/daily-api/pull/2094#discussion_r1701363975 thats why I changed it

PR #2096, opened 2024-08-02
fix: decrypt slack user token and error handling

capJavert:
Forgot to decrypt token in worker which resulted in failed auth on slack. Also added error handling so we can catch future errors in message handler.


capJavert:
Merging so I can test again on prod.

PR #2097, opened 2024-08-02
fix: add domain exposure

rebelchris:
Since we don't want to expose url/canonical to FE, decided to handle it like this on API side.

### DIFF
--- a/src/common/links.ts
+++ b/src/common/links.ts
@@ -59,9 +59,9 @@ export function isValidHttpUrl(link: string): boolean {
 }
 
 export const domainOnly = (url: string): string => {
-  const { origin } = new URL(url);
+  const { hostname } = new URL(url);
 
-  return origin.replace(/(^\w+:|^)\/\//, '');
+  return hostname;
 };
 
 type GetInviteLinkProps = {
--- a/src/common/links.ts
+++ b/src/common/links.ts
@@ -58,6 +58,12 @@ export function isValidHttpUrl(link: string): boolean {
   }
 }
 
+export const domainOnly = (url: string): string => {
+  const { origin } = new URL(url);
+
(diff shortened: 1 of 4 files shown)


src/common/links.ts
```diff
@@ -58,6 +58,12 @@ export function isValidHttpUrl(link: string): boolean {
   }
 }
 
+export const domainOnly = (url: string): string => {
+  const { origin } = new URL(url);
+
+  return origin.replace(/(^\w+:|^)\/\//, '');
+};
```
omBratteng, line 65:
```suggestion
export const domainOnly = (url: string): string => {
  const { host } = new URL(url);

  return host;
};
```

Should work, no?

src/common/links.ts
```diff
@@ -58,6 +58,12 @@ export function isValidHttpUrl(link: string): boolean {
   }
 }
 
+export const domainOnly = (url: string): string => {
+  const { origin } = new URL(url);
+
+  return origin.replace(/(^\w+:|^)\/\//, '');
```
capJavert, line 64: `url.hostname`?

src/common/links.ts
omBratteng: Nvm...

omBratteng: Wait, yeah, it'll work. I read the regex wrong, thought it removed `www.`

capJavert: `hostname` because `host` includes a port

capJavert (approved):

src/common/links.ts
omBratteng: Yeah, changed to hostname 🙈

rebelchris: Yeah hostname will do indeed :)

omBratteng (approved):

PR #2099, opened 2024-08-02
fix: we want to render subdomain, but not www

rebelchris:


src/common/links.ts
```diff
@@ -61,7 +61,7 @@ export function isValidHttpUrl(link: string): boolean {
 export const domainOnly = (url: string): string => {
   const { hostname } = new URL(url);
 
-  return hostname;
+  return hostname.replace(/^www\./, '');
```
omBratteng, line 64:
```suggestion
  return hostname.replace(/^www\d?\./, '');
```
Also catch the rare cases if there's a link to legacy www1 etc

omBratteng (approved):

PR #2102, opened 2024-08-05
feat: vordr on profile replies

omBratteng:
Vordr on Profile replies and Discussion feed


capJavert (commented):
Looks good, let's add some tests for vordr on new queries.

omBratteng:
Added tests

capJavert (approved):
Nice, thanks for adding tests for commentFeed

PR #2106, opened 2024-08-06
fix: replace email index

omBratteng:
https://dailydotdev.slack.com/archives/C03C3JZPRU4/p1722941471628759

### DIFF
--- a/src/entity/user/User.ts
+++ b/src/entity/user/User.ts
@@ -67,6 +67,7 @@ export class User {
   name: string;
 
   @Column({ type: 'text', nullable: true })
+  @Index('IDX_user_email')
   email: string;
 
   @Column({ type: 'text', nullable: true })
--- a/src/entity/user/User.ts
+++ b/src/entity/user/User.ts
@@ -58,6 +58,7 @@ export type UserFlags = Partial<{
 @Entity()
 @Index('IDX_user_lowerusername_username', { synchronize: false })
 @Index('IDX_user_lowertwitter', { synchronize: false })
+@Index('IDX_user_loweremail', { synchronize: false })
 export class User {
   @PrimaryColumn({ length: 36 })
   id: string;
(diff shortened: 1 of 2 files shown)


src/migration/1722942338711-UserEmailIndex.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class UserEmailIndex1722942338711 implements MigrationInterface {
+  name = 'UserEmailIndex1722942338711'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_user_loweremail" ON "user" ((lower(email)))`,
```
vpol, line 8: create index concurrently

src/entity/user/User.ts
```diff
@@ -66,7 +67,6 @@ export class User {
   name: string;
 
   @Column({ type: 'text', nullable: true })
-  @Index('IDX_user_email')
```
capJavert, line 69: Don't we want to keep one of the regular email indexes, probably this one?

src/migration/1722942338711-UserEmailIndex.ts
```diff
@@ -0,0 +1,19 @@
...
+    );
+    await queryRunner.query(`DROP INDEX "public"."IDX_user_email"`);
+    await queryRunner.query(`DROP INDEX "public"."user_email_index"`);
```
idoshamun, line 11: Should we add `if exists` otherwise it will not work locally?

omBratteng: added

src/migration/1722942338711-UserEmailIndex.ts
omBratteng: done

src/entity/user/User.ts
omBratteng: Re-added

rebelchris (approved):

idoshamun (approved):

PR #2108, opened 2024-08-06
feat: vordr on posts

omBratteng:
Marking as ready for review, but I'm working on tests for:
- [x] add vordr check to share post
- [x] set vordr flag on submissions
- [x] hide posts that vordr prevented
- [x] don't send notification if vordr prevented post

### DIFF
--- a/__tests__/comments.ts
+++ b/__tests__/comments.ts
@@ -1474,7 +1474,7 @@ describe('mutation editComment', () => {
   });
 
   describe('vordr', () => {
-    it('should set correct vordr flags on good user', async () => {
+    it('should set correct vordr flags on edited comment by good user', async () => {
       loggedUser = '1';
 
       const res = await client.mutate(MUTATION, {
--- a/__tests__/comments.ts
+++ b/__tests__/comments.ts
@@ -978,6 +978,22 @@ describe('mutation commentOnPost', () => {
       expect(comment.flags).toEqual({ vordr: false });
     });
 
+    it('should set correct vordr flags on good user if vordr filter catches it', async () => {
+      loggedUser = '1';
+
+      const res = await client.mutate(MUTATION, {
+        variables: { postId: 'p1', content: 'VordrWillCatchYou' },
+      });
+
+      expect(res.errors).toBeFalsy();
+
+      const comment = await con.getRepository(Comment).findOneByOrFail({
+        id: res.data.commentOnPost.id,
+      });
+
+      expect(comment.flags).toEqual({ vordr: true });
+    });
+
     it('should set correct vordr flags on bad user', async () => {
       loggedUser = 'vordr';
 
@@ -1206,6 +1222,22 @@ describe('mutation commentOnComment', () => {
       });
 
       expect(comment.flags).toEqual({ vordr: false });
+    });
+
+    it('should set correct vordr flags on good user if vordr filter catches it', async () => {
+      loggedUser = '1';
+
+      const res = await client.mutate(MUTATION, {
+        variables: { commentId: 'c1', content: 'VordrWillCatchYou' },
+      });
+
+      expect(res.errors).toBeFalsy();
+
+      const comment = await con.getRepository(Comment).findOneByOrFail({
+        id: res.data.commentOnComment.id,
+      });
+
+      expect(comment.flags).toEqual({ vordr: true });
     });
 
     it('should set correct vordr flags on bad user', async () => {
@@ -1438,6 +1470,40 @@ describe('mutation editComment', () => {
       id: 'c2',
       lastUpdatedAt: expect.any(String),
       content: 'Edit',
+    });
+  });
+
+  describe('vordr', () => {
+    it('should set correct vordr flags on good user', async () => {
+      loggedUser = '1';
+
+      const res = await client.mutate(MUTATION, {
+        variables: { id: 'c2', content: 'comment' },
+      });
+
+      expect(res.errors).toBeFalsy();
+
+      const comment = await con.getRepository(Comment).findOneByOrFail({
+        id: res.data.editComment.id,
+      });
+
+      expect(comment.flags).toEqual({ vordr: false });
+    });
+
+    it('should set correct vordr flags on good user if vordr filter catches it', async () => {
+      loggedUser = '1';
+
+      const res = await client.mutate(MUTATION, {
+        variables: { id: 'c2', content: 'VordrWillCatchYou' },
+      });
+
+      expect(res.errors).toBeFalsy();
+
+      const comment = await con.getRepository(Comment).findOneByOrFail({
+        id: res.data.editComment.id,
+      });
+
+      expect(comment.flags).toEqual({ vordr: true });
     });
   });
 });
--- a/__tests__/comments.ts
+++ b/__tests__/comments.ts
@@ -978,7 +978,7 @@ describe('mutation commentOnPost', () => {
       expect(comment.flags).toEqual({ vordr: false });
     });
 
-    it('should set correct vordr flags on good user if vordr filter catches it', async () => {
+    it('should set correct vordr flags on comment by good user if vordr filter catches it', async () => {
       loggedUser = '1';
 
       const res = await client.mutate(MUTATION, {
@@ -994,7 +994,7 @@ describe('mutation commentOnPost', () => {
       expect(comment.flags).toEqual({ vordr: true });
     });
 
-    it('should set correct vordr flags on bad user', async () => {
+    it('should set correct vordr flags on comment by bad user', async () => {
       loggedUser = 'vordr';
 
       const res = await client.mutate(MUTATION, {
@@ -1208,7 +1208,7 @@ describe('mutation commentOnComment', () => {
   });
 
   describe('vordr', () => {
-    it('should set correct vordr flags on good user', async () => {
+    it('should set correct vordr flags on comment reply by good user', async () => {
       loggedUser = '1';
 
       const res = await client.mutate(MUTATION, {
@@ -1224,7 +1224,7 @@ describe('mutation commentOnComment', () => {
       expect(comment.flags).toEqual({ vordr: false });
     });
 
-    it('should set correct vordr flags on good user if vordr filter catches it', async () => {
+    it('should set correct vordr flags on comment reply by good user if vordr filter catches it', async () => {
       loggedUser = '1';
 
       const res = await client.mutate(MUTATION, {
@@ -1240,7 +1240,7 @@ describe('mutation commentOnComment', () => {
       expect(comment.flags).toEqual({ vordr: true });
     });
 
-    it('should set correct vordr flags on bad user', async () => {
+    it('should set correct vordr flags on comment reply by bad user', async () => {
       loggedUser = 'vordr';
 
       const res = await client.mutate(MUTATION, {
@@ -1490,7 +1490,7 @@ describe('mutation editComment', () => {
       expect(comment.flags).toEqual({ vordr: false });
     });
 
-    it('should set correct vordr flags on good user if vordr filter catches it', async () => {
+    it('should set correct vordr flags on edited comment by good user if vordr filter catches it', async () => {
       loggedUser = '1';
 
       const res = await client.mutate(MUTATION, {
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -1144,26 +1144,54 @@ describe('query sourceFeed', () => {
   });
 
   describe('vordr', () => {
+    beforeEach(async () => {
+      await saveFixtures(con, ArticlePost, [
+        {
+          id: 'vordr1',
+          shortId: 'svordr1',
+          title: 'vordr1',
+          url: 'http://vordr1.com',
+          image: 'https://daily.dev/image.jpg',
+          score: 10,
+          sourceId: 'b',
+          createdAt: new Date(new Date().getTime() - 4000),
+          tagsStr: 'html,javascript',
+          type: PostType.Article,
+          contentCuration: ['c1', 'c2'],
+          authorId: '2',
+          flags: { vordr: true },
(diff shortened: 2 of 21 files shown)


src/common/feedGenerator.ts
```diff
@@ -536,9 +536,15 @@ export const sourceFeedBuilder = (
...
-  }
+  builder.andWhere(
+    new Brackets((qb) => {
+      return qb
+        .where(`${alias}.authorId = :userId OR ${alias}.scoutId = :userId`, {
+          userId: ctx.userId,
+        })
+        .orWhere(`${alias}.banned = false AND ${alias}.showOnFeed = true`);
+    }),
+  );
```
omBratteng, line 547: @idoshamun just need a sanity check. Would this be enough? It should filter out banned posts unless you're author or scout.

omBratteng: Same from @capJavert if you could do a sanity check 🙈

omBratteng: Could maybe add this to `Post` on graphorm too.

omBratteng:
```ts
qb
  .andWhere(`"${alias}"."deleted" = false`)
  .andWhere(`"${alias}"."visible" = true`)
  .andWhere(
    new Brackets((bqb) => {
      return bqb
        .where(
          `${alias}.authorId = :userId OR ${alias}.scoutId = :userId`,
          {
            userId: ctx.userId,
          },
        )
        .orWhere(
          `${alias}.banned = false AND ${alias}.showOnFeed = true`,
        );
    }),
  ),
```

Though not 100% if it's side-effect free

omBratteng: I think it should work, but it does remove posts on source feed (i.e. `bit`) when the post has `banned` set.

capJavert: Query looks good, If we want to apply it globally adding it in graphorm makes sense to me. As to `banned` thing, let's see what @idoshamun says since I am not sure if it was due to any decision.

idoshamun: Let's not add it to the post because then the post page will not be available and this is not what we want

idoshamun (commented):
Looks good, we need to remember the notification as well

omBratteng:
> Looks good, we need to remember the notification as well

Yeah, working on that part

src/common/feedGenerator.ts
omBratteng: Ah yeah, so I'll adding it to `sourceFeedBuilder` is probably best location.

omBratteng: I went with adding `vordr` flag, as it would block Welcome posts, and I didn't really feel like adding a complex query just to catch edge cases.
It still adds the other flags as before, just for filtering purposes, I added the vordr flag.

omBratteng: That way it shouldn't change anything how it works today

__tests__/feeds.ts
```diff
@@ -1142,6 +1142,30 @@ describe('query sourceFeed', () => {
       'FORBIDDEN',
     );
   });
+
+  describe('vordr', () => {
+    it('should filter out posts that vordr has prevented', async () => {
+      loggedUser = '1';
+      await con
+        .getRepository(ArticlePost)
+        .update({ id: 'p5' }, { authorId: '2', flags: { vordr: true } });
```
capJavert, line 1151: I think this potentially will introduce flaky tests as with commentFeed where due to parallel updates from tests sometimes vordr will get triggered in other tests. So maybe solution is to use different ids/fixtures for vordr test. Did not manage to look at that failing one yet.

__tests__/workers/notifications/postAdded.ts
```diff
@@ -0,0 +1,95 @@
...
+import { Comment, Post, Source, User } from '../../../src/entity';
+import worker from '../../../src/workers/notifications/postAdded';
+import { badUsersFixture, sourcesFixture, usersFixture } from '../../fixture';
+import { postsFixture } from '../../fixture/post';
+import { workers } from '../../../src/workers/notifications';
+import { invokeNotificationWorker, saveFixtures } from '../../helpers';
+
+let con: DataSource;
+
+// jest.mock('../../src/common', () => ({
```
capJavert, line 12: cleanup 🔥

src/common/feedGenerator.ts
```diff
@@ -538,6 +539,16 @@ export const sourceFeedBuilder = (
 
   if (sourceId === 'community') {
```
capJavert, line 540: So for community we ignore vordr?

src/entity/posts/utils.ts
```diff
@@ -340,6 +341,20 @@ export const createSharePost = async (
...
     });
 
+    const vordrStatus = await checkWithVordr(
+      { post: createdPost },
+      { con, userId, req: ctx.req },
+    );
+
+    if (vordrStatus === true) {
+      createdPost.banned = true;
+      createdPost.showOnFeed = false;
```
capJavert, line 351: I think we also have flags.showOnFeed and flags.banned?

src/common/vordr.ts
```diff
@@ -18,29 +18,71 @@ export const validateVordrIPs = (ip: string): boolean =>
...
+    userId,
+    con,
+    req,
+  }: {
+    userId: string;
+    con: DataSource | EntityManager;
+    req: Pick<FastifyRequest, 'ip'>;
+  },
 ): Promise<boolean> => {
+  const { type, id } = getTypeAndId({ comment, post, submission });
```
capJavert, line 63: can we maybe use instanceof instead of mapping types to strings by hand?

src/workers/notifications/postAdded.ts
```diff
@@ -42,6 +42,10 @@ const worker: NotificationWorker = {
       return;
     }
 
+    if (post.flags?.vordr) {
```
capJavert, line 45: Missing test for this?

capJavert (commented):
Vordr getting more power 💪 Some general comments, also missing few tests here and there.

__tests__/comments.ts
```diff
@@ -978,6 +978,22 @@ describe('mutation commentOnPost', () => {
       expect(comment.flags).toEqual({ vordr: false });
     });
 
+    it('should set correct vordr flags on good user if vordr filter catches it', async () => {
```
rebelchris, line 981: Maybe name them explicit for post/comment jest doesn't like same names so much and I actually thought for a second it was duplicated 😅

__tests__/posts.ts
```diff
@@ -105,6 +106,8 @@ beforeEach(async () => {
   await saveFixtures(con, ArticlePost, postsFixture);
   await saveFixtures(con, YouTubePost, videoPostsFixture);
   await saveFixtures(con, PostTag, postTagsFixture);
+  // await saveFixtures(con, User, usersFixture);
```
rebelchris, line 109: This needs to be gone/re-enabled?

src/common/post.ts
```diff
@@ -184,6 +187,20 @@ export const createFreeformPost = async (
...
   });
+
+  const vordrStatus = await checkWithVordr(
+    { post: createdPost },
+    { con, userId: ctx.userId, req: ctx.req },
+  );
+
+  if (vordrStatus === true) {
+    createdPost.banned = true;
+    createdPost.showOnFeed = false;
```
rebelchris, line 198: I think @capJavert was right we also need to set it as flags.
Actually Ante, not sure now which one we use, maybe we should only use either raw or the flags?

rebelchris (changes requested):
Minor comments, good work here!

src/workers/notifications/postAdded.ts
omBratteng: It's here
https://github.com/dailydotdev/daily-api/blob/AS-463-vordr-posts/__tests__/workers/notifications/postAdded.ts

__tests__/workers/notifications/postAdded.ts
omBratteng: removed

src/entity/posts/utils.ts
omBratteng: Also set flags now

src/workers/notifications/postAdded.ts
capJavert: Nice, missed it in diff 🙏

src/common/post.ts
omBratteng: I set both column of flags now

__tests__/feeds.ts
omBratteng: I added specific post for those tests

src/common/vordr.ts
omBratteng: How do you envision it to be?

__tests__/comments.ts
omBratteng: Renamed tests

src/common/post.ts
```diff
@@ -184,6 +187,23 @@ export const createFreeformPost = async (
...
+    { post: createdPost },
+    { con, userId: ctx.userId, req: ctx.req },
+  );
+
+  if (vordrStatus === true) {
+    createdPost.banned = true;
+    createdPost.flags.banned = true;
+
+    createdPost.showOnFeed = false;
+    createdPost.flags.showOnFeed = false;
```
rebelchris, line 201: You might want to use `updateFlagsStatement` here for safety :)

src/entity/posts/utils.ts
```diff
@@ -340,6 +341,23 @@ export const createSharePost = async (
...
     });
 
+    const vordrStatus = await checkWithVordr(
+      { post: createdPost },
+      { con, userId, req: ctx.req },
+    );
+
+    if (vordrStatus === true) {
+      createdPost.banned = true;
+      createdPost.flags.banned = true;
```
rebelchris, line 351: Same here `updateFlagsStatement`

rebelchris (approved):
Minor suggestion, but otherwise no blocking issues from my side.
(Just wait for 2 approvals please)

src/common/post.ts
omBratteng: I get this TS error then

> Type '() => string' has no properties in common with type 'Partial<{ sentAnalyticsReport: boolean; banned: boolean; deleted: boolean; private: boolean; visible: boolean; showOnFeed: boolean; promoteToPublic: number; deletedBy: string; vordr: boolean; }>'.ts(2559)

I think it should be fine to not use it, as it hasn't been saved in the database yet, so it's still a js object.

omBratteng: Can do this instead
```ts
createdPost.flags = {
  ...createdPost.flags,
  banned: true,
  showOnFeed: false,
};
```

rebelchris: CC: @capJavert up to you :)

capJavert: @omBratteng for `updateFlagsStatement` you have to use insert or update methods since save does not support sql statements, since we are just creating a post we can probably use `insert` here.

capJavert: But I think we don't have to use `updateFlagsStatement` here since as @omBratteng said it has not been saved to DB yet.

src/common/vordr.ts
capJavert: Instead of `type === 'comment'` do `comment instanceof Comment`.

capJavert: Also what happens if you pass post and coment to `getTypeAndId` you would always get `type: comment`, so it feels a bit error prone

capJavert: I think better would be to change `checkWithVordr` to accept first argument as entity with generic and then use instanceof inside to decide on type.

src/common/feedGenerator.ts
omBratteng: Probably not 🙈 
Forgot this part I think. Will update query to work for community as well

src/common/post.ts
omBratteng: @capJavert went with this as it's before insertion to DB.
```ts
createdPost.flags = {
  ...createdPost.flags,
  banned: true,
  showOnFeed: false,
};
```

src/common/vordr.ts
omBratteng: Refactored to accept as I mentioned above.

src/common/feedGenerator.ts
omBratteng: Fixed so that it works on community as well. Since we already hide banned posts from community picks, I moved the andWhere statement.

capJavert (approved):

src/common/feedGenerator.ts
```diff
@@ -536,9 +537,19 @@ export const sourceFeedBuilder = (
...
+    new Brackets((qb) => {
+      qb.where(`${alias}.authorId = :userId OR ${alias}.scoutId = :userId`, {
+        userId: ctx.userId,
+      }).orWhere(whereVordrFilter(alias));
+
+      if (sourceId === 'community') {
+        qb.andWhere(`${alias}.banned = false`);
+      }
+
+      return qb;
```
capJavert, line 550: Does this not need to be outside of bracket since the bracket itself is `OR`?

omBratteng: It works when it's inside the bracket

omBratteng: I can test it outside to see if it also works

capJavert: that last bracket then is `"authorId" = :userId OR "scoutId" = :userId OR (VORD) AND "banned" = false`

and it should be `("authorId" = :userId OR "scoutId" = :userId OR (VORD)) AND "banned" = false` no?

omBratteng: No, when it's outside the brackets, I don't see the banned posts that I am the author of that is banned by vordr

capJavert: synced offline

PR #2111, opened 2024-08-07
feat(strict-ts): apply to source schemas

capJavert:

This one focuses on: 
- source schema
- sourceRequest schema

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -1,8 +1,7 @@
 import { ForbiddenError, ValidationError } from 'apollo-server-errors';
 import { IResolvers } from '@graphql-tools/utils';
 import { ConnectionArguments } from 'graphql-relay';
-import { traceResolverObject } from './trace';
-import { Context } from '../Context';
+import { AuthContext, BaseContext, Context } from '../Context';
 import {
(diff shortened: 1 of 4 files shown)


src/schema/sources.ts
```diff
@@ -1404,22 +1450,30 @@ export const resolvers: IResolvers<any, Context> = {
...
       { userId, ...args }: { userId: string } & ConnectionArguments,
-      ctx,
+      ctx: Context,
       info,
     ): Promise<Connection<GQLSourceMember>> => {
       return paginateSourceMembers(
         (queryBuilder, alias) => {
+          queryBuilder = queryBuilder.andWhere(`${alias}."userId" = :user`, {
+            user: userId,
+          });
```
omBratteng, line 1460: Not really related to types
```suggestion
          queryBuilder = queryBuilder.andWhere(`${alias}."userId" = :userId`, {
            userId: userId,
          });
```

omBratteng (approved):

src/schema/sources.ts
capJavert: yeah, ill update

PR #2112, opened 2024-08-07
feat(slack): adjust message format after testing

capJavert:
- now supports private posts as well (since we no longer depend on slack link embeds)
- also refactored WebClient to single place so we can set appropriate log level
- add utm params to links


capJavert:
Merging so I can open future PRs. Message copy verified with product.

PR #2113, opened 2024-08-07
feat(slack): allow join from private post link

capJavert:
- forward squad admin token as `?jt=token&type=squad&source=handle` which frontend can parse and then redirect to proper squad invite page
- also forwards all utm_ params
- commit for frontend implementation of join and redirect https://github.com/dailydotdev/apps/pull/3315/commits/1a616aee258922c2f54163e68bd34eebe1ed3d59
  - reuses referral logic (attributed to squad admin)
  - after squad join redirects back to post from which user joined
  - also preserves all utm_ params


src/workers/postAddedSlackChannelSend.ts
```diff
@@ -38,6 +39,24 @@ export const postAddedSlackChannelSendWorker: TypedWorker<'api.v1.post-visible'>
         ]);
         const source = await post.source;
 
+        let adminJoinToken: string | undefined;
```
sshanzel, line 42: Is it not automatically inferred to be `| undefined` if we don't define a value for it?

src/workers/postAddedSlackChannelSend.ts
```diff
@@ -38,6 +39,24 @@ export const postAddedSlackChannelSendWorker: TypedWorker<'api.v1.post-visible'>
...
+
+        if (source.type === SourceType.Squad) {
+          const admin = await con.getRepository(SourceMember).findOne({
+            where: {
```
sshanzel, line 46: Do we have to add the select statement?

sshanzel (commented):
Looks amazing. I think we can add a test to verify that the token is sent along with the other parameters.

src/workers/postAddedSlackChannelSend.ts
capJavert: nope, it just says string

src/workers/postAddedSlackChannelSend.ts
capJavert: for sure!

capJavert: Added in 30b47a9a41b25ba688cb2a5c58ac8d5914e674cf together with test

__tests__/workers/postAddedSlackChannelSend.ts
```diff
@@ -118,6 +161,49 @@ describe('postAddedSlackChannelSend worker', () => {
...
+    });
+    expect(chatPostMessage).toHaveBeenCalledWith({
+      channel: '1',
+      attachments: [
+        {
+          author_icon: 'https://app.daily.dev/apple-touch-icon.png',
+          author_name: 'daily.dev',
+          image_url: 'https://daily.dev/image.jpg',
+          title: 'Squad Channel Post 1',
+          title_link:
```
rebelchris, line 198: Do you maybe also want to test public squads for the diff or we already do so?

src/workers/postAddedSlackChannelSend.ts
```diff
@@ -38,6 +39,27 @@ export const postAddedSlackChannelSendWorker: TypedWorker<'api.v1.post-visible'>
         ]);
         const source = await post.source;
 
+        let adminJoinToken: string | undefined;
+
+        if (source.type === SourceType.Squad) {
```
rebelchris, line 44: Isn't it only needed for private squads?

rebelchris (approved):
More of a question

src/workers/postAddedSlackChannelSend.ts
capJavert: Yeah good catch, for now actually better to specify the private check as well. I'll do that.

__tests__/workers/postAddedSlackChannelSend.ts
capJavert: When I add the the private check in the worker the response handling will be the same as for source so no need I think. https://github.com/dailydotdev/daily-api/pull/2113#discussion_r1709237010

sshanzel (approved):
Nice. Looks amazing 🚀

src/workers/postAddedSlackChannelSend.ts
capJavert: Solved in bd11f05533a7bd445c602bd8c69e12344ea8d816

__tests__/workers/postAddedSlackChannelSend.ts
capJavert: Solved in bd11f05533a7bd445c602bd8c69e12344ea8d816

capJavert:
@idoshamun if I can get your sanity check as well!

src/workers/postAddedSlackChannelSend.ts
```diff
@@ -63,8 +85,16 @@ export const postAddedSlackChannelSendWorker: TypedWorker<'api.v1.post-visible'>
...
+
+              if (adminJoinToken) {
+                postLinkUrl.searchParams.set('jt', adminJoinToken);
+                postLinkUrl.searchParams.set('source', source.handle);
+                postLinkUrl.searchParams.set('type', source.type);
+              }
+
               const postLink = addNotificationUtm(
-                postLinkPlain,
+                postLinkUrl.toString(),
```
idoshamun, line 97: Let's shorten the link to obscure this info and make it clean?

capJavert: Was thinking about it, it's already obscured in the message markdown itself?

image

if we don't have to add another request to shortening service maybe better, wdyt?

PR #2117, opened 2024-08-08
fix: set vordr flags on comment reply

omBratteng:


omBratteng:
> Should we add tests?

Probably... I see I forgot tests for `commentOnPost` mutation too

idoshamun (approved):

PR #2119, opened 2024-08-08
feat: source category entity

sshanzel:
- Introduce source category entity.
- Add `categoryId` column on the source entity.
- update sources query to filter by category

MI-483


src/workers/postUpdated.ts
```diff
@@ -275,7 +275,7 @@ const createPost = async ({
   data.score = Math.floor(postCreatedAt.getTime() / (1000 * 60));
   data.origin = data?.scoutId
     ? PostOrigin.CommunityPicks
-    : (data.origin ?? PostOrigin.Crawler);
```
sshanzel, line 278: Lint was screaming about this.

src/entity/Source.ts
```diff
@@ -109,6 +111,12 @@ export class Source {
   @Index('IDX_source_flags_featured', { synchronize: false })
   flags: SourceFlagsPublic;
 
+  @Column({ type: 'text', nullable: true, default: null })
```
capJavert, line 114: You don't need default null if nullable is true

src/migration/1723118739478-SourceCategory.ts
```diff
@@ -0,0 +1,39 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceCategory1723118739478 implements MigrationInterface {
+  name = 'SourceCategory1723118739478';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "source_category" ("id" text NOT NULL, "value" text NOT NULL, "enabled" boolean NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "updatedAt" TIMESTAMP NOT NULL DEFAULT now(), CONSTRAINT "UQ_f130e0bac7dea92c2a4084c2f89" UNIQUE ("value"), CONSTRAINT "PK_21e4d5359f2a23fd10053f516e9" PRIMARY KEY ("id"))`,
+    );
+    await queryRunner.query(`ALTER TABLE "source" ADD "categoryId" text`);
```
capJavert, line 10: It cant be set by user? Since we dont validate just to sanity check if we need to at least do some length validation.

src/migration/1723118739478-SourceCategory.ts
```diff
@@ -0,0 +1,39 @@
...
+      ('web', 'Web', true),
+      ('mobile', 'Mobile', true),
+      ('games', 'Games', true),
+      ('devops', 'DevOps', true),
+      ('cloud', 'Cloud', true),
+      ('career', 'Career', true),
+      ('data', 'Data', true),
+      ('fun', 'Fun', true),
+      ('devtools', 'DevTools', true)
+    `);
```
capJavert, line 28: on conflict do nothing?

src/schema/sources.ts
```diff
@@ -1136,6 +1169,12 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   BaseContext
 >({
   Query: {
+    sourceCategories: async (
+      _,
+      __,
+      ctx: Context,
+    ): Promise<GQLSourceCategory[]> =>
+      ctx.con.getRepository(SourceCategory).find(),
```
capJavert, line 1177: We should use graphorm to query so any field resolvers also work. Also can we make the query support pagination as other listings?

src/entity/sources/SourceCategory.ts
```diff
@@ -0,0 +1,19 @@
+import { Column, Entity, PrimaryColumn, UpdateDateColumn } from 'typeorm';
+
+@Entity()
+export class SourceCategory {
+  @PrimaryColumn({ type: 'text' })
```
capJavert, line 5: uuid maybe better for pk, we us it on more places recently?

src/entity/sources/SourceCategory.ts
```diff
@@ -0,0 +1,19 @@
...
+  id: string;
+
+  @Column({ type: 'text', unique: true })
+  value: string;
+
+  @Column()
+  enabled: boolean;
+
+  @Column({ default: () => 'now()' })
```
capJavert, line 14: @CreatedAtColumn (or whatever its called in typeorm)

capJavert (commented):
Looking forward to categories on sources! Some comments to standardize, nothing breaking really.

__tests__/sources.ts
```diff
@@ -105,16 +160,45 @@ beforeEach(async () => {
...
+          id
+          value
+          enabled
+          createdAt
+        }
+      }
+    `);
+    expect(res.errors).toBeFalsy();
+    const categories = getSourceCategories();
+    expect(res.data.sourceCategories).toMatchSnapshot(
```
rebelchris, line 177: Can we maybe avoid snapshot and use something like this? https://jestjs.io/docs/expect#expectarraycontainingarray
Or simply toEqual?

__tests__/sources.ts
```diff
@@ -132,17 +216,34 @@ describe('query sources', () => {
...
     const isPublic = res.data.sources.edges.every(({ node }) => !!node.public);
     expect(isPublic).toBeTruthy();
   });
 
+  it('should filter by category', async () => {
+    console.log('entry');
+    const repo = con.getRepository(Source);
+    await repo.update({ id: 'a' }, { categoryId: 'general' });
+    await repo.update({ id: 'b' }, { categoryId: 'web' });
+    const res = await client.query(QUERY({ first: 10, categoryId: 'web' }));
```
rebelchris, line 240: Do we allow filtering on more categories at once?
(If so we'll need a test, if only 1 it's fine as is)

src/entity/sources/SourceCategory.ts
```diff
@@ -0,0 +1,19 @@
+import { Column, Entity, PrimaryColumn, UpdateDateColumn } from 'typeorm';
+
+@Entity()
+export class SourceCategory {
```
rebelchris, line 4: Actually missed it in RFC, but we should include the SEO metadata I think :)
https://dailydotdev.slack.com/archives/C069XNYMDGV/p1722930423016999?thread_ts=1722870716.569749&cid=C069XNYMDGV

rebelchris (changes requested):
Couple queries besides what Ante logged

__tests__/sources.ts
nensidosari: At least so far from product perspective looks like we filter only by one or get all.

src/entity/Source.ts
sshanzel: Yep. Overkill. Was doing trial and error if the migration would by adding one at a time. Lol.

src/migration/1723118739478-SourceCategory.ts
sshanzel: These categories can only be set by us. I am cool to add a max length, what do you think of 64 characters?

src/migration/1723118739478-SourceCategory.ts
sshanzel: Yep. Let's add that.

src/schema/sources.ts
sshanzel: Yeah, we can do that.

src/entity/sources/SourceCategory.ts
sshanzel: Much better.

__tests__/sources.ts
sshanzel: Yeah, for now, we only do 1.

src/entity/sources/SourceCategory.ts
sshanzel: > Category Page (Template)

The SEO per Category does not change though. It is a template where we just change the value of one word.

__tests__/sources.ts
sshanzel: Sure, let's do that.

src/entity/sources/SourceCategory.ts
rebelchris: Guess we can leave it for now, pretty sure sooner then later product will want full descriptions (like we do on tag/source overviews now)

src/graphorm/index.ts
```diff
@@ -266,6 +266,9 @@ const obj = new GraphORM({
       },
     },
   },
+  SourceCategory: {
+    requiredColumns: ['createdAt'],
```
rebelchris, line 270: Why is this one required only?

src/migration/1723579275223-SourceCategory.ts
```diff
@@ -0,0 +1,40 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "source_category" ("id" uuid NOT NULL DEFAULT uuid_generate_v4(), "title" text NOT NULL, "enabled" boolean NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "updatedAt" TIMESTAMP NOT NULL DEFAULT now(), CONSTRAINT "UQ_66d6dccf282b8104ef9c44c0fb5" UNIQUE ("title"), CONSTRAINT "PK_21e4d5359f2a23fd10053f516e9" PRIMARY KEY ("id"))`,
+    );
+    await queryRunner.query(`ALTER TABLE "source" ADD "categoryId" uuid`);
+    await queryRunner.query(
+      `ALTER TABLE "source" ADD CONSTRAINT "FK_02e1cbb6e33fa90e68dd56de2a9" FOREIGN KEY ("categoryId") REFERENCES "source_category"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`,
+    );
+    await queryRunner.query(`
+      INSERT INTO "source_category"
```
rebelchris, line 15: Can we maybe also add a seed for local just now we working on it?

rebelchris (approved):
Non blocking remarks left :)

src/migration/1723118739478-SourceCategory.ts
capJavert: If its only by use then its fine 👍

capJavert (approved):

src/migration/1723579275223-SourceCategory.ts
sshanzel: Is this not the same thing? Locally, you will have to run migrations, and this will fill up your DB.

src/graphorm/index.ts
sshanzel: The `createdAt` is needed for the pagination, technically you can have everything missing and not break the query.

src/migration/1723579275223-SourceCategory.ts
sshanzel: I can turn this into a seed, if that is more ideal. Wdyt?

rebelchris: Oh sorry I kind of added the one before realizing this.
This will do :)

src/entity/sources/SourceCategory.ts
sshanzel: Yeah, totally agree with that.

PR #2120, opened 2024-08-08
feat: vordr on comment edit and add tests

omBratteng:
Broke it out of #2108 just to keep me sane with all the different tests 😬


omBratteng:
A simple cherry pick didn't work 🙈

PR #2121, opened 2024-08-08
fix: flaky daily digest test

omBratteng:
I think this test is flaky around midnight


__tests__/cron/dailyDigest.ts
```diff
@@ -348,7 +348,7 @@ describe('dailyDigest cron', () => {
       });
 
       it('should not schedule send time on Friday', async () => {
-        jest.setSystemTime(setDay(new Date(), DayOfWeek.Friday));
+        jest.setSystemTime(setHours(setDay(new Date(), DayOfWeek.Friday), 12));
```
capJavert, line 351: If it fails around midnight do we have a bug there then, or you think its just test setup stuff?

rebelchris (approved):
Probably fine to really hardcode jest fake time anyway.

__tests__/cron/dailyDigest.ts
omBratteng: I think it's just the tests, will double check a bit. Always hard when they're dependent on real time :ttiggered:

capJavert: yep, gets complicated

omBratteng: @capJavert think it's just the tests, haven't actually encountered the issue in production

capJavert (approved):

PR #2123, opened 2024-08-09
fix: add allowed_content_curation flag to feed config

rebelchris:
Add `allowed_content_curations` config if user has any blocked items.
Also noticed we didn't even have a test for blocked_content_curations so added one that covers both cases now.

AS-505 #done

### DIFF
--- a/src/integrations/feed/configs.ts
+++ b/src/integrations/feed/configs.ts
@@ -23,6 +23,16 @@ type Options = {
 };
 
 type BaseConfig = Partial<Omit<FeedConfig, 'user_id' | 'page_size' | 'offset'>>;
+const AllowedContentCurationTypes = [
+  'news',
+  'release',
+  'opinion',
+  'listicle',
+  'comparison',
+  'tutorial',
+  'story',
+  'meme',
+] as const;
 
 function getDefaultConfig(
   baseConfig: BaseConfig,
@@ -88,6 +98,9 @@ const addFiltersToConfig = ({
     opts.includeBlockedContentCuration
   ) {
     baseConfig.blocked_content_curations = filters.blockedContentCuration;
+    baseConfig.allowed_content_curations = AllowedContentCurationTypes.filter(
+      (type) => !filters.blockedContentCuration.includes(type),
+    );
   }
 
   return baseConfig;
--- a/__tests__/integrations/feed.ts
+++ b/__tests__/integrations/feed.ts
@@ -179,6 +179,13 @@ describe('FeedPreferencesConfigGenerator', () => {
         defaultEnabledState: true,
         options: { type: PostType.Article },
       },
+      {
+        title: 'News',
+        group: 'content_curation',
+        description: '',
+        defaultEnabledState: true,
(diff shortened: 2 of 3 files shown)


omBratteng (approved):

src/integrations/feed/configs.ts
```diff
@@ -88,6 +98,9 @@ const addFiltersToConfig = ({
     opts.includeBlockedContentCuration
```
capJavert, line 98: Should we add `includeAllowedContentCuration` or rename to `includeContentCuration` if we want to keep them together.

capJavert (approved):

src/integrations/feed/configs.ts
rebelchris: includeContentCuration should just be the default AFAIK when there is blocked ones.
It's more a retro fix at this stage when you do have blocked items we also pass which types are supposed to be included.

Overall don't see a need to have it optional now, but maybe miss some potential context here.

capJavert: Options like `opts. includeBlockedContentCuration` can be passed from feed generators to avoid sending some filtering options to feed request. 

I think we should rename to `includeContentCuration` then so we keep them together.

rebelchris: My main concern is that it's already passed in multiple places now (*and maybe even LOFN not sure)
This way we keep all logic and at very end send this extra part.

I do get your comment, but feels more dangerous to do now.
I'll check as well with @idoshamun if maybe I missed anything.

rebelchris: @capJavert Decided to rename anyway :)

capJavert (approved):

src/integrations/feed/configs.ts
```diff
@@ -83,11 +93,11 @@ const addFiltersToConfig = ({
       baseConfig.allowed_post_types || postTypes
     ).filter((x) => !filters.excludeTypes.includes(x));
   }
-  if (
-    filters.blockedContentCuration?.length &&
-    opts.includeBlockedContentCuration
-  ) {
+  if (filters.blockedContentCuration?.length && opts.includeContentCuration) {
     baseConfig.blocked_content_curations = filters.blockedContentCuration;
```
idoshamun, line 97: I think we don't need this anymore now

rebelchris: What about the lofn provider? (And maybe the one that uses the digest?)

@vpol you probably know best, can we remove it?

vpol: lofn does not use any of the filters at all.

rebelchris: And with your fix nothing else uses the old format right so removing it is fine? 🙏

vpol: we still support it on the feed side, but if you are not going to use it - it's ok.

rebelchris: Removed :)

idoshamun (approved):

PR #2124, opened 2024-08-09
feat: add option to search user suggestions

rebelchris:
Option to search for user suggestions (based on name and username)

## Todo
- [x] Exclude banned users (awaiting Ole-Martin advice)

### DIFF
--- a/__tests__/search.ts
+++ b/__tests__/search.ts
@@ -17,6 +17,7 @@ import { ArticlePost, Keyword, Source, User, UserPost } from '../src/entity';
 import { postsFixture } from './fixture/post';
 import { sourcesFixture } from './fixture/source';
 import { usersFixture } from './fixture/user';
+import { ghostUser } from '../src/common';
 
 let con: DataSource;
 let state: GraphQLTestingState;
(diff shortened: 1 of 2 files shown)


capJavert (approved):

rebelchris:
@capJavert // @omBratteng added vodr exclude now.
Can you please check again?

capJavert (approved):

src/schema/search.ts
```diff
@@ -388,6 +410,37 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      source,
+      { query, limit }: SearchSuggestionArgs,
+      ctx,
+    ): Promise<GQLSearchSuggestionsResults> => {
+      const searchQuery = ctx.con
+        .createQueryBuilder()
+        .select(`id, name as title, username as subtitle, image`)
+        .from('user', 'u')
+        .where('u.infoConfirmed = TRUE')
+        .andWhere("u.id != '404'")
```
omBratteng, line 428: Instead of hardcoding `404`, lets use the `ghostUser.id` variable?

src/schema/search.ts
```diff
@@ -388,6 +410,37 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+            return qb
+              .where(`name ILIKE :query`, {
+                query: `%${query}%`,
+              })
+              .orWhere(`username ILIKE :query`, {
+                query: `%${query}%`,
+              });
+          }),
+        )
+        .andWhere(whereVordrFilter('u'))
```
omBratteng, line 440: Yup, looks good. Should return `COALESCE((u.flags ->> 'vordr')::boolean, false) = false`

__tests__/search.ts
```diff
@@ -528,3 +529,121 @@ describe('query searchSourceSuggestions', () => {
...
+
+  it('should only return vodr false users', async () => {
+    await con.getRepository(User).update(
+      { id: '3' },
+      {
+        flags: updateFlagsStatement<User>({
+          vordr: true,
+        }),
+      },
+    );
```
omBratteng, line 615: Could use the user `vordr` from `badUsersFixture`?

omBratteng (approved):
Looks good, minor comments

src/schema/search.ts
rebelchris: Good point, probably should fix it in some other places as well!

__tests__/search.ts
rebelchris: Ah think this is easier as we can then test again the succesfull test above if that's ok?

PR #2127, opened 2024-08-12
feat: track slack integration connection

capJavert:


capJavert:
@idoshamun if you can check!

src/routes/integrations/slack.ts
```diff
@@ -159,6 +167,25 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
       }
 
+      try {
+        await sendAnalyticsEvent([
+          {
+            event_timestamp: new Date(),
+            event_name: AnalyticsEventName.ConfirmAddingWorkspace,
+            app_platform: 'api',
+            user_id: req.userId,
+            target_id: UserIntegrationType.Slack,
```
capJavert, line 177: only thing different from taxonomy is that it does not make sense to have `extra.source` here since at this point we don't know which source will be connected yet

idoshamun (approved):

PR #2129, opened 2024-08-12
test: attempt fix a flaky test

omBratteng:
This test keeps being flaky locally, so this should fix it hopefully.
Only did this specific one as it was the most flaky and annoyed me

### DIFF
--- a/__tests__/feeds.ts
+++ b/__tests__/feeds.ts
@@ -2328,7 +2328,18 @@ describe('function feedToFilters', () => {
   it('should return filters for tags/sources based on the values from our data', async () => {
     loggedUser = '1';
     await saveFeedFixtures();
-    expect(await feedToFilters(con, '1', '1')).toMatchSnapshot();
+    const filters = await feedToFilters(con, '1', '1');
+    expect(filters.includeTags).toEqual(
+      expect.arrayContaining(['html', 'javascript']),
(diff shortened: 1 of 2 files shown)


__tests__/feeds.ts
```diff
@@ -2276,7 +2276,21 @@ describe('function feedToFilters', () => {
...
     loggedUser = '1';
     await saveFeedFixtures();
-    expect(await feedToFilters(con, '1', '1')).toMatchSnapshot();
+    const filters = await feedToFilters(con, '1', '1');
+    expect(filters.blockedContentCuration).toEqual([]);
+    expect(filters.excludeTypes).toEqual([]);
+    expect(filters.sourceIds).toEqual([]);
+    expect(filters.includeTags).toEqual(
+      expect.arrayContaining(['html', 'javascript']),
+    );
```
capJavert, line 2285: In addition to `arrayContaining`  maybe also check length so we are sure no additional elements are added and test does not catch it?

capJavert: also below

capJavert (approved):
Approving aside from above comment 👍

__tests__/feeds.ts
omBratteng: 👍

PR #2130, opened 2024-08-13
feat: add new entity for ReadingStreakActions

ilasw:
Added new entity for `ReadingStreakActions`

### DIFF
--- a/src/entity/ReadingStreakActions.ts
+++ b/src/entity/ReadingStreakActions.ts
@@ -1,0 +1,24 @@
+import { Column, Entity, OneToOne, PrimaryColumn } from 'typeorm';
+import { UserStreak } from './user';
+
+export enum ReadingStreakActionType {
+  Recover = 'recovered',
+}
+
+@Entity()
+export class ReadingStreakActions {
+  @PrimaryColumn({ length: 36, default: '' })
+  id: string;
+
+  @OneToOne(() => UserStreak, (userStreak) => userStreak.userId, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  userStreak: Promise<UserStreak>;
+
+  @Column({ type: 'timestamptz', default: () => 'now()' })
+  timestamp: Date;
+
+  @Column({ type: 'enum', enum: ReadingStreakActionType })
+  type: ReadingStreakActionType;
+}
--- a/src/entity/ReadingStreakActions.ts
+++ b/src/entity/ReadingStreakActions.ts
@@ -19,6 +19,6 @@ export class ReadingStreakActions {
   @Column({ type: 'timestamptz', default: () => 'now()' })
   timestamp: Date;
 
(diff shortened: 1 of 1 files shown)


src/entity/ReadingStreakActions.ts
```diff
@@ -0,0 +1,24 @@
...
+  @OneToOne(() => UserStreak, (userStreak) => userStreak.userId, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  userStreak: Promise<UserStreak>;
+
+  @Column({ type: 'timestamptz', default: () => 'now()' })
+  timestamp: Date;
+
+  @Column({ type: 'enum', enum: ReadingStreakActionType })
```
rebelchris, line 22: I think we prefer storing as strings but forcing enum on graph side (this makes it a bit more flexible to add/change stuff later on)

src/entity/ReadingStreakActions.ts
```diff
@@ -0,0 +1,24 @@
+import { Column, Entity, OneToOne, PrimaryColumn } from 'typeorm';
+import { UserStreak } from './user';
+
+export enum ReadingStreakActionType {
+  Recover = 'recovered',
+}
+
+@Entity()
+export class ReadingStreakActions {
+  @PrimaryColumn({ length: 36, default: '' })
```
rebelchris, line 10: Why is the default empty? you can just omit it on primary.
```suggestion
  @PrimaryColumn({ type: 'text' })
```

rebelchris (changes requested):
You're also missing a migration for this :)

capJavert:
@ilasw per what RFC or task are you adding this?

rebelchris:
@capJavert This section (i also had to look)
https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1076985865/RFC-055+-+Reading+streaks+recovery#Mutation-recoverStreak (table below)

capJavert:
@ilasw if possible we should bundle this with some concrete queries/mutations and tests, we usually don't just add empty entities since then its easy to miss some nuance. If it helps you can make multiple smaller PRs into this and then merge in one go to `main`. Not sure how your tasks are organized.

src/entity/ReadingStreakActions.ts
```diff
@@ -0,0 +1,24 @@
...
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  userStreak: Promise<UserStreak>;
+
+  @Column({ type: 'timestamptz', default: () => 'now()' })
+  timestamp: Date;
+
+  @Column({ type: 'text' })
+  type: string;
```
capJavert, line 23: typescript type here can still be `ReadingStreakActionType` even though in psql its `text`

capJavert: see `UserAction` for examples

PR #2136, opened 2024-08-13
fix: set correct vordr flags on worker too

omBratteng:
Realised I forgot to add the other flags on scout picks


idoshamun (approved):

capJavert (approved):
If you can add tests 🙏

PR #2137, opened 2024-08-13
test: add missed tests in #2136

omBratteng:
As requested @capJavert, test for #2136 

auto-merge enabled 😈


__tests__/workers/postUpdated.ts
```diff
@@ -1198,6 +1262,94 @@ describe('on post update', () => {
...
+        source_id: COMMUNITY_PICKS_SOURCE,
+        submission_id: uuid,
+      });
+
+      const updatedPost = await con.getRepository(ArticlePost).findOneByOrFail({
+        id: postId,
+      });
+
+      expect(updatedPost.flags.vordr).toEqual(false);
+      expect(updatedPost.banned).toEqual(false);
```
capJavert, line 1347: Can we add expect to make sure if post had some other flags they did not get overwritten.

omBratteng: Added extra test, does that work?

capJavert (approved):

PR #2138, opened 2024-08-13
feat: streaks action entity

ilasw:
- Added new entity for `ReadingStreakActions`;
- Updated check for `checkUserStreak` including `lastRecoverAction`;
- Update `update-current-streak` cron for include `lastRecoverAction`;


sshanzel:
@dailydotdev/web-team should now be ready for review. This covers the flow of resetting the streak in consideration of the streak restoration.

sshanzel:
TODO: once the fix on the other PR is merged, let's see how the tests here would turn out.

__tests__/cron/updateCurrentStreak.ts
```diff
@@ -81,30 +80,156 @@ describe('updateCurrentStreak cron', () => {
...
     await expectSuccessfulCron(cron);
     const streak = await con
       .getRepository(UserStreak)
       .findOneBy({ userId: '1' });
     expect(streak.currentStreak).toBe(1);
   });
+
+  describe('incorporating streaks recovery', () => {
+    beforeEach(() => {
+      nock('http://localhost:5000').post('/e').reply(204);
```
rebelchris, line 93: Why do you need this one here?

__tests__/cron/updateCurrentStreak.ts
```diff
@@ -81,30 +80,156 @@ describe('updateCurrentStreak cron', () => {
...
+    });
+
+    it('should not reset streak if user has recovered date of today', async () => {
+      /*
+      Day today is Wednesday
+      1. Last read was Monday
+      2. User did not read anything on Tuesday
+      3. Wednesday arrived and resets the streak to 0
+      4. User recovers the streak on Wednesday
+      5. When the cron job runs, streak should stay at 1
```
rebelchris, line 103: Can you add a test here where this happens, but user also reads on Wednesday so now streak should be 2?

Maybe we should actually test both these:
- User reads first -> then resets
- User resets first -> then reads

src/common/users.ts
```diff
@@ -497,6 +526,6 @@ export const mastodonSocialUrlMatch =
   /^(?<value>https:\/\/(?:[a-z0-9-]+\.)*[a-z0-9-]+\.[a-z]{2,}\/@[\w-]{2,}\/?)$/;
 
 export const socialUrlMatch =
-  /^(?<value>https:\/\/(?:[a-z0-9-]{1,50}\.){0,5}[a-z0-9-]{1,50}\.[a-z]{2,24}\b([-a-zA-Z0-9@:%_\+.~#?&\/=]*))$/;
+  /^(?<value>https:\/\/(?:[a-z0-9-]{1,50}\.){0,5}[a-z0-9-]{1,50}\.[a-z]{2,24}\b([-a-zA-Z0-9@:%_+.~#?&\/=]*))$/;
```
rebelchris, line 529: Why did this change?

src/cron/updateCurrentStreak.ts
```diff
@@ -16,17 +21,43 @@ const cron: Cron = {
           )
           .addSelect('us.currentStreak', 'current')
           .addSelect('u."weekStart"', 'weekStart')
+          .addSelect(
+            `(date_trunc('day', usa."lastRecoverAt" at time zone COALESCE(u.timezone, 'utc'))::date) - interval '1 day'`,
+            'lastRecoverAt',
+          )
           .from(UserStreak, 'us')
           .innerJoin(User, 'u', 'u.id = us."userId"')
+          .leftJoin(
```
rebelchris, line 30: This whole query is just something we need to closely monitor for performance/optimizations (no direct action needed)

rebelchris (changes requested):
More around test case safety, overall looks good.

__tests__/cron/updateCurrentStreak.ts
sshanzel: The worker fails without this. The existing cron might have been calling this at the end.

src/common/users.ts
sshanzel: Lint says the extra character is not needed and automatically updates this. Probably due to the package upgrade.

__tests__/cron/updateCurrentStreak.ts
sshanzel: Sure, let's add those ones.

sshanzel: > User resets first -> then reads

Added a test now but within the `newView` test suite to make it more realistic in terms of adding a read to update the streak.

> User reads first -> then resets

We should have at least 3 tests on this.

src/cron/updateCurrentStreak.ts
sshanzel: Agreed. This may run in the bg but still, we should keep an eye out if it behaves weirdly after this.

src/entity/user/UserStreakAction.ts
```diff
@@ -0,0 +1,25 @@
+import { Entity, JoinColumn, OneToOne, PrimaryColumn } from 'typeorm';
+import { User } from './User';
+
+export enum UserStreakActionType {
+  Recover = 'recover',
+}
```
capJavert, line 6: Let's move this to common/streak or something, it will help with circular dependencies because entities are imported quite a lot all over the palce.

src/entity/user/UserStreakAction.ts
```diff
@@ -0,0 +1,25 @@
...
+  type: UserStreakActionType;
+
+  @PrimaryColumn({ type: 'timestamptz', default: () => 'now()' })
+  createdAt: Date;
+
+  @OneToOne(() => User, (user) => user.streak, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({ name: 'userId' })
```
capJavert, line 23: You should not need join column right? Typeorm usually infers automatically?

src/common/users.ts
```diff
@@ -441,17 +448,39 @@ export const checkUserStreak = (streak: GQLUserStreakTz): boolean => {
   today.setHours(0, 0, 0, 0);
 
   const day = today.getDay();
-  const difference = differenceInDays(today, lastViewAt);
+  const difference = differenceInDays(today, lastStreakUpdate);
 
   return shouldResetStreak(day, difference, streak.weekStart);
 };
 
+export const getUserLastStreak = async (
```
capJavert, line 456: This is really getLastStreakRecoverDate no?

capJavert (approved):

src/entity/user/UserStreakAction.ts
sshanzel: Agreed. Let's move it.

sshanzel: Just realized, it might actually produce circular dependency if we move it. Since this entity itself needs the enum, then on the same common/users file, we might call the actual entity for fetching.

src/entity/user/UserStreakAction.ts
sshanzel: Just copy-pasted it from UserStreak entity, but we can try.

src/common/users.ts
sshanzel: Yep, that's more accurate.

src/entity/user/UserStreakAction.ts
sshanzel: For some reason, the typeorm did not create the right reference for creating an FK.

src/entity/user/UserStreakAction.ts
capJavert: hmm yeah, let's leave it as is then, you are right

src/entity/user/UserStreakAction.ts
capJavert: I see now, `UserStreak` PK is `userId` and it looks for `id` or `userStreakId`. Good then 👍

rebelchris (approved):

PR #2139, opened 2024-08-13
fix: don't show non visible community picks

rebelchris:
Currently community picks on already added but private/non visible fields would show it exists, on clicking result in 404.
So now we simply show it got added before, and cannot be re-added (as per deleted guidelines)

image

AS-502 #done


omBratteng:
Test?

rebelchris:
@omBratteng Good spot, added now 🙏

capJavert (approved):

PR #2140, opened 2024-08-13
feat(strict-ts): apply to gql schemas 3

capJavert:

This one focuses on: 
- feeds schema
- devcard schema
- notifications schema
- common schema
- some retro fixes for code that was added in the meantime
- remove `any` eslint-disable rules from all schemas (now that context is standardized)

This one concludes work on schemas.

### DIFF
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -72,7 +72,7 @@ export const whereKeyword = (
 
 export const getExcludedAdvancedSettings = async (
   con: DataSource,
-  feedId: string,
+  feedId?: string,
 ): Promise<Partial<AdvancedSettings>[]> => {
   if (!feedId) {
     return [];
@@ -103,8 +103,8 @@ export const getExcludedAdvancedSettings = async (
 
 export const feedToFilters = async (
   con: DataSource,
-  feedId: string,
-  userId: string,
+  feedId?: string,
+  userId?: string,
 ): Promise<AnonymousFeedFilters> => {
   const settings = await getExcludedAdvancedSettings(con, feedId);
   const [tags, excludeSources, memberships] = await Promise.all([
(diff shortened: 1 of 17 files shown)


src/schema/common.ts
```diff
@@ -217,16 +212,16 @@ export const fixedIdsPageGenerator = <
   nodeToCursor: (page, args, node, i, queryParams): string =>
     offsetToCursor(
       page.offset +
-        queryParams.data.findIndex(([postId]) => postId === node.id),
+        queryParams!.data.findIndex(([postId]) => postId === node.id),
```
capJavert, line 215: Using `!` in this file because code so far expected no `undefined` so if anything was to break it should break here as well instead of guarding and then returning wrong data

src/common/feedGenerator.ts
```diff
@@ -103,8 +103,8 @@ export const getExcludedAdvancedSettings = async (
 
 export const feedToFilters = async (
   con: DataSource,
-  feedId: string,
-  userId: string,
+  feedId?: string,
```
rebelchris, line 106: Techically this relies on a feedId right?
Just wondering if:
A. we make it non optional but early return if not existing?
B. we keep it optional, but still early return if not existing?

rebelchris (approved):
Nice work!
Just one remark, also pretty sure you tried something so probably no action needed.

src/common/feedGenerator.ts
capJavert: It technically was optional all the time, you can see checks inside the function to not do some queries when feedId is not defined.

capJavert: I would rather make it non optional in the future but then for anon feeds we would have to implement different functions.

rebelchris: Ah I see, in the last case, let's maybe add a backlog API ticket as non urgent to fix it differently later on.

PR #2143, opened 2024-08-14
fix: add search users 3 char limit

rebelchris:
Added here for adoption purposes to not break FE implementation on old systems.
Also modified the test cases to reflect this limit.


nensidosari (approved):

capJavert (approved):

omBratteng (commented):
lgtm, only thing I would’ve changed is making the 3 into a more easily changeable constant

PR #2146, opened 2024-08-15
fix: add source ID to vordr post slack messages and use short field for username and ID

omBratteng:
Minor change in how messages are sent in slack


rebelchris (approved):
Is it only affecting vodr messages?
(Asking for the other slack messages we have)

omBratteng:
> Is it only affecting vodr messages? (Asking for the other slack messages we have)

Yeah, only Vordr

PR #2150, opened 2024-08-15
fix: enable new post emails for freeform posts

rebelchris:
Enabling post_mention and squad_post_added notifications emails for freeform as well 🤯

### DIFF
--- a/src/common/twitter.ts
+++ b/src/common/twitter.ts
@@ -1,6 +1,6 @@
 import { Post } from '../entity';
 
-export const truncateToTweet = (text: string): string => {
+export const truncateToTweet = (text?: string): string => {
   if (!text) return '';
 
   return text.length <= 130 ? text : `${text.substring(0, 127)}...`;
--- a/src/common/twitter.ts
+++ b/src/common/twitter.ts
@@ -1,9 +1,13 @@
 import { Post } from '../entity';
 
+export const truncateToTweet = (text: string): string => {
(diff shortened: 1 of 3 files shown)


src/common/twitter.ts
```diff
@@ -1,9 +1,13 @@
 import { Post } from '../entity';
 
+export const truncateToTweet = (text: string): string => {
```
capJavert, line 3:
```suggestion
export const truncateToTweet = (text?: string): string => {
```

src/workers/newNotificationV2Mail.ts
```diff
@@ -448,32 +470,51 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
...
       post_link: addNotificationEmailUtm(
         notification.targetUrl,
         notification.type,
       ),
+      user_reputation: author.reputation,
+    };
+
+    if (!sharedPost) {
+      // Freeform post
+      return {
```
capJavert, line 507: Is this right? Regular posts also don't have shared post?

rebelchris: What do you consider a regular post? just article post?
You can't be mentioned on a regular post so it won't come to this part.

capJavert: By mentioned you mean in post's content?

capJavert (approved):
Awesome!

PR #2151, opened 2024-08-15
feat: notification worker when the user loses streak

sshanzel:
NOTE: this should not be merged until we have the feature to restore a streak, otherwise, we will be sending notifications already without having the ability to actually restore the streak.

- When the user's streak gets updated and the current streak becomes 0 from a non-zero value, we should set the cache.
- We will use the cache's value to identify how much was the lost streak.


src/workers/cdc/primary.ts
```diff
@@ -1,85 +1,84 @@
 import {
-  UserState,
```
sshanzel, line 2: Lint annoyingly rearranged the imports on its own 😅

src/workers/cdc/primary.ts
```diff
@@ -795,12 +795,42 @@ const onSquadPublicRequestChange = async (
...
 
+const setRestoreStreakCache = async (
+  userId: string,
+  lastViewAt: Date,
+  previousStreak: number,
+) => {
+  const today = new Date();
+  const lastViewDifference = differenceInDays(today, lastViewAt);
+
+  if (lastViewDifference !== 2) {
```
sshanzel, line 806: - It will never be less than 2 as we don't restore when it has only been a day since user can still read something for the streak today.
- When it is more than 2 days now, it means the user is not eligible anymore to restore it. We only allow it on the day you lose the streak (until the day ends).

sshanzel:
@dailydotdev/web-team feel free to send a review. I just thinking about more tests, but the flow should be as what is intended.

rebelchris (approved):

src/entity/Alerts.ts
```diff
@@ -44,6 +44,9 @@ export class Alerts {
   @Column({ type: 'bool', default: false })
   showStreakMilestone: boolean;
 
+  @Column({ type: 'bool', default: false })
+  showResetStreak: boolean;
```
capJavert, line 48: If we are saving to redis do we need this? Feels like we would need one or the other no point in saving on multiple places?

src/common/utils.ts
```diff
@@ -110,3 +110,7 @@ export const safeJSONParse = <T>(json: string): T | undefined => {
     return undefined;
   }
 };
+
+export function isNumber(value: string | number): boolean {
+  return !!value && !isNaN(Number(value.toString()));
```
capJavert, line 115: this would fail if passed `0`

capJavert: we have `lodash.isnumber` which should be more bulletproof, we can add it to dependencies

src/notifications/generate.ts
```diff
@@ -143,6 +146,15 @@ export const generateNotificationMap: Record<
       .avatarSource(ctx.source)
       .uniqueKey(ctx.bookmark.remindAt.toString())
       .objectPost(ctx.post, ctx.source, ctx.sharedPost),
+  streak_reset_restore: (builder, ctx: NotificationStreakContext) =>
+    builder
+      .icon(NotificationIcon.Streak)
+      .description('Click here if you wish to restore your streak')
+      .uniqueKey(ctx.streak.lastViewAt.toString())
```
capJavert, line 153: could we make it per day (without time portion) eg `19-08-2024` so we don't duplicate notifications for any reason in the same day?

src/common/users.ts
```diff
@@ -431,6 +444,32 @@ export const shouldResetStreak = (
...
+    if (day === lastDayOfWeek) {
+      return FREEZE_DAYS_IN_A_WEEK;
+    }
+
+    if (day === firstDayOfWeek) {
+      return FREEZE_DAYS_IN_A_WEEK + STREAK_RECOVERY_MAX_GAP_DAYS;
+    }
+
+    return STREAK_RECOVERY_MAX_GAP_DAYS;
+  };
```
capJavert, line 468: lets move function to outside so its not inline initiated each time

src/common/users.ts
```diff
@@ -431,6 +444,32 @@ export const shouldResetStreak = (
...
+
+  return difference - getAllowedDays() === 0;
```
capJavert, line 470: this can technically go to negative, just check if thats fine?

src/entity/Alerts.ts
capJavert: Might be missing some context but just feels wrong to have two sources of truth.

src/common/users.ts
sshanzel: Yes, we need exact 0 values.

src/common/utils.ts
sshanzel: Oh, cool. Let me check that.

src/entity/Alerts.ts
ilasw, line 48: Maybe `showRecoverStreak` would be more explicit, what do you think?

sshanzel: Good catch! Also that reset is actually a different thing that happens in our system too.

src/notifications/generate.ts
sshanzel: Yeah, was thinking of that too at first, just realized it is unlikely to happen since reset can only trigger after some days, but I agree, we can do that to ensure duplicates will be avoided.

src/entity/Alerts.ts
sshanzel: Yeah, at first, I thought this wasn't needed too. But the cache is just to hold the value, which can be viewed from the popup many times until it expires.

This one should hold the value of whether the popup should be shown after boot. Then should not be shown again once seen.

src/common/users.ts
sshanzel: Agreed. Should be moved by now.

src/notifications/generate.ts
```diff
@@ -143,6 +146,15 @@ export const generateNotificationMap: Record<
       .avatarSource(ctx.source)
       .uniqueKey(ctx.bookmark.remindAt.toString())
       .objectPost(ctx.post, ctx.source, ctx.sharedPost),
+  streak_reset_restore: (builder, ctx: NotificationStreakContext) =>
+    builder
+      .icon(NotificationIcon.Streak)
+      .description('Click here if you wish to restore your streak')
+      .uniqueKey(new Date(ctx.streak.lastViewAt).toISOString().split('T')[0])
```
capJavert, line 153: let's use `format` from date-fns `dd-MM-YYYY`

capJavert (approved):

src/notifications/generate.ts
sshanzel: Done 🥳 

Just used `yyyy` - console was throwing an error saying I should use it over the uppercase version.

nensidosari (approved):

nensidosari (approved):
Looks good on my side

src/workers/cdc/primary.ts
```diff
@@ -809,12 +817,59 @@ const onSquadPublicRequestChange = async (
...
+
 const onUserStreakChange = async (
   con: DataSource,
   logger: FastifyBaseLogger,
   data: ChangeMessage<UserStreak>,
 ) => {
   if (data.payload.op === 'u') {
+    if (
+      data.payload.after.currentStreak === 0 &&
+      data.payload.before.currentStreak > 2
```
capJavert, line 868: Can we move `2` to some constant, does the day gap one matches?

src/entity/user/UserStreakAction.ts
```diff
@@ -18,7 +18,7 @@ export class UserStreakAction {
   @PrimaryColumn({ type: 'text' })
   type: UserStreakActionType;
 
-  @PrimaryColumn({ type: 'timestamptz', default: () => 'now()' })
+  @PrimaryColumn({ type: 'timestamp', default: () => 'now()' })
```
rebelchris, line 21: More of a request if we can safely modify this and it will not change stored DB values.
Also can we do it for all other `timestamptz` fields?

src/notifications/generate.ts
```diff
@@ -143,6 +147,16 @@ export const generateNotificationMap: Record<
...
+  streak_reset_restore: (builder, ctx: NotificationStreakContext) =>
+    builder
+      .icon(NotificationIcon.Streak)
+      .description('Click here if you wish to restore your streak')
+      .uniqueKey(format(ctx.streak.lastViewAt, 'dd-MM-yyyy'))
+      .targetUrl(notificationsLink)
+      .referenceStreak(ctx.streak)
+      .setTargetUrlParameter([
+        ['streak_restore', ctx.streak.currentStreak.toString()],
+      ]),
```
capJavert, line 159: `streak_restore` is something that triggers popup on the frontend on notifications page?

rebelchris (commented):
Overall good, just safekeeping on the DB before me merge please, and request if we can do all in one go rather?

src/migration/1723670727196-UserStreakAction.ts
```diff
@@ -5,7 +5,7 @@ export class UserStreakAction1723670727196 implements MigrationInterface {
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
-      `CREATE TABLE "user_streak_action" ("userId" character varying(36) NOT NULL, "type" text NOT NULL, "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(), CONSTRAINT "PK_875298ad33fc69f7d704042cdd4" PRIMARY KEY ("userId", "type", "createdAt"))`,
```
capJavert, line 8: Are we gonna manually update this in the DB? ..since this change in migration won't be run on deployment

capJavert (commented):
Few sanity checks, other then that good to go!

src/entity/user/UserStreakAction.ts
sshanzel: There is currently no record for this table, so in this instance, we are safe.

src/common/utils.ts
sshanzel: Ah, yes, forgot to mention. I actually tried it but the test is breaking: https://github.com/dailydotdev/daily-api/pull/2151/commits/a79cdcf82a40fc5828300faf29aa1e859666b010

I think it is due to using `isNaN` vs `Number.isNaN`. The package is probably using the latter, but what we need is the former.

src/notifications/generate.ts
ilasw: Yes, it does!

src/workers/cdc/primary.ts
sshanzel: I think this should be something else, let me update it.

src/common/utils.ts
capJavert: Ok but implementation above is not correct, if we pass `isNumber(0)` it will say `false` while `0` is a number.

src/workers/newView.ts
```diff
@@ -82,8 +84,24 @@ const incrementReadingStreak = async (
...
+        .getRepository(Alerts)
+        .update({ userId }, { showRecoverStreak: false });
+
+      const key = generateStorageKey(
+        StorageTopic.Streak,
+        StorageKey.Reset,
+        userId,
+      );
+
+      await deleteRedisKey(key);
```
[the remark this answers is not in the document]
ilasw: Sure!

src/migration/1723670727196-UserStreakAction.ts
sshanzel: Yeah, this is wrong. I initially thought the migration had not yet run in prod. Let me create a new migration.

src/entity/user/UserStreakAction.ts
sshanzel: For the other fields, we should maybe create a separate ticket, so further investigation can be done and the change must be careful since it involves data changes. Wdyt?

rebelchris: Yeah wasn't aware this was just a new table :)  (empty)
so yeah that makes total sense then.

src/common/utils.ts
capJavert: It is probably related by the fact you do `!!value` which makes your `isNumber` invalid and you are missing a null check.

src/common/users.ts
```diff
@@ -401,6 +416,12 @@ export const Weekends = [Day.Sunday, Day.Saturday];
 export const FREEZE_DAYS_IN_A_WEEK = Weekends.length;
 export const MISSED_LIMIT = 1;
 
+/*
+ * if last streak was Monday, and today is Wednesday, we should allow recovering streak
+ * if last streak was Friday, then the gap is 3 days (the 24-hour period had passed), we should not allow it
```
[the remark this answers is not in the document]
rebelchris: Ah ok, got confused by description then.
If it's the case all is good :)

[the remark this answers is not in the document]
sshanzel: Yeah, thankfully we have tests to prove the cases. The headaches from writing them were worth it 🤣 😭

src/common/utils.ts
capJavert: @sshanzel I added some edge case covers based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#examples and tests to only support number and string types, see 5fcb1255e38a525d4e3e1a0905f1fffeba3d7c35

capJavert (approved):

PR #2152, opened 2024-08-16
fix: sunday as start of week

sshanzel:
When the start of the week is Sunday, the condition is not totally valid on all cases.

We now return early so if the condition doesn't apply, we won't have to unnecessarily check whether it would fit the second part of each remaining conditions.


rebelchris (approved):
Nice find!
But maybe wait for 2 reviews total

sshanzel:
Added two more tests just to be sure all cases are covered explicitly.

capJavert (approved):

omBratteng (approved):

rebelchris (approved):
Nice let's merge it 🙏

PR #2153, opened 2024-08-16
feat(strict-ts): apply to cdc

capJavert:

This one focuses on: 
- cdc
- not the proudest part of the series, mostly used `!` to force unwrap of null values on TS side

This is mainly due to base type `ChangeMessage` being generic for all operations `c | r | u | d`, and we asses the `payload.op` manually

Eg. on `payload.op === u` `before` and `after` will always be not null but on `payload.op === d` `after` will be `null`

In the future implementation strict TS will require us to check for optionality but for legacy stuff it was more efficient to use `!` operator.


rebelchris (approved):
I guess it is what it is :)

PR #2154, opened 2024-08-16
feat(strict-ts): apply to common

capJavert:

This one focuses on: 
- common
- some other files connected to common imports

### DIFF
--- a/src/common/object.ts
+++ b/src/common/object.ts
@@ -34,7 +34,7 @@ export const validateRegex = <T extends ObjectLiteral>(
       return isRequired ? { ...result, [key]: `${key} is required!` } : result;
     }
 
-    const matchResult = value.match(regex);
+    const matchResult = value!.match(regex);
     const isValid = !!matchResult;
 
     if (data && matchResult?.groups?.value) {
--- a/src/common/pubsub.ts
+++ b/src/common/pubsub.ts
@@ -331,13 +331,16 @@ export const notifyContentRequested = async (
 export const notifyFreeformContentRequested = async (
   logger: EventLogger,
   freeform: ChangeMessage<FreeformPost>,
-): Promise<void> =>
-  notifyContentRequested(logger, {
-    id: freeform.payload.after!.id,
-    content: freeform.payload.after!.content,
-    title: freeform.payload.after!.title,
-    post_type: freeform.payload.after!.type,
+): Promise<void> => {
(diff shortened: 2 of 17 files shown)


src/types.ts
```diff
@@ -40,8 +41,8 @@ type IgnoredTypes = Promise<unknown> | ((...args: unknown[]) => unknown);
 export type ChangeObject<Type> = {
   [Property in keyof Type as Exclude<
     Property,
-    Type[Property] extends IgnoredTypes ? Property : never
-  >]: Type[Property] extends Date ? number : Type[Property];
+    Required<Type>[Property] extends IgnoredTypes ? Property : never
+  >]: Required<Type>[Property] extends Date ? number : Type[Property];
```
capJavert, line 45: If propery was declared as `date?: Date` in type the `ChangeObject` util would not cast it to `number`. With this change it does so correctly.

src/common/object.ts
```diff
@@ -34,7 +34,7 @@
       return isRequired ? { ...result, [key]: `${key} is required!` } : result;
     }
 
-    const matchResult = value.match(regex);
+    const matchResult = value!.match(regex);
```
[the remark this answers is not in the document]
capJavert: false positive

rebelchris (approved):
Looks good :)

src/common/pubsub.ts
```diff
@@ -333,10 +333,10 @@ export const notifyFreeformContentRequested = async (
...
   notifyContentRequested(logger, {
-    id: freeform.payload.after.id,
-    content: freeform.payload.after.content,
-    title: freeform.payload.after.title,
-    post_type: freeform.payload.after.type,
+    id: freeform.payload.after!.id,
+    content: freeform.payload.after!.content,
+    title: freeform.payload.after!.title,
+    post_type: freeform.payload.after!.type,
   });
```
omBratteng, line 340:
```ts
  const after = freeform.payload.after!
  notifyContentRequested(logger, {
    id: after.id,
    content: after.content,
    title: after.title,
    post_type: after.type,
  });
```

Could this work? To not having to do `!` each time?

capJavert: yeah, sure 38d60bb170ea246859c834afbe4ecf7a24ed6eac

omBratteng (approved):

PR #2156, opened 2024-08-19
feat: streak recovery query

ilasw:
- Insert query for `streakRecover`;
- Added test case for query;


__tests__/users.ts
```diff
@@ -579,6 +583,72 @@ describe('query userStreaks', () => {
...
+        expect(streakRecover.canDo).toBeFalsy();
+        expect(streakRecover).toHaveProperty('cost');
+        expect(streakRecover).toHaveProperty('oldStreakLength');
+      });
+
+      it('should disallow recover when user is not authenticated', async () => {
+        loggedUser = null;
+        const { data, errors } = await client.query(QUERY);
+        expect(errors).toBeTruthy();
+        expect(data.streakRecover).toBeNull();
```
capJavert, line 613: We use `testQueryErrorCode` for this, see other tests for example

__tests__/users.ts
```diff
@@ -579,6 +583,72 @@ describe('query userStreaks', () => {
...
+      });
+
+      it('should disallow recover when user has no streak', async () => {
+        loggedUser = '2';
+        const { data, errors } = await client.query(QUERY);
+        expect(errors).toBeFalsy();
+        expect(data.streakRecover.canDo).toBeFalsy();
```
capJavert, line 620: again `testQueryErrorCode`

src/entity/user/UserStreakAction.ts
```diff
@@ -5,6 +5,8 @@ export enum UserStreakActionType {
   Recover = 'recover',
 }
 
+export const streakRecoverCost = 25;
```
capJavert, line 8: @sshanzel should we record `ReputationReason` for this when subtracting streak? We can then also move this to `reputationReasonAmount` to have it standardized with other events?

src/schema/users.ts
```diff
@@ -1176,6 +1193,71 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const cantRecoverResult: StreakRecoverQueryResult = {
+        canDo: false,
+        cost: 0,
+        oldStreakLength: 0,
+      };
+
+      const { userId } = ctx;
+      if (!userId) {
+        return cantRecoverResult;
+      }
```
capJavert, line 1210: No need to return this with `@auth` `userId` is always there

src/schema/users.ts
```diff
@@ -1176,6 +1193,71 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        await ctx.con.getRepository(Alerts).update(
+          {
+            userId,
+          },
+          {
+            showRecoverStreak: false,
+          },
+        );
+        return cantRecoverResult;
+      }
```
capJavert, line 1245: I am worried having an `update` inside a query, can we update this somewhere else, eg. from the frontend once user dismisses the modal or worker when streak updates to more then 1?

capJavert (commented):
Looking good, had some questions/comments

src/entity/user/UserStreakAction.ts
ilasw: Yes! Moving there!

__tests__/users.ts
ilasw: This time is different, we need to return the object but is important that `canDo === false`

src/schema/users.ts
ilasw: Yes, I think that moving to a single mutation would be ok, do we have something similar for other Alerts?

src/schema/users.ts
```diff
@@ -1176,6 +1193,68 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const { userId } = ctx;
+
+      const key = generateStorageKey(
+        StorageTopic.Streak,
+        StorageKey.Reset,
+        userId,
+      );
+
+      const oldStreakLength = Number(await getRedisObject(key));
+      const userDontHaveOldStreak =
```
rebelchris, line 1216:
```suggestion
      const userDoesntHaveOldStreak =
```

src/schema/users.ts
```diff
@@ -1176,6 +1193,68 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        !!oldStreakLength && !isNumber(oldStreakLength);
+      if (userDontHaveOldStreak) {
```
rebelchris, line 1218:
```suggestion
      if (userDoesntHaveOldStreak) {
```

src/schema/users.ts
```diff
@@ -1176,6 +1193,68 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        return cantRecoverResult;
+      }
+
+      const streak = await ctx.con.getRepository(UserStreak).findOneBy({
```
rebelchris, line 1222: This should always exist right?

src/schema/users.ts
```diff
@@ -1176,6 +1193,68 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+            userId,
+          },
+          {
+            showRecoverStreak: false,
+          },
+        );
+        return cantRecoverResult;
+      }
+
+      const recoverCount = await ctx.con
```
rebelchris, line 1244: Was there not also some rule to say don't recover if recovered yesterday? (or this above query handles it?)

rebelchris (changes requested):
Minor feedback, overall looks good.

src/schema/users.ts
ilasw: @rebelchris We are handling the checks in the query based on the Redis cache of the user's lost streak. Essentially, from PR #2151 onward, when a user loses their streak, we store a 24-hour value, so we only check for this.

If you recovered yesterday and didn't read anything until today (causing you to lose the streak again), on the streak update cron, the new value will be set in Redis, and you can recover again. Does that logic make sense?

cc @sshanzel

src/schema/users.ts
ilasw: Don't know if every logged user must have a streak, if yes I can remove the if 👀

src/schema/users.ts
rebelchris: Ah got it, yes makes sense :) 🙏

src/schema/users.ts
sshanzel: Yes. We have a trigger to create a streak entity once user is inserted to the DB.

ilasw: Updated! ✅

__tests__/users.ts
capJavert: pardon, seen it wrong thought you were expecting error here, my bad

src/schema/users.ts
capJavert: we have a generic mutation to update alerts, its called `updateUserAlerts` and used from `AuthContext` `updateAlerts` from frontend.

src/schema/users.ts
```diff
@@ -1176,6 +1192,50 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      __,
+      ctx: AuthContext,
+    ): Promise<StreakRecoverQueryResult> => {
+      const { userId } = ctx;
+
+      const streak = await ctx.con.getRepository(UserStreak).findOneBy({
+        userId,
+      });
+      const { currentStreak } = streak;
+      const currentStreakIsTooLongToRecover = currentStreak > 1;
```
nensidosari, line 1206: currentStreakIsTooLongToRecover - what do we mean by it?

ilasw: If user have a current streak of 2 is no possible to recover the old one because at least 2 days have passed since the loss. Maybe I can rename with `currentStreakIsTooLongForRecover` just because the check is "is the current streak too long for trying to recover the old one?". 

But thinking about this I think that if the currentStreak is > 1 we should not have the redis key as mentioned [here](https://github.com/dailydotdev/daily-api/pull/2156#discussion_r1726622461), so we can even remove the check. WDYT?

nensidosari: > But thinking about this I think that if the currentStreak is > 1 we should not have the redis key as mentioned https://github.com/dailydotdev/daily-api/pull/2156#discussion_r1726622461, so we can even remove the check. WDYT?

Yeah with that I think we should be safe, though just since it is here I don't mind keeping it anyways - maybe we can say `timeForRecoveryPassed` to explain the purpose of the check.

src/schema/users.ts
```diff
@@ -508,6 +514,12 @@ export const typeDefs = /* GraphQL */ `
     tags: [TagsReadingStatus]
   }
 
+  type StreakRecoverQuery {
+    canDo: Boolean!
```
rebelchris, line 518: Don't want to spool the whole change, but maybe something like `canRecover` would have been more clear to what it actually means.

No action needed.

src/schema/users.ts
```diff
@@ -1245,6 +1261,55 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          userId,
+          type: UserStreakActionType.Recover,
+        });
+      const cost = recoverCount > 0 ? streakRecoverCost : 0;
+
+      console.log({
+        key,
+        oldStreakLength,
+        streak,
+      });
```
rebelchris, line 1305:
```suggestion
```

rebelchris (approved):
One comment on log, otherwise good to go I think :)

src/schema/users.ts
ilasw: I'm actually pro to change name, we had a similar thread with @nensidosari and if keep raising this change need to be renamed. `canRecover` is okay for both you?

rebelchris: Works for me :)

src/schema/users.ts
```diff
@@ -1245,6 +1261,50 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        userId,
+      });
+      const timeForRecoveryPassed = streak.currentStreak > 1;
+
+      const key = generateStorageKey(
+        StorageTopic.Streak,
+        StorageKey.Reset,
+        userId,
+      );
+      const oldStreakLength = Number(await getRedisObject(key));
```
capJavert, line 1281: We can get `streak` and `oldStreakLength` with `Promise.all` in parallel?

capJavert (approved):
One optimization idea if possible, aside for that good to go 👌

src/schema/users.ts
ilasw: Updated! ✅

src/schema/users.ts
nensidosari: Works for me as well

nensidosari (approved):

PR #2159, opened 2024-08-19
feat: expose post translations

capJavert:
- [x] expose i18n translations through title field
- [x] parse `content-language` header
- [x] save from worker
- [x] add support for required column options in graphorm (so we can load jsonb nested columns)

Basically the flow will be that the frontend will set the language through `gqlClient` header `content-language` and then gql knows how to load it into context and use it in field resolver. 

I will also add global query key on frontend to support language invalidation but more on that in frontend PR.

Setting language during registration and other mutations will be done separate.


src/graphorm/index.ts
```diff
@@ -61,6 +61,26 @@ const nullIfNotSameUser = <T>(
   parent: Pick<User, 'id'>,
 ): T | null => (ctx.userId === parent.id ? value : null);
 
+const createI18nField = (fieldName: string): GraphORMField => {
```
rebelchris, line 64: Nice really like the approach here

src/types.ts
```diff
@@ -120,3 +120,18 @@ export type SlackAuthResponse = {
   };
   error?: string;
 };
+
+export enum ContentLanguage {
```
rebelchris, line 124: More of a generic note do we want to stick with singlular descriptions for some?
(not sure what yggdrasil sets)

but would assume even german would be `de-DE` for instance?

rebelchris (approved):
Nice, love the solution, just one question on the actual abbreviation formatting.

src/graphorm/index.ts
idoshamun: Me too! Very neat

src/migration/1724080448425-ContentLanguage.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ContentLanguage1724080448425 implements MigrationInterface {
+  name = 'ContentLanguage1724080448425';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
```
idoshamun, line 7: Isn't it already stored somewhere when we get it from yggdrasil?

src/workers/postUpdated.ts
```diff
@@ -606,6 +613,12 @@ const fixData = async ({
   }
 
   const duration = data?.extra?.duration / 60;
+  const i18n: PostI18n = {};
+  const titleI18n = data?.meta?.translate_title?.translations;
```
idoshamun, line 617: regarding my previous question, i see that data.meta is stored as is, which means we store twice the titles. can save it only once?
We can consider virtual generated column
https://www.postgresql.org/docs/current/ddl-generated-columns.html

src/types.ts
capJavert: no, this is how yggdrasil assigns it already

src/workers/postUpdated.ts
capJavert: yeah, it is duplicated currently

capJavert: psql only supports stored generated columns, but maybe we can do it on typeorm side or delete it from `meta` when saving

idoshamun: why do we need psql? am I missing something?
Ideally I want to avoid duplication as it's a lot of data

capJavert: you referenced psql generated columns above, and they can't be virtual so if we use them it will be the same as if we are storing double data

I agree on duplication, we need to avoid it

capJavert: Nope it is not supported.

capJavert: We synced offline, basically in 7b4f3b3c5360d128912c851f39b595cf9e95a473 I refactored so it is loaded from `contentMeta`

I also had to add support in graphorm for `requiredColumns` configuration so I can only load part of the `jsonb` field instead of loading all `contentMeta` whenever resolving post entity.

src/graphorm/index.ts
```diff
@@ -120,6 +144,11 @@ const obj = new GraphORM({
       'private',
       'type',
       'slug',
+      {
+        column: `"contentMeta"->'translate_title'->'translations'`,
+        columnAs: 'i18nTitle',
+        isJson: true,
+      },
```
capJavert, line 151: We want to load i18n data all the time so that we can always translate original key to translated one. 

In before implementation I was loading it from new `i18n` column which was a bit smaller but `contentMeta` is big as is so I added support to load `jsonb` nested columns from graphorm.

Now I can just request translated title and then pass it to `createI18nField` so it does the rest.

src/migration/1724080448425-ContentLanguage.ts
capJavert: yes, `contentMeta`

src/graphorm/graphorm.ts
```diff
@@ -338,7 +344,26 @@ export class GraphORM {
...
+      const columnOptions =
+        typeof col === 'object'
+          ? col
+          : {
+              column: col,
+              columnAs: col,
+              isJson: false,
+            };
+
+      if (columnOptions.isJson) {
```
idoshamun, line 356: Why do we need if/else, it looks the same?

idoshamun: nvm i understand now. the if part allows sql injection (kinda)

capJavert: column brackets break with `->` eg. `alias.""contentMeta"->'translate_title'->'translations'"`, and with normal column name its ok `alias."title"`

src/graphorm/index.ts
idoshamun: In the future I'd love to fetch only the desired language and only if needed. let's log it as a point for after the experiment. Again my concern is that the translations are heavy and sending them every time over the network is expensive

idoshamun (approved):

src/graphorm/graphorm.ts
capJavert: yeah, I initially wanted to do `includes('->')` but I think with `isJson` is more safe and we have to consciously in code say this is json column

src/graphorm/index.ts
capJavert: we only send it internally from db to app, we don't send it to user

capJavert: I agree with experiment succeeds we can see to only load specific translation key internally as well

idoshamun: i know. still heavy and inefficient since most of the time you'll need only one language and most users will have english as default

PR #2160, opened 2024-08-20
chore: bin script for missing feed_source follows

rebelchris:
When we migrated from subscribe to follow-notify we missed the fact we need to retro make anyone that's currently subscribed also auto-follow that source.

This script inserts records into `feed_source` if they don't exist, but are subscribed to a source.
(We need to bypass the trigger, else it will auto remove the records from `notification_preference`)

Discussion here:
https://dailydotdev.slack.com/archives/C01473UNK9C/p1724134356244039

### DIFF
--- a/bin/insertMissingSourceFollows.ts
+++ b/bin/insertMissingSourceFollows.ts
@@ -1,0 +1,29 @@
+import '../src/config';
+import createOrGetConnection from '../src/db';
+
+/**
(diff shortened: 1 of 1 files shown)


nensidosari (approved):

idoshamun (approved):

capJavert (approved):

idoshamun:
We can also just run it in the DB if you wish. I don't expect it to overload much

PR #2161, opened 2024-08-20
feat: introduce company and usercompany structure

rebelchris:
Adds the new structure for company and user_company tables.
Also added all necessary mutations and query changes.

## Todo:
- [x] test cases

You should be able to review this part already :)

### DIFF
--- a/src/data-source.ts
+++ b/src/data-source.ts
@@ -16,7 +16,7 @@ export const AppDataSource = new DataSource({
     max: 20,
     idleTimeoutMillis: 120000,
   },
-  logging: true,
+  logging: false,
   entities: ['src/entity/**/*.{js,ts}'],
   migrations: ['src/migration/**/*.{js,ts}'],
   subscribers: ['src/subscriber/**/*.{js,ts}'],
--- a/src/data-source.ts
+++ b/src/data-source.ts
@@ -16,7 +16,7 @@ export const AppDataSource = new DataSource({
     max: 20,
     idleTimeoutMillis: 120000,
   },
-  logging: false,
+  logging: true,
   entities: ['src/entity/**/*.{js,ts}'],
   migrations: ['src/migration/**/*.{js,ts}'],
   subscribers: ['src/subscriber/**/*.{js,ts}'],
--- a/src/entity/Company.ts
+++ b/src/entity/Company.ts
@@ -1,0 +1,19 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class Company {
+  @PrimaryColumn({ type: 'text' })
+  id: string;
+
+  @Column({ type: 'text' })
+  name: string;
+
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ type: 'text', nullable: true })
+  image?: string;
+
+  @Column({ type: 'text', array: true, default: [] })
+  domains: string[];
+}
--- a/src/entity/Company.ts
+++ b/src/entity/Company.ts
@@ -1,4 +1,4 @@
-import { Column, Entity, PrimaryColumn } from 'typeorm';
+import { Column, CreateDateColumn, Entity, PrimaryColumn } from 'typeorm';
 
 @Entity()
 export class Company {
@@ -8,7 +8,7 @@ export class Company {
   @Column({ type: 'text' })
   name: string;
 
-  @Column({ default: () => 'now()' })
+  @CreateDateColumn()
   createdAt: Date;
 
   @Column({ type: 'text', nullable: true })
--- a/src/entity/Company.ts
+++ b/src/entity/Company.ts
@@ -1,4 +1,10 @@
-import { Column, CreateDateColumn, Entity, PrimaryColumn } from 'typeorm';
+import {
+  Column,
+  CreateDateColumn,
+  Entity,
+  PrimaryColumn,
+  UpdateDateColumn,
+} from 'typeorm';
 
 @Entity()
 export class Company {
@@ -10,6 +16,9 @@ export class Company {
 
   @CreateDateColumn()
   createdAt: Date;
+
+  @UpdateDateColumn()
+  updatedAt: Date;
 
   @Column({ type: 'text', nullable: true })
   image?: string;
--- a/src/entity/UserCompany.ts
+++ b/src/entity/UserCompany.ts
@@ -1,21 +1,23 @@
 import {
   Column,
+  CreateDateColumn,
   Entity,
   Index,
   JoinColumn,
   ManyToOne,
   OneToOne,
   PrimaryColumn,
+  UpdateDateColumn,
 } from 'typeorm';
 import { User } from './user';
 import { Company } from './Company';
 
 @Entity()
 export class UserCompany {
-  @Column({ default: () => 'now()' })
+  @CreateDateColumn()
   createdAt: Date;
 
-  @Column({ default: () => 'now()' })
+  @UpdateDateColumn()
   updatedAt: Date;
 
   @Column({ default: false })
@@ -26,7 +28,6 @@ export class UserCompany {
   email: string;
 
   @Column({ type: 'text' })
-  @Index({ unique: true })
   code: string;
 
   @PrimaryColumn()
@@ -47,6 +48,5 @@ export class UserCompany {
     lazy: true,
     onDelete: 'CASCADE',
   })
-  @JoinColumn({ name: 'companyId' })
   company: Promise<Company>;
 }
--- a/src/entity/UserCompany.ts
+++ b/src/entity/UserCompany.ts
@@ -1,0 +1,52 @@
+import {
+  Column,
+  Entity,
+  Index,
+  JoinColumn,
+  ManyToOne,
+  OneToOne,
+  PrimaryColumn,
+} from 'typeorm';
+import { User } from './user';
+import { Company } from './Company';
+
+@Entity()
+export class UserCompany {
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ default: () => 'now()' })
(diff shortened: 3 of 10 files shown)


src/data-source.ts
```diff
@@ -16,7 +16,7 @@ export const AppDataSource = new DataSource({
     max: 20,
     idleTimeoutMillis: 120000,
   },
-  logging: false,
+  logging: true,
```
omBratteng: Accident?

src/schema/users.ts
```diff
@@ -1320,6 +1377,22 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        builder.queryBuilder = builder.queryBuilder
+          .andWhere(`${builder.alias}."userId" = :userId`, {
+            userId: ctx.userId,
+          })
+          .andWhere(`${builder.alias}."verified" = true`);
+
+        return builder;
+      });
+      if (userCompanies.length === 0) {
+        throw new NotFoundError('No companies found');
```
omBratteng, line 1391: Do we need to throw error, or just return empty array or null?

src/schema/users.ts
```diff
@@ -1636,6 +1709,75 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
       return { _: true };
     },
+    addUserCompany: async (
```
omBratteng, line 1712: Does this check if the email already exist?

src/schema/users.ts
```diff
@@ -1320,6 +1377,22 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
       return personalizedDigest;
     },
+    companies: async (_, __, ctx: AuthContext, info) => {
```
rebelchris, line 1380: Missing return type here..

src/schema/users.ts
rebelchris: Ah good point, empty array is fine :)
adjusted (Also are you not on leave?)

src/schema/users.ts
rebelchris: Ah good one, will need to add a test for it, it's a unique index so think the query will already fail on it.

src/schema/users.ts
omBratteng: > (Also are you not on leave?)
🙈

src/schema/users.ts
omBratteng: If the email exists, but claimed by same user and not verified, we should just "ignore" that it exists and just update verification code
In the case when the user stops midway in the verification process (i.e. computer dies or something) and before the cleanup job gets it. [Thread](https://dailydotdev.slack.com/archives/C07EA1E1XME/p1723730384460159?thread_ts=1723727031.217919&cid=C07EA1E1XME)

src/entity/UserCompany.ts
```diff
@@ -0,0 +1,52 @@
...
+import { User } from './user';
+import { Company } from './Company';
+
+@Entity()
+export class UserCompany {
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ default: () => 'now()' })
+  updatedAt: Date;
```
capJavert, line 19: We can do `CreateDateColumn` i `UpdateDateColumn`

src/entity/UserCompany.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  @Column()
+  companyId: string;
+
+  @OneToOne(() => Company, {
+    nullable: true,
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({ name: 'companyId' })
```
capJavert, line 50: Typeorm does not know its `companyId` automatically?

src/entity/UserCompany.ts
```diff
@@ -0,0 +1,52 @@
...
+  @Column({ default: false })
+  verified: boolean;
+
+  @PrimaryColumn({ type: 'text' })
+  @Index({ unique: true })
+  email: string;
+
+  @Column({ type: 'text' })
+  @Index({ unique: true })
+  code: string;
```
capJavert, line 30: Why does code need to be unique?

capJavert (commented):
Nice work. Some comments/questions

src/schema/users.ts
rebelchris: So resubmitting email will just generate a new code?
I'll add that :)

src/entity/UserCompany.ts
rebelchris: What difference do they actually make under the hood?
(I will modify them)

src/entity/UserCompany.ts
rebelchris: Missed it before I added the column definition:
```
@Column()
  companyId: string;
```

src/entity/UserCompany.ts
rebelchris: Hmm yeah actually doesn't have to be :)

src/schema/users.ts
```diff
@@ -1636,6 +1709,75 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       return { _: true };
     },
+    addUserCompany: async (
+      _,
+      { email }: { email: string },
+      ctx: AuthContext,
+    ): Promise<GQLEmptyResponse> => {
+      if (!email?.length) {
+        throw new ValidationError('Email is required');
+      }
```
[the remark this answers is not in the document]
rebelchris: I'll check with tests :)

src/entity/UserCompany.ts
capJavert: `UpdateDateColumn` should update when you do `save` on entity.

src/schema/users.ts
```diff
@@ -1636,6 +1711,79 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          email,
+        });
+
+      if (existingUserCompanyEmail) {
+        if (existingUserCompanyEmail.userId !== ctx.userId) {
+          throw new ValidationError('Email already exists');
+        }
+
+        const updatedRecord = { ...existingUserCompanyEmail, code };
+        await ctx.con.getRepository(UserCompany).save(updatedRecord);
```
omBratteng, line 1740: IIRC if we use upsert, it'll update if exists, or insert if not?

omBratteng (approved):
Looks good, haven't looked at tests, too much for mobile 🙈

src/schema/users.ts
rebelchris: Hmm not sure on typeorm upsert, I vaguely recall it having some issues.
CC: @capJavert you recall anything for upsert?

(We still need the split query anyway for fetching as we don't want to update someone else's entry)

capJavert: save should also do the same but upsert allows providing on conflict part

src/entity/Company.ts
```diff
@@ -0,0 +1,19 @@
...
+@Entity()
+export class Company {
+  @PrimaryColumn({ type: 'text' })
+  id: string;
+
+  @Column({ type: 'text' })
+  name: string;
+
+  @CreateDateColumn()
+  createdAt: Date;
```
capJavert, line 12: Just for the sake of it, do we want updatedAt here as well, useful for most entities?

src/schema/users.ts
```diff
@@ -1646,6 +1721,79 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      _,
+      { email }: { email: string },
+      ctx: AuthContext,
+      info,
+    ): Promise<GQLUserCompany[]> => {
+      await ctx.con
+        .getRepository(UserCompany)
+        .delete({ email, userId: ctx.userId });
+
+      return getUserCompanies(_, ctx, info);
```
capJavert, line 1772: Seems wasteful to fetch all companies on delete? Why do you need it?

src/entity/Company.ts
rebelchris: We could add it yeah, don't see any harm in it.

src/schema/users.ts
```diff
@@ -727,6 +764,11 @@ export const typeDefs = /* GraphQL */ `
     Get integrations for the user
     """
     userIntegrations: UserIntegrationConnection @auth
+
+    """
+    Get companies for user
+    """
+    companies: [UserCompany] @auth
```
capJavert, line 771: Just noticed this allows querying email for other users over `companies` field? Also do we want to limit in customRelation to something like 10 or 50 so we don't return hundreds of rows if someone tries to be naughty

src/graphorm/index.ts
```diff
@@ -119,6 +120,38 @@ const obj = new GraphORM({
...
+          isMany: true,
+          customRelation: (ctx, parentAlias, childAlias, qb): QueryBuilder =>
+            qb
+              .innerJoin(
+                UserCompany,
+                'uc',
+                `"${childAlias}"."id" = uc."companyId"`,
+              )
+              .where('uc.verified = true')
+              .andWhere(`uc."userId" = "${parentAlias}".id`),
```
capJavert, line 134:
```suggestion
              .andWhere(`uc."userId" = "${parentAlias}".id`)
              .limit(50),
```

just so we don't return hundreds of rows if somebody tries to be naughty, we do the same for `privilegedMembers`

capJavert (approved):
Awesome work, just last two comments to check.

src/graphorm/index.ts
rebelchris: For now you can actually only add 1 anyway, we dont support more from fe.

We can tackle this maybe in follow up once we ever do?

capJavert: But from API directly someone could do it 😆

capJavert: Someone naughty xD

PR #2162, opened 2024-08-20
feat: add user language

capJavert:
- [x] add user language column
- [x] add support for language in `newUser` endpoint
- [x] update boot to support language
- [x] users schema updates to support reading/writing to language

https://github.com/dailydotdev/heimdall/pull/98

### DIFF
--- a/__tests__/private.ts
+++ b/__tests__/private.ts
@@ -91,7 +91,6 @@ describe('POST /p/newUser', () => {
         username: 'randomName',
         email: 'randomNewEmail@gmail.com',
         experienceLevel: 'LESS_THAN_1_YEAR',
-        language: ContentLanguage.English,
       })
       .expect(200);
 
@@ -112,7 +111,6 @@ describe('POST /p/newUser', () => {
         username: usersFixture[0].username,
         email: 'randomNewEmail@gmail.com',
         experienceLevel: 'LESS_THAN_1_YEAR',
-        language: ContentLanguage.English,
       })
       .expect(200);
 
@@ -132,7 +130,6 @@ describe('POST /p/newUser', () => {
         username: usersFixture[0].username.toUpperCase(),
         email: 'randomNewEmail@gmail.com',
         experienceLevel: 'LESS_THAN_1_YEAR',
-        language: ContentLanguage.English,
       })
       .expect(200);
 
@@ -152,7 +149,6 @@ describe('POST /p/newUser', () => {
         username: 'disallow',
         email: 'randomNewEmail@gmail.com',
         experienceLevel: 'LESS_THAN_1_YEAR',
-        language: ContentLanguage.English,
       })
       .expect(200);
 
@@ -171,7 +167,6 @@ describe('POST /p/newUser', () => {
         username: 'h-ello',
         email: 'randomNewEmail@gmail.com',
         experienceLevel: 'LESS_THAN_1_YEAR',
-        language: ContentLanguage.English,
       })
       .expect(200);
 
@@ -190,7 +185,6 @@ describe('POST /p/newUser', () => {
         username: 'h/ello',
         email: 'randomNewEmail@gmail.com',
         experienceLevel: 'LESS_THAN_1_YEAR',
-        language: ContentLanguage.English,
       })
       .expect(200);
 
@@ -209,7 +203,6 @@ describe('POST /p/newUser', () => {
         username: 'he',
         email: 'randomNewEmail@gmail.com',
         experienceLevel: 'LESS_THAN_1_YEAR',
-        language: ContentLanguage.English,
       })
       .expect(200);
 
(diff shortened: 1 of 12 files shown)


__tests__/private.ts
```diff
@@ -547,6 +568,48 @@ describe('POST /p/newUser', () => {
...
+        name: usersFixture[0].name,
+        image: usersFixture[0].image,
+        username: usersFixture[0].username,
+        email: usersFixture[0].email,
+        experienceLevel: 'foo',
+        language: 'klingon',
+      })
+      .expect(200);
+
+    expect(body).toEqual({ status: 'failed', reason: 'MISSING_FIELDS' });
```
rebelchris, line 611: We actually need to check maybe with Viktor what happens now if we throw some error to heimdall?
Not really sure it persists back to the end user. (which would mean probably half broken user exists somewhere)

src/entity/user/User.ts
```diff
@@ -239,15 +243,23 @@ export type AddUserData = Pick<
...
 
+const checkLanguage = (language?: string): boolean => {
+  return !!language && validLanguages.includes(language as ContentLanguage);
+};
+
 const checkRequiredFields = (data: AddUserData): boolean => {
-  if (data?.username && !data?.experienceLevel) {
+  if (
+    data?.username &&
+    (!data?.experienceLevel || !checkLanguage(data?.language))
```
rebelchris, line 261: Was just wondering if we should maybe split them into 2 internal ifs?

src/migration/1724168427191-UserLanguage.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UserLanguage1724168427191 implements MigrationInterface {
+  name = 'UserLanguage1724168427191';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "user" ADD "language" text`,
```
rebelchris, line 8: We don't want to set a default?

rebelchris (changes requested):
Minor feedback, think most is fine, but blocking for validation.
Also please sync with Viktor on deployment of heimdall and once done we need a e2e auth check: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/458096761/Critical+flows

src/migration/1724168427191-UserLanguage.ts
capJavert: I was thinking we might want to differentiate users that selected language and those legacy ones that did not yet? So we can (for example) run a campaign to update language at later date?

__tests__/private.ts
capJavert: It should be the same as when we introduced experienceLevel, so should be no problem?

src/migration/1724168427191-UserLanguage.ts
capJavert: Either way in API users will fallback to english when nothing is set

__tests__/private.ts
capJavert: On the other thought, maybe experienceLevel was introduced as required later due to extension, I can just make it optional for now until adoption/experiment is over?

rebelchris: It needs to be optional for flows where we subscribe first -> then ask for it.
I think email is that exception if I remember correct.
 (Actually not sure how we fixed that now for expLevel?) CC: @nensidosari

capJavert: Required experienceLevel was introduced here https://github.com/dailydotdev/daily-api/pull/2034, see thread for more info

capJavert: I think we don't need language to be required until experiment concludes either way?

rebelchris: Oh I see: `if (data?.username` fixes the email users :) 
Hmm yeah just have it optional I guess for now.

src/migration/1724168427191-UserLanguage.ts
rebelchris: Ah yeah that's good :)

src/entity/user/User.ts
capJavert: Split it now

__tests__/private.ts
capJavert: Made optional here 58bc9a9ab43f7f34d3074256a418e27825be06b4 but still validates it if provided

rebelchris (approved):
Looks good!

__tests__/private.ts
```diff
@@ -547,6 +548,48 @@ describe('POST /p/newUser', () => {
...
+      .set('Content-type', 'application/json')
+      .set('authorization', `Service ${process.env.ACCESS_SECRET}`)
+      .send({
+        id: usersFixture[0].id,
+        name: usersFixture[0].name,
+        image: usersFixture[0].image,
+        username: usersFixture[0].username,
+        email: usersFixture[0].email,
+        experienceLevel: 'foo',
+        language: 'klingon',
```
nensidosari, line 587: 😆

nensidosari (approved):

__tests__/private.ts
capJavert: until we introduce it then we will change the test 😆

PR #2165, opened 2024-08-21
feat: streak recover mutation

ilasw:
Mutation checks for:
- User is logged;
- User lost a streak in the past 24h, using Redis cache;
- Current streak is <= 1
- User can afford the cost, find for prev recover using `UserStreakActions` query;

> Not sure if the last 3 updates are the best way to update the user.


src/schema/users.ts
```diff
@@ -1866,6 +1878,77 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       };
     },
+    recoverStreak: async (
+      _,
+      __,
+      ctx: AuthContext,
+      info,
+    ): Promise<GQLUserStreak> => {
+      const { userId } = ctx;
+      const user = await ctx.con
```
rebelchris, line 1888: There's `@auth` directive on graphQL side we shouldn't need to manually verify again.

src/schema/users.ts
```diff
@@ -1866,6 +1878,77 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        StorageTopic.Streak,
+        StorageKey.Reset,
+        userId,
+      );
+      const oldStreakLength = Number(await getRedisObject(key));
+      const userDoesntHaveOldStreak =
+        !oldStreakLength || !isNumber(oldStreakLength);
+      if (userDoesntHaveOldStreak) {
+        throw new ValidationError('No streak to recover');
+      }
```
rebelchris, line 1902: Not 100% sure but isn't this 1-1 the same as for the query, maybe we can combine some logic in a fn?

src/schema/users.ts
```diff
@@ -1866,6 +1878,77 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      const lastUserRecoverAction = await ctx.con
+        .getRepository(UserStreakAction)
+        .findOneBy({
+          userId,
+          type: UserStreakActionType.Recover,
+        });
+      const isFirstRecover = !lastUserRecoverAction;
+      const recoverCost = isFirstRecover ? 0 : 25;
```
rebelchris, line 1918: Can we maybe use `reputationReasonAmount` here somehow?

rebelchris: For the 25 I mean

rebelchris (changes requested):
Minor feedback, but overall looks good.

src/schema/users.ts
ilasw: I still need the `user.reputation` object below, should I change `findOneByOrFail` with `findOneBy`?

rebelchris: Oh right, in that case leave it yes :)

src/schema/users.ts
```diff
@@ -1866,6 +1878,77 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      });
+
+      await ctx.con.getRepository(UserStreak).update(
+        {
+          userId,
+        },
+        {
+          currentStreak: oldStreakLength + streak.current,
+        },
+      );
```
capJavert, line 1948: We can do this in transaction to have changes saved in atomic way in the db.

capJavert (commented):
Good work, appreciate the tests. Had some optimization ideas of data/loading saving.

src/schema/users.ts
```diff
@@ -1866,6 +1878,77 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+    recoverStreak: async (
+      _,
+      __,
+      ctx: AuthContext,
+      info,
+    ): Promise<GQLUserStreak> => {
+      const { userId } = ctx;
+      const user = await ctx.con
+        .getRepository(User)
+        .findOneByOrFail({ id: userId });
```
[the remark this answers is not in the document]
ilasw: Sure! Updated ✅

src/schema/users.ts
```diff
@@ -1866,6 +1878,77 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        .findOneBy({
+          userId,
+          type: UserStreakActionType.Recover,
+        });
+      const isFirstRecover = !lastUserRecoverAction;
+      const recoverCost = isFirstRecover ? 0 : streakRecoverCost;
+      const userCanAfford = user.reputation >= recoverCost;
+
+      if (!userCanAfford) {
+        throw new ValidationError('Not enough reputation to recover streak');
```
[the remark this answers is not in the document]
ilasw: Yes! ✅

src/schema/users.ts
ilasw: Updated ✅

capJavert (approved):

src/schema/users.ts
ilasw: Updated! ✅

src/schema/users.ts
ilasw: Updated! ✅

__tests__/users.ts
```diff
@@ -583,6 +588,126 @@ describe('query userStreaks', () => {
...
+      await con.getRepository(UserStreak).update(
+        { userId: loggedUser },
+        {
+          currentStreak: 0,
+          lastViewAt: subDays(new Date(), 2),
+        },
+      );
+      await con
+        .getRepository(User)
+        .update({ id: loggedUser }, { reputation: 25 });
```
capJavert, line 687: Would we not use a const here instead of `25`?

capJavert: Just so tests go through if we change the points

src/schema/users.ts
```diff
@@ -1866,6 +1880,71 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const [user, lastUserRecoverAction] = await Promise.all([
+        ctx.con.getRepository(User).findOneByOrFail({ id: userId }),
+        await ctx.con.getRepository(UserStreakAction).findOneBy({
+          userId,
+          type: UserStreakActionType.Recover,
+        }),
+      ]);
+      const isFirstRecover = !lastUserRecoverAction;
+      const recoverCost = isFirstRecover ? 0 : streakRecoverCost;
+      const userCanAfford = user.reputation >= recoverCost;
```
capJavert, line 1912: Can't we just use `reputationReasonAmount` here?

src/schema/users.ts
```diff
@@ -1866,6 +1880,71 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      }
+
+      const reputationEvent = {
+        grantToId: userId,
+        targetId: userId,
+        targetType: ReputationType.Streak,
+        reason: isFirstRecover
+          ? ReputationReason.StreakFirstRecovery
+          : ReputationReason.StreakRecover,
+        amount: recoverCost * -1,
```
capJavert, line 1925: Then also here we don't have to do `* -1`

src/schema/users.ts
capJavert: Its fine this way around if you prefer just need to replace `0` with `reputationReasonAmount`

rebelchris (approved):

nensidosari (approved):

PR #2166, opened 2024-08-22
fix: UserStreakAction relation with User must be ManyToOne

ilasw:


src/entity/user/UserStreakAction.ts
```diff
@@ -16,7 +21,7 @@ export class UserStreakAction {
   @PrimaryColumn({ type: 'timestamptz', default: () => 'now()' })
   createdAt: Date;
 
-  @OneToOne(() => User, (user) => user.streak, {
+  @ManyToOne(() => User, (user) => user.streak, {
```
capJavert, line 19: This should also be `streakActions`?

sshanzel: You might have to run migrations too for this one. Let me know if you need any help on that end.

sshanzel (approved):
Thanks for the fix! 🚀

nensidosari (approved):

capJavert (approved):

PR #2167, opened 2024-08-22
feat: updated ReputationEvent entity

ilasw:


src/entity/ReputationEvent.ts
```diff
@@ -23,14 +25,16 @@ export enum ReputationType {
   Source = 'source',
 }
 
-const reputationReasonAmount: Record<ReputationReason, number> = {
+export const reputationReasonAmount: Record<ReputationReason, number> = {
```
capJavert, line 28: Why are we doing this in separate PR, let's bundle it in the one where it is used? https://github.com/dailydotdev/daily-api/pull/2156#discussion_r1726531910

src/entity/ReputationEvent.ts
```diff
@@ -15,6 +15,8 @@ export enum ReputationReason {
   PostReportConfirmed = 'post_report_confirmed',
   SourceRequestApproved = 'source_request_approved',
   CommentDownvoted = 'comment_downvoted',
+  StreakRecover = 'streak_recover',
+  StreakRecoverForFree = 'streak_recover_for_free',
```
sshanzel, line 19: I think it might be better to say `StreakFirstRecovery`. Not a blocker though, I think it is more explicit.

capJavert: agreed

src/entity/ReputationEvent.ts
ilasw: Yes, moving to a separate one because it's needed in mutation too

capJavert (approved):

PR #2169, opened 2024-08-22
feat(strict-ts): apply to common 2

capJavert:

This one focuses on: 
- commons
- some adjustments to existing code that got broken over during process

### DIFF
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -107,7 +107,11 @@ export const feedToFilters = async (
   userId?: string,
 ): Promise<AnonymousFeedFilters> => {
   const settings = await getExcludedAdvancedSettings(con, feedId);
-  const [tags, excludeSources, memberships] = await Promise.all([
+  const [tags, excludeSources, memberships]: [
+    FeedTag[],
+    Source[],
+    { sourceId: SourceMember['sourceId']; hide: boolean }[],
+  ] = await Promise.all([
     feedId
       ? con.getRepository(FeedTag).find({ where: { feedId } })
       : ([] as FeedTag[]),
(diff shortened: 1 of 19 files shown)


src/entity/posts/ArticlePost.ts
```diff
@@ -9,9 +9,9 @@ export class ArticlePost extends Post {
   @Column({ nullable: true })
   publishedAt?: Date;
 
-  @Column({ type: 'text' })
+  @Column({ type: 'text', nullable: true })
   @Index({ unique: true })
-  url: string;
+  url: string | null;
```
capJavert, line 14: it was always nullable was only missing typeorm option on `Column`

src/common/feedGenerator.ts
```diff
@@ -107,7 +107,11 @@ export const feedToFilters = async (
   userId?: string,
 ): Promise<AnonymousFeedFilters> => {
   const settings = await getExcludedAdvancedSettings(con, feedId);
-  const [tags, excludeSources, memberships] = await Promise.all([
+  const [tags, excludeSources, memberships]: [
```
rebelchris, line 110: These are gonna be so useful 🙏

src/common/feedGenerator.ts
```diff
@@ -151,12 +155,15 @@ export const feedToFilters = async (
       }
       return acc;
     },
-    { includeTags: [], blockedTags: [] },
+    { includeTags: [], blockedTags: [] } as {
+      includeTags: string[];
+      blockedTags: string[];
+    },
```
rebelchris, line 161: Wouldn't this already work?
```suggestion
    { includeTags: string[], blockedTags: string[] }
```

rebelchris (approved):
Looks good, non blocking comment/question

src/common/feedGenerator.ts
capJavert: that would be syntax error, its not type its `acc` value of `reduce`, but I moved it to `reduce` generic above. 23504d826da8baaca0c014a0f891afb18d665bce

PR #2170, opened 2024-08-22
feat(strict-ts): apply to integrations

capJavert:

This one focuses on: 
- integrations (mainly feed)

### DIFF
--- a/src/common/slack.ts
+++ b/src/common/slack.ts
@@ -6,6 +6,7 @@ import { DataSource } from 'typeorm';
 import { UserIntegrationSlack } from '../entity/UserIntegration';
 import { createHmac, timingSafeEqual } from 'node:crypto';
 import { FastifyRequest } from 'fastify';
+import { PropsParameters } from '../types';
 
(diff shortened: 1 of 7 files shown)


src/common/slack.ts
```diff
@@ -244,11 +244,13 @@ export const getSlackIntegration = async ({
...
 
-export const getSlackIntegrationOrFail: typeof getSlackIntegration = async ({
+export const getSlackIntegrationOrFail = async ({
   id,
   userId,
   con,
-}) => {
+}: Parameters<
+  typeof getSlackIntegration
+>[0]): Promise<UserIntegrationSlack> => {
```
rebelchris, line 253: Kinda ugly, but guess it's what it is

rebelchris (approved):
Let's gooo

src/common/slack.ts
capJavert: Agreed, added `PropsParameters` util so we can extract more easily also looks better 6f0ad7d65c4763b795315d879f6d1bd424390cf4

PR #2171, opened 2024-08-23
feat: added common function for get old streak to restore

ilasw:


capJavert (changes requested):
Can we just introduce this in the first PR that uses it? I see you are doing it a lot and its hard to reason about the function that is not being used anywhere. 

I understand that sometimes you need a function in multiple other PRs but I think the right way is to merge the first one that uses it and then continue with other ones.

I am happy to help/review the ones you need first if needed.

PR #2174, opened 2024-08-24
fix: notification worker cdc date value

sshanzel:
I was running into some issues while updating the tests due to the fact the `ChangeObject` always expects to receive a `number` for a `Date` column, when in fact, it can receive a `string` when the `Date` column is a `timestamptz`.

Have updated the `ChangeObject` to reflect the actual types that it may contain.

Also, I think having it explicitly defined whether the value is `as number` or `as string` safeguards us from mistakes as you will have to look first to verify which type to use based on the column date type.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -42,7 +42,7 @@ export type ChangeObject<Type> = {
   [Property in keyof Type as Exclude<
     Property,
     Required<Type>[Property] extends IgnoredTypes ? Property : never
-  >]: Required<Type>[Property] extends Date ? number : Type[Property];
+  >]: Required<Type>[Property] extends Date ? number | string : Type[Property];
 };
 
 export type ChangeSchema = {
--- a/src/types.ts
+++ b/src/types.ts
@@ -37,12 +37,13 @@ declare module 'fastify' {
 }
 
 type IgnoredTypes = Promise<unknown> | ((...args: unknown[]) => unknown);
+type CdcDate = number | string;
 
 export type ChangeObject<Type> = {
   [Property in keyof Type as Exclude<
     Property,
     Required<Type>[Property] extends IgnoredTypes ? Property : never
-  >]: Required<Type>[Property] extends Date ? number | string : Type[Property];
+  >]: Required<Type>[Property] extends Date ? CdcDate : Type[Property];
 };
 
 export type ChangeSchema = {
--- a/src/workers/cdc/common.ts
+++ b/src/workers/cdc/common.ts
@@ -88,8 +88,8 @@ export const notifyPostContentUpdated = async ({
     postId: post.id,
     type: post.type,
     title: post.title,
-    createdAt: post.createdAt,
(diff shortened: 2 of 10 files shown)


nensidosari (approved):
Looks good for now :) 
We might need to go after with this idea of unifying what we use https://dailydotdev.slack.com/archives/C02E2C3C13R/p1724477615641729?thread_ts=1724430381.874289&cid=C02E2C3C13R

src/types.ts
```diff
@@ -37,12 +37,13 @@ declare module 'fastify' {
 }
 
 type IgnoredTypes = Promise<unknown> | ((...args: unknown[]) => unknown);
+type CdcDate = number | string;
```
capJavert, line 40: I would rather keep this as `number` and maybe cast on specific parts where `timestampz` is used, because this will introduce a bunch of places where it breaks types. cc: @rebelchris wdyt?

capJavert: It might not be obvious but in strict TS mode this PR adds another 100 type errors in the codebase.

src/workers/cdc/common.ts
```diff
@@ -88,8 +88,8 @@ export const notifyPostContentUpdated = async ({
     postId: post.id,
     type: post.type,
     title: post.title,
-    createdAt: post.createdAt,
-    updatedAt: post.metadataChangedAt,
+    createdAt: post.createdAt as number,
+    updatedAt: post.metadataChangedAt as number,
```
ilasw, line 92: I would prefer to actually convert those values to number if possible, kinda:


```ts
createdAt: +post.createdAt,
```

src/types.ts
rebelchris: I rather would see a move from timestamptz to timestamp then trying to retro fix it with potential of breaking implementations now.

Not sure how everyone feels about that?

capJavert: I am up for that too, alternative is to just handle it in places where timestampz is used, we can explicit cast `value as unknown as string` and thats it (in case we don't want to complicate with DB migrations now).

sshanzel: After checking the build, there seem to be no difference in terms of errors:

branch MI-504
Screenshot 2024-08-27 at 5 04 31 PM

this branch. It actually fixed one more thing.
Screenshot 2024-08-27 at 5 05 08 PM

sshanzel: The branch was not just updated from the main, hence it looked like it introduced strict mode issues. Updated it and now it is even better.

capJavert: @sshanzel oh, my bad on the number of errors part then.

sshanzel: @rebelchris since there was no particular reason for using `timestamptz`, I changed it to `timestamp`. We can probably reopen this PR if we still want the type change, it is still correct under certain circumstances.

Relevant change: https://github.com/dailydotdev/daily-api/pull/2151/commits/91a7d32f0e0338c23f7b862cb45c6dcbf1e5ee00

PR #2175, opened 2024-08-26
fix: some FE related feedback

rebelchris:
Some feedback based on what made sense in FE.

- Image should be mantadory :) (else it's pointless to have a record)
- verifyUserCompanyCode should return the company row to retro input it in FE

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -1744,7 +1744,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       if (existingUserCompanyEmail) {
         if (existingUserCompanyEmail.userId !== ctx.userId) {
           throw new ValidationError(
-            'Oops, this email has already been verified by another user.',
+            'Oops, there was an issue verifying this email. Please use a different one.',
           );
         }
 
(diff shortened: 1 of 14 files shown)


src/schema/users.ts
```diff
@@ -1743,7 +1743,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
       if (existingUserCompanyEmail) {
         if (existingUserCompanyEmail.userId !== ctx.userId) {
-          throw new ValidationError('Email already exists');
+          throw new ValidationError(
+            'Oops, this email has already been verified by another user.',
```
capJavert, line 1747: do we want to give that much info in this case, it should be edge case where someone tries to verify someone else's email and its pretty suspicious case?

capJavert (approved):

src/schema/users.ts
rebelchris: Ah it's just the copy from design I can check with them?
I guess you could find out email addresses this way yeah.

capJavert: We should change it with something more generic like `Oops, this email has can't been verified at this time.`, and we can log something internally to debug it if needed.

capJavert (approved):

nensidosari (approved):

PR #2176, opened 2024-08-26
feat: send company verification email

omBratteng:
Send email verification.
Also added a check to see if the email is already verified, to not generate new codes and sending emails. Edge case, but could be if user has verification open in two windows, verifies in one, then clicks re-send code in another windows.

Required to be merged before https://github.com/dailydotdev/apps/pull/3478

AS-536

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -57,6 +57,7 @@ import {
   portfolioLimit,
   redditSocialUrlMatch,
   roadmapShSocialUrlMatch,
+  sendEmail,
   socialUrlMatch,
   stackoverflowSocialUrlMatch,
   threadsSocialUrlMatch,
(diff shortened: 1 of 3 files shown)


rebelchris (approved):
Looks good to me.
We can also merge this and test it on preview for daily.dev users :)

__tests__/users.ts
```diff
@@ -1159,12 +1169,62 @@ describe('user company', () => {
...
+      expect(sendEmail).toHaveBeenCalledTimes(1);
+      expect(sendEmail).toHaveBeenCalledWith({
+        to: 'u1@com4.com',
+        message_data: {
+          code: expect.any(String),
+        },
+        identifiers: {
+          id: loggedUser,
+        },
+        transactional_message_id: '51',
```
capJavert, line 1201: Use enum value here in case its changed so test dont fail?

src/schema/users.ts
```diff
@@ -1759,6 +1765,18 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         });
       }
 
+      await sendEmail({
+        transactional_message_id:
+          CioTransactionalMessageTemplateId.VerifyCompany,
```
capJavert, line 1770: We don't need sender/subject for email?

capJavert (changes requested):

__tests__/users.ts
omBratteng: Added

src/schema/users.ts
capJavert: I think we need `send_to_unsubscribed: true` so verification is sent to unsubscribed users as well.

omBratteng: Oh, yeah. Added

capJavert (approved):

PR #2177, opened 2024-08-26
fix: add email validation

rebelchris:
Validate email on API side.
FE can already handle the error:

Screenshot 2024-08-26 at 15 52 48

### DIFF
--- a/src/common/object.ts
+++ b/src/common/object.ts
@@ -55,4 +55,6 @@ export const nameRegex = new RegExp(/^(.){1,60}$/);
 export const socialHandleRegex = new RegExp(/^@?([\w-]){1,39}$/i);
 export const handleRegex = new RegExp(/^@?[a-z0-9](\w){2,38}$/i);
 export const descriptionRegex = new RegExp(/^[\S\s]{1,250}$/);
-export const emailRegex = new RegExp(/^[\w-.]+@([\w-]+\.)+[\w-]{2,4}$/g);
+export const emailRegex = new RegExp(
+  /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
+);
--- a/src/common/object.ts
+++ b/src/common/object.ts
@@ -55,6 +55,7 @@ export const nameRegex = new RegExp(/^(.){1,60}$/);
 export const socialHandleRegex = new RegExp(/^@?([\w-]){1,39}$/i);
 export const handleRegex = new RegExp(/^@?[a-z0-9](\w){2,38}$/i);
 export const descriptionRegex = new RegExp(/^[\S\s]{1,250}$/);
+// Originated from: https://emailregex.com
 export const emailRegex = new RegExp(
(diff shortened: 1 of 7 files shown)


src/common/object.ts
```diff
@@ -55,3 +55,4 @@ export const nameRegex = new RegExp(/^(.){1,60}$/);
 export const socialHandleRegex = new RegExp(/^@?([\w-]){1,39}$/i);
 export const handleRegex = new RegExp(/^@?[a-z0-9](\w){2,38}$/i);
 export const descriptionRegex = new RegExp(/^[\S\s]{1,250}$/);
+export const emailRegex = new RegExp(/^[\w-.]+@([\w-]+\.)+[\w-]{2,4}$/g);
```
omBratteng, line 58: Basic email, wouldn't support emails like `name+site@daily.dev`?

rebelchris: Should be ok for this purpose right?

omBratteng: Also wouldn't support `name@domain.solutions` (longer TLDs)

omBratteng: https://emailregex.com

capJavert: https://github.com/colinhacks/zod/blob/8552233c77426f77d3586cc877f7aec1aa0aa45b/src/types.ts#L599-L600

src/common/object.ts
```diff
@@ -55,3 +55,6 @@ export const nameRegex = new RegExp(/^(.){1,60}$/);
 export const socialHandleRegex = new RegExp(/^@?([\w-]){1,39}$/i);
 export const handleRegex = new RegExp(/^@?[a-z0-9](\w){2,38}$/i);
 export const descriptionRegex = new RegExp(/^[\S\s]{1,250}$/);
+export const emailRegex = new RegExp(
```
omBratteng, line 58: Maybe add comment about origin of the regex, so that a poor soul doesn't stumble upon this and wonder wtf is this 😂🙈

omBratteng (approved):

src/common/object.ts
```diff
@@ -55,3 +55,7 @@ export const nameRegex = new RegExp(/^(.){1,60}$/);
 export const socialHandleRegex = new RegExp(/^@?([\w-]){1,39}$/i);
 export const handleRegex = new RegExp(/^@?[a-z0-9](\w){2,38}$/i);
 export const descriptionRegex = new RegExp(/^[\S\s]{1,250}$/);
+// Originated from: https://emailregex.com
+export const emailRegex = new RegExp(
+  /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
+);
```
capJavert, line 61: The above regex supports things like `user/@example.com` or `user*@example.com`. While that is technically valid according to RFC I would prefer the zod one since it is more aligned with what you usually get in web applications. We don't have to support each nuance.

capJavert: I would find it weird to have above emails in the db and common email providers do not support it either way.
image

rebelchris: Moved it :)

src/schema/users.ts
```diff
@@ -1731,6 +1731,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       if (!email?.length) {
         throw new ValidationError('Email is required');
       }
+      if (!email!.match(emailRegex)) {
+        throw new ValidationError('Invalid email');
+      }
```
capJavert, line 1736: sorry for raising multiple times, but just remembered that we had an user that tried pushing really long payloads to fields, maybe we can also add some length check of like 200 characters. Our longest email in the DB is 71 chars.

rebelchris: Done in a very basic way for now

capJavert (approved):

PR #2178, opened 2024-08-26
feat: create cron to cleanup unverified work emails

omBratteng:
Auto-cleanups un-verified work emails that are older than one hour.

AS-538

### DIFF
--- a/__tests__/cron/cleanZombieUserCompany.ts
+++ b/__tests__/cron/cleanZombieUserCompany.ts
@@ -1,0 +1,60 @@
+import { cleanZombieUserCompany as cron } from '../../src/cron/cleanZombieUserCompany';
+import { saveFixtures } from '../helpers';
+import { expectSuccessfulCron } from '../helpers';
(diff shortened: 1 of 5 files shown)


rebelchris (approved):
Looks good to me :)

__tests__/cron/cleanZombieUserCompany.ts
```diff
@@ -0,0 +1,60 @@
...
+    {
+      userId: '1',
+      verified: true,
+      email: 'u1@com1.com',
+      code: '123',
+    },
+  ]);
+});
+
+describe('cleanZombieUserCompany', () => {
```
capJavert, line 29: I see you registered it but add `should be registered` test as well (example in daily digest cron)

capJavert (approved):
🤘

__tests__/cron/cleanZombieUserCompany.ts
omBratteng: Added registered test too

PR #2181, opened 2024-08-27
chore: try to prevent circular import in the future

rebelchris:
Couldn't do all due to `ChildEntity` usage, but this should prevent a lot of circular dep issues in the future already :)


sshanzel (approved):
Wow, this is quite a trick.

Also, great work on minimizing the content for the User entity.

omBratteng (approved):

capJavert (commented):
Nice work on this!

PR #2182, opened 2024-08-28
refactor: create/edit squad mutations

sshanzel:
## Note: done and approved, just awaiting the platform side of things for not including public Squads in the feed unless the user is a member or the Squad is featured.

- Create squad mutation changes.
  - Allow Squad to be public from creation.
  - Accept category id.
- Edit squad mutation changes.
  - Allow Squads to turn to public without approvals.
  - Accept category id.

**Note:** before we merge this, let's coordinate the change needed to not show public Squads (non-featured) in the feed when the user is not part of them.

MI-492 #done


sshanzel:
@dailydotdev/web-team this should be good for review 🚀

nensidosari (approved):
looks good

src/schema/sources.ts
```diff
@@ -1705,18 +1736,14 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-      if (!isNullOrUndefined(isPrivate) && current.private !== isPrivate) {
-        const approved = await ctx.con
-          .getRepository(SquadPublicRequest)
-          .findOne({
-            where: { sourceId, status: SquadPublicRequestStatus.Approved },
-          });
-
-        if (!approved) {
-          throw new ValidationError(SourceRequestErrorMessage.SQUAD_INELIGIBLE);
+      if (!isNullOrUndefined(isPrivate)) {
```
capJavert, line 1739: Can we maybe just case `isPrivate` to always boolean like `const isPrivate = !!args.isPrivate` and then we don't need so much nesting of `if` statements in this and `create` mutation.

capJavert (approved):
Looks good, one comment for potential simplification.

src/schema/sources.ts
sshanzel: Yeah, why not. Less indents are always welcome.

sshanzel: A test failed, I think it is due to the fact that the parameter is not required and that if you send undefined, we are explicitly setting the value, and could be the opposite of what state it needs to be.

PR #2187, opened 2024-08-28
chore: adjust digest limits and ack count

capJavert:
We also ack only single message per worker :hidethepain:

put it to 25 now because this was a ack/s ration for pods so I think each pod should be able to handle it
image


omBratteng (approved):
Let's see if it works 🙈

PR #2188, opened 2024-08-28
feat: download and store code snippets

capJavert:
- [x] download and insert code snippets inside the postUpdated worker
- [x] script to backfill code snippets into PostCodeSnippets entity
- [x] migrations
- [x] tests

Will add query for snippets in following PR.

Verified locally by pulling few snippets from prod. Need to merge this so I can back import with created script.

AS-496

### DIFF
--- a/bin/insertCodeSnippets.ts
+++ b/bin/insertCodeSnippets.ts
@@ -1,0 +1,51 @@
+import fastq from 'fastq';
+import '../src/config';
+import createOrGetConnection from '../src/db';
+import { Post } from '../src/entity';
+import { insertCodeSnippets } from '../src/common/post';
+
+type SnippetRow = { postId: string; filePath: string };
+
+const main = async () => {
+  const queueConcurrency = +process.env.QUEUE_CONCURRENCY || 1;
+  const con = await createOrGetConnection();
+
+  const stream = await con
+    .getRepository(Post)
+    .createQueryBuilder()
+    .select('id', 'postId')
+    .addSelect(`"contentMeta"->>'stored_code_snippets'`, 'filePath')
+    .where(`("contentMeta"->>'stored_code_snippets') is not null`)
+    .stream();
+
+  let insertCount = 0;
+  const insertQueue = fastq.promise(
+    async ({ postId, filePath }: SnippetRow) => {
+      await insertCodeSnippets({
+        entityManager: con.manager,
+        post: {
+          id: postId,
+        },
+        codeSnippetsUrl: filePath,
+      });
+
+      insertCount += 1;
+    },
+    queueConcurrency,
+  );
+
+  stream.on('data', ({ postId, filePath }: SnippetRow) => {
+    insertQueue.push({ postId, filePath });
+  });
+
+  await new Promise((resolve, reject) => {
+    stream.on('error', reject);
+    stream.on('end', resolve);
+  });
+  await insertQueue.drained();
+
+  console.log('insertion finished with a total of: ', insertCount);
+  process.exit();
+};
+
+main();
--- a/bin/insertCodeSnippets.ts
+++ b/bin/insertCodeSnippets.ts
@@ -1,33 +1,94 @@
 import fastq from 'fastq';
 import '../src/config';
-import createOrGetConnection from '../src/db';
 import { Post } from '../src/entity';
+import { downloadFile } from '../src/common/googleCloud';
+import { readdir, readFile, writeFile } from 'fs/promises';
+import createOrGetConnection from '../src/db';
+import path from 'path';
+import { PostCodeSnippetJsonFile } from '../src/types';
 import { insertCodeSnippets } from '../src/common/post';
 
 type SnippetRow = { postId: string; filePath: string };
 
+enum BinInsertCodeSnippetOperation {
+  Download = 'download',
+  Insert = 'insert',
+}
+
 const main = async () => {
   const queueConcurrency = +process.env.QUEUE_CONCURRENCY || 1;
+  const operation = process.argv[2];
+  const savePath = process.argv[3];
+
+  if (
+    !Object.values(BinInsertCodeSnippetOperation).includes(
+      operation as BinInsertCodeSnippetOperation,
+    )
+  ) {
+    throw new Error(`Invalid operation: ${operation}`);
+  }
+
+  if (!savePath) {
+    throw new Error('Path is required');
+  }
+
   const con = await createOrGetConnection();
 
-  const stream = await con
-    .getRepository(Post)
-    .createQueryBuilder()
-    .select('id', 'postId')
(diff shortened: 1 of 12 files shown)


src/common/post.ts
```diff
@@ -252,3 +257,68 @@ export const validatePost = (
...
+  try {
+    const codeSnippetsJson = await downloadJsonFile<PostCodeSnippetJsonFile>({
+      url: codeSnippetsUrl,
+    });
+
+    const codeSnippets = codeSnippetsJson.snippets.reduce(
+      (acc, codeSnippetContent, index) => {
+        const checksum = createHash('sha1');
+        checksum.update(codeSnippetContent);
+        const contentHash = checksum.digest('hex');
```
capJavert, line 283: We use snippet content hash to uniquely identify each snippet. This is because snippets order and presence in article can change at any time when post is updated (eg. author changes or removes snippets). I use `sha1` because of speed and I think we don't need sha256 for security.

src/common/post.ts
```diff
@@ -252,3 +257,68 @@ export const validatePost = (
...
+      contentHash: Not(
+        In(uniqueCodeSnippets.map((snippet) => snippet.contentHash)),
+      ),
+    });
+
+    await entityManager
+      .getRepository(PostCodeSnippet)
+      .upsert(uniqueCodeSnippets, {
+        conflictPaths: ['postId', 'contentHash'],
+      });
```
capJavert, line 315: We always cleanup snippets that are no longer present in article and then upsert the new ones.

src/common/post.ts
```diff
@@ -252,3 +257,80 @@ export const validatePost = (
...
+          }),
+        );
+      }
+
+      return acc;
+    },
+    new Map<string, PostCodeSnippet>(),
+  );
+
+  const uniqueCodeSnippets = Array.from(codeSnippets.values());
```
rebelchris, line 293: We also have uniqueifyArray not sure it works here :)

src/entity/posts/Post.ts
```diff
@@ -12,6 +12,7 @@ import { PostTag } from '../PostTag';
 import { PostKeyword } from '../PostKeyword';
 import { User } from '../user';
 import { PostRelation } from './PostRelation';
+import type { PostCodeSnippet } from './PostCodeSnippet';
```
rebelchris, line 15: 😎

src/common/post.ts
```diff
@@ -252,3 +257,80 @@ export const validatePost = (
...
+  post: Pick<Post, 'id'>;
+  codeSnippetsUrl: string | undefined;
+}) => {
+  try {
+    if (!codeSnippetsUrl) {
+      return;
+    }
+
+    const codeSnippetsJson = await downloadJsonFile<PostCodeSnippetJsonFile>({
+      url: codeSnippetsUrl,
```
rebelchris, line 324: It's one url or can be multiple?
Mainly asking because the other part is called:
`data?.meta?.stored_code_snippets,` (sounds like it could be more than one)

src/common/post.ts
```diff
@@ -252,3 +257,80 @@ export const validatePost = (
...
+  const codeSnippets = codeSnippetsJson.snippets.reduce(
+    (acc, codeSnippetContent, index) => {
+      const checksum = createHash('sha1');
+      checksum.update(codeSnippetContent);
+      const contentHash = checksum.digest('hex');
+
+      if (!acc.has(contentHash)) {
+        acc.set(
+          contentHash,
+          entityManager.getRepository(PostCodeSnippet).create({
```
rebelchris, line 279: We basically push just plain language for now?

rebelchris (changes requested):
Mostly validation questions :)

src/common/post.ts
```diff
@@ -252,3 +257,80 @@ export const validatePost = (
...
+  );
+
+  const uniqueCodeSnippets = Array.from(codeSnippets.values());
+
+  await entityManager.getRepository(PostCodeSnippet).delete({
+    postId: post.id,
+    contentHash: Not(
+      In(uniqueCodeSnippets.map((snippet) => snippet.contentHash)),
+    ),
+  });
```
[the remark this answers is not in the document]
capJavert: no because there could be code snippets that are removed from the article, see https://github.com/dailydotdev/daily-api/pull/2188#discussion_r1736384035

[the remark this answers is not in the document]
rebelchris: cool cool :)

src/common/post.ts
capJavert: it works over set with primitive values and I am doing objects here, I can extend that function to support objects as well?

src/common/post.ts
capJavert: yes, only one URL pointing to file that contains all code snippets, I was confused as well but confirmed with Denis.

src/common/post.ts
capJavert: yes, if we ever had this info or want to process somehow on the API we can easily insert

until then we will fallback to frontend markdown to "detect"

src/common/post.ts
rebelchris: I see :).
Hmm yeah up to you, but `uniqueifyObjectArray` would also be nice to have. think we do it more often.

bin/insertCodeSnippets.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  const con = await createOrGetConnection();
+
+  if (operation === BinInsertCodeSnippetOperation.Download) {
+    const stream = await con
+      .getRepository(Post)
+      .createQueryBuilder()
+      .select('id', 'postId')
+      .addSelect(`"contentMeta"->>'stored_code_snippets'`, 'filePath')
+      .where(`("contentMeta"->>'stored_code_snippets') is not null`)
```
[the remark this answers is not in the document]
capJavert: since we are adding processing to workers this should not be needed to re-run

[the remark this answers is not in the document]
rebelchris: Ah you expect this to run all post in one go?
(if so it's fine, but if it's batched you need some batching right?) (could also be on date for instance)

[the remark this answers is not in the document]
capJavert: its ~8k rows so I think batching with fastq as always should be good, we did bigger amounts with it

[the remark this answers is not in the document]
capJavert: basically two instances of backfill:
- download the code snippet files locally
- insert the snippets into the DB from local files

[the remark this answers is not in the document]
rebelchris: Ah no then it's perfect, more if it was going to be a once-off or not.
In that case let's leave as is :)

rebelchris (approved):
Approving

src/common/post.ts
capJavert: fc2466fb445097bb5b5cec13fdf1ab4696cf1712

PR #2194, opened 2024-08-29
chore: limit query to 10 but return only 3

rebelchris:
To force the index we need to query with limit 10, but only take 3 of the results as response.

### DIFF
--- a/src/schema/search.ts
+++ b/src/schema/search.ts
@@ -458,7 +458,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         )
         .orderBy('u.reputation', 'DESC')
         .andWhere(whereVordrFilter('u'))
-        .limit(getSearchLimit({ limit: Math.min(limit, 10) }));
+        .limit(getSearchLimit({ limit: Math.max(limit, 10) }));
       const hits = await searchQuery.getRawMany();
       return {
         query,


src/schema/search.ts
```diff
@@ -231,7 +231,7 @@ export const typeDefs = /* GraphQL */ `
       """
       Maximum number of users to return
       """
-      limit: Int = ${defaultSearchLimit}
```
capJavert, line 234: Would this not be as is and then we just inject 10 if actual limit < 10

src/schema/search.ts
```diff
@@ -460,10 +460,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         .andWhere(whereVordrFilter('u'))
         .limit(getSearchLimit({ limit }));
```
capJavert, line 461:
```suggestion
        .limit(getSearchLimit({ limit: Math.min(limit, 10) }));
```

src/schema/search.ts
```diff
@@ -460,10 +460,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       const hits = await searchQuery.getRawMany();
-
       return {
         query,
-        hits,
+        hits: hits.slice(0, 3),
```
capJavert, line 465:
```suggestion
        hits: hits.slice(0, limit),
```

src/schema/search.ts
rebelchris: yeah would also work :) 
Doubt we actually ever want to override the limit though.

src/schema/search.ts
```diff
@@ -231,7 +231,7 @@ export const typeDefs = /* GraphQL */ `
       """
       Maximum number of users to return
       """
-      limit: Int = ${defaultSearchLimit}
+      limit: Int = 10
```
rebelchris, line 234:
```suggestion
      limit: Int = ${defaultSearchLimit}
```

src/schema/search.ts
```diff
@@ -422,7 +422,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     },
     searchUserSuggestions: async (
       source,
-      { query, limit }: SearchSuggestionArgs,
+      { query, limit = 10 }: SearchSuggestionArgs,
```
rebelchris, line 425:
```suggestion
      { query, limit }: SearchSuggestionArgs,
```

src/schema/search.ts
```diff
@@ -458,12 +458,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         )
         .orderBy('u.reputation', 'DESC')
         .andWhere(whereVordrFilter('u'))
-        .limit(getSearchLimit({ limit }));
+        .limit(getSearchLimit({ limit: Math.min(limit, 10) }));
```
capJavert, line 461: Math.max I made a typo 😆

capJavert (approved):

PR #2195, opened 2024-08-30
chore: revert scaling to 25 to avoid hitting other services as much

capJavert:
- also lower dead letter to 3 to avoid many respools since probably something is down if it fails more then few times


rebelchris (approved):
Should we also do the 2 hour change?

omBratteng (approved):

capJavert:
@rebelchris it would be a bit more change with tests as well so I think for now just do this.

PR #2196, opened 2024-08-30
feat: source report entity and unified sending of reports

sshanzel:
- Introduced Source Report Entity
- Consolidated the types for reporting
- Send report mutation by supporting two types

MI-496 #done


src/common/reporting.ts
```diff
@@ -0,0 +1,123 @@
+import { DataSource } from 'typeorm';
```
sshanzel, line 1: Everything except `reportSource` was purely taken from `schema/posts`.

src/entity/PostReport.ts
```diff
@@ -14,8 +15,8 @@ export class PostReport {
   @Column({ default: () => 'now()' })
   createdAt: Date;
 
-  @Column({ length: 12 })
-  reason: string;
+  @Column({ length: 12, type: 'varchar' })
+  reason: ReportReason;
```
sshanzel, line 19: Using an `enum` a column, postgresql thinks it is a `text` type. But when you set a max length of the column, it is automatically turned into `varchar`. Using an `enum` with `length` makes the build go crazy. Specifying the type fixes it.

src/schema/posts.ts
```diff
@@ -1302,40 +1231,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-
-      if (reason === 'IRRELEVANT' && !tags?.length) {
-        throw new ValidationError('You must include the irrelevant tags!');
-      }
-
-      const added = await saveHiddenPost(ctx.con, {
-        userId: ctx.userId,
-        postId: id,
-      });
+      await reportPost({ ctx, id, reason, comment, tags });
```
sshanzel, line 1234: To keep backwards compatibility, we are keeping this mutation but using the new function now.

src/common/reporting.ts
```diff
@@ -0,0 +1,123 @@
...
+  ctx,
+  id,
+  reason,
+  comment,
+}: BaseReportArgs) => {
+  if (!sourceReportReasonsMap.has(reason)) {
+    throw new ValidationError('Reason is invalid');
+  }
+
+  await ctx.getRepository(Source).findOneByOrFail({ id });
```
capJavert, line 114: why empty call to this? I think `ensureSourcePermissions` already queries source under the hood.

src/entity/sources/SourceReport.ts
```diff
@@ -0,0 +1,29 @@
...
+  sourceId: string;
+
+  @PrimaryColumn({ length: 36 })
+  @Index('IDX_source_report_user_id')
+  userId: string;
+
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ length: 32, type: 'varchar' })
```
capJavert, line 18: I think we make it text lately since varchar and text are basically the same in psql.

capJavert: same for other columns

src/entity/PostReport.ts
capJavert: 👀

src/entity/sources/SourceReport.ts
sshanzel: Sure, let's update these ones to that.

sshanzel: Updated this one, but for the others, I had to keep it, the migration says it needs to drop the column first to change the type.

src/entity/common.ts
```diff
@@ -0,0 +1,41 @@
+export enum ReportEntity {
```
rebelchris, line 1: Also here, we not fixing comment reports?

src/entity/sources/SourceReport.ts
```diff
@@ -0,0 +1,29 @@
+import { Column, Entity, Index, ManyToOne, PrimaryColumn } from 'typeorm';
+import type { User } from '../user';
+import { ReportReason } from '../common';
+
+@Entity()
```
rebelchris, line 5: Do we want to go seperate route, or stick with childEntities?
(Probably was a discussion I can't recall)

src/schema/users.ts
```diff
@@ -925,6 +939,32 @@ export const typeDefs = /* GraphQL */ `
       vote: Int!
     ): EmptyResponse @auth
 
+    """
+    When a user tries to report an entity. The entity can be either post/comment/source or user when we get there
```
rebelchris, line 943: Here it mentions comment though

rebelchris (changes requested):
Mainly missing comments, but maybe I missed some coms

src/entity/common.ts
sshanzel: Should be updated now.

src/entity/sources/SourceReport.ts
sshanzel: Was near the end of the RFC, but yeah, it was decided to go similar to vote entities.

rebelchris (approved):
Looks good, just note Ante's comment

src/common/reporting.ts
sshanzel: Right. This is just redundant.

nensidosari (approved):

sshanzel:
@capJavert comments should've been addressed by now 🙏

sshanzel:
@capJavert since the comments have now been addressed, I will merge the PR now so we can move forward with this one 🙏

PR #2199, opened 2024-08-30
feat: sync user_company to CIO

rebelchris:
Stream change: https://github.com/dailydotdev/streams/pull/37

Idea is that we listen to `user_company` and if a `company` link is found/changed we sync it to CIO.
We had to introduce a hack to use v2 of the CIO api as the official client doesn't expose it and we need it to insert company with attributes basically. (see [one](https://github.com/customerio/customerio-node/issues/140#issuecomment-2315334534) and [two](https://github.com/customerio/customerio-node/issues/136#issuecomment-1670222531))

We might need to find a way to retro fix the couple companies we have now, but can happen later even. (or retrigger manually)

CIO side receives as such, which is nice because from there we can then target everyone at company x
Screenshot 2024-08-30 at 14 28 53


src/cio.ts
```diff
@@ -9,12 +9,25 @@ import {
   getShortGenericInviteLink,
 } from './common';
 import { FastifyBaseLogger } from 'fastify';
+import { UserCompany } from './entity/UserCompany';
+import { Company } from './entity/Company';
```
rebelchris, line 13:
```suggestion
import type { UserCompany } from './entity/UserCompany';
import type { Company } from './entity/Company';
```

src/common/typedPubsub.ts
```diff
@@ -13,6 +13,7 @@ import {
   pubsub,
 } from './pubsub';
 import { ContentUpdatedMessage } from '@dailydotdev/schema';
+import { UserCompany } from '../entity/UserCompany';
```
rebelchris, line 16:
```suggestion
import type { UserCompany } from '../entity/UserCompany';
```

src/workers/cdc/primary.ts
```diff
@@ -107,6 +107,7 @@ import {
   runReminderWorkflow,
 } from '../../temporal/notifications/utils';
 import { addDays } from 'date-fns';
+import { UserCompany } from '../../entity/UserCompany';
```
rebelchris, line 110:
```suggestion
import type { UserCompany } from '../../entity/UserCompany';
```

src/cio.ts
```diff
@@ -9,12 +9,25 @@ import {
...
+/**
+ * Specific client for using v2 of the Customer.io API.
+ */
+export const cioV2 = new TrackClient(
+  process.env.CIO_SITE_ID,
+  process.env.CIO_API_KEY,
+  { url: 'https://track.customer.io/api/v2' },
+);
+
+const CIO_COMPANY_OBJECT_ID = '4';
```
omBratteng, line 29: What is 4? Magic number?

rebelchris: yes, it's their internal identifier for this object.
That's why I abstracted it to a const declaration for naming it 😂

omBratteng:
```suggestion
// Magic number from customer.io
const CIO_COMPANY_OBJECT_ID = '4';
```

omBratteng: Just a suggestion to add a comment saying it's a magic number from them.

src/workers/userCompanyApprovedCio.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    if (!userCompany.companyId) {
+      return;
+    }
+
+    const company = await con
+      .getRepository(Company)
+      .findOneBy({ id: userCompany.companyId });
+    if (!company) {
+      log.info(
```
capJavert, line 24: `warn`, not sure if we are expecting it?

capJavert (approved):

src/workers/userCompanyApprovedCio.ts
rebelchris: hopefully not :)

rebelchris, line 24:
```suggestion
      log.warn(
```

capJavert: I mean it should be warn if its not supposed to happen often so we se it in logs more easily

PR #2200, opened 2024-08-30
feat: post code snippets query

capJavert:


__tests__/fixture/post.ts
```diff
@@ -397,3 +399,31 @@ export const contentUpdatedPost = {
...
+  postId,
+}: {
+  postId: string;
+}): DeepPartial<PostCodeSnippet>[] => [
+  {
+    postId,
+    contentHash: '865cd55417af5a27ab17ae1fff7510c7acc4f250',
+    order: 0,
+    language: PostCodeSnippetLanguage.Plain,
+    content: 'console.log("Hello World")',
```
rebelchris, line 413: Not related to this specific PR, but we are sanitizing what we crawl right before storing?

rebelchris (approved):

__tests__/fixture/post.ts
capJavert: We are not doing any processing on snippets on API side. What would you sanitize?

rebelchris: More wondering if someone could have malicious snippets that on insert in pg break anything?

capJavert: we are using typeorm and INSERT query it does is parameterized so it should be good

```sql
INSERT INTO "public"."post_code_snippet"("postId", "contentHash", "order", "language", "createdAt", "content") VALUES ($1, $2, $3, DEFAULT, DEFAULT, $4), ($5, $6, $7, DEFAULT, DEFAULT, $8)
```

rebelchris: ah perfect :)

PR #2204, opened 2024-09-02
fix: removed time checks on streak recover;

ilasw:


sshanzel:
I think it would be better to write a test that fails to verify this than removing the checks.

src/workers/cdc/primary.ts
```diff
@@ -822,16 +821,11 @@ const setRestoreStreakCache = async (
...
-
-  if (!shouldAllow) {
-    return;
-  }
 
   const key = generateStorageKey(StorageTopic.Streak, StorageKey.Reset, userId);
+  const today = new Date();
   const now = today.getTime();
-  const nextDay = addDays(now, 1).setHours(0, 0, 0, 0);
+  const nextDay = addDays(now, 1).setHours(23, 59, 59, 0);
```
sshanzel, line 828: I think this is not accurate due to:
- I have read on Monday.
- Missed on Tuesday.
- Streak recovery is triggered.
- With this config, it means you can recover until Thursday 23:59 which is not correct anymore (iirc). It should only last until the day the streak was lost.

sshanzel (changes requested):
Kindly check the comments.

src/workers/cdc/primary.ts
capJavert: Had similar comment, I think it requires a bit of investigation before removing, but @sshanzel you probably have the best view of everything and how it functions.

sshanzel: Yeah, that's why I think it would've helped if we were able to check redis, but now I will try to investigate any existing one, and if we have any data in our redis cache.

ilasw: Right, miss calculated days!

ilasw:
@sshanzel current tests should already cover the actual behaviour. 
The only check needed is if the uses pass from 3+ to 0 need to set a restore entry in redis and the negative case. Wdyt?

sshanzel:
We still can't merge this as it would mean we will show the popup even for ineligible users. Meaning, they would see the popup, but clicking the button would just result to an error saying they cant which might just result to more frustration. I think it is better to put logging instead (I will add it after doing further investigation).

ilasw:
Closing as we discussed

PR #2205, opened 2024-09-02
feat: rate limit company verification requests

capJavert:
5 requests per hour should be more then enough for anyone trying not to abuse


rebelchris (approved):
thanks :D

PR #2209, opened 2024-09-02
fix: streak reset time calculation

sshanzel:
Root cause: the column was `timestamptz`, so we should not use the debezium time function. Returned value was string.

Still added some logging for debugging purposes. This should only log when the user streak was greater than 2 and turned to 0.

### DIFF
--- a/__tests__/workers/cdc/primary.ts
+++ b/__tests__/workers/cdc/primary.ts
@@ -3403,8 +3403,8 @@ describe('user streak change', () => {
     it('should set cache of previous streak even when weekend had passed if it has only been 2 valid days', async () => {
       jest
         .useFakeTimers({ doNotFake })
-        .setSystemTime(new Date('2024-06-24T12:00:00')); // Monday
-      const lastViewAt = new Date('2024-06-20T12:00:00'); // previous Thursday
+        .setSystemTime(new Date('2024-09-02T12:00:00')); // Monday
+      const lastViewAt = new Date('2024-08-29T12:00:00'); // previous Thursday
       const after: ChangeObject<ObjectType> = {
         ...base,
         lastViewAt: lastViewAt.getTime() * 1000,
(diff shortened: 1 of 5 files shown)


__tests__/workers/cdc/primary.ts
```diff
@@ -3259,8 +3259,8 @@ describe('user streak change', () => {
     currentStreak: 2,
     totalStreak: 3,
     maxStreak: 4,
-    lastViewAt: new Date().getTime(),
-    updatedAt: new Date().getTime(),
+    lastViewAt: new Date().toISOString() as never,
+    updatedAt: new Date().toISOString() as never,
```
rebelchris, line 3263: Shouldn't it cast to `string` in this case?
Or TS doesn't like it?

src/common/users.ts
```diff
@@ -9,16 +9,13 @@ import {
...
-  getTimezonedEndOfISOWeek,
-  debeziumTimeToDate,
-} from './utils';
+import { getTimezonedStartOfISOWeek, getTimezonedEndOfISOWeek } from './utils';
 import { GraphQLResolveInfo } from 'graphql';
 import { utcToZonedTime } from 'date-fns-tz';
 import { sendAnalyticsEvent } from '../integrations/analytics';
 import { DayOfWeek, DEFAULT_WEEK_START } from './date';
 import { ChangeObject, ContentLanguage } from '../types';
+import { logger } from '../logger';
```
rebelchris, line 18: Where'd this guy come from?

rebelchris (approved):
Looks alright overall :)

__tests__/workers/cdc/primary.ts
sshanzel: TS doesn't like it.

src/common/users.ts
sshanzel: Oh, the logging was not pushed.

capJavert (approved):
Looks ok changes wise, trust you debugged it from feature perspective, since you have most knowledge of the edge cases.

PR #2210, opened 2024-09-02
feat: add circuit breaker for feed client

capJavert:
- circuit breaker implementation for FeedClient
- can be injected into other feed services like Lofn
- can also be used outside of classes/clients as regular instance/singleton
- possibly we can expose more options and different types of [breaker strategies](https://github.com/connor4312/cockatiel?tab=readme-ov-file#circuitbreakerpolicy--halfopenafter-breaker-initialstate-)

TODO
- [x] check tests (disable breaker in tests)
- [x] return specific error when circuit is broken
- [x] adjust options to match volume expected for feed service

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "apollo-server-errors": "^3.3.0",
     "apollo-server-types": "^3.8.0",
     "cloudinary": "^1.41.3",
+    "cockatiel": "^3.2.1",
     "csv-parse": "^5.5.6",
     "customerio-node": "^4.1.1",
     "dataloader": "^2.2.2",
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -101,6 +101,8 @@ export const removeEmptyValues = <T>(array: T[]): T[] => array.filter(Boolean);
  */
 export const isProd = process.env.NODE_ENV === 'production';
 
+export const isTest = process.env.NODE_ENV === 'test';
+
 export const parseDate = (date: string | Date): Date | undefined => {
   if (!date) {
     return undefined;
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,6 +31,7 @@ import { loadFeatures } from './growthbook';
 import { runInRootSpan } from './telemetry';
 import { loggerConfig } from './logger';
 import { getTemporalClient } from './temporal/client';
+import { BrokenCircuitError } from 'cockatiel';
 
 type Mutable<Type> = {
   -readonly [Key in keyof Type]: Type[Key];
@@ -181,6 +182,11 @@ export default async function app(
                   { body: ctx?.reply?.request?.body },
                   'unknown query',
                 );
+              } else if (error.originalError instanceof BrokenCircuitError) {
+                newError.message = 'Garmr broken error';
+                newError.extensions = {
+                  code: 'GARMR_BROKEN_ERROR',
+                };
               } else if (!error.extensions?.code) {
                 app.log.warn(
                   {
--- a/src/integrations/feed/clients.ts
+++ b/src/integrations/feed/clients.ts
@@ -3,7 +3,7 @@ import { RequestInit } from 'node-fetch';
 import { fetchOptions as globalFetchOptions } from '../../http';
 import { retryFetchParse } from '../retry';
 import { GenericMetadata } from '../lofn';
-import { GarmrService, IGarmrClient } from '../garmr';
+import { GarmrNoopService, IGarmrClient, IGarmrService } from '../garmr';
 
 type RawFeedServiceResponse = {
   data: { post_id: string; metadata: Record<string, string> }[];
@@ -20,25 +20,16 @@ type FeedFetchOptions = RequestInit & {
 export class FeedClient implements IFeedClient, IGarmrClient {
   private readonly url: string;
   private readonly fetchOptions: FeedFetchOptions;
-  private static readonly garmr = new GarmrService({
-    service: FeedClient.name,
-    breakerOpts: {
-      halfOpenAfter: 3 * 1000,
-      threshold: 0.1,
-      duration: 30 * 1000,
-    },
-  });
-
-  get garmr() {
-    return FeedClient.garmr;
-  }
+  readonly garmr: IGarmrService;
 
   constructor(
     url = process.env.INTERNAL_FEED,
     fetchOptions: FeedFetchOptions = globalFetchOptions,
+    garmr: IGarmrService = new GarmrNoopService(),
   ) {
     this.url = url;
     this.fetchOptions = fetchOptions;
+    this.garmr = garmr;
   }
 
   async fetchFeed(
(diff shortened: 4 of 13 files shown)


src/integrations/garmr.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export class GarmrService {
+  instance: IPolicy;
+
+  constructor({
+    handler,
+    breakerOpts,
+    retryOpts,
+    limits,
+  }: {
```
capJavert, line 22: Due to library exposing different options for retries, circuit breakers and concurrent requests handling decided it would be best to put all options under single service.

Garmr or Garm is a wolf or dog associated and described as a blood-stained guardian of Hel (where errors go to die).

src/integrations/feed/clients.ts
```diff
@@ -34,15 +46,18 @@ export class FeedClient implements IFeedClient {
...
-    );
+    const res = await this.garmr.execute(() => {
+      return retryFetchParse<RawFeedServiceResponse>(
+        this.url,
+        {
+          ...this.fetchOptions,
+          method: 'POST',
+          body: JSON.stringify(config),
+        },
+        { retries: 0 },
```
capJavert, line 57: 0 retries now that Garmr handles them. If retries where left to our wrapper they would actually happen internally and retry no matter if breaker was active or not.

src/integrations/garmr.ts
```diff
@@ -0,0 +1,69 @@
...
+    fn: (context: IDefaultPolicyContext) => PromiseLike<T> | T,
+    signal?: AbortSignal,
+  ): Promise<T> {
+    return this.instance.execute(fn, signal);
+  }
+}
+
+export interface IGarmrClient {
+  garmr: GarmrService;
+}
```
capJavert, line 69: Just a common interface for better types and internal usage of Garmr.

src/integrations/feed/clients.ts
```diff
@@ -16,9 +17,20 @@ type FeedFetchOptions = RequestInit & {
 /**
  * Naive implementation of a feed client that fetches the feed from the feed service
  */
-export class FeedClient implements IFeedClient {
+export class FeedClient implements IFeedClient, IGarmrClient {
   private readonly url: string;
   private readonly fetchOptions: FeedFetchOptions;
+  private static readonly garmr = new GarmrService({
```
capJavert, line 23: Static so it is the same instance across all FeedClient instances, important so that breaker stats are consolidated across all calls.

package.json
```diff
@@ -80,6 +80,7 @@
     "apollo-server-errors": "^3.3.0",
     "apollo-server-types": "^3.8.0",
     "cloudinary": "^1.41.3",
+    "cockatiel": "^3.2.1",
```
capJavert, line 83: Decided to use this lib at the end since https://github.com/nodeshift/opossum has quite worse types and options wise is quite limited. For example it would not work with instance based clients as we have now since it requires a function to be passed at the moment of breaker initialization.

src/integrations/feed/clients.ts
```diff
@@ -16,9 +17,20 @@ type FeedFetchOptions = RequestInit & {
...
+    breakerOpts: {
+      halfOpenAfter: 3 * 1000,
+      threshold: 0.1,
+      duration: 30 * 1000,
```
capJavert, line 27: defaults in line what platform projects do, [see](https://dailydotdev.slack.com/archives/C07KEKPVBJ9/p1725281135140569?thread_ts=1725280854.868329&cid=C07KEKPVBJ9), we can tune them as we see fit TBD

halfOpenAfter - time after which circuit goes to half open state and tests service health 
threshold - percentage of requests that need to fail for breaker to open
duration - rolling window in which threshold is counted

src/integrations/garmr.ts
```diff
@@ -0,0 +1,69 @@
...
+      threshold: number;
+      duration: number;
+    };
+    retryOpts?: {
+      maxAttempts: number;
+      timeout: number;
+    };
+    limits?: {
+      maxRequests: number;
+      queuedRequests?: number;
```
capJavert, line 35: Max in progress requests that breaker can handle, eg. if set to 100 on the 101th request that happened in parallel it will break.

src/integrations/garmr.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const circuitBreakerPolicy = circuitBreaker(handler ?? handleAll, {
+      halfOpenAfter: breakerOpts.halfOpenAfter,
+      breaker: new SamplingBreaker({
+        threshold: breakerOpts.threshold,
+        duration: breakerOpts.duration,
+      }),
+    });
+
+    this.instance = wrap(retryPolicy, circuitBreakerPolicy, bulkheadPolicy);
```
capJavert, line 56: Merges all policies into single one, like func compose

vpol:
I can't review the code side, but I'm ok in the logic.

vpol (approved):

src/common/utils.ts
```diff
@@ -101,6 +101,8 @@ export const removeEmptyValues = <T>(array: T[]): T[] => array.filter(Boolean);
  */
 export const isProd = process.env.NODE_ENV === 'production';
 
+export const isTest = process.env.NODE_ENV === 'test';
```
rebelchris, line 104: Oh gosh noooo, i see we did this manual until now 🤦 
https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcSHuZxrhDNbyZzn3YZ4g-0ZzIhXMiUpj3LBnQ&s

src/index.ts
```diff
@@ -181,6 +182,11 @@ export default async function app(
                   { body: ctx?.reply?.request?.body },
                   'unknown query',
                 );
+              } else if (error.originalError instanceof BrokenCircuitError) {
+                newError.message = 'Garmr broken error';
+                newError.extensions = {
+                  code: 'GARMR_BROKEN_ERROR',
```
rebelchris, line 188: 🧑‍🍳

rebelchris (approved):
Looks good to me :) 
I'll pull this locally as well to validate my FE changes

src/common/utils.ts
capJavert: not many places, but yeah, for circuit breaker it makes sense to avoid it in tests

src/integrations/garmr.ts
```diff
@@ -0,0 +1,103 @@
...
+  noop,
+  Policy,
+  retry,
+  SamplingBreaker,
+  wrap,
+} from 'cockatiel';
+import { logger } from '../logger';
+import { isTest } from '../common';
+
+export class GarmrService {
```
idoshamun, line 18: We need metrics here, not just logs

src/integrations/feed/clients.ts
capJavert: we still one reused breaker instances but since we want different configurations per different FeedClients moved it to constructor param

src/integrations/garmr.ts
capJavert: Added metrics 7a5edc92c22e460fbc4503d7c8f4b73bf3fee5d7

PR #2211, opened 2024-09-02
feat: add source request rpc

rebelchris:
Relient on: https://github.com/dailydotdev/schema/pull/33

Introduces a new RPC endpoint to add sourceRequest via a service (for now hardcoded to yggdrasil)
(SourceRequest is not a linked table so just raw insert works fine :))

### DIFF
--- a/src/routes/privateRpc.ts
+++ b/src/routes/privateRpc.ts
@@ -131,6 +131,9 @@ export default function (router: ConnectRouter) {
 
       try {
         req.url = standardizeURL(req.url);
+        if (!isValidHttpUrl(req.url)) {
+          throw new ConnectError('invalid url', Code.InvalidArgument);
+        }
 
         const sourceRequest = await con.getRepository(SourceRequest).insert({
           sourceUrl: req.url,
--- a/src/routes/privateRpc.ts
+++ b/src/routes/privateRpc.ts
@@ -1,6 +1,6 @@
 import { Code, ConnectError, ConnectRouter } from '@connectrpc/connect';
 import { TypeOrmError, TypeORMQueryFailedError } from '../errors';
-import { ArticlePost } from '../entity';
+import { ArticlePost, SourceRequest } from '../entity';
 import { generateShortId } from '../ids';
 import createOrGetConnection from '../db';
 import { isValidHttpUrl, standardizeURL } from '../common/links';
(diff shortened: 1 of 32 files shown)


src/routes/privateRpc.ts
```diff
@@ -116,4 +117,45 @@ export default function (router: ConnectRouter) {
...
+
+      const originalReq = req.clone();
+      const con = await createOrGetConnection();
+
+      try {
+        req.url = standardizeURL(req.url);
+
+        const sourceRequest = await con.getRepository(SourceRequest).insert({
+          sourceUrl: req.url,
+          userId: 'yggdrasil',
```
capJavert, line 137: we have this user?

src/routes/privateRpc.ts
```diff
@@ -1,13 +1,14 @@
...
-import { ArticlePost } from '../entity';
+import { ArticlePost, SourceRequest } from '../entity';
 import { generateShortId } from '../ids';
 import createOrGetConnection from '../db';
 import { isValidHttpUrl, standardizeURL } from '../common/links';
 import { baseRpcContext } from '../common/connectRpc';
 import {
   CreatePostRequest,
   CreatePostResponse,
+  SourceRequestService,
```
capJavert, line 11: Just reminder to publish lib version

capJavert (approved):
nice!

src/routes/privateRpc.ts
rebelchris: It's not needed as sourceRequest is a raw table (see description)
(SourceRequest is not a linked table so just raw insert works fine :))

capJavert: Oh, never noticed it 😆 I would be baffled why it did not work. Ok, that its good for now, but we could add yggdrasil user as we have ghost either way (not blocking for this PR)

rebelchris: I added one first, then thought it would just be spam 😅 (and this user would grift lots of reputation I guess)

src/routes/privateRpc.ts
```diff
@@ -116,4 +117,45 @@ export default function (router: ConnectRouter) {
...
+    async (req, context) => {
+      if (!context.values.get(baseRpcContext).service) {
+        throw new ConnectError('unauthenticated', Code.Unauthenticated);
+      }
+
+      const originalReq = req.clone();
+      const con = await createOrGetConnection();
+
+      try {
+        req.url = standardizeURL(req.url);
```
capJavert, line 133: We should still validate this is defined with `isValidHttpUrl`

capJavert (approved):

PR #2213, opened 2024-09-03
fix: hide alert if streak recover no longer available;

ilasw:
As is, we hide the alert only when the user pass from 0 -> 1 in streak length. 
But if user keep not reading the banner will showing forever at every reload even if he keep click on "Recover my streak" on frontend. This will fix this issue.

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -28,6 +28,7 @@ import {
   UserStreakActionType,
   getAuthorPostStats,
   streakRecoverCost,
+  Alerts,
 } from '../entity';
 import {
   AuthenticationError,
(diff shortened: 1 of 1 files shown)


src/schema/users.ts
```diff
@@ -1983,6 +1984,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
       const oldStreakLength = await getRestoreStreakCache({ userId });
       if (!oldStreakLength) {
+        await ctx.con
+          .getRepository(Alerts)
+          .update({ userId }, { showRecoverStreak: false });
```
capJavert, line 1989: Can we add a test for this case please

ilasw: Sure! Adding shortly, thanks for suggestion!

capJavert: And also reverse test that it does not get updated when there is a valid streak.

PR #2214, opened 2024-09-03
fix: streak reset over the weekend

sshanzel:
I've added three more tests that involve the weekend to solidify the integrity of the process. Found another set of edge cases along the way, but should have been fixed by this PR now.

The edge cases were about, when it is the second day of the week, and the user lost their streak, their last read could've been Friday, Saturday, or Sunday. Have added test cases to each.

Also, to make it easier to maintain, I moved the functions to a separate file.

Ticket: https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31/backlog?epics=visible&selectedIssue=MI-517


src/common/streak.ts
```diff
@@ -0,0 +1,73 @@
...
+const getNextDay = (day: number) => (day === 6 ? 0 : day + 1);
+const getPreviousDay = (day: number) => (day === 0 ? 6 : day - 1);
+
+export const getAllowedDays = ({
+  day: dayToday,
+  firstDayOfWeek,
+  lastDayOfWeek,
+  lastView,
+}: DaysProps) => {
+  // do something
```
rebelchris, line 24: Why is this here?

src/common/streak.ts
```diff
@@ -0,0 +1,73 @@
...
+  day: number;
+  firstDayOfWeek: Day;
+  lastDayOfWeek: Day;
+  lastView: Date;
+}
+
+const getNextDay = (day: number) => (day === 6 ? 0 : day + 1);
+const getPreviousDay = (day: number) => (day === 0 ? 6 : day - 1);
+
+export const getAllowedDays = ({
```
rebelchris, line 18: Find it kind of hard to follow the logic here..
Really wondering if we shouldn't just move to:

- loose streak?
- allowed reset at that time
- set redis

It feels like way less logic

sshanzel: You can't always just allow the reset at that time. Ex:
- I read something on Monday.
- On Wednesday, my streak should reset, but I did not login, so in our system, it has not yet been reset.
- I logged in on Thursday, my streak got cleared, CDC ran, we can set to allow streak recovery as it should only be possible on Wednesday.

src/common/streak.ts
sshanzel: Just forgot to remove when I was writing the function. Should've been removed by now.

src/common/streak.ts
rebelchris: Hmm I see, I guess it's fine its just a ton of logic that is very hard to understand (IMO)

sshanzel: Once we have this verified to be working, we won't probably touch this process in the next 20 years. So I guess we just have to bear the pain for a moment.

sshanzel:
I've added all the same tests but when it is for when the user's start of the week is Sunday. Let me know if you guys want me to create a common function as their differences are the setting of dates. If you prefer to have everything hard-coded on their own, then this is it.

__tests__/workers/cdc/primary.ts
```diff
@@ -3562,6 +3674,387 @@ describe('user streak change', () => {
       expect(alert.showRecoverStreak).toEqual(true);
     });
   });
+
+  describe('streak recovery with Sunday as start of the week', () => {
```
sshanzel, line 3678: Copy-pasted all the other tests, just changed the date values.

src/common/streak.ts
idoshamun: I don't think this assumption will age well knowing how we do things. Probably we'll come up with something new

sshanzel: Ah, yes. That is true. Though this function should stand the test of having different configurations such as different start of weekdays.

rebelchris (approved):
I'll approve it simply because I don't really have a better approach.
Mostly thinking in my head if we could somehow have a weekday-day mapping function, but think it would only complicate readability.

src/common/streak.ts
```diff
@@ -0,0 +1,75 @@
...
+
+interface DaysProps {
+  day: number;
+  firstDayOfWeek: Day;
+  lastDayOfWeek: Day;
+  lastView: Date;
+}
+
+const getNextDay = (day: number) => (day === 6 ? 0 : day + 1);
+const getPreviousDay = (day: number) => (day === 0 ? 6 : day - 1);
```
capJavert, line 16: There are functions in date-fns to get previous and next day from date and then you can get the number of the day from that with `getDay` as usual, feels more bulletproof then doing this helpers.

src/common/streak.ts
capJavert: I agree its really hard to follow the logic with offsets/pluses/minuses, but makes more sense looking at the tests.

src/common/users.ts
```diff
@@ -562,6 +529,23 @@ export enum LogoutReason {
   KratosSessionAlreadyAvailable = 'kratos session already available',
 }
 
+const getAbsoluteDifferenceInDays: typeof differenceInDays = (date1, date2) => {
+  const startOfDay = (date: Date) =>
```
capJavert, line 533: There is a function in date-fns `startOfDay` that you should be use for this

capJavert (commented):
Lot of functions you create are available in our date library so might be better to use them, also you might find some others that can help in the future https://date-fns.org/docs/Getting-Started

src/common/users.ts
sshanzel: Oh, nice. Looks like they literally have almost everything date-related.

src/common/streak.ts
sshanzel: Their implementation of that is a bit different from our requirements. They need a Date to tell the next day, for us, we just need to get the next day if the day today is Saturday (6).

src/common/streak.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const getAllowedDays = ({
+  day: dayToday,
+  firstDayOfWeek,
+  lastDayOfWeek,
+  lastView,
+}: DaysProps) => {
+  const lastViewDay = lastView.getDay();
+  const dayTwo = getNextDay(firstDayOfWeek);
```
nensidosari, line 25: What does dayTwo mean, this is the first unclear thing to me?

sshanzel:
@capJavert should be updated as requested now. We just can't use their getNextDay as it functions differently from what we expect.

src/common/streak.ts
```diff
@@ -0,0 +1,75 @@
...
+    dayToday > dayTwo
+  ) {
+    return STREAK_RECOVERY_MAX_GAP_DAYS;
+  }
+
+  if (dayToday === firstDayOfWeek) {
+    return STREAK_RECOVERY_MAX_GAP_DAYS + FREEZE_DAYS_IN_A_WEEK;
+  }
+
+  const dayOneWeekend = getPreviousDay(lastDayOfWeek);
```
nensidosari, line 39: what does dayOneWeekend refer to?

__tests__/workers/cdc/primary.ts
```diff
@@ -3441,6 +3442,154 @@ describe('user streak change', () => {
...
+      );
+      const lastStreak = await getRestoreStreakCache({ userId: after.userId });
+      expect(triggerTypedEvent).toHaveBeenCalledTimes(1);
+      expect(jest.mocked(triggerTypedEvent).mock.calls[0].slice(1)).toEqual([
+        'api.v1.user-streak-updated',
+        { streak: after },
+      ]);
+      expect(lastStreak).toEqual(3);
+      const alert = await con.getRepository(Alerts).findOneBy({ userId: '1' });
+      expect(alert.showRecoverStreak).toEqual(true);
```
nensidosari, line 3479: I think for a complicated feature like this one it's better to use a describe block with tests for each of the expect test  (for each of the ones that make sense as a final result)- would make it easier for debugging in case one of these things only starts failing - like cache is set but alert is not set to true. 
This makes the title of the test a bit more clear, as for the moment it mentions only the cache but it's also doing a bunch of other things we need like setting the alert to true.

What I think would help is:

```
describe('Reset streak if it has only been 2 valid days even when less than 48 hours'){
beforeEach(... if we can do all the computations we need before here..)

it('should set cache of previous streak to ...')
it('should set alert to showRecoverStreak to true')
...
}
```

src/common/streak.ts
sshanzel: This is to make it work both on Monday and Sunday as the first weekday. `dayOneWeekend` means Saturday having Monday as the first day of the week and Friday having Sunday as the first day of the week.

src/common/streak.ts
sshanzel: Day two of the weekday, similar to the other comment, to make it work across different start of the week. Rather than having the same amount of conditions be copy-pasted to support different start of the week.

__tests__/workers/cdc/primary.ts
sshanzel: It can set one or the other only. If the cache is not set, then the alert was not set. Vice versa, so it wouldn't be too beneficial to have them separated. This would just mean we will multiply our tests by two.

src/common/streak.ts
nensidosari: but the function days getPreviousDay not get first day of week?

sshanzel: Let's say the last day of the week is Sunday if Monday is the start of the week. Then the `lastDayOfWeek` (Sunday) minus one is Saturday.

src/common/streak.ts
nensidosari: So based on the weekend, we would have here either Tuesday for a week that starts on Monday, or Monday for a week that starts on Saturday right?

I think it's clearer if we call this `secondDayOfWeek` or `weekDayTwo`

sshanzel: You mean starts on Sunday? Yes.

Sure, we can always have better variable names.

src/common/streak.ts
nensidosari: Cool then, just this suggestion on my side:

> if it means first day of weekend, let's call it that firstDayOfWeekend or weekendDayOne, it's clearer in my opinion

src/common/streak.ts
nensidosari: > You mean starts on Sunday? Yes.


yup my mistake there, just want to make sure the naming explain itself

nensidosari (approved):

src/common/streak.ts
capJavert: I see you just want to iteratate 0-6 without overflow, you can also do `(new Date().getDay() + 1) % 7` which will go to 0 when day is 6. More typical then doing a exact `=== 6` check.

sshanzel: Ah, right. Yeah, we can do that.

sshanzel: Updated the functions now.

capJavert (approved):

PR #2215, opened 2024-09-03
fix: issue with cdc and streak restore

idoshamun:


omBratteng (approved):
Lgtm, any specific tests we can add?

capJavert (approved):

rebelchris (approved):
Looks ok, not 100% sure what the issue was malformed division?

capJavert:
When diving timestamp (number) with 1000 it can be non integer

PR #2216, opened 2024-09-03
feat: garm on feed api

capJavert:
- add garm to lofn and feed clients used for api
- refactor so clients still have single options argument


rebelchris (approved):
looks good!

PR #2220, opened 2024-09-04
refactor: maxStreak calculation;

ilasw:
test: add test for maxStreak calc;


sshanzel (commented):
Also, in the mutation, we should also clear the cache after successful update. Along with the alerts proeprty being changed.

__tests__/users.ts
```diff
@@ -782,6 +783,78 @@ describe('query userStreaks', () => {
...
+        addDays(new Date(), 1).getTime(),
+      );
+
+      const { data, errors } = await client.mutate(MUTATION);
+      const { recoverStreak } = data;
+      expect(errors).toBeFalsy();
+      expect(recoverStreak).toBeTruthy();
+      expect(recoverStreak.current).toEqual(oldLength);
+      expect(recoverStreak.max).toEqual(20);
+      await deleteRedisKey(redisKey);
```
sshanzel, line 820: Instead of deleting the redis key, we should check here if it was deleted.

sshanzel (approved):
Tests looked amazing. Just update the last part and wait for Ante's approval before merging 🚀

capJavert (approved):

nensidosari (approved):

PR #2221, opened 2024-09-04
feat: send weekly digest per timezone offset

capJavert:
- send weekly digest per timezone offset same as daily digest
- should split the send over many batches throughout the day instead of single one at midnight
- should also increase weekly digest quality (article freshness) since generation is much closer to send
- currently weekly digest is still sent only on Wednesday (but this logic should support configuration change in the future)

### DIFF
--- a/.infra/crons.ts
+++ b/.infra/crons.ts
@@ -42,7 +42,7 @@ export const crons: Cron[] = [
   },
   {
     name: 'personalized-digest',
-    schedule: '15 0 * * *',
+    schedule: '15 * * * *',
     limits: {
       memory: '1Gi',
     },
--- a/src/cron/personalizedDigest.ts
+++ b/src/cron/personalizedDigest.ts
@@ -1,7 +1,8 @@
+import { addHours, startOfHour } from 'date-fns';
 import {
   DEFAULT_TIMEZONE,
   getPersonalizedDigestPreviousSendDate,
-  getPersonalizedDigestSendDate,
+  digestPreferredHourOffset,
   notifyGeneratePersonalizedDigest,
   schedulePersonalizedDigestSubscriptions,
 } from '../common';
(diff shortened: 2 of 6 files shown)


.infra/crons.ts
```diff
@@ -42,7 +42,7 @@ export const crons: Cron[] = [
   },
   {
     name: 'personalized-digest',
-    schedule: '15 0 * * *',
+    schedule: '15 * * * *',
```
capJavert, line 45: Just moved it to every hour at 15th minute for now, its gonna overlap with daily digest load as well but overall load should be smaller 🤞 

also was thinking to move it at 30th minute but as digest grows processing will just rise so they will overlap again, open to ideas

src/cron/personalizedDigest.ts
```diff
@@ -1,32 +1,44 @@
...
       .leftJoin(User, 'u', 'u.id = upd."userId"')
-      .where('upd."preferredDay" = :nextPreferredDay', {
-        nextPreferredDay,
-      })
+      .where(
+        `(EXTRACT(DAY FROM NOW() AT TIME ZONE COALESCE(NULLIF(u.timezone, ''), '${DEFAULT_TIMEZONE}')) - 1) = upd."preferredDay"`,
+        {
+          defaultTimezone: DEFAULT_TIMEZONE,
+        },
+      )
```
capJavert, line 34: basically similar how we are guessing the hour in user's timezone now we are also guessing the day of week in the timezone so we know when we need to send

`-1` is here because psql weekdays are 1-7

idoshamun (approved):

omBratteng (approved):

.infra/crons.ts
rebelchris: My take is probably we want to look eventually at a different way of processing, but not urgent maybe something to chat about during gathering?

src/cron/personalizedDigest.ts
rebelchris: I think you took `day` (1-31) instead of `dow` (0-6)
https://www.postgresql.org/docs/8.1/functions-datetime.html

It probably looked ok because it's only the 5th now.

rebelchris (changes requested):

.infra/crons.ts
capJavert: Its more of a concern if feed service can take it because on our side we can  definitely process more messages quickly.

src/cron/personalizedDigest.ts
capJavert: damn you are righh 100%, this is why we do code reviews, tnx! fixed in f4998457b2ab80cb23515f74bd133790a18f32ea

rebelchris (approved):
Looks good, maybe do we have a test that mimic like middle of month?
(as previous implementation then should have failed?) (not urgent/blocking)

PR #2223, opened 2024-09-04
fix: formatting the streak recovery date to the correct timezone

sshanzel:
When fetching the column, we need to set the right type so that the value will be in UTC format before converting it the user's configured timezone.

Similar to what we were doing in the past with user read history.
https://github.com/dailydotdev/daily-api/blob/350c878774a7838febe0d18f162f0f812c1e82f7/src/schema/users.ts#L1033

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -574,6 +574,35 @@ describe('query userStreaks', () => {
       await expectStreak(5, 5, lastViewAt);
     });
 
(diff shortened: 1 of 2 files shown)


sshanzel:
To prove the case, I reverted the commit that contained the fix. Kindly check this commit and the relevant test case for it.

https://github.com/dailydotdev/daily-api/pull/2223/commits/efd0fb87be1cfe543d3beb50a89ec21c0c34945d

nensidosari (approved):

capJavert (approved):

PR #2224, opened 2024-09-05
fix: reputation event for restoring streak

sshanzel:
Restoring for the 3rd time might introduce an issue. In this case, we should just use when the restoration happened for the target id to provide uniqueness.

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -671,412 +671,6 @@ describe('query userStreaks', () => {
     });
   });
 
-  describe('streak recovery mutation', () => {
-    const MUTATION = `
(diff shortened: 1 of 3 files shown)


capJavert (approved):

sshanzel:
Let me just add the tests before merging. We should catch the error with the tests.

sshanzel:
Added all the tests in this commit: https://github.com/dailydotdev/daily-api/pull/2224/commits/111e7d5f3d1d2def9385c8d8fb393308a1f0b69d

The other commits about the tests, I just moved it out of unnecessary describe nesting.

PR #2226, opened 2024-09-05
fix: re-enable tracing

omBratteng:


.infra/Pulumi.prod.yaml
```diff
@@ -106,7 +106,7 @@ config:
       secure: AAABAI/bDI+xz5PMlMZseI2kW5oanHk3z1KcAOZJb+/G8arhgSP1EMdqdOTk0kku1t6KTWtl9+oSSDKWYNH1CG5D+Ac1ZPFAWc/ldmEb/iU=
     onesignalAppId:
       secure: AAABALTizkxZkCfsq6/fx2v7nXxsVP7psWBfSUct0zeyrX0ABxPpD30zIsyx5r5wqt1UJIIXDIVOnhM7W3vzrzaatrE=
-    otelEnabled: false
+    otelEnabled: true
```
capJavert, line 109: What was the issue again when we disabled?

omBratteng: I think it was user search going haywire, causing lots of logs because it never really finished a span

omBratteng: It might still occur, haven't had a chance to check it locally. Enabling it so @vdubc can get traces from API

capJavert (approved):

.infra/Pulumi.prod.yaml
capJavert: Ignore the test fail, looking at it

omBratteng: aight 👍

PR #2227, opened 2024-09-05
fix: flaky digest test

capJavert:
Simulation of "timezone" in the test was not right, I was just manually calculating offset and since weekly digest takes day into account it would no longer match.

Now I set the value in proper timezone globally inside cron during test so everything executes in the said timezone.


omBratteng (approved):

rebelchris (approved):
Fakers for the win 🙏

PR #2231, opened 2024-09-06
fix: setting of cache when timezone is involved

sshanzel:
Most of our streaks data are queried and the correct timezones are fetched. However, in CDC, the data being passed is in UTC, hence, in JS we need to convert it to the right timezone for a proper calculation.

Added to tests:
- Timezone > UTC
- Timezone < UTC

Validated the entire CDC process relating to the setting of cache, every end considers timezone now.


src/common/users.ts
```diff
@@ -546,8 +546,9 @@ export const shouldAllowRestore = async (
 ) => {
   const { userId, lastViewAt: lastViewAtDb } = streak;
   const user = await con.getRepository(DbUser).findOneBy({ id: userId });
-  const today = new Date();
-  const lastView = new Date(lastViewAtDb);
+  const timezone = user.timezone || 'utc';
```
capJavert, line 549: we have `DEFAULT_TIMEZONE` const probably better to use it

src/common/users.ts
```diff
@@ -546,8 +546,9 @@ export const shouldAllowRestore = async (
...
+  const today = utcToZonedTime(new Date(), timezone);
+  const lastView = utcToZonedTime(new Date(lastViewAtDb), timezone);
```
capJavert, line 551: `lastViewAtDb` is not `Date` already?

capJavert (approved):
Looks good, some questions mostly

src/common/users.ts
sshanzel: Oh, nice, yeah. Let's use that.

src/common/users.ts
sshanzel: No, this one came from CDC. So either a number or a string. You never know :hide_the_pain:

PR #2232, opened 2024-09-09
fix: submission limit

rebelchris:
Revert submission limit back to 3, but allow infinite submissions for team members.

### DIFF
--- a/src/schema/submissions.ts
+++ b/src/schema/submissions.ts
@@ -134,32 +134,30 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       ctx: AuthContext,
       info,
     ): Promise<GQLSubmitArticleResponse> => {
-      const user = await ctx
-        .getRepository(User)
-        .findOneByOrFail({ id: ctx.userId });
-
-      const isTeamMember = await ctx.con
-        .getRepository(Feature)
-        .findOneBy({ feature: FeatureType.Team, userId: ctx.userId });
+      const [user, isTeamMember] = await Promise.all([
+        ctx.getRepository(User).findOneByOrFail({ id: ctx.userId }),
+        ctx.con
+          .getRepository(Feature)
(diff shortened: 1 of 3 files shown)


omBratteng (approved):

idoshamun (approved):

src/schema/submissions.ts
```diff
@@ -132,21 +138,28 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
 
-      if (!hasSubmissionAccess(user)) {
-        return {
-          result: 'rejected',
-          reason: SubmissionFailErrorMessage.ACCESS_DENIED,
-        };
-      }
+      const isTeamMember = await ctx.con
+        .getRepository(Feature)
+        .findOneBy({ feature: FeatureType.Team, userId: ctx.userId });
```
capJavert, line 143: We can fetch this in parallel with user?

src/schema/submissions.ts
```diff
@@ -132,21 +138,28 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-          result: 'rejected',
-          reason: SubmissionFailErrorMessage.LIMIT_REACHED,
-        };
+      if (!isTeamMember) {
+        if (!hasSubmissionAccess(user)) {
+          return {
+            result: 'rejected',
+            reason: SubmissionFailErrorMessage.ACCESS_DENIED,
+          };
+        }
```
capJavert, line 153: What if instead of nested if we have `const limit = isTeamMember ? Infinity : submissionLimit`, the rest of the logic would be the same?

Also I think we should still check `hasSubmissionAccess` even if user is a team member.

rebelchris: The check just happens on 2 sides making it a bit more tricky and it's more of a hack at the moment.
I'll check though if I can easily do it :)

PR #2234, opened 2024-09-09
fix: temporarily disable trace

omBratteng:
I see there's lots of span errors, need to investigate why it happens before I re-enable tracing


capJavert (approved):
Lets not enabled again then until it gets looked at, can you create a task for it?

Also not sure if Valerii needs it still for his task?

PR #2236, opened 2024-09-09
fix: flaky comment feed test

capJavert:
Hopefully fixes flaky test with comments feed. https://app.circleci.com/pipelines/github/dailydotdev/daily-api/9579/workflows/ed62130f-b122-451c-bd8a-4350e8035d4a/jobs/38531/tests

I found that `vordr flag` suite saves fixture for 4 comments to have `vordr: true` which is exact difference of test case failing:
image

This happens because we run our tests in parallel (over multiple workers) so even though its in describe they will execute in the same time against the db, hence the flakiness.

### DIFF
--- a/__tests__/triggers/user.ts
+++ b/__tests__/triggers/user.ts
@@ -2,7 +2,13 @@ import { DataSource } from 'typeorm';
 import createOrGetConnection from '../../src/db';
 import { saveFixtures } from '../helpers';
 import { badUsersFixture, usersFixture } from '../fixture/user';
-import { Post, Source, User, UserStreak, Comment } from '../../src/entity';
+import {
+  Source,
+  User,
+  UserStreak,
+  Comment,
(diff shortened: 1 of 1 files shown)


__tests__/triggers/user.ts
```diff
@@ -41,38 +47,46 @@ describe('user', () => {
...
+          url: 'http://pvr1.com',
+          canonicalUrl: 'http://pvr1c.com',
+        },
+      ]);
       await saveFixtures(con, Comment, [
         {
-          id: 'c1',
-          postId: 'p1',
+          id: 'cvr1',
+          postId: 'pvr1',
```
capJavert, line 62: We should probably strive to avoid using fixtures directly and make unique ids across different files to avoid this kind of issues (we do it in recent features either way).

omBratteng (approved):

PR #2237, opened 2024-09-09
chore: support for manual read replica

rebelchris:
Support for 1 read replica slave.
Set the default to still use master, as we often want direct results on mutations from the master.

This way we can manually query the slave if we want it.
Example duplicates a endpoint for testing it: (modified data for testing purposes)

Screenshot 2024-09-09 at 16 46 20
Screenshot 2024-09-09 at 16 46 28

### DIFF
--- a/.infra/Pulumi.adhoc.yaml
+++ b/.infra/Pulumi.adhoc.yaml
@@ -80,7 +80,7 @@ config:
     typeormDatabase: api
     typeormHost: postgres
     typeormPassword: 12345
-    typeormUser: postgres
+    typeormUsername: postgres
     tz: UTC
     urlPrefix: http://api.local.com
     vordrIps: ""
--- a/.infra/Pulumi.adhoc.yaml
+++ b/.infra/Pulumi.adhoc.yaml
@@ -80,7 +80,7 @@ config:
     typeormDatabase: api
     typeormHost: postgres
     typeormPassword: 12345
-    typeormUsername: postgres
+    typeormUser: postgres
     tz: UTC
     urlPrefix: http://api.local.com
     vordrIps: ""
--- a/src/data-source.ts
+++ b/src/data-source.ts
@@ -18,7 +18,7 @@ export const AppDataSource = new DataSource({
     master: {
       host: process.env.TYPEORM_HOST || 'localhost',
       port: 5432,
-      username: process.env.TYPEORM_USER || 'postgres',
+      username: process.env.TYPEORM_USERNAME || 'postgres',
       password: process.env.TYPEORM_PASSWORD || '12345',
       database:
         process.env.TYPEORM_DATABASE ||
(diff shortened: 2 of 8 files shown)


src/data-source.ts
```diff
@@ -20,4 +13,25 @@ export const AppDataSource = new DataSource({
...
+        process.env.TYPEORM_DATABASE ||
+        (process.env.NODE_ENV === 'test' ? 'api_test' : 'api'),
+    },
+    slaves: [
+      {
+        host: process.env.TYPEORM_READ_HOST || 'localhost',
+        port: 5432,
+        username: process.env.TYPEORM_READ_USERNAME || 'postgres',
+        password: process.env.TYPEORM_READ_PASSWORD || '12345',
+        database: process.env.TYPEORM_READ_DATABASE,
```
omBratteng, line 33: Fallback for adhoc?

rebelchris: We can set fallback to test db? (we need some fallback else it will fail to spool.
Let me know what you think?

capJavert: Just specify master as slave in adhoc (so we can still use `slave` in query build?

capJavert: (to avoid having another db locally as you said)

rebelchris: Yeah that also works :)

.infra/Pulumi.adhoc.yaml
```diff
@@ -80,7 +80,7 @@ config:
     typeormDatabase: api
     typeormHost: postgres
     typeormPassword: 12345
-    typeormUser: postgres
+    typeormUsername: postgres
```
rebelchris, line 83: Noticed this needed to be `username` 🫠

src/schema/sources.ts
```diff
@@ -367,6 +367,66 @@ export const typeDefs = /* GraphQL */ `
       categoryId: String
     ): SourceConnection!
 
+    """
```
idoshamun, line 370: That's for debugging right?

.infra/Pulumi.adhoc.yaml
capJavert: probably was defaulting to `postgres`?

src/schema/sources.ts
rebelchris: Yeah just a duplication but using the slave specifically :)

idoshamun (approved):

omBratteng (approved):

src/schema/sources.ts
```diff
@@ -367,6 +367,66 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      Filter by category
+      """
+      categoryId: String
+    ): SourceConnection!
+
+    """
+    Get all available sources from read replica
+    """
+    sourcesReadReplica(
```
capJavert, line 403: did you add double definition or is diff just funky?

rebelchris: @capJavert Yeah it's just a duplicate of sources to verify on production.
Idea is to merge -> test via postman -> revert this change and make the main one use the RR.

Just wanted to be safe and not break prod on deployment if that makes sense?

capJavert: no i get it but sourcesReadReplica query is defined twice?

rebelchris: Oh sorry, missed that somehow.

capJavert (approved):

PR #2238, opened 2024-09-09
feat(strict-ts): apply to env variables

capJavert:

This one focuses on: 
- env files
- all required env files are typed as `string` so that they are not optional (`string | undefined`) as default `process.env` type

### DIFF
--- a/src/temporal/config.ts
+++ b/src/temporal/config.ts
@@ -25,7 +25,10 @@ export const getTemporalServerOptions = () => {
   const address = TEMPORAL_ADDRESS;
   const namespace = TEMPORAL_NAMESPACE;
 
-  if (!process.env.TEMPORAL_CLIENT_CERT_PATH) {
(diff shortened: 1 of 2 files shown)


omBratteng (approved):
LGTM!

Side note, do you think there are any ESLint rules that could throw an error if env var is unknown | any, i.e. not defined in our types?
Would make it easier to ensure we update types when we add new variables

capJavert:
> Side note, do you think there are any ESLint rules that could throw an error if env var is unknown | any, i.e. not defined in our types?
Would make it easier to ensure we update types when we add new variables

With strict mode actually if you try to use the variable which is not in types it will make it `string | undefined`. Which will force you to use fallback or type it out. Its not perfect but its better.

As to actually requiring variables to be defined when starting up, no, we need config manager that will check that. We could use zod or schema manager sa joi for that.

PR #2240, opened 2024-09-10
feat: sort sources by members count

sshanzel:
Order the sources by members count depending if the parameter sent says true.

### DIFF
--- a/__tests__/__snapshots__/sources.ts.snap
+++ b/__tests__/__snapshots__/sources.ts.snap
@@ -128,19 +128,6 @@ Object {
 }
 `;
 
-exports[`query source current member should return current member as blocked 1`] = `
-Object {
-  "source": Object {
-    "currentMember": Object {
-      "permissions": Array [],
-      "role": "blocked",
-      "roleRank": -1,
-    },
-    "id": "a",
-  },
-}
-`;
-
 exports[`query source current member should return current member as member 1`] = `
 Object {
   "source": Object {
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -514,6 +514,9 @@ describe('query mostRecentSources', () => {
   `;
 
   it('should return most recent sources', async () => {
+    await con
+      .getRepository(Source)
+      .update({ id: In(['a', 'b']) }, { type: SourceType.Machine });
     const res = await client.query(QUERY);
     expect(res.errors).toBeFalsy();
     expect(res.data).toMatchObject({
@@ -773,7 +776,7 @@ query Source($id: ID!) {
       .getRepository(SourceMember)
       .update({ userId: '1' }, { role: SourceMemberRoles.Blocked });
     const res = await client.query(QUERY, { variables: { id: 'a' } });
-    expect(res.data).toMatchSnapshot();
+    expect(res.data.source).toBeNull();
   });
 
   it('should not return post permission in case memberPostingRank is set above user roleRank', async () => {
@@ -1115,9 +1118,7 @@ query Source($id: ID!) {
   it('should return source url', async () => {
     const res = await client.query(QUERY, { variables: { id: 'a' } });
     expect(res.errors).toBeFalsy();
-    expect(res.data.source.permalink).toEqual(
-      'http://localhost:5002/sources/a',
-    );
+    expect(res.data.source.permalink).toEqual('http://localhost:5002/squads/a');
   });
 
   it('should return squad url', async () => {
@@ -1431,10 +1432,10 @@ describe('query mySourceMemberships', () => {
     const res = await client.query(createQuery(SourceType.Squad));
     expect(res.errors).toBeFalsy();
     expect(res.data.mySourceMemberships).toBeDefined();
-    expect(res.data.mySourceMemberships.edges).toHaveLength(1);
+    expect(res.data.mySourceMemberships.edges).toHaveLength(2);
     expect(
       res.data.mySourceMemberships.edges.map(({ node }) => node.source.id),
-    ).toEqual(['squad']);
+    ).toEqual(expect.arrayContaining(['a', 'squad']));
   });
 });
 
@@ -2871,14 +2872,12 @@ describe('mutation joinSource', () => {
 
   it('should throw error when joining non squad source', async () => {
     loggedUser = '1';
+    await con
+      .getRepository(Source)
+      .update({ id: 'a' }, { type: SourceType.Machine });
     return testMutationErrorCode(
       client,
-      {
-        mutation: MUTATION,
-        variables: {
-          sourceId: 'a',
-        },
-      },
+      { mutation: MUTATION, variables: { sourceId: 'a' } },
       'FORBIDDEN',
     );
   });
@@ -3462,12 +3461,14 @@ describe('Source flags field', () => {
     expect(res.errors).toBeFalsy();
     expect(res.data.source.flags).toEqual({
       ...defaultPublicSourceFlags,
+      totalMembers: 2,
       featured: true,
     });
   });
 
   it('should return default values for unset flags', async () => {
     loggedUser = '1';
+    await con.getRepository(Source).update({ id: 'a' }, { flags: {} });
     const res = await client.query(QUERY);
     expect(res.data.source.flags).toEqual(defaultPublicSourceFlags);
   });
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -237,6 +237,7 @@ describe('query sources', () => {
     featured: boolean;
     filterOpenSquads: boolean;
     categoryId: string;
+    sortByMembersCount: boolean;
   }
 
   const QUERY = ({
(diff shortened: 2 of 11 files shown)


src/schema/sources.ts
```diff
@@ -1252,6 +1258,13 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
             );
           }
 
+          if (args.sortByMembersCount) {
+            builder.queryBuilder.orderBy(
+              `${builder.alias}."membersCount"`,
```
capJavert: this is could be heavy query since we order over `count(*)`, not ideal

sshanzel: My thoughts too initially. Maybe we can store it in the flags column then add some triggers. Wdyt?

src/migration/1725988464760-SourceMembersIndex.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceMembersIndex1725988464760 implements MigrationInterface {
+  name = 'SourceMembersIndex1725988464760';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_source_flags_total_members" ON post USING HASH (((flags->'totalMembers')::integer))`,
```
capJavert, line 8: Checked index is used with `EXPLAIN ANALYZE`?

src/entity/Source.ts
```diff
@@ -29,13 +29,15 @@ export interface SourceFlagsPublic {
...
   totalUpvotes: number;
+  totalMembers: number;
 }
 
 export const defaultPublicSourceFlags: SourceFlagsPublic = {
   featured: false,
   totalViews: 0,
   totalPosts: 0,
   totalUpvotes: 0,
+  totalMembers: 0,
```
capJavert, line 40: We are gonna have to backfill too?

src/migration/1725988464760-SourceMembersIndex.ts
sshanzel: Haven't tried that, let me check.

src/entity/Source.ts
sshanzel: Yes. We should have a script to fix this.

src/migration/1725988464760-SourceMembersIndex.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceMembersIndex1725988464760 implements MigrationInterface {
+  name = 'SourceMembersIndex1725988464760';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_source_flags_total_members" ON source USING BTREE (((flags->>'totalMembers')::integer))`,
```
sshanzel, line 8: @capJavert I updated the index, but I don't seem to see any line that says about the index being utilized:

Screenshot 2024-09-11 at 5 47 32 PM

rebelchris: Index is ::integer sort is using ::text)::integer

With JSON it comes very precise.
Also make sure you have enough records locally so it actually hits the index 🙏

sshanzel: I actually tried that locally too but no luck 😭 
Screenshot 2024-09-11 at 5 56 50 PM

sshanzel: This was the result. Could it be that the `DESC` part makes it messed up?

Screenshot 2024-09-11 at 5 58 20 PM

capJavert: Let me take a look.

capJavert: @sshanzel for me it shows up 
image

capJavert: did you apply the index migration locally on your db?

__tests__/__snapshots__/sources.ts.snap
```diff
@@ -128,19 +128,6 @@ Object {
 }
 `;
 
-exports[`query source current member should return current member as blocked 1`] = `
```
sshanzel, line 131: Replaced with a better expect statement.

__tests__/sources.ts
```diff
@@ -695,7 +776,7 @@ query Source($id: ID!) {
       .getRepository(SourceMember)
       .update({ userId: '1' }, { role: SourceMemberRoles.Blocked });
     const res = await client.query(QUERY, { variables: { id: 'a' } });
-    expect(res.data).toMatchSnapshot();
```
sshanzel, line 698: As seen from the snapshot, there was actually some data to be returned, but it was actually a false return. When a source has members, then it is definitely a Squad. Now that we turned the Squad requested here, the query returned the expected behavior which is not to return anything since you are blocked from the Squad.

bin/syncMembersCount.ts
```diff
@@ -0,0 +1,53 @@
+import fastq from 'fastq';
```
sshanzel, line 1: Tested locally. Worked as intended.

__tests__/sources.ts
```diff
@@ -421,6 +434,71 @@ describe('query sources', () => {
...
+
+    const [cSquad, aSquad, bSquad] = res.data.sources.edges;
+    expect(cSquad.node.id).toEqual('c');
+    expect(aSquad.node.id).toEqual('a');
+    expect(bSquad.node.id).toEqual('b');
+    const isBiggest = res.data.sources.edges
+      .filter(({ node }) => node.id !== 'c')
+      .every(
+        ({ node }) => cSquad.node.flags.totalMembers > node.flags.totalMembers,
+      );
```
capJavert, line 479: This is a bit clunky evaluation because you are only checking if first not is the biggest but interim notes can still be out of order, better to do something like `expect(res.data.sources.edges.map(item => item.id)).toBe(['a', 'b'])`

capJavert: Same for below test.

src/migration/1725988464760-SourceMembersIndex.ts
sshanzel: @capJavert as discussed, the index worked on certain instances, let's see how it behaves in prod data (we will run the explain analyze after merge).

src/graphorm/index.ts
```diff
@@ -377,14 +383,6 @@ const obj = new GraphORM({
             base64(`time:${new Date(node.createdAt).getTime()}`),
         },
       },
-      membersCount: {
```
capJavert, line 380: This was not used in queries from FE?

src/schema/sources.ts
```diff
@@ -2034,5 +2048,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
       return referralUrl;
     },
+    membersCount: (source: GQLSource): number =>
+      source?.flags?.totalMembers ?? 0,
```
capJavert, line 2052: Why did we moved it here from graphorm?

src/graphorm/index.ts
capJavert: https://github.com/dailydotdev/daily-api/pull/2240#discussion_r1754221733

src/schema/sources.ts
sshanzel: Good point. Let's revert it.

__tests__/sources.ts
sshanzel: I created a function to determine whether the list is sorted or not. Would love to hear what you think of it.

capJavert: I think it over complicates, in tests you don't want abstractions, you want a point blank comparison, so if you are expecting 4 elements in order we should just check that. 

```
expect(res.data.sources.edges.map(({ node }) => node.id)).toEqual([
  'c',
  'a',
  'b',
  'squad',
])
```

sshanzel: Cool. Let's do that.

sshanzel: My worry is that it might produce a flaky test, hence introduced a strict function.

sshanzel: Yep, I am getting it now locally on the unordered list.

capJavert: > My worry is that it might produce a flaky test, hence introduced a strict function.

Its not really strict but its a function that evaluates what it gets vs what test expects. If its flaky then we need to fix the fixtures. For example you can introduce uniquely IDed members as here https://github.com/dailydotdev/daily-api/pull/2236#discussion_r1750173450

src/schema/sources.ts
capJavert: We can remove from here

sshanzel: Done 🎉

__tests__/sources.ts
sshanzel: I mean it is strict that the list must not be ordered which is one of the ways we can check whether they received the props or not that we are trying to identify. Although, tests should be deterministic, and the root cause it fails is yes, the fixtures.

capJavert (approved):

__tests__/__snapshots__/sources.ts.snap
rebelchris: NICE! I love this :D

rebelchris (approved):
Nice looks good :)

PR #2243, opened 2024-09-10
feat: content updated protobuf format

capJavert:
- content updated topic will now get messages in protobuf format (Viktor is the only subscriber currently)
- added support to send Buffer messages to `publishEvent`
- also fixes sending ns/ms instead of seconds as timestamp

### DIFF
--- a/src/common/date.ts
+++ b/src/common/date.ts
@@ -26,3 +26,9 @@ export const isWeekend = (
       return day === DayOfWeek.Saturday || day === DayOfWeek.Sunday;
   }
 };
+
+export const getSecondsTimestamp = (ms: number | Date): number => {
+  const msValue = ms instanceof Date ? ms.getTime() : ms;
+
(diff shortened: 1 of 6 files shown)


src/common/date.ts
```diff
@@ -26,3 +26,9 @@ export const isWeekend = (
       return day === DayOfWeek.Saturday || day === DayOfWeek.Sunday;
   }
 };
+
+export const getSecondsTimestamp = (ms: number | Date): number => {
```
rebelchris, line 30: Isn't this just `debeziumTimeToDate`
https://github.com/dailydotdev/daily-api/blob/e20022838c55b6a5bd79bde372e6e15d7867bfc9/src/common/utils.ts#L138

rebelchris (approved):

vpol:
in general looks good to me 🚀

vpol (approved):

PR #2245, opened 2024-09-11
chore: fix deprecation warnings and re-enable tracing with 0 sampling

omBratteng:
I have verified on local environment if I set the traceparent header to force sampling, it will let us trigger tracing when sampling ratio is 0.

### DIFF
--- a/src/telemetry/common.ts
+++ b/src/telemetry/common.ts
@@ -3,7 +3,8 @@ import dc from 'node:diagnostics_channel';
 import { SemanticAttributes as SemAttr } from '@opentelemetry/semantic-conventions';
 import type { MetricOptions } from '@opentelemetry/api';
 
-export const channel = dc.channel('fastify.initialization');
+export { dc };
+export const channelName = 'fastify.initialization';
 
 // Try to get the app version from the header, then query param, then default to unknown
 export const getAppVersion = (req: FastifyRequest): string => {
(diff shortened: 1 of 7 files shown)


rebelchris (approved):

src/telemetry/common.ts
```diff
@@ -1,19 +1,16 @@
 import { FastifyRequest } from 'fastify';
-import dc from 'node:diagnostics_channel';
-import { SemanticAttributes as SemAttr } from '@opentelemetry/semantic-conventions';
+export { default as dc } from 'node:diagnostics_channel';
+export * from '@opentelemetry/semantic-conventions';
```
capJavert, line 3: Why are we re-exporting 3party packages?

capJavert: If we were to do it I would rather make it an abstraction layer/class/function that hides implementation and allows easy change without refactor. If its just due shorter lines I would rather keep it as is.

capJavert: If we were to do it I would rather make it an abstraction layer/class/function that hides implementation and allows easy change without refactor. Just changing imports I would rather leave it as is.

omBratteng: I'll remove the re-export of these two

omBratteng: :triggered: it doesn't want to give me a suggestion for auto-import of them
image

omBratteng: Needed to be a direct dependency for auto suggestions to work 😬

capJavert (approved):

PR #2246, opened 2024-09-11
feat(strict-ts): apply to workers

capJavert:

This one focuses on: 
- workers
- handling email/notification worker
- stuff that accumulated in the meantime

### DIFF
--- a/src/common/twitter.ts
+++ b/src/common/twitter.ts
@@ -6,7 +6,9 @@ export const truncateToTweet = (text?: string): string => {
   return text.length <= 130 ? text : `${text.substring(0, 127)}...`;
 };
 
-export const truncatePostToTweet = (post: Pick<Post, 'title'>): string => {
+export const truncatePostToTweet = (
+  post: Pick<Post, 'title'> | undefined,
+): string => {
   if (!post || !post.title?.length) return '';
 
   return truncateToTweet(post.title);
--- a/src/workers/newNotificationV2Mail.ts
+++ b/src/workers/newNotificationV2Mail.ts
@@ -97,7 +97,7 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
       .getRepository(Submission)
       .findOneBy({ id: notification.referenceId });
     if (!submission) {
-      return;
+      return null;
     }
     return {
       submitted_at: formatMailDate(submission.createdAt),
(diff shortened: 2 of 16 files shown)


src/common/twitter.ts
```diff
@@ -6,7 +6,9 @@ export const truncateToTweet = (text?: string): string => {
   return text.length <= 130 ? text : `${text.substring(0, 127)}...`;
 };
 
-export const truncatePostToTweet = (post: Pick<Post, 'title'>): string => {
+export const truncatePostToTweet = (
+  post: Pick<Post, 'title'> | undefined,
```
rebelchris, line 10: Why not `Partial`? (more of a question not really blocking)

src/workers/newNotificationV2Mail.ts
```diff
@@ -113,12 +113,12 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
       .findOneBy({ id: notification.referenceId })) as ArticlePost;
     const submission = await con
       .getRepository(Submission)
-      .findOneBy({ url: post.url, userId: user.id });
+      .findOneByOrFail({ url: post.url!, userId: user.id });
```
rebelchris, line 116: Do we want hard fails here?
(not sure how CDC will handle it vs if it's manually early return?)

src/workers/postAddedSlackChannelSend.ts
```diff
@@ -61,7 +61,7 @@ export const postAddedSlackChannelSendWorker: TypedWorker<'api.v1.post-visible'>
         );
 
         if (source.private && source.type === SourceType.Squad) {
-          const admin: Pick<SourceMember, 'referralToken'> = await con
+          const admin: Pick<SourceMember, 'referralToken'> | null = await con
```
rebelchris, line 64: Potential Partial candidate

rebelchris (approved):
nice work, non blocking comments.
Let's hope we can quickly turn on strict mode so all new stuff is good from the gecko.

src/workers/postAddedSlackChannelSend.ts
capJavert: its not object syntax, we get entity or null from the db

src/common/twitter.ts
capJavert: Its not object syntax

src/workers/newNotificationV2Mail.ts
capJavert: Below it uses `submission.createdAt` so it would fail either way if submission was null, I was thinking to avoid `findOneByOrFail` and use `!` (force unwrap) but think this is better.

From what I can see it will just log `failed to send notification email` as usual, so it should not go into spiral of retrying.

rebelchris: Alright, let's keep like this, just keep an eye out on merging :)

capJavert: I will 🫡

PR #2247, opened 2024-09-11
feat(strict-ts): apply to graphorm

capJavert:

This one focuses on: 
- graphorm


rebelchris (approved):
Nice :D

PR #2249, opened 2024-09-12
fix: members count trigger when creating a squad

sshanzel:
There is an instance where the user does not start being a member, that is when creating a new Squad. To make it more accurate, we check whether the user is not blocked.

### DIFF
--- a/src/migration/1726147441577-SourceMembersCountFix.ts
+++ b/src/migration/1726147441577-SourceMembersCountFix.ts
@@ -1,0 +1,43 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceMembersCountFix1726147441577 implements MigrationInterface {
+  name = 'SourceMembersCountFix1726147441577';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    queryRunner.query(`
+      CREATE OR REPLACE FUNCTION increment_squad_members_count()
+        RETURNS TRIGGER
+        LANGUAGE PLPGSQL
+        AS
+      $$
+      BEGIN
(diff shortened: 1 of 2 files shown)


src/migration/1726147441577-SourceMembersCountFix.ts
```diff
@@ -0,0 +1,43 @@
...
+        RETURN NEW;
+      END;
+      $$
+    `);
+    queryRunner.query(
+      `
+      CREATE OR REPLACE TRIGGER increment_squad_members_count
+      AFTER INSERT ON "source_member"
+      FOR EACH ROW
+      WHEN (NEW.role != 'blocked')
```
sshanzel, line 31: The only difference from the previous one is this condition.

https://github.com/dailydotdev/daily-api/blob/main/src/migration/1725981526888-SourceFlagMembersIncrement.ts

nensidosari (approved):

sshanzel:
@capJavert let me know what you think 👀

src/migration/1726147441577-SourceMembersCountFix.ts
capJavert: down should have the old function and trigger definition :hidethepain: .. its so hard to update triggers in migrations

capJavert (approved):
All good, just one comment to make a valid migration.

src/migration/1726147441577-SourceMembersCountFix.ts
sshanzel: Ah, right. Yep. Definitely needs to created on the down function 😅

PR #2250, opened 2024-09-12
feat: content preference

capJavert:
AS-577

### DIFF
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -5,10 +5,8 @@ import {
 } from '../entity/contentPreference/types';
 import { ContentPreferenceUser } from '../entity/contentPreference/ContentPreferenceUser';
 import { NotificationPreferenceUser } from '../entity/notifications/NotificationPreferenceUser';
-import {
-  NotificationPreferenceStatus,
-  NotificationType,
-} from '../notifications/common';
+import { NotificationType } from '../notifications/common';
+import { EntityManager, In } from 'typeorm';
 
 type FollowEntity = ({
   ctx,
@@ -35,6 +33,28 @@ const entityToNotificationTypeMap: Record<
   [ContentPreferenceType.User]: [NotificationType.UserPostAdded],
 };
 
+const cleanContentNotificationPreference = async ({
+  ctx,
+  entityManager,
+  id,
+  notificationTypes,
+}: {
+  ctx: AuthContext;
+  entityManager?: EntityManager;
+  id: string;
+  notificationTypes: NotificationType[];
+}) => {
+  const notificationRepository = (entityManager ?? ctx.con).getRepository(
+    NotificationPreferenceUser,
+  );
+
+  await notificationRepository.delete({
+    userId: ctx.userId,
+    referenceId: id,
+    notificationType: In(notificationTypes),
+  });
+};
+
 const followUser: FollowEntity = async ({ ctx, id, status }) => {
   await ctx.con.transaction(async (entityManager) => {
     const repository = entityManager.getRepository(ContentPreferenceUser);
@@ -48,29 +68,12 @@ const followUser: FollowEntity = async ({ ctx, id, status }) => {
 
     await repository.save(contentPreference);
 
-    const notificationRepository = entityManager.getRepository(
-      NotificationPreferenceUser,
-    );
-
-    if (status === ContentPreferenceStatus.Subscribed) {
-      const notificationPreferences = entityToNotificationTypeMap.user.map(
-        (notificationType) => {
-          return notificationRepository.create({
-            userId: ctx.userId,
-            referenceId: id,
-            referenceUserId: id,
-            notificationType,
-            status: NotificationPreferenceStatus.Subscribed,
-          });
-        },
-      );
-
-      await notificationRepository.save(notificationPreferences);
-    } else {
-      await notificationRepository.delete({
-        userId: ctx.userId,
-        referenceUserId: id,
-        referenceId: id,
+    if (status !== ContentPreferenceStatus.Subscribed) {
+      cleanContentNotificationPreference({
+        ctx,
+        entityManager,
+        id,
+        notificationTypes: entityToNotificationTypeMap.user,
       });
     }
   });
@@ -86,14 +89,11 @@ const unfollowUser: UnFollowEntity = async ({ ctx, id }) => {
       referenceId: id,
     });
 
-    const notificationRepository = entityManager.getRepository(
-      NotificationPreferenceUser,
-    );
-
-    await notificationRepository.delete({
-      userId: ctx.userId,
-      referenceUserId: id,
-      referenceId: id,
+    cleanContentNotificationPreference({
+      ctx,
+      entityManager,
+      id,
+      notificationTypes: entityToNotificationTypeMap.user,
     });
   });
 };
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -1,0 +1,127 @@
+import { AuthContext } from '../Context';
+import {
+  ContentPreferenceStatus,
+  ContentPreferenceType,
+} from '../entity/contentPreference/types';
+import { ContentPreferenceUser } from '../entity/contentPreference/ContentPreferenceUser';
+import { NotificationPreferenceUser } from '../entity/notifications/NotificationPreferenceUser';
+import {
+  NotificationPreferenceStatus,
+  NotificationType,
+} from '../notifications/common';
+
+type FollowEntity = ({
+  ctx,
+  id,
+  status,
+}: {
+  ctx: AuthContext;
+  id: string;
+  status: ContentPreferenceStatus.Follow | ContentPreferenceStatus.Subscribed;
+}) => Promise<void>;
+
+type UnFollowEntity = ({
+  ctx,
+  id,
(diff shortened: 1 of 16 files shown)


src/common/contentPreference.ts
```diff
@@ -0,0 +1,135 @@
...
+}: {
+  ctx: AuthContext;
+  id: string;
+}) => Promise<void>;
+
+const entityToNotificationTypeMap: Record<
+  ContentPreferenceType,
+  NotificationType[]
+> = {
+  [ContentPreferenceType.User]: [NotificationType.UserPostAdded],
```
rebelchris, line 35: 🤟  this looks neat and scaleable :)

src/common/contentPreference.ts
```diff
@@ -0,0 +1,135 @@
...
+            referenceUserId: id,
+            notificationType,
+            status: NotificationPreferenceStatus.Subscribed,
+          });
+        },
+      );
+
+      await notificationRepository.save(notificationPreferences);
+    } else {
+      await notificationRepository.delete({
```
rebelchris, line 70: Nothing major, but wondering if we should abstract this into a in file function, just for the sake if next month you need to add another field here.

src/entity/contentPreference/ContentPreference.ts
```diff
@@ -0,0 +1,34 @@
+import {
+  Column,
+  Entity,
+  ManyToOne,
+  PrimaryColumn,
+  TableInheritance,
+} from 'typeorm';
+import type { User } from '../user/User';
+import { ContentPreferenceStatus, ContentPreferenceType } from './types';
```
rebelchris, line 9:
```suggestion
import type { ContentPreferenceStatus, ContentPreferenceType } from './types';
```

src/migration/1726149087071-ContentPreference.ts
```diff
@@ -0,0 +1,45 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ContentPreference1726149087071 implements MigrationInterface {
+  name = 'ContentPreference1726149087071';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "content_preference" ("referenceId" text NOT NULL, "userId" character varying NOT NULL, "type" text NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "status" text NOT NULL, "referenceUserId" text, CONSTRAINT "PK_3f50987656b7555fa8195d6f05b" PRIMARY KEY ("referenceId", "userId"))`,
```
rebelchris, line 8: Shouldn't referenceId be uniquely indexed?

src/schema/contentPreference.ts
```diff
@@ -0,0 +1,285 @@
...
+    referenceId: ID!
+
+    user: User!
+
+    type: ContentPreferenceType!
+
+    createdAt: DateTime!
+
+    status: ContentPreferenceStatus!
+  }
```
rebelchris, line 40: Not sure if it was intentional or you missed it, non blocking, but this doesn't have descriptions

src/schema/contentPreference.ts
```diff
@@ -0,0 +1,285 @@
...
+      """
+      Entity type (user, source..)
+      """
+      entity: ContentPreferenceType!
+    ): ContentPreference @auth
+
+    """
+    Who follows user
+    """
+    userFollowers(
```
rebelchris, line 74: This can be multiple entities right?
Just raising it because below is called:

- userFollowing
- this should maybe be `userFollows`
- And have description: `What follows user`

Or certain entities should not be allowed here if its only users.

src/schema/contentPreference.ts
```diff
@@ -0,0 +1,285 @@
...
+      """
+      Id of user
+      """
+      userId: ID!
+      """
+      Entity to list (user, source..)
+      """
+      entity: ContentPreferenceType!
```
rebelchris, line 82: Here I mean.

src/schema/contentPreference.ts
```diff
@@ -0,0 +1,285 @@
...
+
+const contentPreferencePageGenerator =
+  offsetPageGenerator<GQLContentPreference>(10, 50);
+
+export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
+  unknown,
+  BaseContext
+>({
+  Query: {
+    contentPreferenceStatus: async (
```
rebelchris, line 158: Just more wondering where do we actually use this singular call?

src/schema/users.ts
```diff
@@ -1158,25 +1170,36 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-          .addSelect('sum(comment.upvotes)', 'numCommentUpvotes')
-          .from(Comment, 'comment')
-          .where({ userId: id })
-          .innerJoin(Post, 'p', `comment.postId = p.id`)
-          .andWhere('p.visible = true')
-          .andWhere('p.deleted = false')
-          .getRawOne(),
-      ]);
+      const [postStats, commentStats, numFollowing, numFollowers] =
+        await Promise.all([
```
rebelchris, line 1174: I guess at one stage this would make a good `user_stats` mv?

rebelchris (commented):
Nice some minor questions, but think most are ok, just let me know on re-request and will approve :)

src/schema/contentPreference.ts
nensidosari: > Or certain entities should not be allowed here if its only users.

👍 agree with this

nensidosari: Do we ever plan in our roadmap to show which users follows a specific source? just thinking as if so, this could be another abstraction where we don't specify user, rather entityFollowers(this would be any type of entity to users) and userFollowing can stay the same as it doesn't really make sense for any other entity to be following rather then just user.

src/common/contentPreference.ts
```diff
@@ -0,0 +1,127 @@
...
+          return notificationRepository.create({
+            userId: ctx.userId,
+            referenceId: id,
+            notificationType,
+            status: NotificationPreferenceStatus.Subscribed,
+          });
+        },
+      );
+
+      await notificationRepository.save(notificationPreferences);
```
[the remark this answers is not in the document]
nensidosari: This question still relevant?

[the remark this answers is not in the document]
capJavert: yes cc: @rebelchris @nensidosari

src/migration/1726149087071-ContentPreference.ts
capJavert: referenceId is generic field, referenceUserId is the one which has FK, same as NotificationPreference

this is because you can't have FK to multiple tables

src/schema/contentPreference.ts
capJavert: We usually don't add descriptions to types, I think self explanatory?

src/schema/contentPreference.ts
capJavert: In some cases like search page or post page author/scout we have singular follow buttons so might be easier to lazy load state like this instead of nested user resolve.

src/schema/users.ts
capJavert: Yeah probably.

src/common/contentPreference.ts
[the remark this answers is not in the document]
capJavert: If ContentPreference row has status `Subscribed` we automatically then allow all types of follow notifications.

Currently we only notify  user followers (subscribes) when user adds a new post.

Let’s say in a future we also want to notify when he comments. 

User X already follows and is subscribed to user Y. But he only has NotificationPreference entry for user_post_added type. So in case of adding NotificationPreference rows with `Subscribed` we would have to backfill. 

With above idea we treat absence of NotificationPreference row in case of ContentPreference as user is subscribed to everything.

We still allow adding `NotificationPreference.Muted` which would block specific notification.

So its just a matter of optimization of storage and amount of rows we have for each user when they follow different entities.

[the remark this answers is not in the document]
capJavert: Mentioned also in [RFC](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1126432769/RFC-067+-+Follow?focusedCommentId=1145536546)

src/schema/contentPreference.ts
capJavert: I am don't understand the naming issue, we have two queries userFollowers (followers of the user), userFollowing (entities that user is following)?

nensidosari: All good on my side

rebelchris: followers just sounds like it's always `users` 
only `userFollows` but that could again by confusing, so just leave it for now :)

src/schema/contentPreference.ts
rebelchris: Got it :)

src/common/contentPreference.ts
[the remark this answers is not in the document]
nensidosari: Yeah sounds good to me as well

nensidosari (approved):

src/common/contentPreference.ts
capJavert: done

PR #2255, opened 2024-09-16
fix: only add span labels if span is still open

omBratteng:
Ensure that we only set span labels while the span is still open.
In cases where the request is cancelled while it is still running (i.e. user navigated away)


capJavert (approved):

capJavert:
@omBratteng do you want to increase sampling from zero as well so you can test?

omBratteng:
Not yet, will test via cURL first

PR #2258, opened 2024-09-16
feat: source category search by title

sshanzel:
Search the category by title.


nensidosari (approved):

src/schema/sources.ts
```diff
@@ -1232,8 +1232,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       graphorm.queryOneOrFail(ctx, info, (builder) => ({
         ...builder,
         queryBuilder: builder.queryBuilder.where(
-          `"${builder.alias}"."id" = :id`,
-          { id },
+          `(CAST("${builder.alias}"."id" AS text) = :id OR LOWER(REPLACE("${builder.alias}".title, ' ', '')) = :id)`,
```
capJavert, line 1235: We need to make sure proper index is added for this

sshanzel: Would we want it even if this is just the categories of 10 items? Might grow more though most likely less than a hundred.

capJavert: Yes and no, I don't know how large it will grow in the future, but either way all the casting here looks unnecessary? The `LIKE` would not work here?

capJavert: With above it feels like many queries won't really work when searching?

sshanzel: Yeah, let's index it just to be sure.

> The LIKE or ILIKE would not work here?

I found that `lower` works better than ILIKE. Using `=` should be enough for us since most of our categories are single-worded. And if it is two, then the replace lower catches that.

capJavert: See my comment https://github.com/dailydotdev/apps/pull/3541#discussion_r1761686670

sshanzel:
@capJavert @rebelchris have updated the flow now and uses slug.

src/schema/sources.ts
```diff
@@ -1232,8 +1232,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       graphorm.queryOneOrFail(ctx, info, (builder) => ({
         ...builder,
         queryBuilder: builder.queryBuilder.where(
-          `"${builder.alias}"."id" = :id`,
-          { id },
+          `(CAST("${builder.alias}".id AS text) = :id OR ${builder.alias}.slug = :id)`,
+          { id: id.toLowerCase() },
         ),
```
rebelchris, line 1237: I think you can also keep the id as is, but add `orWhere()` to not have to do cast.

rebelchris (approved):
Non breaking comment.

src/schema/sources.ts
capJavert: Also would say no need to lower case, we expect lowercase from frontend, everything else is invalid slug.

src/migration/1726556744920-SourceCategorySlug.ts
```diff
@@ -0,0 +1,40 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourceCategorySlug1726556744920 implements MigrationInterface {
+  name = 'SourceCategorySlug1726556744920';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "source_category" ADD "slug" text GENERATED ALWAYS AS (trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(LEFT(source_category.title,100),'')||'-'||source_category.id)), '[^a-z0-9-]+', '-', 'gi'))) STORED NOT NULL`,
```
capJavert, line 8: We need to expose slug in graphql type as well?

sshanzel: Yes. Should be part of the Schema now.

src/schema/sources.ts
```diff
@@ -1231,10 +1232,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     ): Promise<GQLSourceCategory> =>
       graphorm.queryOneOrFail(ctx, info, (builder) => ({
         ...builder,
-        queryBuilder: builder.queryBuilder.where(
-          `"${builder.alias}"."id" = :id`,
-          { id },
-        ),
+        queryBuilder: builder.queryBuilder
+          .where(`CAST("${builder.alias}".id AS text) = :id`, { id })
```
rebelchris, line 1236: You don;t need cast here anymore?

src/entity/sources/SourceCategory.ts
```diff
@@ -22,4 +23,16 @@ export class SourceCategory {
...
   updatedAt: Date;
+
+  @Column({
+    type: 'text',
+    update: false,
+    insert: false,
+    nullable: false,
+    unique: true,
+    generatedType: 'STORED',
+    asExpression: `trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(LEFT(source_category.title,100),'')||'-'||source_category.id)), '[^a-z0-9-]+', '-', 'gi'))`,
```
rebelchris, line 34: You should remove it here as well :)

rebelchris (changes requested):

src/schema/sources.ts
sshanzel: We still need it. When the value it compared against is not UUID. It would throw an error.

src/entity/sources/SourceCategory.ts
sshanzel: Oh, right. Good catch. Let me re-run a migration after deleting just to be sure.

src/schema/sources.ts
rebelchris: Oh right it's to not fail on slugs :) no problem then

src/entity/sources/SourceCategory.ts
```diff
@@ -22,4 +23,16 @@ export class SourceCategory {
...
   updatedAt: Date;
+
+  @Column({
+    type: 'text',
+    update: false,
+    insert: false,
+    nullable: false,
+    unique: true,
+    generatedType: 'STORED',
+    asExpression: `trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(LEFT(source_category.title,100),''))), '[^a-z0-9-]+', '-', 'gi'))`,
```
rebelchris, line 34: Do we need the COALESCE and LEFT still? 
Think just lower trim should do right?

rebelchris (changes requested):

src/entity/sources/SourceCategory.ts
capJavert: LEFT needed to still keep max 100 chars in a slug I think

sshanzel: Cool. Wouldn't really change anything without COALESCE. Actually better to have it than not.

sshanzel:
@capJavert the slug should now have the title only.

capJavert (approved):

PR #2260, opened 2024-09-17
feat: order by priority column

sshanzel:
- Introduce a new column to handle ordering based on our demand.
- Apparently, there are some changes with the categories. Rather than reusing an existing entity and just updating the title, it might become inaccurate. Hence, deleting the rows instead. Good thing that we also have no Squads with any category yet. Although they should be safe as cascading would take effect and null would be set in any case.
  - Another option is to turn the `enabled` to `false` values.
  - ~~Removed `['General', 'DevOps', 'Cloud', 'Data']`.~~
  - ~~Added `['Basics', 'DevOps & Cloud', 'AI', 'Open Source', 'DevRel']`~~
  - Removed every category (good thing no one is using any of the items yet). This is to make it easier to manage in the future and better devex. We will run a once off script to insert the data along with the Squad updates. This would make seeding our local env consistent.
- Created the migration for the column and did the necessary insert/delete/update commands at the same time.

### DIFF
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -58,7 +58,7 @@ beforeAll(async () => {
 
 const getSourceCategories = () => [
   {
-    title: 'General',
+    title: 'Basics',
     enabled: true,
   },
   {
@@ -74,11 +74,11 @@ const getSourceCategories = () => [
     enabled: true,
   },
   {
-    title: 'DevOps',
+    title: 'DevOps & Cloud',
     enabled: true,
   },
   {
-    title: 'Cloud',
+    title: 'Open Source',
     enabled: true,
   },
   {
@@ -86,7 +86,7 @@ const getSourceCategories = () => [
     enabled: true,
   },
   {
-    title: 'Data',
+    title: 'AI',
     enabled: true,
   },
   {
@@ -95,6 +95,10 @@ const getSourceCategories = () => [
   },
   {
     title: 'DevTools',
+    enabled: true,
+  },
+  {
+    title: 'DevRel',
     enabled: true,
   },
 ];
@@ -252,6 +256,43 @@ describe('query sourceCategories', () => {
     );
     expect(isAllFound).toBeTruthy();
   });
+
+  it('should return categories ordered by priority', async () => {
+    await con.createQueryRunner().query(`
+      DO $$
+      DECLARE
+          categories TEXT[] := ARRAY['Basics','Web','Mobile','DevOps & Cloud','AI','Games','DevTools','Career','Open Source','DevRel','Fun'];
+          i INT;
+      BEGIN
+          -- Iterate over the array and update the table
+          FOR i IN 1..array_length(categories, 1) LOOP
+              UPDATE source_category
+              SET priority = i
+              WHERE slug = trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(LEFT(categories[i],100),''))), '[^a-z0-9-]+', '-', 'gi'));
+          END LOOP;
+      END $$;
+    `);
+    const expected = [
+      'Basics',
+      'Web',
+      'Mobile',
+      'DevOps & Cloud',
+      'AI',
+      'Games',
+      'DevTools',
+      'Career',
+      'Open Source',
+      'DevRel',
+      'Fun',
+    ];
+    loggedUser = '1';
+    const res = await client.query(QUERY);
+    expect(res.errors).toBeFalsy();
+    const mapped = res.data.sourceCategories.edges.map(
+      ({ node }) => node.title,
+    );
+    expect(mapped).toEqual(expected);
+  });
 });
 
 describe('query sources', () => {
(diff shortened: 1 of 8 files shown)


__tests__/sources.ts
```diff
@@ -252,6 +256,43 @@ describe('query sourceCategories', () => {
     );
     expect(isAllFound).toBeTruthy();
   });
+
+  it('should return categories ordered by priority', async () => {
+    await con.createQueryRunner().query(`
+      DO $$
```
sshanzel, line 262: Copy pasted the function I used in the migration.

src/schema/sources.ts
```diff
@@ -1043,6 +1042,11 @@ const membershipsPageGenerator = offsetPageGenerator<GQLSourceMember>(100, 500);
 
 const sourcePageGenerator = offsetPageGenerator<GQLSource>(100, 500);
 
+const categoriesPageGenerator = offsetPageGenerator<GQLSourceCategory>(
+  100,
+  500,
```
capJavert, line 1047: let's put lower default, something like 10 and 50

src/migration/1726584940063-SourceCategoryPriority.ts
```diff
@@ -0,0 +1,66 @@
...
+            -- Iterate over the array and update the table
+            FOR i IN 1..array_length(categories, 1) LOOP
+                UPDATE source_category
+                SET priority = i
+                WHERE slug = trim(BOTH '-' FROM regexp_replace(lower(trim(COALESCE(LEFT(categories[i],100),''))), '[^a-z0-9-]+', '-', 'gi'));
+            END LOOP;
+        END $$;
+      `,
+    );
+  }
```
capJavert, line 40: Just thinking can we do this inserts as a once off, if we ever add new categories it will mean we need to again fight with migrations. Non blocking, up to you whatever you think its best.

sshanzel: Right. Yeah, really hard to decide if want to drop everything here, then just prepare an insert script along with the ordering one.

sshanzel: That might actually be the better idea as I can export what will be created in prod then make those data as the seeds to make it consistent if we want to do more seedings that involve this table. Rather than generating new ids every time we run migrations locally.

src/entity/sources/SourceCategory.ts
```diff
@@ -18,6 +18,9 @@ export class SourceCategory {
   @Column()
   enabled: boolean;
 
+  @Column({ default: null, nullable: true })
```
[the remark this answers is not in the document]
sshanzel: Yeah, been meaning to ask if we want this. Forgot in the description. Lets add it.

src/schema/sources.ts
sshanzel: I put 15 - 30. Wdyt? Just so we can easily cover the initial amount of 11 categories we have.

bin/updateSquadsCategory.ts
```diff
@@ -0,0 +1,71 @@
+import fastq from 'fastq';
```
sshanzel, line 1: Script to update existing Squads from this [JSON file](https://github.com/dailydotdev/daily-api/blob/e483f23e32dab8b22412c573d49b6f32f1475b02/bin/categorizedSquads.ts). Originally from a CSV [provided by product team](https://dailydotdev.slack.com/archives/C069XNYMDGV/p1726575033390519?thread_ts=1725703282.452569&cid=C069XNYMDGV).

bin/updateSquadsCategory.ts
```diff
@@ -0,0 +1,54 @@
...
+interface Result {
+  title: string;
+  id: string;
+}
+
+(async () => {
+  const con = await createOrGetConnection();
+  await con.transaction(async (manager) => {
+    console.log('initializing stream');
+    const categories = await manager.getRepository(SourceCategory).find();
```
capJavert, line 18: Would `find({ title: In(categorizedSquads.map(item => item.category) })` would work here better to not query everything?

src/schema/sources.ts
capJavert: sure

bin/insertCategories.sql
```diff
@@ -0,0 +1,28 @@
+INSERT INTO "source_category"
+(title, enabled)
+VALUES
+('Languages', true),
+('Web', true),
+('Mobile', true),
```
capJavert, line 6: You will run this once to insert initial categories?

capJavert (approved):

bin/updateSquadsCategory.ts
sshanzel: We do this pull to ensure everything listed from their data was found and errors to be fixed. Thankfully we did, found some issues with their data and have been fixed now.

bin/insertCategories.sql
sshanzel: Yes. Just to have a reference on what we are trying to run :shipit:

bin/updateSquadsCategory.ts
sshanzel: We can add this condition on the actual update that we will run. Let me add it quickly.

PR #2261, opened 2024-09-18
fix: allowing boot and graphql endpoints to bots

ilasw:
Removing `/boot` and `/graphql` from disallowed paths. 

But maybe we want do the the reverse at some point disabling only the crucial ones and allowing the rest [like gitlab do](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/39520).

### DIFF
--- a/src/routes/index.ts
+++ b/src/routes/index.ts
@@ -52,7 +52,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
   fastify.get('/robots.txt', (req, res) => {
     return res.type('text/plain').send(`User-agent: *
 Allow: /devcards/
-Allow: /graphql/
+Allow: /graphql
 Allow: /boot
 Disallow: /`);
   });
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Yeah agreed on the notion of rather disallowing only specifics.
But for now this is perfect.

capJavert:
We are doing this in order to try and fix SEO?

ilasw:
@capJavert Yes, [this is the ticket](https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31?selectedIssue=MI-531)

src/routes/index.ts
```diff
@@ -52,6 +52,8 @@ export default async function (fastify: FastifyInstance): Promise<void> {
   fastify.get('/robots.txt', (req, res) => {
     return res.type('text/plain').send(`User-agent: *
 Allow: /devcards/
+Allow: /graphql/
```
omBratteng, line 55: IIRC I think it's only `/graphql`
```suggestion
Allow: /graphql
```

ilasw: Yes, thanks!

capJavert (approved):

nensidosari (approved):

PR #2263, opened 2024-09-18
feat: add new notification columns to `User`

omBratteng:
Extends user table to add two new columns, `followingEmail` and `followNotifications`

AS-584

### DIFF
--- a/src/migration/1726691862710-User.ts
+++ b/src/migration/1726691862710-User.ts
@@ -1,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class User1726691862710 implements MigrationInterface {
+  name = 'User1726691862710'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "user" ADD "followingEmail" boolean NOT NULL DEFAULT true`);
+    await queryRunner.query(`ALTER TABLE "user" ADD "followNotifications" boolean NOT NULL DEFAULT true`);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "user" DROP COLUMN "followNotifications"`);
+    await queryRunner.query(`ALTER TABLE "user" DROP COLUMN "followingEmail"`);
+  }
+}
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -384,6 +384,8 @@ const getUser = (con: DataSource, userId: string): Promise<User | null> =>
       'cover',
       'experienceLevel',
       'language',
+      'followingEmail',
+      'followNotifications',
     ],
   });
 
--- a/__tests__/boot.ts
+++ b/__tests__/boot.ts
@@ -110,6 +110,8 @@ const LOGGED_IN_BODY = {
     company: null,
     cover: null,
     experienceLevel: null,
+    followNotifications: true,
+    followingEmail: true,
     isTeamMember: false,
     roadmap: null,
     threads: null,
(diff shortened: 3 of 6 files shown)


src/routes/boot.ts
```diff
@@ -384,6 +384,8 @@ const getUser = (con: DataSource, userId: string): Promise<User | null> =>
       'cover',
       'experienceLevel',
       'language',
+      'followingEmail',
+      'followNotifications',
```
omBratteng, line 388: I was thinking we could maybe start moving some of these out of boot and into a separate query. We don't really need all of these at boot

rebelchris: Yeah these for sure we only need only on the specific settings screen right?
Don't we do a separate query there already maybe?

src/migration/1726691862710-User.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class User1726691862710 implements MigrationInterface {
+  name = 'User1726691862710'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "user" ADD "followingEmail" boolean NOT NULL DEFAULT true`);
```
rebelchris, line 7: Just thinking is there any case where we want to consider false now?
I do think when someone turned off all emails for instance or all push this should be false... Maybe something to consider in the migration (should be a easy query)

From FE it should work the same.
Turning the main email off should turn off all sub emails immediatly.

CC: @capJavert fyi

rebelchris (changes requested):

src/routes/boot.ts
capJavert: I think settings still look at boot, but @omBratteng can check that, I agree with moving them out but we are waiting for product to define new notif settings feature.

src/migration/1726691862710-User.ts
capJavert: I agree, we can do once of UPDATE query or script but we need to check with product because they usually don't like emails off 🙈

src/routes/boot.ts
omBratteng: It still uses boot

src/migration/1726691862710-User.ts
omBratteng: @capJavert do we only want to check for the Activity and Community updates, or also personalized digest?
Because when email notifications are "Off", all of them are off. But if you enable digest, it technically back on.

capJavert: I think the first one is fine for now

omBratteng: For emails it's as simple as
```sql
UPDATE
	"user"
SET
	"followingEmail" = FALSE
WHERE
	"user"."acceptedMarketing" = FALSE
	AND "user"."notificationEmail" = FALSE;
```

It would update 120k rows

omBratteng: For push notifications it's either on or it's not. So there we could just keep it on as default

omBratteng: We should probably run it manually in batches. Once when this is merged, and once just as we go live to catch any changes that might've been done while UI wasn't in place

omBratteng: Anything else missing from the PR?

omBratteng: Easier to run a query to turn it off for 120k users, then the other 870k that has either enabled

rebelchris (approved):

capJavert (approved):

capJavert:
Some more tests need to be updated I think.

PR #2267, opened 2024-09-19
fix: content preference state on follow queries

capJavert:
Fixes following (no pun intended 😆 ) [issue](https://dailydotdev.slack.com/archives/C07ATUPBPJ9/p1726756727350719)

Basically following and followers queries already define `:userId` parameter which means that inside graphorm `contentPreference` field resolver it gets overwritten and then it always takes the state from parent entity.

This fixes it by naming the param inside the field resolver to unique name.


omBratteng (approved):

rebelchris (approved):

rebelchris:
Ah this similar case caused me many headaches before 😅

PR #2268, opened 2024-09-20
fix: allow content preference on search

rebelchris:
@capJavert Please let me know what you think before I proceed.
The idea is that for the page query we can enable this and for the suggestion query we omit it.

Since the contentpreference is needed for more then 1 entity this seemed ok, but let me know if you are against it.

### DIFF
--- a/src/schema/search.ts
+++ b/src/schema/search.ts
@@ -31,7 +31,8 @@ import {
 import { getOffsetWithDefault } from 'graphql-relay';
 import { Brackets } from 'typeorm';
 import { whereVordrFilter } from '../common/vordr';
-import type { ContentPreference } from '../entity/contentPreference/ContentPreference';
+import { ContentPreference } from '../entity/contentPreference/ContentPreference';
+import { ContentPreferenceType } from '../entity/contentPreference/types';
 
 type GQLSearchSession = Pick<SearchSession, 'id' | 'prompt' | 'createdAt'>;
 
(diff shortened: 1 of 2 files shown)


src/schema/search.ts
```diff
@@ -39,6 +44,7 @@ interface GQLSearchSuggestion {
   title: string;
   subtitle?: string;
   image?: string;
+  contentpreferencestatus?: ContentPreferenceStatus;
```
capJavert, line 47: `contentPreferenceStatus`?

src/schema/search.ts
```diff
@@ -455,7 +467,20 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        searchQuery.addSelect('cpu.status as contentPreferenceStatus');
+        searchQuery.leftJoin(
+          ContentPreferenceUser,
+          'cpu',
+          'cpu.referenceId = u.id AND cpu.type = :cpuType AND cpu.userId = :cpuUserId',
+          {
+            cpuType: ContentPreferenceType.User,
+            cpuUserId: ctx.userId,
+          },
+        );
```
capJavert, line 481: Yeah this sounds good, I was thinking if we maybe also want to expose usual `ContentPreference` columns instead of just status, its basically the same since you already join/select?

capJavert (approved):

src/schema/search.ts
capJavert: so:
- createdAt
- referenceId
- type

capJavert: Just to keep it inline with what frontend types expect.

src/schema/search.ts
rebelchris: I had it like that, but due to graphorm lowercasing it need to be like this in return.
not 100% sure if we can fix it? 😅

src/schema/search.ts
rebelchris: Hmm we can try? would be ideal I guess if it's just a extendable entity.
Let me try some magic

src/schema/search.ts
capJavert: graphorm lowercasing? first time I hear about it, how does the other stuff work then like `contentPreference`

src/schema/search.ts
```diff
@@ -422,8 +430,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     },
     searchUserSuggestions: async (
       source,
-      { query, limit }: SearchSuggestionArgs,
-      ctx,
+      { query, limit, includeContentPreference }: SearchSuggestionArgs,
+      ctx: AuthContext,
```
capJavert, line 434: This should be just `ctx: Context` since its not `@auth` query.

src/schema/search.ts
```diff
@@ -455,7 +463,16 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
                 query: `%${query}%`,
               });
           }),
-        )
+        );
+      if (includeContentPreference) {
```
capJavert, line 467:
```suggestion
      if (includeContentPreference && ctx.userId) {
```

capJavert (approved):

capJavert (approved):

PR #2274, opened 2024-09-22
feat: notify on slack when source is reported

sshanzel:
When a Squad is reported, we should send a Slack notification to the same channel where we send post and comment reports.


src/common/slack.ts
```diff
@@ -157,6 +157,38 @@ export const notifyNewVordrPost = async (
...
+  userId: string,
+  source: Source,
+  reason: string,
+  comment?: string,
+): Promise<void> => {
+  await webhooks.content.send({
+    text: 'Source/Squad was just reported!',
+    attachments: [
+      {
+        title: source.name ?? `Post ${source.id}`,
```
idoshamun, line 170:
```suggestion
        title: source.name ?? `Source ${source.id}`,
```

idoshamun (approved):
Just fix the minor typo and it's ready to go :)

rebelchris (approved):
Besides ido nothing to add.
We also have one for report user?

capJavert (approved):

PR #2282, opened 2024-09-25
feat: public threshold for source

sshanzel:
- New property in the source.flags column.
- Updated query to accept a parameter to filter threshold.
- Cron job to update the flag every 10 minutes.

### DIFF
--- a/src/cron/updateSourcePublicThreshold.ts
+++ b/src/cron/updateSourcePublicThreshold.ts
@@ -13,9 +13,12 @@ export const updateSourcePublicThreshold: Cron = {
       .set({ flags: updateFlagsStatement({ publicThreshold: true }) })
       .where(
         `
-          (flags->>'totalMembers')::int >= 3 AND
-          (flags->>'totalPosts')::int >= 3 AND
-          (flags->>'publicThreshold')::boolean IS NOT TRUE
+          s.type = 'squad' AND
+          s.image IS NOT NULL AND
+          s.description IS NOT NULL AND
+          (s.flags->>'totalMembers')::int >= 3 AND
+          (s.flags->>'totalPosts')::int >= 3 AND
+          (s.flags->>'publicThreshold')::boolean IS NOT TRUE
       `,
       )
       .execute();
--- a/src/cron/updateSourcePublicThreshold.ts
+++ b/src/cron/updateSourcePublicThreshold.ts
@@ -1,4 +1,3 @@
-import { JsonContains, MoreThan, Not } from 'typeorm';
 import { Cron } from './cron';
 import { Source } from '../entity';
 import { updateFlagsStatement } from '../common';
@@ -7,16 +6,19 @@ export const updateSourcePublicThreshold: Cron = {
   name: 'update-source-public-threshold',
   handler: async (con, logger) => {
     logger.info('updating public threshold value for valid sources...');
-    const result = await con.getRepository(Source).update(
-      {
-        flags: JsonContains({
-          publicThreshold: Not(true),
-          totalMembers: MoreThan(2),
-          totalPosts: MoreThan(2),
-        }),
-      },
-      { flags: updateFlagsStatement({ publicThreshold: true }) },
-    );
+    const result = await con
+      .getRepository(Source)
+      .createQueryBuilder('s')
+      .update()
+      .set({ flags: updateFlagsStatement({ publicThreshold: true }) })
+      .where(
+        `
+          (flags->>'totalMembers')::int >= 3 AND
+          (flags->>'totalPosts')::int >= 3 AND
+          (flags->>'publicThreshold')::boolean IS NOT TRUE
+      `,
+      )
+      .execute();
     logger.info(
       { count: result.affected },
       'public threshold updated for valid sources! 🚀',
--- a/src/cron/updateSourcePublicThreshold.ts
+++ b/src/cron/updateSourcePublicThreshold.ts
@@ -13,12 +13,12 @@ export const updateSourcePublicThreshold: Cron = {
       .set({ flags: updateFlagsStatement({ publicThreshold: true }) })
       .where(
         `
-          s.type = 'squad' AND
-          s.image IS NOT NULL AND
-          s.description IS NOT NULL AND
-          (s.flags->>'totalMembers')::int >= 3 AND
-          (s.flags->>'totalPosts')::int >= 3 AND
-          (s.flags->>'publicThreshold')::boolean IS NOT TRUE
+          "type" = 'squad' AND
+          "image" IS NOT NULL AND
+          "description" IS NOT NULL AND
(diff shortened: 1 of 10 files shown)


src/cron/updateSourcePublicThreshold.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Source } from '../entity';
+import { updateFlagsStatement } from '../common';
+
+export const updateSourcePublicThreshold: Cron = {
+  name: 'update-source-public-threshold',
+  handler: async (con, logger) => {
+    logger.info('updating public threshold value for valid sources...');
+    const result = await con.getRepository(Source).update(
+      {
+        flags: JsonContains({
```
rebelchris, line 12: TIL

rebelchris (approved):
Nothing really to add.

src/cron/updateSourcePublicThreshold.ts
```diff
@@ -0,0 +1,25 @@
...
+          publicThreshold: Not(true),
+          totalMembers: MoreThan(2),
+          totalPosts: MoreThan(2),
```
idoshamun, line 15: @sshanzel let's add index for these fields please

src/graphorm/index.ts
```diff
@@ -381,6 +382,7 @@ const obj = new GraphORM({
             totalViews: value?.totalViews ?? totalViews,
             totalUpvotes: value?.totalUpvotes ?? totalUpvotes,
             totalMembers: value?.totalMembers ?? totalMembers,
+            publicThreshold: value?.publicThreshold ?? publicThreshold,
```
idoshamun, line 385: It's not user-facing, so we don't need it

src/schema/sources.ts
```diff
@@ -361,6 +362,11 @@ export const typeDefs = /* GraphQL */ `
       """
       featured: Boolean
 
+      """
```
idoshamun, line 365: It shouldn't be user facing

idoshamun (changes requested):
Neat implementation but added some comments :)

src/schema/sources.ts
sshanzel: Right. Just used it for testing. We can remove it.

src/cron/updateSourcePublicThreshold.ts
sshanzel: All three should have their own indexes now :shipit:

src/schema/sources.ts
sshanzel: This one is for filtering the squad directory when asking for the sources.

src/cron/updateSourcePublicThreshold.ts
sshanzel: Unfortunately, it doesn't work with `MoreThan`, `LessThan`, etc. 😞

src/entity/Source.ts
```diff
@@ -111,8 +115,9 @@ export class Source {
 
   @Column({ type: 'jsonb', default: {} })
   @Index('IDX_source_flags_featured', { synchronize: false })
+  @Index('IDX_source_flags_posts_members_threshold', { synchronize: false })
```
sshanzel, line 118: @idoshamun the index had to be declared here. Unlike the other composite indexes we have, we will use jsonb columns on this one, and the top-level index does not accept `synchronize: false` value.

src/cron/updateSourcePublicThreshold.ts
```diff
@@ -0,0 +1,27 @@
...
+    logger.info('updating public threshold value for valid sources...');
+    const result = await con
+      .getRepository(Source)
+      .createQueryBuilder('s')
+      .update()
+      .set({ flags: updateFlagsStatement({ publicThreshold: true }) })
+      .where(
+        `
+          (flags->>'totalMembers')::int >= 3 AND
+          (flags->>'totalPosts')::int >= 3 AND
```
rebelchris, line 17: I'm missing squad image and description
https://dailydotdev.slack.com/archives/C069XNYMDGV/p1726986254903039?thread_ts=1726896915.493839&cid=C069XNYMDGV

src/entity/Source.ts
```diff
@@ -111,8 +115,9 @@ export class Source {
...
   @Index('IDX_source_flags_total_members', { synchronize: false })
-  flags: SourceFlagsPublic;
+  flags: SourceFlagsPublic & SourceFlagsPrivate;
```
rebelchris, line 120: Normally we have `SourceFlags` and `SourceFlagsPublic` is a different one.

src/migration/1727345240782-SourceFlagsIndexes.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class SourceFlagsIndexes1727345240782 implements MigrationInterface {
+  name = 'SourceFlagsIndexes1727345240782';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_source_flags_posts_members_threshold" ON source(((flags->'totalPosts')::integer), ((flags->'totalMembers')::integer), ((flags->'publicThreshold')::boolean))`,
+    );
+    await queryRunner.query(
+      `UPDATE "public"."source" SET flags = flags || '{"publicThreshold": false}' WHERE type = 'squad'`,
```
rebelchris, line 11: Should we not set everything to true for now?

src/schema/sources.ts
```diff
@@ -1312,6 +1313,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         (builder) => {
           builder.queryBuilder
             .andWhere(filter)
+            .andWhere(
+              `(${builder.alias}.flags->'publicThreshold')::boolean IS TRUE`,
+            )
```
rebelchris, line 1318: But this is on sources right?
Do we only use it for squads now?

rebelchris (changes requested):
Minor feedback

src/migration/1727345240782-SourceFlagsIndexes.ts
sshanzel: We discussed to update this a bit in the earlier phase. I guess it is better to put the right value now rather than wait for the cron job.

src/cron/updateSourcePublicThreshold.ts
sshanzel: Should be part of the conditions now.

src/schema/sources.ts
sshanzel: Yes, though let's maybe add this filter only when `filterOpenSquads` is present.

src/migration/1727345240782-SourceFlagsIndexes.ts
sshanzel: Updated now, would love to hear what you think.

PR #2283, opened 2024-09-25
chore: add query read replica implementation

rebelchris:
Add implementations for read replica on all `graphorm.query` calls that can use it.

### DIFF
--- a/src/schema/actions.ts
+++ b/src/schema/actions.ts
@@ -82,12 +82,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   Query: {
     actions: (_, __, ctx: AuthContext, info): Promise<GQLUserAction[]> =>
       graphorm.query<GQLUserAction>(ctx, info, (builder) => {
-        builder.queryBuilder = builder.queryBuilder.where(
-          {
-            userId: ctx.userId,
-          },
-          true,
-        );
+        builder.queryBuilder = builder.queryBuilder.where({
+          userId: ctx.userId,
+        });
 
         return builder;
       }),
--- a/src/schema/actions.ts
+++ b/src/schema/actions.ts
@@ -82,9 +82,12 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   Query: {
     actions: (_, __, ctx: AuthContext, info): Promise<GQLUserAction[]> =>
       graphorm.query<GQLUserAction>(ctx, info, (builder) => {
-        builder.queryBuilder = builder.queryBuilder.where({
-          userId: ctx.userId,
-        });
+        builder.queryBuilder = builder.queryBuilder.where(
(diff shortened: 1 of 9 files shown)


omBratteng (approved):

src/schema/actions.ts
```diff
@@ -82,9 +82,12 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   Query: {
     actions: (_, __, ctx: AuthContext, info): Promise<GQLUserAction[]> =>
       graphorm.query<GQLUserAction>(ctx, info, (builder) => {
-        builder.queryBuilder = builder.queryBuilder.where({
-          userId: ctx.userId,
-        });
+        builder.queryBuilder = builder.queryBuilder.where(
```
capJavert, line 85: actions might be tricky because for some we rely to not show popups when action is done, so people could see popups multiple times if read replica does not get the action update

src/schema/comments.ts
```diff
@@ -549,12 +549,17 @@ const getCommentById = async (
   ctx: Context,
   info: GraphQLResolveInfo,
 ): Promise<GQLComment> => {
-  const res = await graphorm.query<GQLComment>(ctx, info, (builder) => {
-    builder.queryBuilder = builder.queryBuilder
-      .andWhere(`${builder.alias}.id = :id`, { id })
-      .limit(1);
-    return builder;
-  });
+  const res = await graphorm.query<GQLComment>(
```
capJavert, line 552: So here we can't use read replica because you are trying to return modified data

src/schema/posts.ts
```diff
@@ -1068,13 +1068,20 @@ const getPostById = async (
   info: GraphQLResolveInfo,
   id: string,
 ): Promise<GQLPost> => {
-  const res = await graphorm.query<GQLPost>(ctx, info, (builder) => ({
-    ...builder,
-    queryBuilder: builder.queryBuilder.where(
-      `"${builder.alias}"."id" = :id AND "${builder.alias}"."deleted" = false`,
-      { id },
+  const res = await graphorm.query<GQLPost>(
```
capJavert, line 1071: also used in mutations

src/schema/sources.ts
```diff
@@ -1093,12 +1093,17 @@ const getSourceById = async (
   info: GraphQLResolveInfo,
   id: string,
 ): Promise<GQLSource> => {
-  const res = await graphorm.query<GQLSource>(ctx, info, (builder) => {
-    builder.queryBuilder = builder.queryBuilder
-      .andWhere('(id = :id or handle = :id)', { id })
-      .limit(1);
-    return builder;
-  });
+  const res = await graphorm.query<GQLSource>(
```
capJavert, line 1096: also mutations

capJavert (approved):

PR #2286, opened 2024-09-27
chore: add read replica support to queryPaginated

rebelchris:
Read replica support for queryPaginated.
Pretty sure it's the most safe one to implement on as it should only be singular reads without mutations happening before.


idoshamun (approved):

src/common/datePageGenerator.ts
```diff
@@ -93,5 +93,7 @@ export function queryPaginatedByDate<
       }
       return builder;
     },
+    undefined,
```
capJavert, line 96: not sure about this syntax, its gonna look bad

capJavert: i might have an idea for something else, we can try monday

rebelchris: Decided to do as separate PR

capJavert (approved):
Good from my end, just check the test.

PR #2288, opened 2024-09-27
chore: optimize comments schema

rebelchris:
Related to read replica usage, but went wider to see where we could optimize the comments schema by using more narrow selects.


src/graphorm/graphorm.ts
```diff
@@ -677,6 +677,7 @@ export class GraphORM {
    * @param nodeToCursor A function that creates a cursor from a node
    * @param beforeQuery A callback function that is called before executing the query
    * @param transformNodes Apply any transformation on the nodes before adding page info
+   * @param readReplica Whether to use the read replica instance
```
rebelchris, line 680: This one is basically the same as: https://github.com/dailydotdev/daily-api/pull/2286

idoshamun (approved):

omBratteng (approved):

src/schema/comments.ts
```diff
@@ -628,6 +630,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
             return builder;
           },
+          readReplica: true,
```
capJavert, line 633: For postComments just to make sure we do client side optimistic vs refetching postComments because that would mean read replica might not have a comment change yet.

capJavert (approved):
Only one comment (pun intended) to check.

PR #2289, opened 2024-09-29
feat: create new function that updates vordr flags on posts

omBratteng:
Just a quick attempt at a db trigger that also sets the correct vordr flags on posts when a user gets vordr'ed.

### DIFF
--- a/src/migration/1727578711381-VordrUser.ts
+++ b/src/migration/1727578711381-VordrUser.ts
@@ -4,6 +4,8 @@ export class VordrUser1727578711381 implements MigrationInterface {
   name = 'VordrUser1727578711381'
 
   public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE INDEX "IDX_post_flags_vordr" ON post USING HASH (((flags->'vordr')::boolean))`);
+
     await queryRunner.query(`CREATE OR REPLACE FUNCTION vordr_update_flags()
 RETURNS TRIGGER AS $$
 BEGIN
@@ -62,6 +64,8 @@ EXECUTE FUNCTION vordr_update_flags();`);
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "IDX_post_flags_vordr"`);
+
     await queryRunner.query(`CREATE OR REPLACE FUNCTION comment_flags_update_vordr()
 RETURNS TRIGGER AS $$
 BEGIN
--- a/src/migration/1727578711381-VordrUser.ts
+++ b/src/migration/1727578711381-VordrUser.ts
@@ -36,7 +36,8 @@ BEGIN
         '{vordr}', 'false', true),        -- set vordr to false
         '{banned}', 'false', true)        -- set banned to false
     WHERE
-      ("authorId" = NEW.id OR "scoutId" = NEW.id);
+      ("authorId" = NEW.id OR "scoutId" = NEW.id)
+      AND flags @> '{"vordr": true}';
 
     UPDATE
       comment
--- a/src/migration/1727578711381-VordrUser.ts
+++ b/src/migration/1727578711381-VordrUser.ts
@@ -4,7 +4,7 @@ export class VordrUser1727578711381 implements MigrationInterface {
   name = 'VordrUser1727578711381'
 
   public async up(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(`CREATE INDEX "IDX_post_flags_vordr" ON post USING HASH (((flags->'vordr')::boolean))`);
+    await queryRunner.query(`CREATE INDEX "IDX_post_flags_vordr" ON post (((flags->'vordr')::boolean))`);
 
     await queryRunner.query(`CREATE OR REPLACE FUNCTION vordr_update_flags()
 RETURNS TRIGGER AS $$
--- a/src/migration/1727578711381-VordrUser.ts
+++ b/src/migration/1727578711381-VordrUser.ts
@@ -17,7 +17,7 @@ BEGIN
         '{vordr}', 'true', true),         -- set vordr to true
         '{banned}', 'true', true)         -- set banned to true
     WHERE
-      "authorId" = NEW.id;
+      ("authorId" = NEW.id OR "scoutId" = NEW.id);
 
     UPDATE
       comment
@@ -36,7 +36,7 @@ BEGIN
         '{vordr}', 'false', true),        -- set vordr to false
         '{banned}', 'false', true)        -- set banned to false
     WHERE
-      "authorId" = NEW.id;
+      ("authorId" = NEW.id OR "scoutId" = NEW.id);
 
     UPDATE
       comment
--- a/src/migration/1727578711381-VordrUser.ts
+++ b/src/migration/1727578711381-VordrUser.ts
@@ -57,7 +57,7 @@ $$ LANGUAGE plpgsql;`);
     await queryRunner.query(`CREATE OR REPLACE TRIGGER user_flags_vordr
 AFTER UPDATE OF flags ON "user"
 FOR EACH ROW
-WHEN ((OLD.flags ->> 'vordr')::boolean IS DISTINCT FROM ((NEW.flags ->> 'vordr'))::boolean)
+WHEN (OLD.flags -> 'vordr' IS DISTINCT FROM NEW.flags -> 'vordr')
 EXECUTE FUNCTION vordr_update_flags();`);
 
     await queryRunner.query(`DROP FUNCTION comment_flags_update_vordr`);
--- a/src/migration/1727578711381-VordrUser.ts
+++ b/src/migration/1727578711381-VordrUser.ts
@@ -14,7 +14,7 @@ BEGIN
       post
     SET
       "banned" = TRUE,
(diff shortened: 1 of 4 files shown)


omBratteng:
@idoshamun perhaps we need to add something to the where statements so we don't accidentally revert some banned/showOnFeed when a user gets un-vordr'ed

src/migration/1727578711381-VordrUser.ts
```diff
@@ -0,0 +1,87 @@
...
+      post
+    SET
+      "showOnFeed" = TRUE,
+      "banned" = FALSE,
+      "flags" = jsonb_set(jsonb_set(jsonb_set(flags,
+        '{showOnFeed}', 'false', true),   -- set showOnFeed to true
+        '{vordr}', 'false', true),        -- set vordr to false
+        '{banned}', 'false', true)        -- set banned to false
+    WHERE
+      "authorId" = NEW.id;
```
omBratteng, line 43: Probably check where vordr is already set to true?

src/migration/1727578711381-VordrUser.ts
```diff
@@ -0,0 +1,87 @@
...
+BEGIN
+  IF NEW.flags @> '{"vordr": true}' THEN
+    -- If vordr is true, set showOnFeed to false and banned to true in post table
+    UPDATE
+      post
+    SET
+      "showOnFeed" = FALSE,
+      "banned" = TRUE,
+      "flags" = jsonb_set(jsonb_set(jsonb_set(flags,
+        '{showOnFeed}', 'false', true),   -- set showOnFeed to false
```
idoshamun, line 18: `banned` is enough we don't need `showOnFeed`

src/migration/1727578711381-VordrUser.ts
idoshamun: Yes, it's better to check that it's vordr

omBratteng: 👍

src/migration/1727578711381-VordrUser.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    UPDATE
+      comment
+    SET
+      "flags" = jsonb_set(flags, '{vordr}', 'true', true)
+    WHERE
+      "userId" = NEW.id;
+
+  ELSIF NEW.flags @> '{"vordr": false}' THEN
+    -- If vordr is false, set showOnFeed to true and banned to false in post table
```
capJavert, line 32: Could it be that the post was banned outside of vordr and then with this we retro unban?

omBratteng: Yeah, so should probably add to the where statement where vordr flag is true. So it would only undo vordr'ed posts

omBratteng: Added `AND flags @> '{"vordr": true}'`

capJavert: does `@>` support JSON fields? eg. if flags are `{"deleted":true,"vordr":false}` will it also match?

omBratteng: Yeah
> ### jsonb @> jsonb → boolean
>
> > Does the first JSON value contain the second? (See [Section 8.14.3](https://www.postgresql.org/docs/current/datatype-json.html#JSON-CONTAINMENT) for details about containment.)
> >
> > '{"a":1, "b":2}'::jsonb @> '{"b":2}'::jsonb → t

omBratteng: I guess alternative statement would be `AND flags->>'vordr' = 'true'`

omBratteng: OR `AND (flags->>'vordr')::boolean = true`

src/migration/1727578711381-VordrUser.ts
```diff
@@ -0,0 +1,90 @@
...
+    UPDATE
+      comment
+    SET
+      "flags" = jsonb_set(flags, '{vordr}', 'true', true)
+    WHERE
+      "userId" = NEW.id;
+
+  -- If vordr on user is set to false, and vordr on post is true
+  -- Set banned and vordr to false in post table, and set vordr to false in comment table
+  ELSIF NEW.flags @> '{"vordr": false}' THEN
```
capJavert, line 34: Just interested, currently we rarely/never set vordr back to false?

src/migration/1727578711381-VordrUser.ts
```diff
@@ -0,0 +1,90 @@
...
+      "userId" = NEW.id;
+  END IF;
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;`);
+
+    await queryRunner.query(`CREATE OR REPLACE TRIGGER user_flags_vordr
+AFTER UPDATE OF flags ON "user"
+FOR EACH ROW
+WHEN ((OLD.flags ->> 'vordr')::boolean IS DISTINCT FROM ((NEW.flags ->> 'vordr'))::boolean)
```
capJavert, line 60: You don't have to cast to string `->>` and then `boolean` again? Using only `((OLD.flags -> 'vordr')::boolean` should be fine?

capJavert: Also `IS DISTINCT` works when flags->vordr is null/empty?

src/migration/1727578711381-VordrUser.ts
capJavert: Not sure if this is covered again? https://github.com/dailydotdev/daily-api/pull/2289#discussion_r1780797883, if possible at test case for it

omBratteng: It's covered in the test `should update all posts the user has made when the vordr flag is set to false`
But I could add an explicit test for it

src/migration/1727578711381-VordrUser.ts
omBratteng: > Also IS DISTINCT works when flags->vordr is null/empty?

Pretty sure, could add test case for it

src/migration/1727578711381-VordrUser.ts
omBratteng: Added explicit test

src/migration/1727578711381-VordrUser.ts
```diff
@@ -0,0 +1,91 @@
...
+    UPDATE
+      post
+    SET
+      "banned" = TRUE,
+      "flags" = jsonb_set(jsonb_set(flags,
+        '{vordr}', 'true', true),         -- set vordr to true
+        '{banned}', 'true', true)         -- set banned to true
+    WHERE
+      ("authorId" = NEW.id OR "scoutId" = NEW.id)
+      AND "banned" = FALSE;
```
omBratteng, line 24: @idoshamun I added a check here to not set vordr on already banned posts. Since it would basically negate the check when un-vordrering, since banned posts would then match the `(flags->'vordr')::boolean = true` statement.

src/migration/1727578711381-VordrUser.ts
omBratteng: > You don't have to cast to string ->> and then boolean again? Using only ((OLD.flags -> 'vordr')::boolean should be fine?

Changed it to just be `OLD.flags -> 'vordr' IS DISTINCT FROM NEW.flags -> 'vordr'`

omBratteng: > Also IS DISTINCT works when flags->vordr is null/empty?
Yes.

It is already in the tests when changing userId 1, it doesn't have vordr flag set. Or do you want explicit test for it?

capJavert: That's fine if works 👍 Was more concerned for this case https://github.com/dailydotdev/daily-api/pull/2289#discussion_r1784296841

capJavert (approved):
Looks ok to me, maybe wait for Ido as well.

omBratteng:
Yup, will wait for @idoshamun

rebelchris (approved):
Looks ok to me :) 
We might have to add squads to this function soon as well :)

src/migration/1727578711381-VordrUser.ts
idoshamun: The only issue I can think of now is that banning the posts will result in penalization lost which will make this process very visible. wdyt?

omBratteng: Penalization of the old posts by them?

idoshamun: yes. every time we ban a post it penalized the author/scout

omBratteng: But this is when we vordr an entire account?

idoshamun: yes, instead of banning let's just set `showOnFeed` false as well as `vordr`

omBratteng: 👍 will change

Perhaps we also could adjust the API so that vordr flag has the same effect as showOnFeed, so that we don't need to set that too?

idoshamun: We will need to change the feed as well. It's just too many places I think

omBratteng: @idoshamun changed to be `showOnFeed=false` instead of `banned=true`

idoshamun (approved):

PR #2291, opened 2024-09-29
fix: reduce members limit for source

idoshamun:
By default we used to return 50 members and we show only 5.
 That's a lot of unused data. and I mean A LOT! 🤯

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -408,8 +408,8 @@ const obj = new GraphORM({
               }),
         },
         pagination: {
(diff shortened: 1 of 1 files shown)


omBratteng (approved):

omBratteng:
Can we control it from frontend? I'm thinking we want to get more users when viewing the member list

idoshamun:
@omBratteng when we need pagination we use a different query `SourceMembers`

omBratteng:
Ah, then it's fine

PR #2292, opened 2024-09-30
test: timezone edge case on clearing streak

sshanzel:
- Test a specific edge case. Reference: https://github.com/dailydotdev/daily/issues/1464
- Add a column to better understand when the clearing happened.


sshanzel:
@capJavert updated to use logger instead.

capJavert (approved):

nensidosari (approved):

sshanzel:
@capJavert one last thing before I merge, just wanted it to be know. I think the average amount of users getting cleared of their streak per hour is around 1k. Would that be too much of logging?

capJavert:
@sshanzel let's merge and see how today goes, I would not say its a lot for a limited time.

rebelchris (approved):

PR #2296, opened 2024-10-01
feat: remote config key as env

capJavert:
Just to make it easier when working locally to replace config key with something else without doing any code changes that can be committed accidentally.

### DIFF
--- a/src/remoteConfig.ts
+++ b/src/remoteConfig.ts
@@ -11,7 +11,6 @@ type RemoteConfigValue = {
 
 class RemoteConfig {
   private gb?: GrowthBook;
-  private readonly configKey = 'api-config';
 
   async init(): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/remoteConfig.ts
```diff
@@ -43,7 +42,10 @@ class RemoteConfig {
       throw new Error('remote config not initialized');
     }
 
-    const result = this.gb.getFeatureValue(this.configKey, null);
+    const result = this.gb.getFeatureValue(
+      process.env.API_CONFIG_FEATURE_KEY,
```
omBratteng, line 46: Add a check like with `GROWTHBOOK_API_CONFIG_CLIENT_KEY` env?

omBratteng (approved):

src/remoteConfig.ts
capJavert: no need, it will just be null

omBratteng: 👍

PR #2297, opened 2024-10-01
fix: user post duplicate notification

capJavert:
So the issue was that on join with `notification_preference` table I did not include `notificationType`, which would mean for users that subscribed to some other notificationType their rows would be returned inside the worker.
image

With added condition on join now we only get single row per user (NULL if no preference or subscribed if user resubscribed to type)
image


omBratteng (approved):

rebelchris (approved):
Looks safe with the additional type.

rebelchris (approved):
Looks safe, with the type addition

PR #2298, opened 2024-10-01
fix: non feed-service feeds ignored public threshold

idoshamun:
Add public threshold support for API generated feeds


__tests__/__snapshots__/compatibility.ts.snap
```diff
@@ -224,18 +224,6 @@ Object {
       "title": "P1",
       "url": "http://p1.com",
     },
-    Object {
```
idoshamun, line 227: Not sure why this post was even part of the feeds as it was part of a private source. This is true everywhere

__tests__/fixture/post.ts
```diff
@@ -137,6 +137,7 @@ export const postsFixture: DeepPartial<ArticlePost | SharePost>[] = [
     type: PostType.Article,
     origin: PostOrigin.Squad,
     yggdrasilId: 'd1053f05-4d41-4fc7-885c-c0f7c841a7b6',
+    private: true,
```
idoshamun, line 140: Source is private so I add to set it here as well

src/common/feedGenerator.ts
```diff
@@ -248,31 +249,33 @@ export const applyFeedWhere = (
   postTypes: string[],
   removeHiddenPosts = true,
   removeBannedPosts = true,
-  allowPrivateSources = true,
+  allowPrivatePosts = true,
```
idoshamun, line 252: Renaming for better terminology

src/schema/bookmarks.ts
```diff
@@ -379,7 +379,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         bookmarksFeedBuilder(ctx, unreadOnly, listId, builder, alias),
       bookmarkPageGenerator,
       applyBookmarkPaging,
-      { removeHiddenPosts: false, removeBannedPosts: false },
+      {
+        removeHiddenPosts: false,
+        removeBannedPosts: false,
+        removeNonPublicThresholdSquads: false,
```
rebelchris, line 385: Hmm I see :)

src/schema/feeds.ts
```diff
@@ -1203,6 +1205,8 @@ const feedResolverCursor = feedResolver<
         ...(args?.refresh && { refresh: true }),
       }),
     warnOnPartialFirstPage: true,
+    // Feed service should take care of this
+    removeNonPublicThresholdSquads: false,
```
rebelchris, line 1209: But still why do we ship it here then?

rebelchris (approved):
Not much to add.
Just becoming a bit messy with which config actually does something I personally find. (Eg does api config do anything, lofn, feed itself)

src/schema/feeds.ts
idoshamun: I disabled this condition for the feed to make the query leaner on the API side. It is enabled by default

idoshamun:
@rebelchris these are all different configs for different purposes. we can consider offloading more things to feed or vice versa. biggest drawback of feed service is that it's not always up to date as the api

PR #2299, opened 2024-10-01
feat(strict-ts): apply to notifications

capJavert:

This one focuses on: 
- notifications
- notifications workers


src/notifications/generate.ts
```diff
@@ -34,7 +34,7 @@ const systemTitle = () => undefined;
 
 export const notificationTitleMap: Record<
   NotificationType,
-  (ctx: NotificationBaseContext) => string | undefined
+  (ctx: never) => string | undefined
```
capJavert, line 37: I've cast this as never because that is the only way (that I know of) to get dynamic types on `ctx` params in the `notificationTitleMap` below. Only caveat to this is that we have to cast to `never` in `generateNotificationV2` but think its better then casting in this file since `generateNotificationV2` is more internal function

src/notifications/index.ts
```diff
@@ -157,6 +157,10 @@ export async function generateAndStoreNotificationsV2(
   entityManager: EntityManager,
   args: NotificationHandlerReturn,
 ): Promise<void> {
+  if (!args) {
+    return;
+  }
```
capJavert, line 162: This never gets called with `args = undefined` because worker checks for it but for safety its here

src/notifications/builder.ts
```diff
@@ -299,7 +301,7 @@ export class NotificationBuilder {
     if (source.type === SourceType.Squad && addSquadAvatar) {
       newBuilder = newBuilder.avatarSource(source);
     }
-    if (post.type === PostType.Share) {
+    if (post.type === PostType.Share && sharedPost) {
```
rebelchris, line 304: Nice this might actually catch some edge-case fails 😅

rebelchris (approved):
Looks ok to me :)

src/notifications/builder.ts
capJavert: Yep, hopefully catch a lot more with strict enabled 🙏

PR #2301, opened 2024-10-01
refactor: milestone should not include the value 1

sshanzel:


rebelchris (approved):
Pro if we can add a test 😉

omBratteng (approved):

nensidosari (approved):

sshanzel:
> Pro if we can add a test 😉

Yep. Definitely beneficial. Let me add one.

PR #2303, opened 2024-10-02
feat: send traces to opentelemetry-collector

omBratteng:
Not ready for merge yet, but can be reviewed

### DIFF
--- a/.infra/Pulumi.prod.yaml
+++ b/.infra/Pulumi.prod.yaml
@@ -110,8 +110,7 @@ config:
     onesignalAppId:
       secure: AAABALTizkxZkCfsq6/fx2v7nXxsVP7psWBfSUct0zeyrX0ABxPpD30zIsyx5r5wqt1UJIIXDIVOnhM7W3vzrzaatrE=
     otelEnabled: true
-    otelTracesSampler: parentbased_traceidratio
-    otelTracesSamplerArg: 0.05
+    otelTracesSampler: parentbased_always_on
     personalizedDigestFeed:
       secure: AAABABnHXr0d1pbDX7KRjoBXvFnr7ImTUJ7Y88GlhlaC7EiSG/nxVLHtnQL0hgczHqt1bPn25TN/R8Au6CEAQpq6dGXLwCfA+gWPypTVHg==
     personalizedDigestSecret:
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -46,8 +46,8 @@ export class Context {
     return this.req.trackingId;
   }
 
-  get isTeamMember(): boolean | undefined {
-    return this.req.isTeamMember;
+  get isTeamMember(): boolean {
+    return !!this.req.isTeamMember;
   }
 
   get premium(): boolean {
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -46,6 +46,10 @@ export class Context {
     return this.req.trackingId;
   }
 
+  get isTeamMember(): boolean | undefined {
+    return this.req.isTeamMember;
+  }
+
   get premium(): boolean {
     return !!this.req.premium;
   }
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -585,9 +585,12 @@ export const checkIfUserIsTeamMember = async (
   con: DataSource,
   userId: string,
 ): Promise<boolean> => {
-  const isTeamMember = await con.getRepository(Feature).findOneBy({
-    feature: FeatureType.Team,
-    userId,
+  const isTeamMember = await con.getRepository(Feature).findOne({
+    select: ['value'],
+    where: {
+      feature: FeatureType.Team,
+      userId,
+    },
   });
 
   return isTeamMember?.value === 1;
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -5,8 +5,6 @@ import {
   UserStreak,
   UserStreakAction,
   UserStreakActionType,
-  Feature,
-  FeatureType,
 } from '../entity';
 import { differenceInDays, isSameDay, max, startOfDay } from 'date-fns';
 import { DataSource, EntityManager, In, Not } from 'typeorm';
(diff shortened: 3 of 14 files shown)


src/common/users.ts
```diff
@@ -571,6 +573,18 @@ export const shouldAllowRestore = async (
   });
 };
 
+export const checkIfUserIsTeamMember = async (
+  con: DataSource,
+  userId: string,
+): Promise<boolean> => {
+  const isTeamMember = await con.getRepository(Feature).findOneBy({
```
rebelchris, line 580: Maybe add specific select on value.

rebelchris (approved):
Can't say much about the telemetry stuff, but overall looks good, just one optimization.

src/telemetry/opentelemetry.ts
```diff
@@ -137,6 +141,15 @@ export const tracer = (serviceName: string) => {
       req.span = api.trace.getSpan(api.context.active());
     });
 
+    fastify.addHook('preHandler', async (req) => {
+      if (req?.span?.isRecording()) {
+        const con = await createOrGetConnection();
+        const isTeamMember = await checkIfUserIsTeamMember(con, req.userId);
```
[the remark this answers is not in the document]
omBratteng: From what I gather, when using diagnostic channels, it shouldn't really affect response times much. But before I went to bed, I noticed that `isTeamMember` is in the boot, so probably in JWT as well. If not, I'll add it to the JWT

src/common/users.ts
omBratteng: Changed to only select value

omBratteng: Though function isn't used that much for telemetry, since I moved that to the JWT instead. But still keeping it in since it's being re-used other places

src/telemetry/opentelemetry.ts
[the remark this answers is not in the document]
omBratteng: Refactored to use JWT instead since we're already checking for `isTeamMember` during boot.

.infra/Pulumi.prod.yaml
```diff
@@ -110,8 +110,7 @@ config:
     onesignalAppId:
       secure: AAABALTizkxZkCfsq6/fx2v7nXxsVP7psWBfSUct0zeyrX0ABxPpD30zIsyx5r5wqt1UJIIXDIVOnhM7W3vzrzaatrE=
     otelEnabled: true
-    otelTracesSampler: parentbased_traceidratio
-    otelTracesSamplerArg: 0.05
```
capJavert, line 114: Where did sampling go?

omBratteng: `parentbased_always_on`
So API always samples, and sends it to the collector, which then decides if out should be sampled or not (tail sampling)
That way we can enable 100% sampling for staff or errors, all others have a percentage rate

omBratteng: See here https://github.com/dailydotdev/daily-infra/blob/7dad00fd2c19cc45f7999b48591f58b318b020a0/tracing.ts#L71-L87

capJavert: my next question was what `parentbased_traceidratio` does 😆 , cool!

omBratteng: `parentbased` it takes into account what the sampling rate is set on the parent. I.e. if API was called from a different service that was also doing tracing, but it decided to not sample this request. That would make it respect that.

`traceidratio` just sampling rate.

`always_on` ... always on

`parentbased_traceidratio` Respects parent decision, but if none, sets sampling rate

`parentbased_always_on` Respects parent decision, but if none, always on

src/telemetry/opentelemetry.ts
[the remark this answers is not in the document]
rebelchris: It was a discussion at the time to add to JWT, but we decided not to as it was a once-off.
Now it makes fully sense so thanks for adding it :)

src/schema/feeds.ts
```diff
@@ -1407,10 +1406,12 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-          const isTeamMember = await ctx.con
-            .getRepository(Feature)
-            .findOneByOrFail({ feature: FeatureType.Team, userId: ctx.userId });
-          if (!isTeamMember || isTeamMember.value !== 1) {
+          const isTeamMember = await checkIfUserIsTeamMember(
+            ctx.con,
+            ctx.userId,
+          );
+
+          if (!isTeamMember) {
```
rebelchris, line 1414: This can now actually read it from the JWT right?

src/schema/submissions.ts
```diff
@@ -136,9 +134,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     ): Promise<GQLSubmitArticleResponse> => {
       const [user, isTeamMember] = await Promise.all([
         ctx.getRepository(User).findOneByOrFail({ id: ctx.userId }),
-        ctx.con
-          .getRepository(Feature)
-          .findOneBy({ feature: FeatureType.Team, userId: ctx.userId }),
+        await checkIfUserIsTeamMember(ctx.con, ctx.userId),
```
rebelchris, line 137: Here as well

rebelchris (approved):
Since it's in JWT we can remove the extra queries i think :)

src/Context.ts
```diff
@@ -46,6 +46,10 @@ export class Context {
     return this.req.trackingId;
   }
 
+  get isTeamMember(): boolean | undefined {
+    return this.req.isTeamMember;
+  }
```
capJavert, line 51:
```suggestion
  get isTeamMember(): boolean {
    return !!this.req.isTeamMember;
  }
```

rebelchris (approved):
Looks good to me!

capJavert (commented):
Looks good, but let's merge on Monday to avoid any issues over weekend.

PR #2304, opened 2024-10-02
feat: allow filtering of squads

rebelchris:
Added filtering option to set which sourceTypes to return.

Related Frontend: https://github.com/dailydotdev/apps/pull/3612

AS-625 #done

### DIFF
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -29,7 +29,7 @@ import graphorm from '../graphorm';
 import { mapArrayToOjbect } from './object';
 import { runInSpan } from '../telemetry';
 import { whereVordrFilter } from './vordr';
-import { baseFeedConfig } from '../integrations/feed/generators';
+import { baseFeedConfig } from '../integrations/feed';
 
 export const WATERCOOLER_ID = 'fd062672-63b7-4a10-87bd-96dcd10e9613';
 
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -373,7 +373,6 @@ export function feedResolver<
       (el) =>
         !(queryParams as AnonymousFeedFilters).excludeSourceTypes.includes(el),
     );
-    console.log('qp', sourceTypes);
 
     const result = await runInSpan('feedResolver.queryPaginated', async () =>
       graphorm.queryPaginated<GQLPost>(
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -29,6 +29,7 @@ import graphorm from '../graphorm';
 import { mapArrayToOjbect } from './object';
 import { runInSpan } from '../telemetry';
 import { whereVordrFilter } from './vordr';
+import { baseFeedConfig } from '../integrations/feed/generators';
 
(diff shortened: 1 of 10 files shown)


src/common/feedGenerator.ts
```diff
@@ -355,6 +369,12 @@ export function feedResolver<
       return excludedTypes ? !excludedTypes.includes(type) : true;
     });
 
+    const sourceTypes = baseFeedConfig.source_types.filter(
+      (el) =>
+        !(queryParams as AnonymousFeedFilters).excludeSourceTypes.includes(el),
+    );
+    console.log('qp', sourceTypes);
```
rebelchris, line 376:
```suggestion
```

src/common/personalizedDigest.ts
```diff
@@ -216,6 +217,9 @@ export const getPersonalizedDigestEmailPayload = async ({
     blocked_tags: feedConfig.blockedTags,
     blocked_sources: feedConfig.excludeSources,
     feed_config_name: feature.feedConfig,
+    source_types: baseFeedConfig.source_types.filter(
+      (el) => !feedConfig.excludeSourceTypes.includes(el),
+    ),
```
capJavert, line 222: Digest config supports it, I know in the past we had to add/enabled specifically, maybe check with Vas?

src/common/feedGenerator.ts
```diff
@@ -278,6 +286,12 @@ export const applyFeedWhere = (
     );
   }
 
+  if (sourceTypes.length > 0) {
+    newBuilder = newBuilder.andWhere(`source.type IN (:...sourceTypes)`, {
+      sourceTypes,
+    });
+  }
```
capJavert, line 293: Index might be needed for this?

src/common/feedGenerator.ts
```diff
@@ -162,23 +163,28 @@ export const feedToFilters = async (
...
-        if (curr.group === 'content_types') {
-          acc.excludeTypes.push(curr.options.type);
-        }
-        if (curr.group === 'content_curation') {
-          acc.blockedContentCuration.push(curr.options.type);
+  const { excludeTypes, blockedContentCuration, excludeSourceTypes } =
+    settings.reduce<{
+      excludeTypes: string[];
+      blockedContentCuration: string[];
+      excludeSourceTypes: string[];
```
capJavert, line 170: We need to add this to `FiltersInput` gql type?

capJavert (commented):
Some questions mostly

src/common/personalizedDigest.ts
rebelchris: Yeah checked, he's adding it to all:
https://dailydotdev.slack.com/archives/C04UN615UAF/p1727855255223979

src/common/feedGenerator.ts
rebelchris: They actually never really come from frontend though, but think we might use it via internal graph resolver on API

src/common/feedGenerator.ts
rebelchris: Index already exists.

capJavert (approved):
Looks alright to me 👍

omBratteng (approved):
LGTM, just need to be sure it actually gets passed on to feed.
When I added the video filtering, I only had it working for feed v1 🙈

PR #2305, opened 2024-10-02
feat: enable strict type checking

capJavert:
This is the final of the PRs in a series to bring our TypeScript types to comply with [strict](https://www.typescriptlang.org/tsconfig/#strict) config option. This will allow us better type checking during builds and development and allow us to ship with more confidence.

Left comments on things that need attention.


src/workers/notifications/userStreakResetNotification.ts
```diff
@@ -32,7 +32,7 @@ const worker = generateTypedNotificationWorker<'api.v1.user-streak-updated'>({
       streak: {
         ...streak,
         currentStreak: parseInt(lastStreak, 10),
-        lastViewAt: new Date(streak.lastViewAt!).getTime(),
+        lastViewAt: new Date(streak.lastViewAt!),
```
capJavert, line 35: @sshanzel, for sanity check, strict type requires Date, from what I see in notification handler `format` accepts `Date | number | string` so it should be good?

src/schema/users.ts
```diff
@@ -2116,7 +2125,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         amount: recoverCost,
       };
 
-      const currentStreak = oldStreakLength + streak.current;
+      const currentStreak = oldStreakLength + streak.current!;
```
capJavert, line 2128: @sshanzel for you sanity check, here `streak.current` could be `undefined`, I unwrapped it with `!` operator, but we might want to return early if `streak.current` does not exist? Maybe cause of some bugs?

src/schema/users.ts
```diff
@@ -1904,7 +1909,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         throw new ValidationError('We can only verify unique company domains');
       }
 
-      const company = await ctx.con.getRepository(Company).findOneBy({
+      const company = await ctx.con.getRepository(Company).findOneByOrFail({
```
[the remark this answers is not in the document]
rebelchris: Yeah should be fine indeed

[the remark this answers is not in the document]
capJavert: ah nvm, it actually is supposed to be nullable at least based on db type? I changed it back.

src/workers/notifications/userStreakResetNotification.ts
capJavert: test also fails for it because it expects number https://app.circleci.com/pipelines/github/dailydotdev/daily-api/9926/workflows/6d3c9906-2592-4ff0-a78d-33267a4a1751/jobs/39988/tests

src/schema/users.ts
sshanzel: Should always exist. When a user is created, we trigger a user streak insert then the default value for this property is 0.

src/schema/users.ts
```diff
@@ -1985,16 +1990,20 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       await ctx.con.getRepository(UserCompany).save(updatedRecord);
 
-      return await graphorm.queryOne<GQLUserCompany>(ctx, info, (builder) => {
-        builder.queryBuilder = builder.queryBuilder
-          .andWhere(`${builder.alias}."userId" = :userId`, {
-            userId: ctx.userId,
-          })
-          .andWhere(`${builder.alias}."email" = :email`, { email })
-          .andWhere(`${builder.alias}."verified" = true`);
+      return await graphorm.queryOneOrFail<GQLUserCompany>(
```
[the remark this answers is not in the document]
sshanzel: Yes.

src/workers/notifications/userStreakResetNotification.ts
sshanzel: I vaguely remember if there was an issue with `format` accepting the value. But if it works, then yeah, we should be good here too.

sshanzel: Maybe we can manually update the expected value from int to string since that is what will happen now.

src/schema/users.ts
capJavert: I made `current` not optional in, can you check if ok @sshanzel? 1191e9cbf75a52a1ea21c45791987350daa57e6a

rebelchris (approved):
Wow massive kudos! 🎉

src/schema/users.ts
sshanzel: Yes. Having it required is the way to go 🙌

sshanzel (approved):
Wow, is this the final PR for fully enabling type-script on API? Well done! 🎉

PR #2308, opened 2024-10-03
feat: add option to shuffle tags

omBratteng:
Adds option to shuffle tags during tag selection

- [x] Add tests

AS-619

### DIFF
--- a/src/schema/tags.ts
+++ b/src/schema/tags.ts
@@ -111,6 +111,11 @@ export const typeDefs = /* GraphQL */ `
       Tags which should be excluded from recommendations
       """
       excludedTags: [String]!
+
+      """
+      Whether to shuffle the tags
+      """
+      shuffle: Boolean = false
     ): TagResults!
   }
 `;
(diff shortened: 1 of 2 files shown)


rebelchris (approved):
This part looks good to me :)

src/schema/tags.ts
```diff
@@ -192,13 +206,17 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         .orderBy('value', 'ASC')
         .execute();
 
+      if (shuffle) {
+        hits.sort(() => Math.random() - 0.5);
+      }
```
capJavert, line 211: Hmm while implementation is fine, I would ask is it better to randomize on the frontend? Feels like a niche thing which would not much sense to do on API because you would get random result each time, while on frontend you can actually control the seed and when you randomize. Does that make sense?

omBratteng: I am fine with either. Thoughts @rebelchris ?

rebelchris: Moved to frontend:
https://github.com/dailydotdev/apps/pull/3617/commits/cd68200837abfb16e434d29749f1250c58d603fc

Kept the recommendation the same as they ranked by probability and that seemed odd to shuffle.

omBratteng: Should I close this then?

PR #2310, opened 2024-10-04
fix: user post duplicate user notification

capJavert:
AS-602

Iteration on https://github.com/dailydotdev/daily-api/pull/2306, still using constraint approach but we now introduce `uniqueKey` for `user_notification` so we can dedup over notification type and other params. 

Unique key is built from notification type, referenceId and referenceType.

We unique constraint over `userId + uniqueKey` because we still want to have notifications across different users. 

By default `uniqueKey` is `NULL` and ignore by constraint.

### DIFF
--- a/src/entity/notifications/UserNotification.ts
+++ b/src/entity/notifications/UserNotification.ts
@@ -9,6 +9,14 @@ import { NotificationV2 } from './NotificationV2';
   'createdAt',
 ])
 @Index('IDX_user_notification_user_id_read_at', ['userId', 'readAt'])
+@Index(
+  'IDX_user_notification_userId_uniqueKey_unique',
+  ['userId', 'uniqueKey'],
+  {
+    unique: true,
+    where: `"uniqueKey" IS NOT NULL`,
+  },
+)
 export class UserNotification {
   @PrimaryColumn({ type: 'uuid' })
   notificationId: string;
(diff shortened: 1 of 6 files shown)


src/notifications/common.ts
```diff
@@ -235,3 +236,33 @@ export const getUnreadNotificationsCount = async (
...
+};
+
+export const generateUserNotificationUniqueKey = ({
+  type,
+  referenceId,
+  referenceType,
+}: {
+  type: NotificationType;
+  referenceId?: string;
+  referenceType?: NotificationReferenceType;
```
capJavert, line 259: I don't think we need `notification_v2.uniqueKey` here because `notificationId` on `user_notification` already dedups over it.

src/entity/notifications/UserNotification.ts
```diff
@@ -9,6 +9,14 @@ import { NotificationV2 } from './NotificationV2';
   'createdAt',
 ])
 @Index('IDX_user_notification_user_id_read_at', ['userId', 'readAt'])
+@Index(
+  'IDX_user_notification_userId_uniqueKey_unique',
+  ['userId', 'uniqueKey'],
+  {
+    unique: true,
+    where: `"uniqueKey" IS NOT NULL`,
```
rebelchris, line 17: 🤤

rebelchris (approved):
Looks good to me :)

omBratteng (approved):

PR #2311, opened 2024-10-04
fix: properly sync all topics to cio

capJavert:
> When sending `cio_subscription_preferences` from `cio.identify` (for user) we have to include all existing topics in the object. If not CIO will take what it has and make everything else a default value (that is set in cio topics area) and for stuff like digest and follow it will become "subscribed" again.

Above was hotfixed in https://github.com/dailydotdev/daily-api/pull/2314

In addition to above, this PR also sync other subs in the same way. 

In addition it also syncs digest after subscribe/unsubscribe mutations through `identifyUserPersonalizedDigest` util.

More discussion in [thread](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1728038592931229)

### DIFF
--- a/src/cio.ts
+++ b/src/cio.ts
@@ -17,6 +17,7 @@ import { FastifyBaseLogger } from 'fastify';
 import type { UserCompany } from './entity/UserCompany';
 import type { Company } from './entity/Company';
 import { DataSource } from 'typeorm';
+import { logger } from './logger';
 
 export const cio = new TrackClient(
   process.env.CIO_SITE_ID,
@@ -172,6 +173,31 @@ export async function identifyUserCompany(
   } catch (err) {
     if (err instanceof CustomerIORequestError && err.statusCode === 400) {
       log.warn({ err }, 'failed to update user company in cio');
+      return;
+    }
+    throw err;
+  }
+}
+
+export async function identifyUserPersonalizedDigest({
+  userId,
+  subscribed,
+}: {
+  cio: TrackClient;
+  userId: string;
+  subscribed: boolean;
+}): Promise<void> {
+  try {
+    await cio.identify(userId, {
+      [`cio_subscription_preferences.topics.topic_${CioUnsubscribeTopic.Digest}`]:
+        subscribed,
+    });
+  } catch (err) {
+    if (err instanceof CustomerIORequestError && err.statusCode === 400) {
+      logger.warn(
+        { err },
+        'failed to update user personalized digest subscription in cio',
+      );
       return;
     }
     throw err;
--- a/src/cio.ts
+++ b/src/cio.ts
@@ -13,7 +13,6 @@ import {
   getFirstName,
   getShortGenericInviteLink,
 } from './common';
-import { FastifyBaseLogger } from 'fastify';
 import type { UserCompany } from './entity/UserCompany';
 import type { Company } from './entity/Company';
 import { DataSource } from 'typeorm';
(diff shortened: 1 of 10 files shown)


rebelchris (approved):
Nice fix!

src/cio.ts
```diff
@@ -81,18 +87,34 @@ export async function identifyUserStreak(
...
+  con,
+  log,
+  cio,
+  user,
+}: {
+  con: DataSource;
+  log: FastifyBaseLogger;
+  cio: TrackClient;
+  user: ChangeObject<User>;
+}): Promise<void> {
```
capJavert, line 100: Refactored to props syntax because I had to add `con`, we prefer it either way.

rebelchris: Nice yeah we should just use object notation by default 🤦

src/schema/users.ts
```diff
@@ -1790,7 +1790,14 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
 
-      await resubscribeUser(cio, ctx.userId);
+      await Promise.all([
+        resubscribeUser(cio, ctx.userId),
+        identifyUserPersonalizedDigest({
+          userId: ctx.userId,
+          cio,
+          subscribed: true,
+        }),
+      ]);
```
capJavert, line 1800: Subscribes to digest in CIO

src/schema/users.ts
```diff
@@ -1810,6 +1817,12 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         });
       }
 
+      await identifyUserPersonalizedDigest({
+        userId: ctx.userId,
+        cio,
+        subscribed: false,
+      });
```
capJavert, line 1824: Unsubscribes from digest in CIO

src/schema/users.ts
```diff
@@ -1790,7 +1790,14 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         flags,
       });
 
-      await resubscribeUser(cio, ctx.userId);
+      await Promise.all([
+        resubscribeUser(cio, ctx.userId),
+        identifyUserPersonalizedDigest({
```
rebelchris, line 1795: Doesn't `identifyUser` already do all of it?

rebelchris (approved):
Overall good, just wondering if we should just only have the 1 generic function that does all of them at once?

src/schema/users.ts
capJavert: it does the digest part but I would have to fetch user data here to be able to use it so to me it made more sense to create a separate helper to also be able to jjst set digest part of subscription.

rebelchris: Ah ok let's start lean then :)

src/workers/userUpdatedCio.ts
```diff
@@ -24,7 +24,12 @@ const worker: TypedWorker<'user-updated'> = {
       await resubscribeUser(cio, user.id);
     }
 
-    await identifyUser(log, cio, user);
+    await identifyUser({
+      con,
+      log,
```
omBratteng, line 29: we can remove log, and just import logger. So we don't have to keep passing it around

capJavert: yeah, I used it in new function, will also add here while I am refactoring it either way

capJavert: Done in 9fe14dff7e5a4186290b734209f74fe4200368a5

omBratteng (approved):

idoshamun (approved):

PR #2312, opened 2024-10-04
feat: MI-580 - Added moderation required property to squads

AmarTrebinjac:
Introduces column "moderationRequired" which lets squads require approval on posts by users.

### DIFF
--- a/__tests__/sources.ts
+++ b/__tests__/sources.ts
@@ -1,14 +1,8 @@
-import {
-  disposeGraphQLTesting,
-  GraphQLTestClient,
-  GraphQLTestingState,
-  initializeGraphQLTesting,
-  MockContext,
-  saveFixtures,
-  testMutationError,
-  testMutationErrorCode,
-  testQueryErrorCode,
-} from './helpers';
+import { randomUUID } from 'crypto';
+import { DataSource, In, Not } from 'typeorm';
+import { updateFlagsStatement, WELCOME_POST_TITLE } from '../src/common';
+import { isNullOrUndefined } from '../src/common/object';
+import createOrGetConnection from '../src/db';
 import {
   defaultPublicSourceFlags,
   NotificationPreferenceSource,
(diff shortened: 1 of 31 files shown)


ilasw (commented):
Maybe we want to  add tests to make sure that no one can add both approval and only moderator option 👀

__tests__/sources.ts
```diff
@@ -1859,6 +1859,22 @@ describe('mutation createSquad', () => {
...
+      client,
+      {
+        mutation: MUTATION,
+        variables: {
+          ...variables,
+          moderationRequired: true,
+          memberPostingRole: SourceMemberRoles.Moderator,
+        },
+      },
+      'GRAPHQL_VALIDATION_FAILED',
```
capJavert, line 1874: Since other cases can throw `GRAPHQL_VALIDATION_FAILED` can you use `testMutationError` to check the error message specifically in addition to code, so we are sure correct error is returned?

__tests__/sources.ts
```diff
@@ -1859,6 +1859,22 @@ describe('mutation createSquad', () => {
     );
   });
 
+  it('Should throw error when post moderation is required, but posting is restricted to moderators only', async () => {
```
capJavert, line 1862: we usually use lowercase to keep it readable in outputs
```suggestion
  it('should throw error when post moderation is required, but posting is restricted to moderators only', async () => {
```

src/entity/Source.ts
```diff
@@ -165,4 +165,7 @@ export class SquadSource extends Source {
 
   @Column({ default: 0 })
   memberInviteRank: number;
+
+  @Column({ default: false })
```
capJavert, line 169: Missing this part of RFC in regarding default value when squad is created? https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1151500290/RFC-069+-+Admin+approval+on+Squads+posts#Squad-creation

src/schema/sources.ts
```diff
@@ -232,6 +232,11 @@ export const typeDefs = /* GraphQL */ `
     """
     memberInviteRole: String
 
+    """
+    Enable post moderation for the squad
+    """
+    moderationRequired: Boolean
```
capJavert, line 238: We need a test for query as well.

__tests__/sources.ts
AmarTrebinjac: Done!

rebelchris (approved):
I think we only missing a test to change the value of this field to verify it persists in database?
Otherwise no blockers so approving already.

PR #2319, opened 2024-10-07
fix: use subapp instead on api-bg

omBratteng:
This is to prevent the HPA for api-bg to scale up when digest subscription is filled

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -112,7 +112,7 @@ if (isPersonalizedDigestEnabled) {
     name,
     isAdhocEnv,
     addLabelsToWorkers(personalizedDigestWorkers, {
-      app: name,
+      app: `${name}-personalized-digest`,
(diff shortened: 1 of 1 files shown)


.infra/index.ts
```diff
@@ -112,7 +112,7 @@ if (isPersonalizedDigestEnabled) {
     name,
     isAdhocEnv,
     addLabelsToWorkers(personalizedDigestWorkers, {
-      app: name,
+      app: `${name}-personalized-digest`,
       subapp: 'personalized-digest',
```
omBratteng, line 116: Could probably remove the subapp label now

omBratteng: Nvm, I changed api-bg HPA instead so that it selects `subapp: background`

rebelchris (approved):

idoshamun (approved):

PR #2322, opened 2024-10-09
fix: integration changes

rebelchris:
Added two parts here:
- Redirect with current integration id (it's only needed for new squad via this integration) (@capJavert happy to hear your thoughts)
- Singular integration query, we can use this from FE to query above returned element :)

### DIFF
--- a/__tests__/integrations/slack.ts
+++ b/__tests__/integrations/slack.ts
@@ -150,7 +150,7 @@ describe('GET /integrations/slack/auth/callback', () => {
     ).expect(307);
 
     expect(body).toEqual({});
-    expect(headers.location).toBe('http://localhost:5002/squads/test');
+    expect(headers.location).toContain('http://localhost:5002/squads/test');
 
     const integration = await con
       .getRepository(UserIntegration)
@@ -184,7 +184,7 @@ describe('GET /integrations/slack/auth/callback', () => {
   });
 
   it('should update integration if it already exists', async () => {
-    await con.getRepository(UserIntegration).save([
+    const existingIntegration = await con.getRepository(UserIntegration).save([
       {
         userId: '1',
         type: UserIntegrationType.Slack,
@@ -219,7 +219,9 @@ describe('GET /integrations/slack/auth/callback', () => {
     ).expect(307);
 
     expect(body).toEqual({});
-    expect(headers.location).toBe('http://localhost:5002/squads/test');
+    expect(headers.location).toBe(
+      `http://localhost:5002/squads/test&iid=${existingIntegration[0].id}`,
+    );
 
     const integration = await con
       .getRepository(UserIntegration)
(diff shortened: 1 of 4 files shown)


omBratteng (approved):

src/schema/users.ts
```diff
@@ -1630,6 +1635,30 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         },
       );
     },
+    userIntegration: async (
```
capJavert, line 1638: We do fetch all integrations either way for dropdown so we can just maybe search inside `useIntegrationsQuery` or is it used somewhere where `useIntegrationsQuery` is not already triggered?

src/routes/integrations/slack.ts
```diff
@@ -187,7 +192,10 @@ export default async function (fastify: FastifyInstance): Promise<void> {
         );
       }
 
-      return redirectResponse({ res, path: redirectPath });
+      return redirectResponse({
+        res,
+        path: `${redirectPath}&iid=${integrationId}`,
```
capJavert, line 197: Let's use URL to assemble because we can't know if `redirectPath` already contains any search params or not

__tests__/integrations/slack.ts
```diff
@@ -150,7 +150,7 @@ describe('GET /integrations/slack/auth/callback', () => {
     ).expect(307);
 
     expect(body).toEqual({});
-    expect(headers.location).toBe('http://localhost:5002/squads/test');
+    expect(headers.location).toContain('http://localhost:5002/squads/test');
```
capJavert, line 153: We can parse it as URL and then check pathname + `iid` query, this feels like it can miss something with `toContain`

rebelchris: Yeah happy with that :)

src/routes/integrations/slack.ts
rebelchris: Good point :) I'll address

src/schema/users.ts
rebelchris: Yeah it's used quicker and singular actually.
So create squad with slack -> enter `new squad` page and here we need this singular integration only.

capJavert: 👍

capJavert: Then only if you can add a test for this query please

src/routes/integrations/slack.ts
```diff
@@ -52,6 +50,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
     const redirectPath = redirectPathCookie?.startsWith('/')
       ? redirectPathCookie
       : '/';
+    const redirectUrl = new URL(redirectPath, process.env.COMMENTS_PREFIX);
```
capJavert, line 53:
```suggestion
    const redirectUrl = new URL(`${process.env.COMMENTS_PREFIX}${redirectPath}`);
```

capJavert: passing `//somedomain.com` would ignore second argument of `new URL`

capJavert (approved):
Nice work 👌

PR #2324, opened 2024-10-09
fix: changed the default value of moderationRequired in createSquad

AmarTrebinjac:
Fix breaking changes


src/schema/sources.ts
```diff
@@ -1689,6 +1711,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         description,
         memberPostingRole = SourceMemberRoles.Member,
         memberInviteRole = SourceMemberRoles.Member,
+        moderationRequired,
```
AmarTrebinjac, line 1714: There was an issue with the fact that "moderationRequired" was set to true as a default value, but it cannot be changed in the frontend yet. This means, if one were to create a squad set with posting permission to moderator, it would cause a validation error.

src/entity/Source.ts
```diff
@@ -165,4 +165,7 @@ export class SquadSource extends Source {
 
   @Column({ default: 0 })
   memberInviteRank: number;
+
+  @Column({ default: true })
```
capJavert, line 169: Is it not default false per RFC as well?

capJavert: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1151500290/RFC-069+-+Admin+approval+on+Squads+posts#Table

sshanzel: Yeah, this should be `false` to be backwards compat.

__tests__/sources.ts
```diff
@@ -1713,8 +1746,8 @@ query RelatedTags($sourceId: ID!) {
 
 describe('mutation createSquad', () => {
   const MUTATION = `
-  mutation CreateSquad($name: String!, $handle: String!, $description: String, $postId: ID!, $commentary: String!, $memberPostingRole: String, $memberInviteRole: String, $categoryId: ID, $isPrivate: Boolean) {
-  createSquad(name: $name, handle: $handle, description: $description, postId: $postId, commentary: $commentary, memberPostingRole: $memberPostingRole, memberInviteRole: $memberInviteRole, categoryId: $categoryId, isPrivate: $isPrivate) {
+  mutation CreateSquad($name: String!, $handle: String!, $description: String, $postId: ID!, $commentary: String!, $memberPostingRole: String, $memberInviteRole: String, $categoryId: ID, $isPrivate: Boolean, $moderationRequired: Boolean) {
```
sshanzel, line 1749: I think we are missing one more test to ensure the property was saved on creation.

src/entity/Source.ts
AmarTrebinjac: > https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1151500290/RFC-069+-+Admin+approval+on+Squads+posts#Table



> Is it not default false per RFC as well?

@capJavert I made this change in response to your comment on my previous PR. Although looking back at it now, I understand that I misunderstood your comment, which is the entire reason this error originated in the first place! 🤦🏼

capJavert: 😆 Oh I see it now, yeah I was referring to the fact that we will need a script to update and `moderationRequired=true` will just be sent from frontend of course.

__tests__/sources.ts
AmarTrebinjac: @sshanzel I updated the existing "should create squad" test to check if it's saved on creation!

sshanzel: Yeah, I was hoping to keep existing tests as is and introduce a new one, so we know previous calls would still work as intended.

src/entity/Source.ts
rebelchris: Ah I guess it doesn't really mater what the default is right?
It will be based on user input + state of other field anyway, we can just do this for backward compat and then ensure we set it right for the new incoming requests.

__tests__/sources.ts
```diff
@@ -928,6 +929,38 @@ query Source($id: ID!) {
       'NOT_FOUND',
     ));
 
+  it('squad should have post moderation disabled', async () => {
```
capJavert, line 932: These test only tests typeorm really and what we want to test is mutations which actually change the data. It would be better that this test is inside createSquad mutation and tests that moderationRequired is disabled by default. It would also confirm backward compatibility.

capJavert: You can also add other case that Lee mentioned that when you pass `moderationRequired=true` its set correctly.

__tests__/sources.ts
```diff
@@ -928,6 +929,38 @@ query Source($id: ID!) {
...
+      public: true,
+      moderationRequired: false,
+    });
+
+    const res = await client.query(QUERY, { variables: { id: '12345' } });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.source.moderationRequired).toEqual(false);
+  });
+
+  it('squad should have post moderation enabled after update', async () => {
```
capJavert, line 946: This one should be inside editSquad and test that when passing moderationRequired it changes the field value in the db.

capJavert (commented):
Looks good implementation wise, we just need tests that actually test added mutation vs queries through typeorm.

AmarTrebinjac:
@capJavert Think I've updated all tests accordingly now to meet your requests! 😄

src/schema/sources.ts
```diff
@@ -975,6 +990,12 @@ const validateSquadData = async (
     throw new ValidationError('Invalid member invite role');
   }
 
+  if (moderationRequired && memberPostingRole !== SourceMemberRoles.Member) {
```
capJavert, line 993: One edge case which comes to mind, if client does not send `memberPostingRole` it will act as if moderator only because `undefined !== SourceMemberRoles.Member`. Maybe we should check the actual value set on squad in the db if memberPostingRole is missing in mutation?

On the other hand I don't see why this would even be a requirement on API, to simplify we can just accept moderationRequired always, or is it something I am missing?

AmarTrebinjac: @capJavert I mean you're not entirely wrong that it wouldn't really matter, as we are going to check if the user is admin and if moderation is required when posts are created to determine if we need to create a SourcePostModeration entry or not. Having the validation on frontend only should suffice, as the purpose of the validation is purely an UX one.

Want me to remove the check?

__tests__/sources.ts
```diff
@@ -1755,6 +1774,7 @@ describe('mutation createSquad', () => {
     expect(newSource?.memberInviteRank).toEqual(
       sourceRoleRank[SourceMemberRoles.Member],
     );
+    expect(newSource?.moderationRequired).toEqual(false);
```
sshanzel, line 1777: Nice 🚀

src/schema/sources.ts
sshanzel: Yeah, I think it will be easier if we just set it with whatever that user passes, as long as the user has the right user privilege for the Squad.

capJavert, line 993:
```suggestion
  if (moderationRequired && memberPostingRole && memberPostingRole !== SourceMemberRoles.Member) {
```

capJavert: Yeah remove it or check for existance of `memberPostingRole` as well.

capJavert (approved):

src/schema/sources.ts
capJavert: https://github.com/dailydotdev/daily-api/pull/2324#discussion_r1798963928

rebelchris (approved):

PR #2325, opened 2024-10-09
feat: add http duration metrics

omBratteng:
Enable metrics for some auto instrumentation


capJavert:
Should not be needed since we added https://github.com/dailydotdev/daily-api/pull/2326

PR #2330, opened 2024-10-10
feat: you cant follow ghosts

capJavert:


__tests__/helpers.ts
```diff
@@ -178,7 +178,7 @@ export const testQueryErrorCode = async (
   code: string,
 ): Promise<void> =>
   testQueryError(client, query, (errors) => {
-    expect(errors.length).toEqual(1);
+    expect(errors?.length || 0).toEqual(1);
```
capJavert, line 181: Just noticed it breaks sometimes if errors is undefined so fixed it so our test fails are more clean

omBratteng (approved):

PR #2332, opened 2024-10-11
fix: null notifications

rebelchris:
There's an [issue](https://github.com/dailydotdev/daily/issues/1488) where if you bookmark shared post without title it will show `null`.
This was actually a deeper lying issue in notification generation.

So decided to fallback in several places to sharedPost.title if it exists.

Screenshot 2024-10-11 at 14 42 33

### DIFF
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -39,7 +39,7 @@ const getPostOrSharedPostTitle = (
     return ctx.post.title;
   }
 
-  if (ctx.sharedPost?.title && ctx.sharedPost?.title?.length > 0) {
+  if (ctx.sharedPost?.title?.length > 0) {
     return ctx.sharedPost.title;
   }
 
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -96,7 +96,7 @@ export const notificationTitleMap: Record<
     `The collection "<b>${ctx.post.title}</b>" just got updated with new details`,
   dev_card_unlocked: () => 'DevCard unlocked!',
   post_bookmark_reminder: (ctx: NotificationPostContext) =>
-    `Reading reminder! <b>${ctx.post.title}</b>`,
+    `Reading reminder! <b>${ctx.post.title ? ctx.post.title : ctx.sharedPost?.title ? ctx.sharedPost?.title : undefined}</b>`,
   source_post_added: (
     ctx: NotificationPostContext & NotificationDoneByContext,
   ) => `New post from <b>${ctx.source.name}</b>, check it out now!`,
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -32,6 +32,20 @@ import { format } from 'date-fns';
 
 const systemTitle = () => undefined;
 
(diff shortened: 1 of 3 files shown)


rebelchris:
Maybe we should run manual update query in DB to fix notification_v2.title
From:
Reading reminder! <b>null</b> 
To:
Reading reminder!
(just the quickest fix I guess)

Only 15 rows

src/notifications/generate.ts
```diff
@@ -96,7 +96,7 @@ export const notificationTitleMap: Record<
     `The collection "<b>${ctx.post.title}</b>" just got updated with new details`,
   dev_card_unlocked: () => 'DevCard unlocked!',
   post_bookmark_reminder: (ctx: NotificationPostContext) =>
-    `Reading reminder! <b>${ctx.post.title}</b>`,
+    `Reading reminder! <b>${ctx.post.title ? ctx.post.title : ctx.sharedPost?.title ? ctx.sharedPost?.title : undefined}</b>`,
```
capJavert, line 99: Can we do this with if-else, double ternary makes my head spin 😆

sshanzel: Yeah, I agree, maybe a generic function to fetch title that tries to fallback on possible ones.

src/notifications/generate.ts
```diff
@@ -32,6 +32,20 @@ import { format } from 'date-fns';
 
 const systemTitle = () => undefined;
 
+const getPostOrSharedPostTitle = (
+  ctx: NotificationPostContext,
+): string | null | undefined => {
+  if (ctx.post.title && ctx.post.title?.length > 0) {
```
capJavert, line 38:
```suggestion
  if (ctx.post?.title?.length > 0) {
```

src/notifications/generate.ts
```diff
@@ -32,6 +32,20 @@ import { format } from 'date-fns';
...
+    return ctx.post.title;
+  }
+
+  if (ctx.sharedPost?.title && ctx.sharedPost?.title?.length > 0) {
```
capJavert, line 42:
```suggestion
  if (ctx.sharedPost?.title?.length > 0) {
```

capJavert (approved):

src/notifications/generate.ts
```diff
@@ -32,6 +32,20 @@ import { format } from 'date-fns';
 
 const systemTitle = () => undefined;
 
+const getPostOrSharedPostTitle = (
+  ctx: NotificationPostContext,
+): string | null | undefined => {
+  if (ctx.post?.title?.length > 0) {
```
rebelchris, line 38:
```suggestion
  if (ctx.post?.title?.length) {
```

src/notifications/generate.ts
```diff
@@ -32,6 +32,20 @@ import { format } from 'date-fns';
...
+    return ctx.post.title;
+  }
+
+  if (ctx.sharedPost?.title?.length > 0) {
```
rebelchris, line 42:
```suggestion
  if (ctx.sharedPost?.title?.length) {
```

PR #2333, opened 2024-10-11
feat: add source moderation count

ilasw:
Added `moderationPostCount` field to Source query only if _source member_ and _moderation is required_ .
Updated also tests including moderated squad to fixtures.


sshanzel (approved):
Looks good to me! 🚀

sshanzel:
I think we can start merging this one now once we get the approval since the entity work has been merged.

__tests__/sources.ts
```diff
@@ -1082,6 +1095,50 @@ query Source($id: ID!) {
...
+  it('should not return moderationPostCount when moderation is not required', async () => {
+    loggedUser = '1';
+    const res = await client.query(QUERY, { variables: { id: 'squad' } });
+    console.log({ res });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.source.moderationRequired).toEqual(false);
+    expect(res.data.source.moderationPostCount).toBeFalsy();
+  });
+
+  it('should return moderationPostCount when user is admin', async () => {
```
sshanzel, line 1126: I think we are missing one more test. This is for when the user is not a moderator.

ilasw: Updated with more test cases ✅

src/schema/sources.ts
```diff
@@ -1543,7 +1554,27 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      }
+
+      const isModerator =
+        sourceRoleRank[source.currentMember.role] >= sourceRoleRank.moderator;
+      const status = SourcePostModerationStatus.Pending;
+      const query: FindOptionsWhere<SourcePostModeration> = {
+        status,
+        sourceId: source.id,
+        ...(!isModerator && { createdById: ctx.userId }),
+      };
```
capJavert, line 1570: This will need an index for sure over sourceId + createdById

ilasw: Added indexes!

src/migration/1730992663679-SourceModerationIndexes.ts
```diff
@@ -0,0 +1,14 @@
...
+export class SourceModerationIndexes1730992663679 implements MigrationInterface {
+    name = 'SourceModerationIndexes1730992663679'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE INDEX "IDX_852c087ea5d00671312fb71c2a" ON "source_post_moderation" ("sourceId", "createdById") `);
+        await queryRunner.query(`CREATE INDEX "IDX_532c94738c6b1334e4bc27c41c" ON "source_post_moderation" ("sourceId") `);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+    }
```
capJavert, line 12: Missing drop index

src/schema/sources.ts
capJavert: We usually add the fields through graphorm, but since this is count I am fine with adding it here for now. Probably ideal to add this to flags as well, but let's see.

src/schema/sources.ts
```diff
@@ -1543,7 +1554,26 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      const moderationPostCount = await ctx.con
+        .getRepository(SourcePostModeration)
+        .countBy(query);
+
+      return { ...source, moderationPostCount };
```
capJavert, line 1576: Realized just now what you were doing. 

We usually add the fields through graphorm, the difference is that if you request GQLSource anywhere graphorm will resolve the field, like this it only works on this specific query.

But since this is count I am fine with adding it here for now. Probably ideal to add this to flags as well so its not counted each time but let's not complicate for now since it was defined like this in RFC.

capJavert (approved):
Just missing down in mutations but good to merge after that.

PR #2334, opened 2024-10-11
feat: add follow source support

capJavert:
AS-648


rebelchris (approved):
We don't do any backwards filling in these phases right?
(From the old table to contentPreference)

capJavert:
@rebelchris backfill is the last step of the phase 2, but yeah will be done separate

PR #2335, opened 2024-10-12
feat: New SquadPostModeration Entity

AmarTrebinjac:
Created the new SquadPostModeration entity and migration.


src/entity/SquadPostModeration.ts
```diff
@@ -0,0 +1,83 @@
...
+  sourceId: string;
+
+  @ManyToOne('Source', (source: Source) => source.id, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  source: Promise<Source>;
+
+  @Column({
+    type: 'enum',
```
sshanzel, line 33: Nice. We should actually start using this enum. How does it behave if we add more items to the enum? I assume we would have to run another migration? We might want to check if we want that maintenance overhead.

src/entity/SquadPostModeration.ts
```diff
@@ -0,0 +1,83 @@
...
+    enum: SquadPostModerationStatus,
+    default: SquadPostModerationStatus.Pending,
+  })
+  status: SquadPostModerationStatus;
+
+  @Column({ type: 'text' })
+  createdById: string;
+
+  @Column({ type: 'text', nullable: true })
+  moderatedById: string | null;
```
sshanzel, line 43: We should create the FK indexes for these two columns similar to the `sourceId`.

src/entity/SquadPostModeration.ts
```diff
@@ -0,0 +1,83 @@
...
+  content: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  contentHtml: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  image: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  sharedPostId: string | null;
```
sshanzel, line 82: Same to the `postId`, let's introduce the FK here by defining the relation.

sshanzel (commented):
Looks amazing! We just need to introduce the right FKs.

src/entity/SquadPostModeration.ts
AmarTrebinjac: @sshanzel I thought so too, but the specs didn't mention it, and I looked at some other entities we have, and there's plenty of these references to other entities that don't have explicit FK's. Is there a specific reason for that? 🤔

sshanzel: Yeah, the previous ones were just missed, but as much as we can, it is better to have them FKs defined for faster query. Especially we will be filtering these per user for their requested posts.

AmarTrebinjac: @sshanzel Gotchu. I thought there was maybe a reason for it, so since it wasn't specified, I didn't apply it either. I'll fix it up! 😄

src/entity/SquadPostModeration.ts
AmarTrebinjac: @sshanzel Yes, I do think we would need to run another migration to update it:
```
await queryRunner.query(
 `CREATE TYPE "public"."squad_post_moderation_status_enum" AS ENUM('approved', 'rejected', 'pending')`,
);
```

and then:
```
await queryRunner.query(
 `CREATE TABLE "status" "public"."squad_post_moderation_status_enum" NOT NULL DEFAULT 'pending'
);
```

also trying to add a non specified value in TablePlus gives an error:
Screenshot 2024-10-13 at 13 13 30


Personally I think it's good practice to ensure only the desired values can be put into the database.

sshanzel: Yeah, I agree, the stricter, the better. We can maybe raise this to the web guild as it has its pros and cons, and since I think this is probably the first usage of it. So that, moving forward, we can utilize the same type again if agreed upon.

capJavert: @AmarTrebinjac we usually validate enums on gql side. While psql enum is also good it requires a lot of effort to add new ones once the table grows. You can see on other examples how gql enum validation can be done.

capJavert: From safety perspective its the same since users always interact with entities through gql.

src/entity/SquadPostModeration.ts
```diff
@@ -0,0 +1,102 @@
...
+import { User } from './user';
+
+export enum SquadPostModerationStatus {
+  Approved = 'approved',
+  Rejected = 'rejected',
+  Pending = 'pending',
+}
+
+@Entity()
+export class SquadPostModeration {
```
capJavert, line 20: I am missing tests, queries for this entity? 

While we can merge this most of the time when we just add an entity in single PR we then in implementation find out that we forgot about something and then we have to add more migrations. 

So if possible we should merge this with a task of adding this entity in gql, adding queries and so on. Hope it makes sense?

AmarTrebinjac: That's fair. I thought testing would be better paired with [this](https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31?selectedIssue=MI-598) task, but what you are saying makes sense, so I could add a few here.

capJavert: @AmarTrebinjac yeah, you can just join the two tasks on this branch if you want, that would be perfect 👍

ilasw: @capJavert @AmarTrebinjac we should add query/mutation according to RFC to this entity for:
- get the list of pending posts of a specific squad
- approve/reject 1+ posts 

I think that this PR should contain this feature and we should raise a new one for post approval creation.  Wdyt?

AmarTrebinjac: I'm a bit beyond this point unfortunately. It's gonna be this + [MI-598](https://dailydotdev.atlassian.net/browse/MI-598)

PR #2337, opened 2024-10-14
fix: prefer optimized select

rebelchris:
Already properly indexed and not much we can win here, besides doing a slightly better select (only omits 1 field)

Alternative extra ticket could be that we could potentially store this config in Redis as we always know when it updates?
CC: @capJavert for your evaluation if it's worth it.

Screenshot 2024-10-14 at 12 57 07

WT-2226 #done


omBratteng (approved):

capJavert (approved):

capJavert:
Let's see how it performs.

PR #2339, opened 2024-10-14
fix: reading streak query

rebelchris:
Only a very minor optimization found in adding a 2 key index.
Couldn't use the third key as we use `at time zone` which is not [immutable](https://stackoverflow.com/questions/51333687/how-do-i-create-an-index-in-postgresql-on-an-expression-that-uses-the-date-trunc)

However it hits the new index on production which does help a little bit.
Also moved this to ReadReplica as it's never that close to any view anyway.

Screenshot 2024-10-14 at 14 59 18

### DIFF
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -519,7 +519,7 @@ export const getLastStreakRecoverDate = async (
 };
 
 export const checkAndClearUserStreak = async (
-  con: DataSource | EntityManager,
+  con: DataSource,
   info: GraphQLResolveInfo,
   streak: GQLUserStreakTz,
 ): Promise<boolean> => {
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -499,7 +499,7 @@ export const checkUserStreak = (
 };
 
 export const getLastStreakRecoverDate = async (
-  con: DataSource | EntityManager,
+  con: DataSource,
   userId: string,
 ) => {
   const lastRecoverAction = await queryReadReplica(con, ({ queryRunner }) =>
@@ -521,7 +521,7 @@ export const getLastStreakRecoverDate = async (
 };
 
 export const checkAndClearUserStreak = async (
-  con: DataSource | EntityManager,
+  con: DataSource,
   info: GraphQLResolveInfo,
   streak: GQLUserStreakTz,
 ): Promise<boolean> => {
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -7,7 +7,7 @@ import {
   UserStreakActionType,
 } from '../entity';
 import { differenceInDays, isSameDay, max, startOfDay } from 'date-fns';
-import { DataSource, EntityManager, In, Not } from 'typeorm';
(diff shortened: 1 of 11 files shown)


src/common/users.ts
```diff
@@ -499,27 +499,29 @@ export const checkUserStreak = (
...
-    .select(
-      `MAX(date_trunc('day', usa."createdAt"::timestamptz at time zone COALESCE(u.timezone, 'utc'))::date - interval '1 day')`,
-      'createdAt',
-    )
-    .from(UserStreakAction, 'usa')
-    .innerJoin(DbUser, 'u', 'u.id = usa."userId"')
-    .where(`usa."userId" = :userId`, { userId })
-    .andWhere(`usa.type = :type`, { type: UserStreakActionType.Recover })
-    .getRawOne<UserStreakAction>();
+  const lastRecoverAction = await queryReadReplica(con, ({ queryRunner }) =>
```
capJavert, line 505: Since this helper is now used in restore logic, it could potentially make problems due to replica being out of sync. This is called directly in CDC so actually it could be outdated easily.

rebelchris: But unlikely it's calculating your streak while you just read is what I meant?
Happy to move it back to main worker though.

capJavert: I am worried for `onUserStreakChange` cdc callback, there on streak change we read from replica which could be outdated due to change just happened on `master`. I agree read replica can be used on gql side but we can maybe fix it by using different helper in CDC.

Maybe in `userStreak` query we can wrap `checkAndClearUserStreak` call in `queryReadReplica` instead of using it inside the helper.

capJavert: Something like this:

```ts
const hasClearedStreak = await queryReadReplica(
  ctx.con,
  ({ queryRunner }) => {
    return checkAndClearUserStreak(queryRunner.connection, info, streak);
  },
);
```

src/schema/users.ts
```diff
@@ -1353,11 +1354,13 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         };
       }
 
-      const hasClearedStreak = await checkAndClearUserStreak(
+      const hasClearedStreak = await queryReadReplica(
         ctx.con,
-        info,
-        streak,
+        ({ queryRunner }) => {
+          return checkAndClearUserStreak(queryRunner.connection, info, streak);
```
capJavert, line 1360: Tnx for the changes, from my testing it might be better to pass `queryRunner`, see https://github.com/dailydotdev/daily-api/pull/2340, but not blocking. Let's see tomorrow.

rebelchris: @capJavert It makes only a slight difference.
(See the last before update now doesn't use a new `pg.connect` but due to needing both the RR and master it's a bit tricky to omit even more.

Let me know what you think.

| Before | After |
|--------|--------|
| Screenshot 2024-10-15 at 09 24 52 | Screenshot 2024-10-15 at 09 37 41 |

capJavert: yeah, not blocking here since its a one call either way

capJavert (approved):

PR #2340, opened 2024-10-14
feat: reuse manager and pool connect calls

capJavert:
I found the reason why we have multiple `pg-pool.connect` in boot traces. Even though pool manages connection reuse each call to `queryReadReplica` still has to query a connection from a pool, which has some cost, usually few ms, but sometimes as seen in production few 10s or 100s of ms.

I optimized boot so all read replica calls are bundled inside single `queryReadReplica` call, which reuses `QueryRunner`, which then does a single `pg-pool.connect` call under the hood.

After optimization the number of calls fell from 8 to 4. 

Looking further I realized that same thing happens with `con.query`, `con.find` or other calls. So I added a `queryDataSource` abstraction which allows us to also reuse single manager for `master` calls. After this it fell to only 2 calls for `loggedInBoot`, one for `master` and one for `slave` (read replica).

We need to deploy to see the difference it will make and if it will eliminate the high `pg-pool.connect` calls in production but looks much better already locally. 

Other changes include types refactor from `con: DataSource` to `con: DataSource | QueryRunner` to support passing a runner instance from parent calls.

Example traces before/after (from local jaeger):

Before:
image

After:
image

We can also see smaller number of spans due to pool calls being reused:

Before:
image

After:
image

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -1,6 +1,6 @@
 import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';
 import createOrGetConnection from '../db';
-import { DataSource, Not } from 'typeorm';
+import { DataSource, Not, QueryRunner } from 'typeorm';
 import { clearAuthentication, dispatchWhoami } from '../kratos';
 import { generateUUID } from '../ids';
 import { generateSessionId, setTrackingId } from '../tracking';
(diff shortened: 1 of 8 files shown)


idoshamun (approved):
Amazing find! This should be documented as a guideline somewhere

omBratteng (approved):

src/routes/boot.ts
```diff
@@ -425,30 +419,31 @@ const loggedInBoot = async ({
...
-      getAlerts(con, userId),
-      getSettings(con, userId),
-      getUnreadNotificationsCount(con, userId),
-      getSquads(con, userId),
-      getAndUpdateLastBannerRedis(con),
-      getExperimentation({ userId, con }),
-      getMarketingCta(con, log, userId),
       middleware ? middleware(con, req, res) : {},
-      getFeeds({ con, userId }),
+      queryDataSource(con, ({ queryRunner }) => {
```
rebelchris, line 430: Nice, this is super neat!

rebelchris (approved):
Wow nice find!

PR #2341, opened 2024-10-15
feat: increase pool size

capJavert:
https://github.com/dailydotdev/daily-api/pull/2340 fixed amount of pool calls that boot does but I still see some longer calls. 

Now I can confirm that calls that take longer are the new connections that need to be added to pool, while just querying existing connection from pool is "cheap".

To try and mitigate I increased `max` connections allowed inside pool which should lower amount of times the pool needs to create new connections. Our max con on pg instances is 1000.

Trace below from production, below connect call takes 80% of total request time, which makes boot latency spike.

Screenshot 2024-10-15 at 12 36 46

[WT-2230](https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11?selectedIssue=WT-2230)

[WT-2230]: https://dailydotdev.atlassian.net/browse/WT-2230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/data-source.ts
+++ b/src/data-source.ts
@@ -6,7 +6,7 @@ export const AppDataSource = new DataSource({
   schema: 'public',
   synchronize: false,
   extra: {
-    max: 20,
+    max: 30,
     idleTimeoutMillis: 120000,
   },
   logging: false,


nensidosari (approved):

rebelchris (approved):
If max is 1000 can't we spool even further safely?

omBratteng (approved):

src/data-source.ts
```diff
@@ -6,7 +6,7 @@ export const AppDataSource = new DataSource({
   schema: 'public',
   synchronize: false,
   extra: {
-    max: 20,
+    max: 30,
```
capJavert, line 9: > If max is 1000 can't we spool even further safely?

@rebelchris its max connections on instance, and since we have multiple pods we can't say 1000 per pods. For example API can scale to max 25, so 30x25 is 750, but we also have bg and other stuff.

Our connections usually spike to 200 in read replica and 300 in master.

capJavert: So this is `30` connections per pool (and each pod has its own pool)

idoshamun (approved):

PR #2342, opened 2024-10-15
fix: optimizations for reading tag query

rebelchris:
Decided not to introduce the uniqueKey index, see discussion [here](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1728989949471359).

The query format changes a little bit (mostly order optimization that PG prefers)
And making the types of `userId` match to help reduce another 20/30ms delay.

| Before | After |
|--------|--------|
| Screenshot 2024-10-15 at 13 09 11 | Screenshot 2024-10-15 at 13 10 04 | 

> Note: This doesn't include the varchar changes yet.
> Note: It's averages, ran about 10 queries on each type this was average

WT-2229 #done

### DIFF
--- a/src/common/users.ts
+++ b/src/common/users.ts
@@ -293,16 +293,17 @@ export const getUserReadingTags = (
   con: DataSource,
   { userId, dateRange: { start, end }, limit = 8 }: ReadingDaysArgs,
 ): Promise<TagsReadingStatus[]> => {
+  console.log(userId, start, end, limit);
   return con.query(
     `
       WITH filtered_view AS (
         SELECT
-          v.*,
+          v. "postId",
           CAST(v.timestamp AT TIME ZONE COALESCE(u.timezone,
               'UTC') AS DATE) AS day
         FROM
           "view" v
-          JOIN "user" u ON u.id = v."userId"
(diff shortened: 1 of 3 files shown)


src/common/users.ts
```diff
@@ -387,6 +393,8 @@ export const getUserReadingRank = async (
     });
   };
 
+  console.log(req.getQueryAndParameters());
+
```
omBratteng, line 397:
```suggestion
```

src/common/users.ts
```diff
@@ -320,24 +320,29 @@ export const getUserReadingTags = (
...
+        tag,
+        "readingDays",
         tr."readingDays" * 1.0 / dd.total_days AS percentage,
         dd.total_days AS total
-      FROM
-        tag_readings tr
+      FROM (
+        SELECT
+          tr.tag,
+          tr. "readingDays"
```
capJavert, line 338: Nested select faster?

capJavert (approved):

src/common/users.ts
rebelchris: According to result and everSQL yeah 👍
Screenshot 2024-10-15 at 13 58 06

PR #2345, opened 2024-10-15
feat(boot): move unread notifications to read replica

capJavert:


omBratteng (approved):

rebelchris (approved):
Isn't this a high change of being out of sync?
Due to bg worker doing insert, we might miss a count here and there on load?

Don't think it will make much difference though.

PR #2347, opened 2024-10-15
fix: should not return empty title posts

rebelchris:
not a 100% sure what Meili has indexed for shared posts, but maybe it's the related post?
This should fix it on API side to not return those.

### DIFF
--- a/src/schema/search.ts
+++ b/src/schema/search.ts
@@ -331,6 +331,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         .where('post.id IN (:...ids)', {
           ids: hits.length ? hits.map((x) => x.post_id) : ['nosuchid'],
         })
+        .andWhere('post.title IS NOT NULL')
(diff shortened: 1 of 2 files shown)


src/schema/search.ts
```diff
@@ -331,6 +331,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         .where('post.id IN (:...ids)', {
           ids: hits.length ? hits.map((x) => x.post_id) : ['nosuchid'],
         })
+        .andWhere('post.title IS NOT NULL')
```
capJavert, line 334: Wondering if this will slow down the query since we don't have the index on `title`, but maybe we can achive the same with just `filter` on JS side.

rebelchris: That's also fine with me, we can do it in the map below if you prefer?

capJavert: I check with `explain analyze`, it should be good since we filter by id either way which is PK.

capJavert (approved):

PR #2353, opened 2024-10-18
fix: make sure release is called in case of error

capJavert:
I noticed that in case of query failed `release` would not be throw. Don't think it had impact because we don't have failed queries on these but good to be safe in those cases.


omBratteng (approved):

rebelchris (approved):
Ah good find, I guess at first this worked fine with the singular .finally

This should work :)

PR #2358, opened 2024-10-21
fix: temporary fix to ensure debezium is deployed to old nodes

omBratteng:
A temporary fix while I move stuff over to new nodes. The debezium deployments requires a bit of manual intervention, so will do that later


capJavert (approved):

omBratteng:
Nvm, don't actually need this. I see all debezium instances have a `topology.kubernetes.io/zone` node selector that ensure they all select the old nodes

PR #2361, opened 2024-10-25
feat: allow scheduling api onto spot nodes

omBratteng:
Weight is `n of 100`, where the remainder is the weight on non-spot nodes.
So if weight is 50, it should do 50/50 split when scheduling new pods as long as there are available nodes for it.

I have enabled it for all variations of our deployments, but with different weights based on what think would work as a nice balance.


.infra/package.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "api",
   "devDependencies": {
-    "@types/node": "^18.17.0"
+    "@types/node": "20.x"
```
capJavert, line 4: let's put our version of node here?

.infra/index.ts
```diff
@@ -408,6 +422,10 @@ if (isAdhocEnv) {
         },
         targetAverageValue: 100_000,
       },
+      spot: {
+        enabled: true,
+        weight: 70,
+      },
```
capJavert, line 428: Since personalized digest is quite bursty might be an issue if spot node gets killed mid processing it could cause spikes, so maybe better to have it a bit lower or turn of until we see how it works? At least it looks like that from my understanding. WDYT?

.infra/index.ts
```diff
@@ -323,8 +327,10 @@ if (isAdhocEnv) {
       livenessProbe,
       metric: { type: 'memory_cpu', cpu: 85 },
       disableLifecycle: true,
-      // ports: [{ containerPort: 9464, name: 'metrics' }],
-      // servicePorts: [{ targetPort: 9464, port: 9464, name: 'metrics' }],
+      spot: {
+        enabled: true,
+        weight: 70,
```
capJavert, line 332: `100`? 😆

.infra/.nvmrc
```diff
@@ -1 +1 @@
-18
+20
```
capJavert, line 1: shoudl we do same as one in api? `20.12`

capJavert (approved):

.infra/index.ts
```diff
@@ -306,6 +306,10 @@ if (isAdhocEnv) {
         { targetPort: 3000, port: 80, name: 'http' },
         { targetPort: 9464, port: 9464, name: 'metrics' },
       ],
+      spot: {
+        enabled: true,
+        weight: 50,
+      },
```
capJavert, line 312: Looks good on 50% as a start, we just need to be carful on evictions so load is handled throughout the day.

.infra/index.ts
omBratteng: Can set it to 50/50. If a spot node is evicted, they'll just get rescheduled onto other nodes, and if no spot nodes are available, it'll use regular nodes.
It's just a best-effort selection

.infra/package.json
omBratteng: Set it to 20.12.x

.infra/index.ts
omBratteng: Haha, maybe once I see it works 😂

.infra/.nvmrc
omBratteng: Set it to 20.12

.infra/index.ts
omBratteng: Yeah, my thoughts too.
I might want to try at one point with two deployments of api, one that's always 3 replicas with no auto scaling to keep a minimum, on regular nodes. Then a 2nd deployment that's "only" scheduled on spot nodes, that can scale to 0

omBratteng: Our current spot nodes that host the redis replicas, has an age of 30 days. So evictions doesn't happen too often. But it can happen
image

.infra/index.ts
capJavert: I am worried about time it takes from eviction -> node is back up again, since digest processing is time constrained.

omBratteng: As I mentioned, it's a best-effort. If no spot-nodes are schedulable, it'll just fallback to regular nodes

omBratteng: Also, I am going to add a antiPodAffinity later on, so that it tries to spread pods more evenly, so that not all of the pods ends up on one node.

capJavert: I understand that, worried about the time it takes from node being evicted and then replaced with whatever node is available, is it seconds, minutes...?

omBratteng: Seconds

rebelchris (approved):

PR #2367, opened 2024-10-29
fix: make websocket pods start with different command without tracing

omBratteng:
Something in websockets is causing tracing to go haywire and collected thousands of spans, so it exceeds the queue size.
So as a fix for now, we just create a new command for websocket pods to start on, that does not have tracing enabled.

My first thought was adding a check for `ENABLE_SUBSCRIPTIONS` variable, but we set it in adhoc, so it would disable tracing in adhoc 🙈

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -391,14 +391,6 @@ if (isAdhocEnv) {
       createService: true,
       serviceType: 'ClusterIP',
       disableLifecycle: true,
-      // ports: [
-      //   { containerPort: 3000, name: 'http' },
-      //   { containerPort: 9464, name: 'metrics' },
-      // ],
-      // servicePorts: [
-      //   { targetPort: 3000, port: 80, name: 'http' },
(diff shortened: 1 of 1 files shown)


.infra/index.ts
```diff
@@ -224,10 +224,10 @@ if (isAdhocEnv) {
...
         { targetPort: 9464, port: 9464, name: 'metrics' },
       ],
-      podAnnotations: {
-        'prometheus.io/scrape': 'true',
-        'prometheus.io/port': '9464',
-      },
+      // podAnnotations: {
+      //   'prometheus.io/scrape': 'true',
+      //   'prometheus.io/port': '9464',
+      // },
```
capJavert, line 230: let's just remove it 🔥

omBratteng: Removed it

omBratteng:
> Looks ok logic wise, deploys ok locally?
Yeah, it starts as expected

PR #2368, opened 2024-10-30
feat: settings flag for sidebar settings

rebelchris:
Decided with Sab we'll go ahead with the changer for now as the change in order of elements is important for product.
We will still evaluate further approaches to making sidebar better in follow up ticket.

This PR introduces a flag ability to settings where we can store any type of JSON.

AS-665


src/entity/Settings.ts
```diff
@@ -19,6 +19,21 @@ export enum ChecklistViewState {
   Hidden = 'hidden',
 }
 
+export type SettingsFlags = Partial<{
```
capJavert, line 22: This is usually enum and then we can reuse it with toGqlEnum for schema definition.

src/schema/settings.ts
```diff
@@ -322,7 +357,10 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         const settings = await repo.findOneBy({ userId });
 
         if (!settings) {
-          return repo.save({ ...data, userId });
+          return repo.save({
+            ...data,
+            userId,
```
capJavert, line 362: I think we will need `updateFlagsStatement` because if request does not send everything the whole `flags` field will get overwritten.

capJavert (commented):
Nice! Just minor comments to check.

src/schema/settings.ts
```diff
@@ -53,6 +56,20 @@ interface GQLUpdateSettingsInput extends Partial<GQLSettings> {
 export const typeDefs = /* GraphQL */ `
   ${toGQLEnum(ChecklistViewState, 'ChecklistViewState')}
 
+  type SettingsFlagsPublic {
+    sidebarSquadExpanded: Boolean
+    sidebarCustomFeedsExpanded: Boolean
+    sidebarOtherExpanded: Boolean
+    sidebarResourcesExpanded: Boolean
+  }
```
capJavert, line 64: This can be https://github.com/dailydotdev/daily-api/pull/2368#discussion_r1822516038

src/entity/Settings.ts
rebelchris: @capJavert I don't this this is true, can't find an example of enum base flags?

src/schema/settings.ts
rebelchris: Let me change this and add failing test first, good catch 🥳

src/entity/Settings.ts
capJavert: Ignore this 😆 I was looking at completely wrong place, you can't really make flags from enum either way.

src/schema/settings.ts
capJavert: also ignore

src/schema/settings.ts
```diff
@@ -321,11 +356,26 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         const repo = manager.getRepository(Settings);
         const settings = await repo.findOneBy({ userId });
 
+        const dataWithAdjustedFlags = {
+          ...data,
+          flags: updateFlagsStatement<Settings>({
+            ...data.flags,
+          }),
```
capJavert, line 363:
```suggestion
          flags: updateFlagsStatement<Settings>(data.flags),
```

capJavert (approved):

PR #2369, opened 2024-10-30
refactor: use a single query for feed filters

idoshamun:
In an effort to reduce the load on the database, I merged together the different queries used to generate the feed filters object.

It also includes some minor optimizations.

If this will not reduce the load we can consider redis. Now with a single query it can be easier to cache it. I may also consider moving the javascript logic to the query itself but let's see.

### DIFF
--- a/__tests__/__snapshots__/feeds.ts.snap
+++ b/__tests__/__snapshots__/feeds.ts.snap
@@ -7,9 +7,7 @@ Object {
   ],
   "blockedTags": Array [],
   "excludeSourceTypes": Array [],
-  "excludeSources": Array [
-    "fd062672-63b7-4a10-87bd-96dcd10e9613",
-  ],
+  "excludeSources": Array [],
   "excludeTypes": Array [],
   "includeTags": Array [],
   "sourceIds": Array [],
@@ -23,9 +21,7 @@ Object {
   "excludeSourceTypes": Array [
     "squad",
   ],
-  "excludeSources": Array [
-    "fd062672-63b7-4a10-87bd-96dcd10e9613",
-  ],
+  "excludeSources": Array [],
   "excludeTypes": Array [],
   "includeTags": Array [],
   "sourceIds": Array [],
@@ -37,9 +33,7 @@ Object {
   "blockedContentCuration": Array [],
   "blockedTags": Array [],
   "excludeSourceTypes": Array [],
-  "excludeSources": Array [
-    "fd062672-63b7-4a10-87bd-96dcd10e9613",
-  ],
+  "excludeSources": Array [],
   "excludeTypes": Array [],
   "includeTags": Array [],
   "sourceIds": Array [
--- a/__tests__/workers/__snapshots__/personalizedDigestEmail.ts.snap
+++ b/__tests__/workers/__snapshots__/personalizedDigestEmail.ts.snap
@@ -187,9 +187,7 @@ Object {
 exports[`personalizedDigestEmail worker should generate personalized digest for user in timezone ahead UTC 2`] = `
 Object {
   "allowed_tags": Array [],
-  "blocked_sources": Array [
-    "fd062672-63b7-4a10-87bd-96dcd10e9613",
-  ],
+  "blocked_sources": Array [],
   "blocked_tags": Array [],
   "date_from": Any<String>,
   "date_to": Any<String>,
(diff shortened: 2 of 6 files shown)


src/common/feedGenerator.ts
```diff
@@ -204,20 +228,32 @@ export const feedToFilters = async (
     { hide: [], show: [] },
   );
 
-  // If the user is not a member of the watercooler, hide it
-  if (!membershipsByHide.show.includes(WATERCOOLER_ID)) {
```
idoshamun, line 208: This is already being taken care of by the feed

src/common/feedGenerator.ts
```diff
@@ -73,24 +78,78 @@ export const whereKeyword = (
...
-  const [settings, feedAdvancedSettings] = await Promise.all([
-    con.getRepository(AdvancedSettings).find(),
-    con.getRepository(FeedAdvancedSettings).findBy({ feedId }),
-  ]);
+const getRawFiltersData = async (
+  con: DataSource,
+  feedId: string,
+  userId: string,
+): Promise<RawFiltersData | undefined> => {
+  const selects = [
```
idoshamun, line 108: I use GraphORM approach of aggregating subqueries into json array

src/common/feedGenerator.ts
```diff
@@ -73,24 +78,78 @@ export const whereKeyword = (
...
+    rawFilterSelect(con, 'memberships', (qb) =>
+      qb
+        .select('"sourceId"')
+        .addSelect("COALESCE((flags->'hideFeedPosts')::boolean, FALSE)", 'hide')
+        .from(SourceMember, 't')
+        .where('"userId" = $2'),
+    ),
+  ];
+  const query =
+    'select ' + selects.map((select) => `(${select.getQuery()})`).join(', ');
```
capJavert, line 138: Hmm I don't like this part honestly there is however `queryDataSource` helper that will minimize pg-pool connect calls like we do for boot, but it will still use multiple queries. 

Would be interested to load test and see if these 3 options is the best:
- current implementation
- `queryDataSource`
- your implementation with single query

idoshamun: yap. i want to generate a report only for this part based on the traces. since it's a child trace i still haven't found a way. i'm not sure it will improve myself either.

reason i use raw queries is that typeorm doesn't allow to build queries without a from statement, but i don't need it here

capJavert: Talking because I am not sure which is the bottleneck here, the multiple queries or just our data being saved in inefficient format across multiple tables.

idoshamun: issue is i need to deploy it to test but first i'll try to generate a report

capJavert: As for report, I think its more of a limitation on google trace explorer side no, some other trace interfaces are more powerful.

idoshamun: The report always analyzes the parent :triggered:
https://console.cloud.google.com/traces/tasks/2994973240782792936?project=devkit-prod

__tests__/__snapshots__/feeds.ts.snap
```diff
@@ -37,9 +33,7 @@ Object {
   "blockedContentCuration": Array [],
   "blockedTags": Array [],
   "excludeSourceTypes": Array [],
-  "excludeSources": Array [
-    "fd062672-63b7-4a10-87bd-96dcd10e9613",
-  ],
```
capJavert, line 42: Just to sanity check, is this change is tests alright?

__tests__/workers/__snapshots__/personalizedDigestEmail.ts.snap
```diff
@@ -300,9 +298,7 @@ Object {
 exports[`personalizedDigestEmail worker should generate personalized digest for user in timezone behind UTC 2`] = `
 Object {
   "allowed_tags": Array [],
-  "blocked_sources": Array [
-    "fd062672-63b7-4a10-87bd-96dcd10e9613",
-  ],
+  "blocked_sources": Array [],
   "blocked_tags": Array [],
```
capJavert, line 302: Here as well?

capJavert (approved):

__tests__/__snapshots__/feeds.ts.snap
idoshamun: Yap removing watercooler from being hard coded. Taken care by the feed service already

PR #2370, opened 2024-10-30
feat: UserTopReader entity and GraphQL queries

omBratteng:
TODO:

- [x] Add tests

AS-657


src/schema/users.ts
```diff
@@ -1659,6 +1696,29 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         },
       );
     },
+    topReaderBadge: async (
+      _,
+      { id }: { id: string },
+      ctx: AuthContext,
+      info: GraphQLResolveInfo,
+    ) => {
+      const res = await graphorm.query<GQLUserTopReader>(
```
capJavert, line 1705: you can use `queryOneOrFail`

src/graphorm/index.ts
```diff
@@ -165,6 +165,16 @@ const obj = new GraphORM({
         },
         transform: nullIfNotLoggedIn,
       },
+      topReader: {
+        relation: {
+          isMany: true,
+          customRelation: (_, parentAlias, childAlias, qb): QueryBuilder =>
+            qb
+              .where(`${childAlias}."userId" = ${parentAlias}.id`)
+              .orderBy(`${childAlias}."issuedAt"`, 'DESC')
```
capJavert, line 174: might need an index over `userId` + `issuedAt`

omBratteng: Will add that

omBratteng: Added

src/schema/users.ts
omBratteng: Changed to use that

capJavert (approved):

PR #2373, opened 2024-10-30
Squad post moderation flow

AmarTrebinjac:
Adds new flow for handling squad post moderation. Mostly new code, but does also add additional checks in ensureSourcePermissions function, which is used in the various existing create/edit functions.


AmarTrebinjac:
To be honest I'm not quite sure what happened to the commit history here.

Tried to move to new branch to fix it, but alas 🤷

src/schema/posts.ts
```diff
@@ -1521,6 +1677,63 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        sourceId,
+        image,
+        title,
+        content,
+        type,
+        sharedPostId,
+      }: CreateSquadPostModerationArgs,
+      ctx: AuthContext,
+      info,
+    ): Promise<GQLSquadPostModeration> => {
```
ilasw, line 1692: I think that we should not add an API endpoint for this. 
We should keep the create post endpoint for not change any logic on frontend post creation/share as [stated here](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1151500290/RFC-069+-+Admin+approval+on+Squads+posts#Post-creation.1)

src/schema/posts.ts
```diff
@@ -1101,6 +1219,44 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const isModerator = await isPrivilegedMember(ctx, sourceId);
+      if (!isModerator) throw new ForbiddenError('Access denied!');
+
+      return graphorm.query<GQLSquadPostModeration>(ctx, info, (builder) => ({
+        ...builder,
+        queryBuilder: builder.queryBuilder.where(
+          `"${builder.alias}"."sourceId" = :sourceId`,
+          { sourceId },
+        ),
+      }));
```
ilasw, line 1237: Both users and mods can fetch this endpoint.
- If **user**: check if joined squad, then retrieve **my** pending posts
- if **admin/mod**: check permission, then return **all** pending posts

src/entity/SquadPostModeration.ts
```diff
@@ -0,0 +1,98 @@
...
+  id: string;
+
+  @Column({ type: 'text' })
+  sourceId: string;
+
+  @ManyToOne('Source', (source: Source) => source.id, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  source: Promise<Source>;
```
ilasw, line 31: Do we need both `sourceId` and `source`? I think that We can go for the relation only (same for post and sharedPost).
cc @capJavert

ilasw (changes requested):
Looks good, just raised few points around specifications

src/schema/posts.ts
sshanzel: I think the core issue with the old one is when you create a freeform post without moderation, everything is as intended, but when you integrate moderation, the expected returned object would not be a post anymore. Either we return a union type or a `null` value. I am more inclined to the `null` value personally.

src/entity/SquadPostModeration.ts
sshanzel: Yes, that is true for the schema. However, for the entity we need the FK which is the source id.

src/schema/posts.ts
AmarTrebinjac: @ilasw I discussed this with @sshanzel and we pretty much concluded what he wrote up here. I personally prefer not to return null because it isn't really clear why that is being returned and is quite flaky design wise.

I am not familiar with union types in graphql, but if we really want to keep it in the same endpoint, then I'd prefer to go that way.

src/schema/posts.ts
AmarTrebinjac: I think the user getting their own moderation posts should be a different endpoint from the admin getting all of the ones for the squad

src/schema/posts.ts
AmarTrebinjac: @capJavert Do you have any thoughts on this?

src/entity/SquadPostModeration.ts
capJavert: Yeah, source is not in the db, its just a logical relation on a class.

src/schema/posts.ts
capJavert: Not 100% taking all logic into account, but I would rather have queries/mutations return the same thing always, oppose to union type which I agree would just complicate. Things like this are hard to visual in the RFC sometimes so changes in implementation sometimes do happen. 

So I am fine with leaving it as separate for now as long as we are able to implement all the other requirements.

ilasw: @sshanzel can you please point where I can find this logic in code? Thanks! 🙏

src/entity/SquadPostModeration.ts
```diff
@@ -0,0 +1,98 @@
...
+  image: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  sharedPostId?: string | null;
+
+  @ManyToOne('Post', (post: Post) => post.id, {
+    lazy: true,
+    onDelete: 'SET NULL',
+  })
+  sharedPost: Promise<Post>;
```
capJavert, line 97: Why not just cascade on delete? (same for other above)

src/routes/boot.ts
```diff
@@ -205,6 +209,7 @@ const getSquads = async (
         permalink: getSourceLink(source),
         currentMember: {
           permissions: postingPermissions,
+          role,
```
capJavert, line 212: Do we need this or accidental?

src/schema/posts.ts
```diff
@@ -646,6 +702,30 @@ export const typeDefs = /* GraphQL */ `
...
+
+    """
+    Get squad post moderations by source id
+    """
+    squadPostModerationsBySourceId(
+      """
+      Id of the source
+      """
+      sourceId: ID!
+    ): [SquadPostModeration!]!
```
capJavert, line 727: This should be behind `@auth`?

src/schema/posts.ts
```diff
@@ -646,6 +702,30 @@ export const typeDefs = /* GraphQL */ `
...
+    squadPostModeration(
+      """
+      Id of the requested post moderation
+      """
+      id: ID!
+      """
+      Id of the source
+      """
+      sourceId: ID!
+    ): SquadPostModeration!
```
capJavert, line 717: This should be behind `@auth`?

src/schema/posts.ts
```diff
@@ -646,6 +702,30 @@ export const typeDefs = /* GraphQL */ `
...
+
+    """
+    Get squad post moderations by source id
+    """
+    squadPostModerationsBySourceId(
```
capJavert, line 722: This should have pagination support? `graphorm.query` already supports it you just need to expose after/first.

src/entity/SquadPostModeration.ts
```diff
@@ -0,0 +1,98 @@
...
+import { User } from './user';
+
+export enum SquadPostModerationStatus {
+  Approved = 'approved',
+  Rejected = 'rejected',
+  Pending = 'pending',
+}
+
+@Entity()
+export class SquadPostModeration {
```
capJavert, line 20: I see its named differently in  diferent places of RFC but I think it should be named `SourcePostModeration` because squads are just sources under the hood so this can potentially be applied to any post (not just squad posts). 

That said its more of a naming thing so lowest prio.

src/schema/posts.ts
```diff
@@ -1101,6 +1219,44 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   BaseContext
 >({
   Query: {
+    squadPostModerationsBySourceId: async (
+      _,
+      { sourceId }: { sourceId: string },
+      ctx: Context,
+      info,
+    ): Promise<GQLSquadPostModeration[]> => {
+      const isModerator = await isPrivilegedMember(ctx, sourceId);
```
capJavert, line 1228: you can use `ensureSourcePermission` here by adding a new permission to `moderatorPermissions` and `adminPermissions`. That way you don't need to create new `isPrivilegedMember` function

capJavert (commented):
Logic wise looks good 👌 , some minor stuff on usages of existing functions and naming.

src/routes/boot.ts
capJavert: Ok, need to update `BootSquadSource` type as well then 👍

src/entity/SquadPostModeration.ts
AmarTrebinjac: True, this one should probably cascade.

@capJavert Which do you mean "other above" ? The one for user?

If that's the case, I think that even if the user gets deleted, the post may be interesting for the squad and they might want to keep it, I talked with Lee about potentially reassigning those to Ghost user.

What do you think ?

capJavert: yeah but here you are describing the relation of what happens wtih SquadPostModeration when post gets deleted, which should always delete moderation items when post is deleted?

capJavert: I don't think user would want anything of his posted if you delete an account as well, so maybe better to just keep it simple and avoid this case if it was not mentioned in the RFC or product spec.

src/schema/posts.ts
AmarTrebinjac: I'll look into this

src/schema/posts.ts
ilasw: PS: @sshanzel @AmarTrebinjac do you have calculated the potential effort impact of this change on frontend? If yes please can you add a section on RFC so we can track this activity? Thanks!

src/schema/posts.ts
```diff
@@ -1101,6 +1219,52 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   BaseContext
 >({
   Query: {
+    squadPostModerationsBySourceId: async (
+      _,
+      { sourceId }: { sourceId: string },
```
ilasw, line 1224: From RFC we should have both `sourceID` and `status`, plus pagination mentioned already by Ante

```graphql
query SourceModerationPosts {
  sourcePendingPosts(
    sourceId: ID!, 
    status: 'pending',
    """
    Pagination
    """
    after: String,
    first: Int
    ): [Post]
}
```

src/schema/posts.ts
sshanzel: > I would rather have queries/mutations return the same thing always, oppose to union type which I agree would just complicate (same for null).

I agree. It would be harder to catch when the form of the object is either A or B.

> do you have calculated the potential effort impact of this change on frontend?

In frontend it should be just a simple ternary operator on which to call whether the squad requires moderation or not. We should have something similar in the codebase. Same amount of parameters, same requirements, just a different endpoint.

src/schema/posts.ts
sshanzel: @AmarTrebinjac let me know if you need any help how to use our pagination.

src/schema/posts.ts
sshanzel: Should we make the `status` an array? Just so for regular members that posted something and we should pull both rejected and pending posts. Wdyt?

sshanzel:
I think we are missing two columns:
- reason: Reason for the rejection.
- remarks: if there are any comments about the rejection.

AmarTrebinjac:
@sshanzel Would that not be "moderatorMessage" ?

src/schema/posts.ts
AmarTrebinjac: I have already mostly completed the frontend task for this !

src/schema/posts.ts
AmarTrebinjac: @capJavert I don't think using ensureSourcePermission is the right call here, mainly because it throws error if they don't have permission, which would make the moderation required step a bit awkward. I think this little isPrivilegedMember function works well.

However, I will redo my change in boot, and remove the role property that I added, and instead add a permission "Moderate", to check what mutation to call from the frontend!

AmarTrebinjac:
Added properties "url" and "imageUrl". 

When users upload an external link, the image is already an url, and not a file upload. We also need to save the link to the item they are sharing since we don't have an existing post to reference to.

src/entity/SourcePostModeration.ts
```diff
@@ -0,0 +1,104 @@
...
+  sharedPostId?: string | null;
+
+  @ManyToOne('Post', (post: Post) => post.id, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  sharedPost: Promise<Post>;
+
+  @Column({ type: 'text', nullable: true })
+  url?: string | null;
```
AmarTrebinjac, line 103: Added properties "url" and "imageUrl".

When users upload an external link, the image is already an url, and not a file upload. We also need to save the link to the item they are sharing since we don't have an existing post to reference to.

src/schema/posts.ts
```diff
@@ -1101,6 +1224,56 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      }));
+    },
+    sourcePostModeration: async (
+      _,
+      { id, sourceId }: { id: string; sourceId: string },
+      ctx: Context,
+      info,
+    ): Promise<GQLSourcePostModeration> => {
+      const isModerator = await isPrivilegedMember(ctx, sourceId);
+      if (!isModerator) throw new ForbiddenError('Access denied!');
```
capJavert, line 1263: @AmarTrebinjac for this one though `ensureSourcePermission` is exactly what we want no?

src/schema/posts.ts
capJavert: Ok, we should still check however that user can't create a post directly if its not allowed without moderation on the squad eg. someone just takes the API calls and does them manually through postman.

src/schema/sources.ts
```diff
@@ -934,7 +950,11 @@ export const canPostToSquad = (
     return false;
   }
 
-  return sourceRoleRank[sourceMember.role] >= squad.memberPostingRank;
+  const memberRank = sourceRoleRank[sourceMember.role];
+
+  if (memberRank >= sourceRoleRank.moderator) return true;
+
+  return !squad.moderationRequired && memberRank >= squad.memberPostingRank;
```
capJavert, line 957: Just sanity check if this will work for current squads for backward compatibility? `moderationRequired` is false by default?

src/schema/posts.ts
capJavert: See this is done in `canPostToSquad` 🟢

src/schema/posts.ts
capJavert: Just make sure its de-dupped so if someone sends `['pending', 'pending', 'rejected']` it becomes `['pending', 'rejected']` and it only accepts valid enums.

src/schema/posts.ts
```diff
@@ -1521,6 +1694,72 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        content,
+        commentary,
+        type,
+        url,
+        sharedPostId,
+        imageUrl,
+      }: CreateSourcePostModerationArgs,
+      ctx: AuthContext,
+      info,
+    ): Promise<GQLSourcePostModeration> => {
```
capJavert, line 1712: We should validate that user is a member of this source?

sshanzel (commented):
Looking great with the progress! Just had a few comment, but I think we are on the right track 🛤️

src/schema/sources.ts
AmarTrebinjac: Haha, yep! It already caused a problem 😅

https://github.com/dailydotdev/daily-api/pull/2324

src/schema/posts.ts
rebelchris: @sshanzel 
> In frontend it should be just a simple ternary operator

This calls for mistakes as well though, i can already envision we forget to add it in one specific place (let's say mobile via settings posting flow) and only realize weeks down the line.

In general it would be great if we could uniform some actions somehow?
I'm still ok with splitting, but maybe in FE it can be 1 call and that call determines in a hook what one to publish?

CC: @capJavert // @AmarTrebinjac would love to hear thoughts on how we can minimize risk with 2 flows.

sshanzel: > This calls for mistakes as well though, i can already envision we forget to add it in one specific place (let's say mobile via settings posting flow) and only realize weeks down the line.

The condition should lie in a single place (a hook, ideally) where all decisions are made, not on the component level. Otherwise, it would be a mess to put conditions all over the place in each flow.

src/schema/sources.ts
```diff
@@ -815,6 +815,7 @@ export enum SourcePermissions {
   Delete = 'delete',
   Edit = 'edit',
   ConnectSlack = 'connect_slack',
+  Moderate = 'moderate',
```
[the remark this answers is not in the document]
AmarTrebinjac: Updated in https://github.com/dailydotdev/daily-api/pull/2388

src/schema/posts.ts
```diff
@@ -1521,6 +1694,72 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        type,
+        sharedPostId,
+        url,
+        createdById: userId,
+      };
+
+      if (commentary) {
+        await validateCommentary(commentary);
+        pendingPost.titleHtml = generateTitleHtml(commentary, mentions);
+      }
```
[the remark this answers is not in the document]
AmarTrebinjac: We could explicitly check if its a freeform post, but freeform does not have the commentary property, so it should not be a problem.

PR #2375, opened 2024-10-31
feat: create notification for user top reader badges

omBratteng:
AS-667

### DIFF
--- a/src/notifications/builder.ts
+++ b/src/notifications/builder.ts
@@ -17,6 +17,7 @@ import {
   type UserTopReader,
 } from '../entity';
 import {
+  defaultImage,
   getDiscussionLink,
   getSourceLink,
   getUserPermalink,
@@ -286,7 +287,7 @@ export class NotificationBuilder {
       name: ctx.keyword.flags.title,
       targetUrl: '',
       referenceId: ctx.userTopReader.id,
-      image: pickImageUrl({ createdAt: new Date() }),
+      image: defaultImage.placeholder,
     });
     return this;
   }
--- a/src/notifications/builder.ts
+++ b/src/notifications/builder.ts
@@ -27,6 +27,7 @@ import {
   NotificationBundleV2,
   NotificationStreakContext,
   Reference,
+  type NotificationUserTopReaderContext,
 } from './types';
 import { NotificationIcon } from './icons';
 import { SourceMemberRoles } from '../roles';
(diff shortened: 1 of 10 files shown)


src/notifications/builder.ts
```diff
@@ -267,6 +278,19 @@ export class NotificationBuilder {
     return this;
   }
 
+  avatarTopReaderBadge(
+    ctx: NotificationUserTopReaderContext,
+  ): NotificationBuilder {
+    this.avatars.push({
+      type: 'top_reader_badge',
+      name: ctx.keyword.flags.title,
+      targetUrl: '',
```
rebelchris, line 287: This needs something like notification?topreader&badgeId no?

src/notifications/generate.ts
```diff
@@ -127,12 +128,27 @@ export const notificationTitleMap: Record<
...
 export const generateNotificationMap: Record<
   NotificationType,
   (builder: NotificationBuilder, ctx: never) => NotificationBuilder
 > = {
+  user_given_top_reader: (builder, ctx: NotificationUserTopReaderContext) =>
+    builder
+      .icon(NotificationIcon.TopReaderBadge)
+      .referenceUserTopReader(ctx.userTopReader)
+      .targetUrl(notificationsLink)
+      .setTargetUrlParameter([
```
rebelchris, line 146: Oh I see you do it here, maybe we should do this in the builder rather?

rebelchris (changes requested):
Minor comments

src/notifications/builder.ts
omBratteng: I don't use that URL in frontend, but it has to be set in DB, se empty string it is 😬

src/notifications/generate.ts
omBratteng: Not sure what you mean? Doing it like this for `streak_reset_restore`, `promoted_to_admin` and `promoted_to_moderator` notifications.

rebelchris: Oh right, I thought we set it in the builder already, that case never mind me :)

rebelchris (approved):

omBratteng:
Forgot to add email notification 🙈

omBratteng:
Added test

sshanzel (approved):
Looks good to me! :shipit: 

Just a question: where do we trigger the generation of this notification?

omBratteng:
> Just a question: where do we trigger the generation of this notification?

@sshanzel black magic? 😈

It's going to be triggered via a cronjob I'm working on

PR #2376, opened 2024-10-31
feat: add `topReaderBadge` to alerts and create new query to fetch top reader badges

omBratteng:
AS-662

### DIFF
--- a/src/entity/Alerts.ts
+++ b/src/entity/Alerts.ts
@@ -60,8 +60,8 @@ export class Alerts {
 
   bootPopup?: boolean;
 
-  @Column({ type: 'uuid', nullable: true })
-  topReaderBadge?: string | null;
+  @Column({ type: 'bool', default: false })
+  showTopReader?: boolean;
 
   @OneToOne('User', {
     lazy: true,
@@ -86,5 +86,5 @@ export const ALERTS_DEFAULT: Omit<Alerts, 'userId' | 'flags' | 'user'> = {
   lastBootPopup: null,
   bootPopup: false,
   showRecoverStreak: false,
-  topReaderBadge: null,
+  showTopReader: false,
 };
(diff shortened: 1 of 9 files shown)


src/entity/Alerts.ts
```diff
@@ -60,6 +60,9 @@ export class Alerts {
 
   bootPopup?: boolean;
 
+  @Column({ type: 'uuid', nullable: true })
+  topReaderBadge?: string | null;
```
omBratteng, line 64: @capJavert I wanted to make this reference the actual ID in the database, but not sure what relation to use with typeorm. Or should I just add it as a manual query?

omBratteng: My reasoning is to put the `id` of `user_top_reader` into alerts, then in frontend I can just use the `topReaderBadge` query based on that ID to fetch it, instead of `topReader` on user and then pick the latest

capJavert: Relation would be OneToOne with onDelete: `SET NULL`, basically FK if I see correctly. 

> Or should I just add it as a manual query?

Since user is fetched on most places I would probably lean to just doing a query, because else we need to maintain and update this alert which gives an opportunity of it getting out of sync. When we take the latest one we don't have to fetch anything else, we just show the data?

omBratteng: > Relation would be OneToOne with onDelete: SET NULL, basically FK if I see correctly.

I tried that, but the migration maker didn't output anything for it 😬

omBratteng: But I can change alert to be boolean, and then use the whoami query with topReader selector, but update it to accept a limit, so we don't need to fetch 5, when we only want the latest one anyway

capJavert: Yeah, boolean would be more simple, when would we remove the alert then, when users sees the popup?

omBratteng: When the user closes the modal

rebelchris (approved):
Looks good to me as we don't need pagination on it.

PR #2377, opened 2024-10-31
feat: add block mutations

capJavert:
AS-651

- add block mutations for all supported entities so far (with backward compatibility)
- squad entities (from other #2378)

### DIFF
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -4,8 +4,11 @@ import {
   ContentPreferenceType,
 } from '../entity/contentPreference/types';
 import { ContentPreferenceUser } from '../entity/contentPreference/ContentPreferenceUser';
-import { NotificationType } from '../notifications/common';
-import { EntityManager, EntityTarget, In } from 'typeorm';
+import {
+  NotificationPreferenceStatus,
+  NotificationType,
+} from '../notifications/common';
+import { EntityManager, EntityTarget, In, Not } from 'typeorm';
 import { ConflictError } from '../errors';
 import { ContentPreferenceFeedKeyword } from '../entity/contentPreference/ContentPreferenceFeedKeyword';
 import { ContentPreferenceSource } from '../entity/contentPreference/ContentPreferenceSource';
@@ -91,6 +94,7 @@ export const cleanContentNotificationPreference = async ({
     userId,
     referenceId: id,
     notificationType: In(notificationTypes),
+    status: Not(NotificationPreferenceStatus.Muted),
   });
 };
 
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -55,6 +55,7 @@ export const entityToNotificationTypeMap: Record<
   [ContentPreferenceType.Source]: [
     NotificationType.SourcePostAdded,
     NotificationType.SquadPostAdded,
+    NotificationType.SquadMemberJoined,
   ],
 };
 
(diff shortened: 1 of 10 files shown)


src/common/contentPreference.ts
```diff
@@ -278,3 +373,37 @@ export const unfollowEntity = ({
...
+export const unblockEntity = async ({
+  ctx,
+  id,
+  entity,
+}: {
+  ctx: AuthContext;
+  id: string;
+  entity: ContentPreferenceType;
+}): Promise<void> => {
+  // currently unblock is just like unfollow, eg. remove everything from db
```
rebelchris, line 407: Why remark, is it going to change?

src/schema/contentPreference.ts
```diff
@@ -304,6 +337,28 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     ): Promise<GQLEmptyResponse> => {
       await unfollowEntity({ ctx, id, entity });
 
+      return {
+        _: true,
+      };
+    },
+    block: async (
```
rebelchris, line 344: Not 100% sure, but do we also need to introduce the other side backward compatibility?
So the current block flow to set contentPreference?

(Maybe you did this in other PR already, can't recall)

rebelchris (approved):
Nothing blocking

src/common/contentPreference.ts
capJavert: Not sure currently, but thought note is due here since it might look as an error to use `unfollow`.

src/schema/contentPreference.ts
capJavert: I did add it, `blockKeyword` also saves to `Feedtag` with `blocked: true` and `blockSource` saves to `FeedSource`. 

Do we do anything additional with blocking?

sshanzel (approved):
Looks good to me. One question though, why does blocking and unblocking looks similar in terms of `notificationPreference`. They both have 0 entry on that entity, shouldn't we add an entry with `blocked = true` so that even default notifications won't be sent as they are blocked, if I understand correctly.

capJavert:
@sshanzel in both cases we wont to remove all notification preferences so user is not subscribed to anything, does that make sense?

> They both have 0 entry on that entity, shouldn't we add an entry with blocked = true so that even default notifications won't be sent as they are blocked, if I understand correctly.

We do add `blocked: true` for `FeedTag` and `FeedSource` for backward compatibility.

sshanzel:
> @sshanzel in both cases we wont to remove all notification preferences so user is not subscribed to anything, does that make sense?
> 
> > They both have 0 entry on that entity, shouldn't we add an entry with blocked = true so that even default notifications won't be sent as they are blocked, if I understand correctly.
> 
> We do add `blocked: true` for `FeedTag` and `FeedSource` for backward compatibility.

Thank you for the clarity, yeah, I was just wondering if it should also be the case for `User` entity. This is so, for example, when you are mentioned by the blocked user, you won't get notified, since your preference is `blocked`.

src/entity/contentPreference/ContentPreferenceSource.ts
```diff
@@ -14,11 +20,9 @@ export class ContentPreferenceSource extends ContentPreference {
   @Column({ type: 'text', default: null })
   feedId: string;
 
-  @Column({ type: 'text', default: SourceMemberRoles.Member })
-  role: SourceMemberRoles;
```
capJavert, line 18: Leaving role in db for now so I can copy data

PR #2378, opened 2024-10-31
As 652 squad content preference

capJavert:
AS-652

- add support for referralToken (managed to add it to base ContentPreferenceSource to not complicate with another ContentPreferenceType of `squad`)
- backward compatibility for mutations in sources schema
- few more relevant comments in code

### DIFF
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -17,6 +17,7 @@ import {
   NotificationPreferenceUser,
 } from '../entity';
 import { ghostUser } from './utils';
+import { randomUUID } from 'crypto';
 
 type FollowEntity = ({
   ctx,
@@ -195,9 +196,16 @@ const followSource: FollowEntity = async ({ ctx, id, status }) => {
       sourceId: id,
       feedId: ctx.userId,
       status,
+      referralToken: randomUUID(),
     });
 
-    await repository.save(contentPreference);
+    await repository
+      .createQueryBuilder()
+      .insert()
+      .into(ContentPreferenceSource)
+      .values(contentPreference)
+      .orUpdate(['status'], ['referenceId', 'userId'])
+      .execute();
 
     if (status !== ContentPreferenceStatus.Subscribed) {
       cleanContentNotificationPreference({
@@ -309,9 +317,16 @@ const blockSource: BlockEntity = async ({ ctx, id }) => {
       sourceId: id,
       feedId: ctx.userId,
       status: ContentPreferenceStatus.Blocked,
+      referralToken: randomUUID(),
     });
 
-    await repository.save(contentPreference);
+    await repository
+      .createQueryBuilder()
+      .insert()
+      .into(ContentPreferenceSource)
+      .values(contentPreference)
+      .orUpdate(['status'], ['referenceId', 'userId'])
+      .execute();
 
     cleanContentNotificationPreference({
       ctx,
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -45,7 +45,7 @@ type BlockEntity = ({
   id: string;
 }) => Promise<void>;
 
-const entityToNotificationTypeMap: Record<
+export const entityToNotificationTypeMap: Record<
   ContentPreferenceType,
   NotificationType[]
 > = {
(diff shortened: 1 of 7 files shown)


src/common/contentPreference.ts
```diff
@@ -309,16 +324,24 @@ const blockSource: BlockEntity = async ({ ctx, id }) => {
...
     });
 
-    await repository.save(contentPreference);
+    await repository
+      .createQueryBuilder()
+      .insert()
+      .into(ContentPreferenceSource)
+      .values(contentPreference)
+      .orUpdate(['status'], ['referenceId', 'userId'])
+      .execute();
```
capJavert, line 336: Changed this to upsert so that `referralToken` is preserved when moving from `follow` -> `subscribed`

src/entity/SourceMember.ts
```diff
@@ -54,7 +50,3 @@ export class SourceMember {
   @Column({ type: 'jsonb', default: {} })
   flags: SourceMemberFlags;
 }
-
-const TOKEN_BYTES = 32;
-export const generateMemberToken = async (): Promise<string> =>
```
capJavert, line 59: Removed this in favour of `crypto.randomUUID`, its faster, sync and more simple and I don't think we get anything from using custom alphabet.

src/schema/sources.ts
```diff
@@ -1095,10 +1101,25 @@ const addNewSourceMember = async (
   con: DataSource | EntityManager,
   member: Omit<DeepPartial<SourceMember>, 'referralToken'>,
 ): Promise<void> => {
+  const referralToken = randomUUID();
```
capJavert, line 1104: Existing tokens are as is, new ones will be UUID, it does not really change anything DB wise and length wise its almost the same (urls are shortened either way)

src/schema/sources.ts
```diff
@@ -1095,10 +1101,25 @@ const addNewSourceMember = async (
...
+    referralToken,
   });
+
+  await con.getRepository(ContentPreferenceSource).insert(
+    con.getRepository(ContentPreferenceSource).create({
+      userId: member.userId,
+      referenceId: member.sourceId,
+      sourceId: member.sourceId,
+      feedId: member.userId,
+      status: ContentPreferenceStatus.Subscribed,
```
capJavert, line 1117: When user joins a squad they are subscribed by default

src/schema/sources.ts
```diff
@@ -1917,9 +1977,32 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        if (isBlock) {
+          await cleanContentNotificationPreference({
+            ctx,
+            entityManager,
+            id: sourceId,
+            notificationTypes: entityToNotificationTypeMap.source,
+            notficationEntity: NotificationPreferenceSource,
+            userId: memberId,
+          });
+        }
```
capJavert, line 2004: Blocked users should not get notifications

src/entity/contentPreference/ContentPreferenceSource.ts
```diff
@@ -20,6 +20,10 @@ export class ContentPreferenceSource extends ContentPreference {
   @Column({ type: 'jsonb', default: {} })
   flags: SourceMemberFlags;
 
+  @Column({ type: 'text' })
+  @Index({ unique: true })
+  referralToken: string;
```
rebelchris, line 25: Why do we need it on content preference though?
Wouldn't it make more sense to keep it under source_member ship?

rebelchris (commented):
Just one question around the token placement, not quite sure I get the need for it.

src/entity/contentPreference/ContentPreferenceSource.ts
capJavert: Since joining a source (squad) will now be the same as follow all of it is now saved to `content_preference` table. Source member will be removed when follow migration is done.

rebelchris: What about `role` and flags in that case?

capJavert: they are also added (but in previous PR)

rebelchris (approved):

src/entity/contentPreference/ContentPreferenceSource.ts
capJavert: @rebelchris as concluded on slack I refactored to flags in 557f1cd8c7441244e24ea2137236867e5dea5f28

rebelchris (approved):
Looks good to me :)

PR #2380, opened 2024-10-31
feat: squad post approval background workers

sshanzel:
TODO:
- [ ] Tests for the content of the generated notification.
- [ ] Templated IDs for the relevant email notifications (waiting for product).
- [ ] Tests for the email templates.


src/entity/SquadPostModeration.ts
```diff
@@ -95,4 +95,9 @@ export class SquadPostModeration {
     onDelete: 'SET NULL',
   })
   sharedPost: Promise<Post>;
+
+  @Column({ type: 'text', nullable: true })
+  reason?: string;
+  @Column({ type: 'text', nullable: true })
+  comment?: string;
```
sshanzel, line 102: Just temporary here, this should be added to the other PR for entity.

src/notifications/icons.ts
```diff
@@ -10,5 +10,6 @@ export enum NotificationIcon {
   Star = 'Star',
   DevCard = 'DevCard',
   BookmarkReminder = 'BookmarkReminder',
+  Timer = 'Timer',
```
sshanzel, line 13: This needs to reflect in the frontend accordingly.

.infra/common.ts
```diff
@@ -297,6 +297,18 @@ export const workers: Worker[] = [
     topic: 'api.v1.post-visible',
     subscription: 'api.post-added-user-notification',
   },
+  {
+    topic: 'api.v1.squad-post-moderation-submitted',
```
rebelchris, line 301: You also did streams?

src/workers/notifications/squadPostModerationApproved.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const worker =
+  generateTypedNotificationWorker<'api.v1.squad-post-moderation-submitted'>({
+    subscription: 'api.v1.squad-post-moderation-submitted-notification',
+    handler: async ({ post: data }, con) => {
+      if (!data?.postId) {
+        return;
+      }
+
+      if (!data?.createdById) {
```
rebelchris, line 14: What's the use-case where this is null?

src/workers/notifications/squadPostModerationRejected.ts
```diff
@@ -0,0 +1,31 @@
...
+import { User } from '../../entity';
+import { SquadPostModerationStatus } from '../../entity/SquadPostModeration';
+
+const worker =
+  generateTypedNotificationWorker<'api.v1.squad-post-moderation-submitted'>({
+    subscription: 'api.v1.squad-post-moderation-rejected-notification',
+    handler: async ({ post }, con) => {
+      if (
+        !post?.createdById ||
+        post.status !== SquadPostModerationStatus.Rejected
```
rebelchris, line 13: Don't we need this check for the approved one as well? (to say it's not approved yet?)
Also maybe split it like the above one, i like how readable it is.

src/workers/notifications/squadPostModerationRejected.ts
```diff
@@ -0,0 +1,31 @@
...
+      ) {
+        return;
+      }
+
+      await con
+        .getRepository(User)
+        .findOneOrFail({ where: { id: post.createdById } });
```
rebelchris, line 20: Just to fail if user doesn't exist?

rebelchris (approved):
Nice work, nothing blocking.
Left some generic questions though, and please wait for Ante as well

.infra/common.ts
capJavert:
```suggestion
    topic: 'api.v1.source-post-moderation-submitted',
```

should we name it `source` so its generic? (also for others)

src/workers/notifications/squadPostModerationApproved.ts
```diff
@@ -0,0 +1,28 @@
+import { generateTypedNotificationWorker } from './worker';
+import { NotificationType } from '../../notifications/common';
+import { NotificationPostContext } from '../../notifications';
+import { buildPostContext } from './utils';
+
+const worker =
```
capJavert, line 6: Just naming wise again, should we name `source` instead of squad, so its generic and squads are just sources under the hood

src/workers/notifications/squadPostModerationRejected.ts
capJavert: Maybe better to query and early return since this will log errors from notification worker

capJavert (commented):
Looking good so far, just some sanity checks.

.infra/common.ts
sshanzel: Will create a PR for these 3 👀

src/workers/notifications/squadPostModerationApproved.ts
sshanzel: Just that TS strict thing. Either we early return it or use the `!` notation. I should remove this once the entity is updated from the other PR that `createdById` is NOT nullable anymore.

src/workers/notifications/squadPostModerationRejected.ts
sshanzel: Right, let me check any missing checks.

src/workers/notifications/squadPostModerationApproved.ts
sshanzel: Yeah, we can refactor to that to keep it consistent.

src/workers/notifications/squadPostModerationRejected.ts
sshanzel: Yeah, let's do early return instead.

src/workers/notifications/squadPostModerationApproved.ts
capJavert: Once entity is not nullable it will auto resolve 👍 Let's avoid `!`

sshanzel: Yeah, I tend to avoid them, too, as much as I can.

PR #2387, opened 2024-11-04
feat: content preference data copy scripts

capJavert:
AS-672

- copy data from old tables to content_preference
- section on data mapping from RFC https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1126432769/RFC-067+-+Follow#FeedTag
- only change from RFC is that we put `role` and `referralToken` in `ContentPreferenceSource` entity under `flags`
- added classic limit/offset and trigger disable as we do in these kind of scripts


bin/copyFeedSourceContentPreference.ts
```diff
@@ -0,0 +1,44 @@
...
+  const con = await createOrGetConnection();
+
+  await con.transaction(async (manager) => {
+    await manager.query(`SET session_replication_role = replica;`);
+
+    await manager.query(`
+      INSERT INTO content_preference ("userId", "referenceId", "type", "createdAt", "status", "sourceId", "feedId")
+      SELECT f."userId", fs."sourceId", 'source', NOW(), CASE WHEN fs."blocked" = True THEN 'blocked' ELSE 'follow' END, fs."sourceId", fs."feedId"
+      FROM feed_source fs
+      LEFT JOIN feed f ON f."id" = fs."feedId"
```
idoshamun, line 33: why `left join` and not join? this may cause weird edge cases

bin/copyFeedSourceContentPreference.ts
```diff
@@ -0,0 +1,44 @@
...
+      LIMIT ${limit} OFFSET ${offset}
+      ON CONFLICT ("referenceId", "userId") DO UPDATE
```
idoshamun, line 35: There could be foreign key violations, maybe we should take it into account here?

bin/copyFeedTagContentPreference.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (Number.isNaN(offset)) {
+    throw new Error('offset argument is invalid, it should be a number');
+  }
+
+  const con = await createOrGetConnection();
+
+  await con.transaction(async (manager) => {
+    await manager.query(`SET session_replication_role = replica;`);
+
+    await manager.query(`
```
idoshamun, line 29: All my comments apply here as well

bin/copySourceMemberContentPreference.ts
```diff
@@ -0,0 +1,44 @@
...
+  const con = await createOrGetConnection();
+
+  await con.transaction(async (manager) => {
+    await manager.query(`SET session_replication_role = replica;`);
+
+    await manager.query(`
+      INSERT INTO content_preference ("userId", "referenceId", "type", "createdAt", "status", "sourceId", "flags")
+      SELECT sm."userId", sm."sourceId", 'source', NOW(), CASE WHEN sm."role" = 'blocked' THEN 'blocked' ELSE 'subscribed' END, sm."sourceId", jsonb_build_object('role', sm."role", 'referralToken', sm."referralToken")
+      FROM source_member sm
+      LIMIT ${limit} OFFSET ${offset}
```
idoshamun, line 33: Do we really need batching here?

idoshamun: If we do, make sure to order by time otherwise we may get corrupted data

capJavert: can't really order by time because not all entities have `createdAt`, but I will add to at least preserve order somehow

bin/copyFeedSourceContentPreference.ts
```diff
@@ -0,0 +1,44 @@
...
+  const offset = +offsetArgument;
+
+  if (Number.isNaN(offset)) {
+    throw new Error('offset argument is invalid, it should be a number');
+  }
+
+  const con = await createOrGetConnection();
+
+  await con.transaction(async (manager) => {
+    await manager.query(`SET session_replication_role = replica;`);
```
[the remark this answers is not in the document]
capJavert: disabling triggers, but since we don't have those on content_preference yet I can remove indeed

bin/copyFeedSourceContentPreference.ts
capJavert: Only column that does not have FK is `feed_tag.tag` (missing one to `keyword` and from what I have seen it does contain valid values, will double check

good thing nothing reads from content_preference for those entities yet so I can easily run all of this multiple times and adjust the data

bin/copySourceMemberContentPreference.ts
capJavert: I think we need batching because tables are quite bit and from doing it before its easier to grasp what is happening like that

bin/copyFeedSourceContentPreference.ts
capJavert: updated

bin/copyFeedSourceContentPreference.ts
idoshamun: yes we don't need fk to tag anyway. so no worries then, ignore my comment

bin/copySourceMemberContentPreference.ts
idoshamun: just noticed we need order by for all queries

capJavert: yes

capJavert: I can do:
- source_member.createdAt
- feed_source.sourceId
- feed_tag.tag

?

idoshamun: my concern is that it's not an immutable order. it may change as we get more data.
in this case i'd suggest to snapshot the data to a temporary table and batch from there or no batching

capJavert: but why snapshot, its not breaking anything? We can always clear content preference since its not used

capJavert: @idoshamun also one thing that might not be obvious, but API is already writing to this new tables (in addition to old ones) so we just have to do it once to have data parity

idoshamun: My concern is that some data will be missed due to immutable order during this migration. But if new data is being added already then it shouldn't be an issue

capJavert: yep, my bad for not highlighting that in description of the PR as well

idoshamun (approved):

rebelchris (approved):

PR #2388, opened 2024-11-05
feat: Source post moderation entity and queries

AmarTrebinjac:
After some discussions with @ilasw and @sshanzel we decided to split the Source Post Moderation and the Query into its own PR again. Coupling it with the creation of posts was a mistake on my part and has held up some tasks unnecessarily long.

So this is a split of [Squad post moderation flow](https://github.com/dailydotdev/daily-api/pull/2373), taking out the entity, migration and the query into their own PR to get it reviewed and approved faster.


src/entity/SourcePostModeration.ts
```diff
@@ -0,0 +1,104 @@
...
+  postId: string | null;
+
+  @ManyToOne('Post', (post: Post) => post.id, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  post: Promise<Post>;
+
+  @Column({ type: 'text' })
+  type: string;
```
ilasw, line 76: Do we have some more specific type to apply? Something like `PostType`?

I see that in Post entity we are using this:

```ts
@Column({ default: PostType.Article })
type: PostType;
```

Maybe we can have a strict version only with squad-like posts.

src/entity/SourcePostModeration.ts
```diff
@@ -0,0 +1,104 @@
...
+  title?: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  titleHtml?: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  content?: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  contentHtml: string | null;
```
ilasw, line 88: I would expect that both `content` and `contentHtml` are required/optional, there is some reason for this?

src/entity/SourcePostModeration.ts
```diff
@@ -0,0 +1,104 @@
...
+  image: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  sharedPostId?: string | null;
+
+  @ManyToOne('Post', (post: Post) => post.id, {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  sharedPost: Promise<Post>;
```
ilasw, line 100: Maybe this should be optional? 👀

src/schema/posts.ts
```diff
@@ -1101,6 +1223,117 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   BaseContext
 >({
   Query: {
+    sourcePostModeration: async (
+      _,
+      { id, sourceId }: { id: string; sourceId: string },
+      ctx: Context,
+      info,
+    ): Promise<GQLSourcePostModeration> => {
+      await ensureSourcePermissions(ctx, sourceId);
```
ilasw, line 1232: This function ensure that we can view this source post, but we need a deep check on this. 

Only `Moderators/Admins` or `post author` can query a specific post.

ilasw (changes requested):
Seems good overall, just few questions

src/entity/SourcePostModeration.ts
```diff
@@ -0,0 +1,104 @@
...
+export enum SourcePostModerationStatus {
+  Approved = 'approved',
+  Rejected = 'rejected',
+  Pending = 'pending',
+}
+
+@Entity()
+export class SourcePostModeration {
+  @PrimaryColumn({ type: 'text' })
+  id: string;
```
rebelchris, line 22: Isn't this just generated uuid?

src/entity/SourcePostModeration.ts
```diff
@@ -0,0 +1,104 @@
...
+  titleHtml?: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  content?: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  contentHtml: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  image: string | null;
```
rebelchris, line 91: Should be optional I guess here as well

src/schema/posts.ts
rebelchris: Ah good point, we can probably extend the function if moderation is on it should check that part as well?

src/schema/posts.ts
```diff
@@ -1101,6 +1223,117 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        }),
+      );
+    },
+    sourcePostModerations: async (
+      _,
+      args: SourcePostModerationArgs,
+      ctx: Context,
+      info,
+    ): Promise<ConnectionRelay<GQLSourcePostModeration>> => {
+      const isModerator = await isPrivilegedMember(ctx, args.sourceId);
```
rebelchris, line 1252: I think we need sourcePermissions here as well right?

src/schema/posts.ts
capJavert: it should be a new `SourcePermissions` value which can be passed as a 3rd argument to `ensureSourcePermissions`

src/entity/SourcePostModeration.ts
capJavert: Good catch, it can be of type `PostType` because we want this to be generic not just for squads
```suggestion
  type: PostType;
```

src/entity/SourcePostModeration.ts
capJavert: content can be empty for shared posts

src/entity/SourcePostModeration.ts
AmarTrebinjac: This is how we define the primary column in most of our entities, is it not? 🤔

Example post:

```
export class Post {
  id: string;
```

rebelchris: Post is kind of legacy old. it also generates it's own ID which is odd.
You can use something like this maybe?
https://github.com/dailydotdev/daily-api/blob/bf166a5159205b029bf32450000cdb47086791d2/src/entity/DevCard.ts#L23

src/entity/SourcePostModeration.ts
AmarTrebinjac: @rebelchris Do we not always insert a placeholder where there is no image?

rebelchris: No for freeform we allow no image for instance.

src/schema/posts.ts
AmarTrebinjac: > This function ensure that we can view this source post, but we need a deep check on this.
> 
> Only `Moderators/Admins` or `post author` can query a specific post.

I originally used isPrivileged here, but assumed that maybe an user can also look up their individual post. I'll update it

src/entity/SourcePostModeration.ts
AmarTrebinjac: Alright, will update

src/schema/posts.ts
AmarTrebinjac: No. The isPrivilegedMember function gets the source user by using userId + sourceId, so it will only get members of the source. I should modify the findOneBy to findOneByOrFail, so that if you're not a member at all it should error.

If it actually finds the user it checks if the user is of rank moderator or higher.

capJavert: If this will now also throw, this should be ensureSourcePermissions, if not let's leave it as is, it was brought by multiple times already in other PR

AmarTrebinjac: Clarified in DMs 🤝

capJavert: 🤝

__tests__/fixture/source.ts
```diff
@@ -69,4 +69,13 @@ export const sourcesFixture: DeepPartial<MachineSource>[] = [
     active: true,
     private: true,
   },
+  {
+    id: 'm',
```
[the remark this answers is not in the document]
AmarTrebinjac: I'll move this stuff around a little in the original PR if that's okay with you? 

Just want the entity and query functions in ASAP so it's not blocking 😅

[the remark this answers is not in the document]
rebelchris: Sure that's fine with me :)

src/entity/SourcePostModeration.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export enum SourcePostModerationStatus {
+  Approved = 'approved',
+  Rejected = 'rejected',
+  Pending = 'pending',
+}
+
+@Entity()
+export class SourcePostModeration {
+  @PrimaryGeneratedColumn()
```
rebelchris, line 21: @capJavert Does he need `uuid` here? (not sure it's needed since it's handled on migration)

src/migration/1730799822114-SourcePostModeration.ts
```diff
@@ -0,0 +1,45 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SourcePostModeration1730799822114 implements MigrationInterface {
+  name = 'SourcePostModeration1730799822114';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "source_post_moderation" ("id" SERIAL NOT NULL, "sourceId" text NOT NULL, "status" text NOT NULL, "createdById" character varying, "moderatedById" character varying, "moderatorMessage" text, "rejectionReason" text, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "updatedAt" TIMESTAMP NOT NULL DEFAULT now(), "postId" text, "type" text NOT NULL, "title" text, "titleHtml" text, "content" text, "contentHtml" text, "image" text, "sharedPostId" text, "externalLink" text, CONSTRAINT "PK_5198be0eb3f1ebe81e75ee07865" PRIMARY KEY ("id"))`,
```
rebelchris, line 8: Not sure SERIAL is the right one here though

rebelchris (changes requested):
Two comments on the ID, but requires @capJavert opinion I think.

src/entity/SourcePostModeration.ts
capJavert: yeah we usually use `@PrimaryGeneratedColumn('uuid')`

src/schema/posts.ts
```diff
@@ -1101,6 +1223,121 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      _,
+      { id, sourceId }: { id: string; sourceId: string },
+      ctx: Context,
+      info,
+    ): Promise<GQLSourcePostModeration> => {
+      await ensureSourcePermissions(
+        ctx,
+        sourceId,
+        SourcePermissions.ModeratePost,
+      );
```
ilasw, line 1236: I think that if we are going to update this line we should update even the `ensureSourcePermissions` function and put an additional if `permission === SourcePermissions.ModeratePost` check (if user is mod or is the current author). 

Plus: maybe `ModeratePost` is not the best name for this. What do you think about something like `ViewModerationPost`?

ilasw (changes requested):

src/schema/posts.ts
AmarTrebinjac: I think that's going to be very clunky because it's going to require us to pass the Moderation Entity into ensureSourcePermissions as well, or fetch it first and then send the author id.

I think in this case it would then be better to again check isPrivilegedMember, and if not, try add an additional "andWhere userId" clause on the graphorm query.

Thoughts @capJavert  ?

ilasw: I'm open for both approaches ✔️ 

My main concern is that we added a value to enum that is not reflecting action (view post) nor what the method is doing.

src/entity/SourcePostModeration.ts
AmarTrebinjac: Will update to specify UUID as in Ante's comment

src/schema/posts.ts
capJavert: Do agree with Luca that enum value does not match.

Since this is just read operation again, it looks more simple to me to use `isPrivilegedMember` in this case indeed. 

Let's focus on what needs to be done:
- user should not be able to get post moderation when he is not a member of source
- user should be able to get its post moderation
- admin/moderator should be able to get all post moderations for source

Anything else?

If above is implemented and covered with tests to me its good.

ilasw (approved):
Looks good to me! Great work 🎉

capJavert (commented):
Some final comments, overall pretty clean implementation, good job swimming through all the changes.

src/schema/posts.ts
```diff
@@ -1101,6 +1223,129 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+              .offset(page.offset);
+
+            if (args.status) {
+              builder.queryBuilder.andWhere(
+                `"${builder.alias}"."status" IN (:...status)`,
+                {
+                  status: Array.from(new Set(args.status)),
+                },
+              );
+            }
```
ilasw, line 1298: Maybe we want to set `pending` as default status, wdyt?

AmarTrebinjac: I think if no status is set we don't query on status

capJavert (approved):

src/schema/posts.ts
capJavert: if no statuse were sent then the result would always be empty, because `Array.from(new Set(undefined))` is `[]`

src/schema/posts.ts
```diff
@@ -1101,6 +1223,129 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+              })
+              .orderBy(`${builder.alias}.updatedAt`, 'DESC')
+              .limit(page.limit)
+              .offset(page.offset);
+
+            if (args.status) {
+              builder.queryBuilder.andWhere(
+                `"${builder.alias}"."status" IN (:...status)`,
+                {
+                  status: Array.from(new Set(args.status)),
```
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert  Doing this and making it undefined by default so it doesn't pass as true on empty array

PR #2392, opened 2024-11-05
feat: create feed on user creation

capJavert:
Most users get the feed as soon as they register while going through onboarding, but when quitting or going through other signup flows feed gets created only when they add some filters to feed.

This makes it not consistent and some of our FKs that rely on feed existing for users sometimes fail. 

There is about 40k user rows that don't have feed (I'll run a separate query to update those after this is merged).


rebelchris (approved):
Looks good to me :)

PR #2393, opened 2024-11-05
fix: notification on thread flow

rebelchris:
We had to seperate issues we are solving here:

1. Currently all users in a thread get notified on each comment (perceived as spam and removed) -> https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1730805899671609?thread_ts=1729434869.622669&cid=C01L0QXQJNB
2. Muting only affected sub comment, which you would mute, but since we don't even support third level comments it did literally nothing 🤦 , So now we do some black magic to check if comments have a parent comment and rather mute that :) 
3. To resolve point 2 we decided on adding both retro fix on read and actual mutate to showcase the parent instead of sub reference.

Adjusted test cases to resolve these changes as well.

Solves: https://github.com/dailydotdev/daily/issues/1504


src/schema/notifications.ts
```diff
@@ -422,6 +423,22 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         });
       }
 
+      const newCommentComments = data.filter(({ notificationType }) =>
```
capJavert, line 426: I know its not a part of this pr but just noticed that user can send unrestricted amount of `data` here, can we do just a quick `if (data.length) > 100) throw`, to avoid doing large `IN()` statements?

src/workers/notifications/commentReply.ts
```diff
@@ -37,30 +37,24 @@ const worker: NotificationWorker = {
...
-    const threadFollowers = await con
-      .getRepository(Comment)
-      .createQueryBuilder()
-      .select('"userId"')
-      .distinct(true)
-      .where('"parentId" = :parentId', { parentId: parent.id })
-      .andWhere('"userId" not in (:...exclude)', {
-        exclude: [parent.userId, comment.userId],
-      })
-      .getRawMany();
```
capJavert, line 57: Where did this go?

src/schema/notifications.ts
```diff
@@ -422,6 +423,22 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      if (newCommentComments.length) {
+        const commentIds = newCommentComments.map(
+          ({ referenceId }) => referenceId,
+        );
+        const commentComments = await ctx
+          .getRepository(Comment)
+          .find({ select: ['id', 'parentId'], where: { id: In(commentIds) } });
+        commentComments.forEach(({ id, parentId }) => {
+          const param = params.find(({ referenceId }) => referenceId === id);
+          param!.referenceId = parentId || id;
```
capJavert, line 438:
```suggestion
          if (!param) {
              return
          }

          param.referenceId = parentId || id;
```

capJavert (commented):
Just some sanity checks

src/schema/notifications.ts
capJavert: (also fine to just open a task for this)

capJavert: https://dailydotdev.atlassian.net/browse/AS-673

src/workers/notifications/commentReply.ts
rebelchris: This is point a, its spammy to receive these if you simply in the thread 👍

capJavert: Ah right!

capJavert (approved):

src/schema/notifications.ts
rebelchris: This doesn't work, we need to always set it regardless of it existing or not.
In your version we basically never set the parent.

src/schema/notifications.ts
rebelchris: I also fixed this one :)

src/schema/notifications.ts
capJavert: But you should not need to force coercion with `!.` If `param` is `undefined` the code will break.

PR #2400, opened 2024-11-06
feat: create source post moderation

AmarTrebinjac:
Created a new PR for creating source post moderation because [the previous one became a mess](https://github.com/dailydotdev/daily-api/pull/2373). Between the weird initial commit history and splitting off half the PR into another one and renaming a bunch of stuff, I thought it'd be easier to just make a new PR and avoid dupes and merge conflicts. Still accounting for the feedback given in the previous PR.


__tests__/posts.ts
```diff
@@ -3473,6 +3471,94 @@ describe('query sourcePostModeration', () => {
   });
 });
 
+describe('mutation createSourcePostModeration', () => {
```
ilasw, line 3474: I think we should add some tests for user (member/mod/admin) trying to post in squad directly. 
This is an important point for this PR imo 👀

src/routes/boot.ts
```diff
@@ -195,16 +195,17 @@ const getSquads = async (
...
         { memberPostingRank },
       );
-      // we only send posting permissions from boot to keep the payload small
-      const postingPermissions = permissions.filter(
-        (item) => item === SourcePermissions.Post,
+      // we only send posting and moderation permissions from boot to keep the payload small
+      const essentialPermissions = permissions.filter(
+        (item) =>
+          item === SourcePermissions.Post ||
+          item === SourcePermissions.ModeratePost,
```
ilasw, line 202: Why are we changing this? If we are Moderator we can also post, right?

src/schema/posts.ts
```diff
@@ -1582,6 +1635,91 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      }
+
+      await canAccessSource(
+        ctx,
+        source,
+        sourceMember,
+        SourcePermissions.Post,
+        sourceId,
+      );
+      const id = generateUUID();
```
ilasw, line 1674: I think `id` should be autogenerated in insert, do we need to do it here?

ilasw (changes requested):
Seems good code wise, mainly we should add some more tests

src/schema/posts.ts
AmarTrebinjac: This is how we do it in the other requests where we return the entity that was just created with graphql as far as I can tell

__tests__/posts.ts
```diff
@@ -3473,6 +3471,94 @@ describe('query sourcePostModeration', () => {
...
+      variables: { ...params, sharedPostId: 'p1', type: PostType.Share },
+    });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.createSourcePostModeration).toBeTruthy();
+    expect(res.data.createSourcePostModeration.type).toEqual(PostType.Share);
+    expect(res.data.createSourcePostModeration.contentHtml).toBeDefined();
+    expect(res.data.createSourcePostModeration.sharedPostId).toEqual('p1');
+  });
+
+  it('should successfully create a squad post moderation entry of type article', async () => {
```
sshanzel, line 3535: It should not actually be possible to create a type of `article` as that only comes from the scraper.

__tests__/posts.ts
```diff
@@ -3473,6 +3471,94 @@ describe('query sourcePostModeration', () => {
...
+    const res = await client.mutate(MUTATION, {
+      variables: { ...params, type: PostType.Freeform },
+    });
+    expect(res.errors).toBeFalsy();
+    expect(res.data.createSourcePostModeration).toBeTruthy();
+    expect(res.data.createSourcePostModeration.type).toEqual(PostType.Freeform);
+    expect(res.data.createSourcePostModeration.contentHtml).toBeDefined();
+  });
+
+  it('should successfully create a squad post moderation entry of type share', async () => {
```
sshanzel, line 3523: For `share` type, there is to types, one when there is a `sharePostId`, and one when the `externalLink` is present.

sshanzel (commented):
Looks amazing overall, I think we just missing a few more tests but we are almost there 🤩

__tests__/posts.ts
AmarTrebinjac: @sshanzel Are you sure? the RFC mentions the type can be article, share or freeform. I assumed Article was a shared link. Looking at createExternalLink , it adds a post of type article:

```
 await entityManager.getRepository(ArticlePost).insert({
```

src/routes/boot.ts
rebelchris: But moderator can always bypass which you already know based on currentUser when posting.
I also don't think we need this change.

src/schema/posts.ts
```diff
@@ -1582,6 +1635,91 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const sourceMember = await con.getRepository(SourceMember).findOne({
+        where: { sourceId: sourceId, userId: userId },
+        relations: ['source'],
+      });
+
+      const source = await sourceMember?.source;
+
+      if (!source) {
+        throw new ForbiddenError('Access denied!');
+      }
```
rebelchris, line 1665: Doesn't `canAccessSource` handle this already?

src/schema/posts.ts
rebelchris: Leaving this up to @capJavert we do use both ways.
Think uuid only for article related stuff, but it's more for yggdrasil I think.

src/schema/posts.ts
```diff
@@ -1582,6 +1635,91 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+            id,
+            upload.createReadStream(),
+            UploadPreset.PostBannerImage,
+          );
+          pendingPost.image = coverImageUrl;
+        }
+        await createSourcePostModeration(manager, pendingPost);
+      });
+
+      return graphorm.queryOneOrFail<GQLSourcePostModeration>(
```
rebelchris, line 1711: If we going to query anyway, you might as well just get the return value from the insert?

rebelchris (commented):
Very minor feedback, overall good approach.

__tests__/fixture/source.ts
```diff
@@ -79,3 +85,18 @@ export const sourcesFixture: DeepPartial<MachineSource>[] = [
     private: false,
   },
 ];
+
+export const squadFixture: DeepPartial<SquadSource>[] = [
```
capJavert, line 89: I would just define this in the file you are using it, we want to move away from fixtures probably since they often create flaky tests

src/schema/posts.ts
capJavert: We don't have to generate UUID, psql will automatically generate it since its a primary column.

src/schema/sources.ts
```diff
@@ -955,7 +957,11 @@ export const canPostToSquad = (
     return false;
   }
 
-  return sourceRoleRank[sourceMember.role] >= squad.memberPostingRank;
+  const memberRank = sourceRoleRank[sourceMember.role];
+
+  if (memberRank >= sourceRoleRank.moderator) return true;
```
capJavert, line 962: why is this needed specifically, does below case not solve it?

src/schema/posts.ts
capJavert: +1

also do we want to check type of source as well, and throw if its not a squad, just to make sure no accidental moderations for sources if someone is playing with API directly, you can see example in `joinSource`

src/common/post.ts
```diff
@@ -220,6 +225,41 @@ export const createFreeformPost = async (
...
+export const createSourcePostModeration = async (
+  con: DataSource | EntityManager,
+  args: CreateSourcePostModeration,
+) => {
+  const newPost = con.getRepository(SourcePostModeration).create({
+    ...args,
+    status: SourcePostModerationStatus.Pending,
+  });
+  await con.getRepository(SourcePostModeration).save(newPost);
+  return newPost;
```
capJavert, line 249: We should return the result from `save` because it will have PK and other defaults from the DB.

src/schema/posts.ts
AmarTrebinjac: We do this on other post creation as well. Asked @capJavert about it and the reason we do it this way is to retain the GraphQL schema

src/schema/posts.ts
AmarTrebinjac: @rebelchris it doesn't because source is not an optional parameter

src/schema/sources.ts
AmarTrebinjac: @capJavert  We need it so moderators can bypass the moderation required check. If we remove this line, the next one will return false even if user is moderator or higher.

src/schema/posts.ts
rebelchris: Yeah aware of that, meant more you could maybe return the inserted ID, then return the graphORM layer?

src/routes/boot.ts
rebelchris: @capJavert / @ilasw  I guess since we decided to handle deciding if it's moderated post on frontend, they need it in boot, as we have flows (context -> share to squad) that rely soley on boot squads now.

(Just FYI, as I struggled to understand the case at first)
Thanks @AmarTrebinjac for explaining the edge-case your facing.

capJavert: ok from my side 👍

AmarTrebinjac: The TL:DR is basically: 

Post flows use the squads provided by boot, and not the source query

src/schema/posts.ts
```diff
@@ -1582,6 +1636,95 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      if (commentary) {
+        await validateCommentary(commentary);
+        pendingPost.titleHtml = generateTitleHtml(commentary, mentions);
+      }
+      if (content) {
+        pendingPost.contentHtml = markdown.render(content, { mentions });
+      }
+
+      let newPostId: string;
```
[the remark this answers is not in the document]
AmarTrebinjac: @rebelchris the await is in a transaction, and the return is outside that scope. But looking on it, there's not really any need for the transaction, so I'll remove it and it should be good.

__tests__/posts.ts
sshanzel: Yes. We may create the article post behind the scenes but we don't tag it as created by the user. We will create another post with `Share` type that is then referenced to the logged user.

sshanzel: This means, your request is of `share` type than `article.

rebelchris (approved):
Not much to add from my side.
Nice work

sshanzel (approved):
Don't have anything blocking! Looks amazing already, just want to see the same tests be applied where applicable.

sshanzel (approved):
Amazing work. Once @capJavert approves, I think we should be safe to merge.

src/schema/sources.ts
capJavert: I get it but it should still compare with `squad.memberPostingRank`, because if we introduce squads where only admins can post this would no longer work.

Below is more correct I would say (please check):

```suggestion
  if (memberRank >= sourceRoleRank.moderator && memberRank >= squad.memberPostingRank) {
  return true;  
}
```

capJavert (approved):

PR #2402, opened 2024-11-06
feat: calculate top readers

omBratteng:
TODO:
- [x] Add test for worker

AS-668


.infra/crons.ts
```diff
@@ -106,4 +106,8 @@ export const crons: Cron[] = [
     schedule: '*/5 * * * *',
     activeDeadlineSeconds: 4 * 60,
   },
+  {
+    name: 'calculate-top-readers',
+    schedule: '0 2 1 * *',
```
omBratteng, line 111: First of the month at 02:00.

The cron `update-highlighted-views` that runs at 04:15 each day handles the refresh of the UserStats MV, so no point really in running it in this cron.

rebelchris (approved):
Overall looks good to me :) 
Just note we blocked by the magic URL

src/migration/1730903564613-UserTopReader.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class UserTopReader1730903564613 implements MigrationInterface {
+  name = 'UserTopReader1730903564613'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE INDEX IF NOT EXISTS "IDX_post_keyword_status" ON "post_keyword" ("status")`);
+    await queryRunner.query(`CREATE UNIQUE INDEX IF NOT EXISTS "IDX_user_top_reader_userId_issuedAt_keywordValue" ON "user_top_reader" ("userId", "issuedAt", "keywordValue") `);
```
capJavert, line 8: This intended, don't see it defined anywhere?

src/cron/calculateTopReaders.sql
```diff
@@ -0,0 +1,75 @@
+WITH
+views AS MATERIALIZED (
```
capJavert, line 2: Do we need it materialized since its run only once?

src/cron/calculateTopReaders.ts
```diff
@@ -0,0 +1,119 @@
...
+          const { userId } = topReader;
+
+          // We need to exit the loop once we've reached the limit of top readers per keyword
+          if (counter > LIMIT_PER_KEYWORD) {
+            logger.debug({ keyword, counter }, 'limit reached');
+            break; // Break out of the loop and continue to the next keyword
+          }
+
+          // We need to ensure that we don't assign the same userId to multiple keywords
+          if (userIds.includes(userId)) {
```
capJavert, line 81: This could be a Map or object, more efficient to lookup existing then array

src/cron/calculateTopReaders.ts
```diff
@@ -0,0 +1,119 @@
...
+            logger.debug({ userId, keyword }, 'duplicate userId');
+            continue; // Continue to the next user
+          } else {
+            userIds.push(userId);
+          }
+
+          topReaders[keyword].push(userId);
+          counter++;
+
+          await con.getRepository(UserTopReader).upsert(
```
capJavert, line 91: Maybe open a transaction top level, not sure how much time upsert would be called on average run?

src/cron/calculateTopReaders.ts
```diff
@@ -0,0 +1,119 @@
...
+              keywordValue: keyword,
+              issuedAt: issuedAt,
+            },
+            {
+              conflictPaths: ['userId', 'issuedAt', 'keywordValue'],
+              skipUpdateIfNoValuesChanged: true,
+            },
+          );
+
+          continue; // Continue to the next user
```
capJavert, line 103: Why do you need a continue here, would it not go to next iteration either way?

src/workers/notifications/userTopReaderAdded.ts
```diff
@@ -0,0 +1,84 @@
...
+          image: uploadedImage.url,
+        },
+      );
+
+      await con.getRepository(Alerts).update(
+        { userId },
+        {
+          showTopReader: true,
+        },
+      );
```
capJavert, line 71: Could do it in a single transaction as well

src/cron/calculateTopReaders.sql
omBratteng: I think it improved the query time by a few seconds each run

src/migration/1730903564613-UserTopReader.ts
omBratteng: Forgot to add it. Done now.

src/cron/calculateTopReaders.sql
omBratteng: actually no, scratch that. I'd worsened it a bit. Will remove 🙈

src/entity/user/UserTopReader.ts
```diff
@@ -4,6 +4,9 @@ import type { Keyword } from '../Keyword';
 
 @Entity()
 @Index('IDX_user_top_reader_userId_issuedAt', { synchronize: false })
+@Index('IDX_user_top_reader_userId_issuedAt_keywordValue', {
+  synchronize: false,
+})
```
capJavert, line 9:
```suggestion
@Index(
  'IDX_user_top_reader_userId_issuedAt_keywordValue',
  ['userId', 'issuedAt', 'keywordValue'],
  { unique: true },
)
```

this would be equivalent?

src/workers/cdc/primary.ts
```diff
@@ -971,6 +972,19 @@ const onUserCompanyCompanyChange = async (
...
+  _: DataSource,
+  logger: FastifyBaseLogger,
+  data: ChangeMessage<UserTopReader>,
+) => {
+  if (data.payload.op !== 'c') {
+    return;
+  }
+  await triggerTypedEvent(logger, 'api.v1.user-top-reader', {
+    userTopReader: data.payload.after!,
+  });
```
capJavert, line 985: We only want `u` no, for `d` we would not want to run it?

src/entity/user/UserTopReader.ts
omBratteng: Yup

src/cron/calculateTopReaders.ts
omBratteng: You're right. It was a leftover from when I tried `Array.every`

src/cron/calculateTopReaders.ts
omBratteng: Running in transaction now

src/cron/calculateTopReaders.ts
omBratteng: Refactored to be Map

src/cron/calculateTopReaders.ts
```diff
@@ -0,0 +1,119 @@
...
+    const keywords: Record<string, Omit<TopReaderQueryResult, 'keyword'>[]> =
+      {};
+
+    try {
+      const sql = await readFile(
+        path.join(__dirname, 'calculateTopReaders.sql'),
+        'utf8',
+      );
+
+      const result = await con.query<TopReaderQueryResult[]>(sql, [
```
capJavert, line 41: run it on read replica?

capJavert (approved):
Code wise looks good, is it possible to somehow trigger a test run before first "official month"?

src/cron/calculateTopReaders.ts
capJavert: can use `queryReadReplica`

omBratteng: Ideally, yes. but read replica has issue with long queries, as it'll terminate them.
We can adjust this setting, but as we get more users and views, the query duration will likely increase, and we'll need to adjust this

So I would prefer running it on master

capJavert: ok, let's leave it on master for now, but in the future would like it to be on rr

PR #2403, opened 2024-11-07
AS-615 reduce freeform post minimum content length

denisb0:
Reduced `FREEFORM_POST_MINIMUM_CONTENT_LENGTH` to process most of the posts. Probably could set to 0 as well?
Did not change `FREEFORM_POST_MINIMUM_CHANGE_LENGTH`.

### DIFF
--- a/src/entity/posts/FreeformPost.ts
+++ b/src/entity/posts/FreeformPost.ts
@@ -2,7 +2,7 @@ import { ChildEntity, Column } from 'typeorm';
 import { Post, PostType } from './Post';
 
 // Minimun content length required for new posts to trigger content-requested
-export const FREEFORM_POST_MINIMUM_CONTENT_LENGTH = 50;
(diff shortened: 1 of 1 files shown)


src/entity/posts/FreeformPost.ts
```diff
@@ -2,7 +2,7 @@ import { ChildEntity, Column } from 'typeorm';
 import { Post, PostType } from './Post';
 
 // Minimun content length required for new posts to trigger content-requested
-export const FREEFORM_POST_MINIMUM_CONTENT_LENGTH = 50;
+export const FREEFORM_POST_MINIMUM_CONTENT_LENGTH = 5;
```
omBratteng, line 5: @capJavert should we change this to use GB so we can adjust it on the fly?

capJavert: Yeah, could be good

capJavert (approved):

PR #2405, opened 2024-11-07
feat: source post moderation cdc - post creation

sshanzel:
Changes:
- CDC that manages the creation of posts.
- The RFC may have specified on post delete, we need to delete the relevant requests, but we won't need it anymore since we cascade on delete on their relation.
- Refactored some functions to make it reusable to creating posts.


sshanzel:
So glad the tests have passed 😌 

Time to write a ton of tests now for the CDC.

src/common/post.ts
```diff
@@ -329,3 +344,125 @@ export const insertCodeSnippetsFromUrl = async ({
...
+): Promise<ModeratedPostCdc | null> => {
+  if (!moderated?.postId) {
+    logger.error('unable to update moderated post', { moderated });
+    return null;
+  }
+
+  const postParam = { id: moderated.postId };
+  const repo = con.getRepository(Post);
+
+  await repo.findOneOrFail({ where: { id: moderated.postId } });
```
rebelchris, line 362: This is just for failing if doesn't exist? (maybe add explicit id select to narrow scope)

src/common/post.ts
```diff
@@ -329,3 +344,125 @@ export const insertCodeSnippetsFromUrl = async ({
...
+
+  const updatedPost: Partial<FreeformPost> = {
+    title: moderated.title,
+    titleHtml: moderated.titleHtml,
+  };
+
+  if (moderated.content) {
+    updatedPost.content = moderated.content;
+    updatedPost.contentHtml = moderated.contentHtml!;
```
rebelchris, line 371: Don't we do stuff with markdown or tags here? (can't recall so top of mind)

src/common/post.ts
```diff
@@ -329,3 +344,125 @@ export const insertCodeSnippetsFromUrl = async ({
...
+  }
+
+  if (moderated.image) {
+    updatedPost.image = moderated.image;
```
rebelchris, line 375: Same here if cloudinary upload etc

src/common/post.ts
```diff
@@ -329,3 +344,125 @@ export const insertCodeSnippetsFromUrl = async ({
...
+    image,
+    createdById,
+    sourceId,
+    titleHtml,
+    externalLink,
+    sharedPostId,
+  } = moderated;
+
+  if (moderated.type === PostType.Freeform) {
+    const id = await generateShortId();
```
rebelchris, line 420: Is this was we do now?
I think @ilasw mentioned it in another thread as well.

src/entity/posts/utils.ts
```diff
@@ -285,20 +305,24 @@ export const createExternalLink = async (
         visible: isVisible,
       },
     });
-    await createSharePost(
-      entityManager,
+    const post = await createSharePost({
```
rebelchris, line 308: Why do we need it returned now?

src/entity/posts/utils.ts
```diff
@@ -351,27 +386,29 @@ export const createSharePost = async (
...
+          content: createdPost.title || undefined,
+          type: VordrFilterType.Post,
+        },
+        { con, userId, req: ctx.req },
+      );
 
-    if (vordrStatus === true) {
-      createdPost.banned = true;
-      createdPost.showOnFeed = false;
+      if (vordrStatus) {
```
rebelchris, line 399: @omBratteng was there specific case you checked === true?

rebelchris (commented):
Overall minor comments so looking good.

I see we have a ton of `create post` kind of functions is there no way we can somehow combine them into seperated utility functions?
I find it very hard to follow why there's multiple things that all look similar.

CC: @capJavert for your opinion as well?

src/common/post.ts
```diff
@@ -329,3 +344,125 @@ export const insertCodeSnippetsFromUrl = async ({
...
+        url: cleanUrl,
+        sourceId,
+        authorId: createdById,
+        commentary: content,
+      },
+    });
+    return { ...moderated, postId: post.id };
+  }
+
+  logger.error(`unable to process moderated post: ${moderated.id}`);
```
capJavert, line 465: Let's use object style syntax as we use in other workers, it will give us better visibility

```suggestion
  logger.error({ moderated }, 'unable to process moderated post');
```

capJavert (approved):
Looks good, some sanity checks, but good to merge after!

src/common/post.ts
```diff
@@ -329,3 +344,125 @@ export const insertCodeSnippetsFromUrl = async ({
...
+  if (moderated.image) {
+    updatedPost.image = moderated.image;
+  }
+
+  await repo.update(postParam, updatedPost);
+
+  return moderated;
+};
+
+export const getExistingPost = async (
```
capJavert, line 383: @rebelchris I agree with duplicated functions, but I think it just comes to the fact we often add utility functions in different files without really looking if they are used or if there is something similar. 

Would say this is an example, its used in a single place and for sure if I need to check existing post I'll just write it myself again.

sshanzel:
> I see we have a ton of create post kind of functions is there no way we can somehow combine them into seperated utility functions?

Totally agree with that, although I tried to make them generic as much as I can, but we have three different processes for 3 different kind of submissions. For Freeform, we just create everything the user adds, for share, we have two kinds. One is sharing already existing post, the other accepts external link. All 3 of them have different requirements, for example, Share does not need `content`, then on external link we have a `title` but going to be used on the article post generated then generate another share post with the `content`.

src/common/post.ts
```diff
@@ -204,7 +219,7 @@ export const createFreeformPost = async (
     { con, userId: ctx.userId, req: ctx.req },
   );
 
-  if (vordrStatus === true) {
+  if (vordrStatus) {
```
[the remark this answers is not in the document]
sshanzel: The function returns seem to be consistent to be a boolean type. Feel free to let me know if I have to revert.

src/entity/posts/utils.ts
sshanzel: We now need to return it. This makes us reuse the same function. Previous usages should not matter whether this returns anything or not.

src/common/post.ts
sshanzel: This is for the actual post in the `post` table, for the moderation table, we generate it in DB level. If that is what you mean as I don't have reference to the thread.

src/common/post.ts
sshanzel: I'm not sure I understand. We have `id: moderated.postId` already right?

src/common/post.ts
sshanzel: Yes, and since from the post submission we generate it already. Hence, we just pass it here.

rebelchris: Ah ok, guess just very hard to follow from review which part actually does something 😢 
One day we need to clean this up a bit I feel.

src/common/post.ts
rebelchris: Ah maybe yes, this one we did now with shortId?

src/common/post.ts
sshanzel: Oh, didn't know you actually pull everything if you don't specify. Yeah, sure let's do that.

src/common/post.ts
sshanzel: Yeah, totally agree. Feels like we need some sort of process flow that will handle everything when it comes to creating post.

src/common/post.ts
sshanzel: I just saw from the other implementation where we create post we use this id. Wdyt, should we use uuid?

src/entity/posts/utils.ts
sshanzel: Because the function is now used on two parts, and the other one needs to utilize some prop from the created post. Specifically the id.

src/common/post.ts
```diff
@@ -195,6 +206,10 @@ export const createFreeformPost = async (
     },
   });
 
+  if (!ctx) {
+    return con.getRepository(FreeformPost).save(createdPost);
+  }
```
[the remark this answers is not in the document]
sshanzel: Yeah, just wanted to avoid huge chunks of indent, but we can do that too.

src/common/post.ts
sshanzel: Cool. Didn't wanna flood or logs with might be unnecessary props which can just be pulled.

src/common/post.ts
rebelchris: Yeah probably good idea, it's not blocking for now, we can log it for web guild meeting :) 
It's good to think about these systems to avoid re-building stuff.

src/common/post.ts
rebelchris: I personally don't mind more up to @capJavert I think

src/common/post.ts
rebelchris: Let's log it for backlog, don't want to keep this project blocked by it.

rebelchris: @sshanzel 

> Totally agree with that, although I tried to make them generic as much as I can, but we have three different processes for 3 different kind of submissions. For Freeform, we just create everything the user adds, for share, we have two kinds. One is sharing already existing post, the other accepts external link. All 3 of them have different requirements, for example, Share does not need content, then on external link we have a title but going to be used on the article post generated then generate another share post with the content.

This is a very good point, ideally this is just wrapped in 3 functions then:
- `createFreeformPost`
- `createShareExistingPost`
- `createShareExternalLinkPost`

And those rule all the things, but it's more tech debt clean up as this point.

src/common/post.ts
capJavert: This should basically never happen, so if when deploy it starts to happen we need to adjust the worker so it handles the cases which end up on this

ilasw (approved):
Nice, looks really good :)

src/common/post.ts
sshanzel: Makes sense, yeah. I'll update quickly.

src/common/post.ts
sshanzel: Yeah, that would be best. We actually apply that now here, hopefully, we can apply it on the legacy code too. 

Ticket created: https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11/backlog?selectedIssue=WT-2245

rebelchris (approved):

PR #2406, opened 2024-11-07
feat: add inline sql extension

capJavert:
- for vscode users

PR #2407, opened 2024-11-07
fix: add missing SQL file

omBratteng:
SQL did not get bundles 🙈
Created a new directory called `queries` and add that to image


rebelchris (approved):

src/config.ts
```diff
@@ -46,3 +46,5 @@ export const generateStorageKey = (
 export const FEED_SURVEY_INTERVAL = 30;
 
 export const WEBAPP_MAGIC_IMAGE_PREFIX = `/image-generator`;
+
+export const SQL_QUERIES_PATH = '/opt/app/queries';
```
capJavert, line 50: now it won't work locally? should we just add it inline and lint with https://github.com/dailydotdev/daily-api/pull/2406

omBratteng: It works inside the container. But I'm going to adjust this to be better later, just want to get it out and test it.

PR #2410, opened 2024-11-07
feat: post moderation mutation

ilasw:
What changed from RFC:
- added missing the reject message
- added `sourceId` because it makes easier to check posts (otherwise we have to detect sourceId from posts )


src/schema/posts.ts
```diff
@@ -2092,6 +2122,76 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      await ensureSourcePermissions(
+        ctx,
+        sourceId,
+        SourcePermissions.MemberRemove,
+      );
+
+      const pendingPosts = await ctx.con
+        .getRepository(SourcePostModeration)
+        .find({
+          where: {
```
rebelchris, line 2166: Probably add `select: id` here to avoid requesting the whole object.

rebelchris (commented):
Nice, some minor comments/questions.

src/schema/posts.ts
```diff
@@ -1123,6 +1127,32 @@ export const typeDefs = /* GraphQL */ `
...
+      sourceId: ID!
+      """
+      Rejection reason for the post
+      """
+      rejectionReason: String
+      """
+      Moderator message for the post
+      """
+      moderatorMessage: String
+    ): [SourcePostModeration]! @auth
```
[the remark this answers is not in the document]
ilasw: We can use it in moderation tab, on update we will update the related query ✅

src/schema/posts.ts
```diff
@@ -2092,6 +2122,76 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          SourcePostModerationStatus.Rejected,
+        ].includes(status)
+      ) {
+        throw new ValidationError('Invalid status provided');
+      }
+
+      await ensureSourcePermissions(
+        ctx,
+        sourceId,
+        SourcePermissions.MemberRemove,
```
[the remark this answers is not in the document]
AmarTrebinjac: Should be "ModeratePost" role that we also use when checking if the submission is create post or create post moderation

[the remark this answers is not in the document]
ilasw: Updated using `isPrivilegedMember` function ✔️

src/schema/posts.ts
ilasw: Updated ✔️

rebelchris (approved):

src/schema/posts.ts
```diff
@@ -277,6 +281,38 @@ export const typeDefs = /* GraphQL */ `
...
+    """
+    createdAt: DateTime!
+    """
+    Author of the post
+    """
+    createdBy: User
+    """
+    Author id
+    """
+    createdById: ID
```
sshanzel, line 307: We should not need this anymore since we already have the `createdBy` object.

src/schema/posts.ts
```diff
@@ -277,6 +281,38 @@ export const typeDefs = /* GraphQL */ `
...
+    """
+    Moderator of the post
+    """
+    moderatedBy: User
+    """
+    Moderator id
+    """
+    moderatedById: ID
```
sshanzel, line 315: Same goes here.

sshanzel (approved):
Looks amazing! Don't have anything blocking.

src/schema/posts.ts
```diff
@@ -2093,6 +2147,78 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+            id: In(uniquePostIds),
+            sourceId,
+            status: SourcePostModerationStatus.Pending,
+          },
+          select: ['id'],
+        });
+
+      if (pendingPosts.length !== uniquePostIds.length) {
+        throw new ValidationError('Some posts are not pending');
+      }
```
capJavert, line 2202: Don't want to add scope, but should we just ignore those that are not pending. It could break due to race conditions and it feels bad it fails the whole request because one of the post was not pending. For example it could have been another moderator approves one of the posts just before.

src/schema/posts.ts
```diff
@@ -2093,6 +2147,78 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          moderatedById,
+          ...(isRejectedWithReason && { rejectionReason, moderatorMessage }),
+        },
+      );
+
+      return await ctx.con.getRepository(SourcePostModeration).findBy({
+        id: In(uniquePostIds),
+        status,
+      });
+    },
```
capJavert, line 2221: We should use gql like we do in #2400

src/schema/posts.ts
ilasw: Updated ✅

src/schema/posts.ts
ilasw: Updated ✅

capJavert (approved):

PR #2412, opened 2024-11-07
feat: add search source endpoints

rebelchris:
Introduces 2 new endpoints needed for source selection in onboarding.

- searchSources => Used for when you use the search box, needed slight different result from the suggestions one we have, so decided to do custom one.
- sourceRecommendationByTags => Initial page load, we pass the tags you selected on previous screen and find highest matching sources. (Needed to do 2 queries to avoid messy graphORM logic we couldn't deduplicate, and since it's once off this is fine)

AS-678

### DIFF
--- a/src/schema/sources.ts
+++ b/src/schema/sources.ts
@@ -75,6 +75,7 @@ import {
 } from '../common/contentPreference';
 import { MIN_SEARCH_QUERY_LENGTH } from './tags';
 import { categorizedSquads } from '../../bin/categorizedSquads';
+import { getSearchLimit } from '../common/search';
 
 export interface GQLSourceCategory {
   id: string;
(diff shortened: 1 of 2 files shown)


src/schema/sources.ts
```diff
@@ -1495,6 +1527,69 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      return await graphorm.query<GQLSource>(
+        ctx,
+        info,
+        (builder) => {
+          builder.queryBuilder
+            .where(`name ILIKE :query OR handle ILIKE :query`, {
+              query: `%${query}%`,
+            })
+            .andWhere({ active: true, private: false })
+            .limit(limit);
```
capJavert, line 1549: also limit

capJavert (approved):

capJavert (approved):

PR #2414, opened 2024-11-07
infra: use image proxy

idoshamun:
Override the image urls to use the image proxy. This implementation is partial just to try it out.

Some aspects are not covered here, for example markdown images, boot, etc.


rebelchris (commented):
Preload wont work now, something to keep an eye out for

idoshamun:
@rebelchris once I finish everything I'll fix it as well. Good note

PR #2415, opened 2024-11-07
infra: use image proxy p2

idoshamun:
Add the missing bits:
* boot
* comments
* post content
* uploaded images
* top reader
* fallback image

### DIFF
--- a/src/common/cloudinary.ts
+++ b/src/common/cloudinary.ts
@@ -108,8 +108,11 @@ export const uploadPostFile = (
   preset: PostPreset,
 ) => uploadFile(name, preset, stream);
 
-export const mapCloudinaryUrl = (url?: string): string | undefined =>
-  url?.replace(
+export function mapCloudinaryUrl(url: string): string;
+export function mapCloudinaryUrl(url: undefined): undefined;
+export function mapCloudinaryUrl(url?: string): string | undefined {
+  return url?.replace(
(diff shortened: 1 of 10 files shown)


src/common/cloudinary.ts
```diff
@@ -108,8 +108,12 @@ export const uploadPostFile = (
   preset: PostPreset,
 ) => uploadFile(name, preset, stream);
 
-export const mapCloudinaryUrl = (url?: string): string | undefined =>
-  url?.replace(
+export function mapCloudinaryUrl(url: string): string;
+export function mapCloudinaryUrl(url: undefined): undefined;
+export function mapCloudinaryUrl(url?: string): string | undefined;
+export function mapCloudinaryUrl(url?: string): string | undefined {
```
rebelchris, line 114: Is this intentional/correct?
If so why is it needed?

rebelchris (changes requested):
One thing that seems off.

src/common/cloudinary.ts
rebelchris: Surely you can do the last one alone and early return if no value is given?
If you accept string|undefined?

idoshamun: But type wise I have to return string | undefined which violates ts strict. That it is how it was before I added the new use case

rebelchris: Ah I see, not the biggest fan of overloads in JS world, but guess it's ok if it works 😂

rebelchris (approved):
Looks ok

PR #2422, opened 2024-11-09
chore: bin script to update source post moderation status

sshanzel:
Update existing Squads that allows member to post to have moderation required.

MI-586 #done

### DIFF
--- a/bin/updateSourceModeration.ts
+++ b/bin/updateSourceModeration.ts
@@ -1,0 +1,32 @@
+import '../src/config';
+import createOrGetConnection from '../src/db';
+import { SquadSource } from '../src/entity';
+import { runInQueueStream } from './common';
+
+interface Result {
+  title: string;
+  id: string;
+}
+
+(async () => {
+  const con = await createOrGetConnection();
+  await con.transaction(async (manager) => {
+    console.log('initializing stream');
+
+    const builder = manager
+      .getRepository(SquadSource)
+      .createQueryBuilder('ss')
+      .select('ss.id', 'id')
+      .where('ss."memberPostingRank" = 0');
+
+    const stream = await builder.stream();
+
+    await runInQueueStream<Result>(stream, async ({ id }: Result) => {
+      await manager
+        .getRepository(SquadSource)
+        .update({ id }, { moderationRequired: true });
+    });
+  });
+
+  process.exit();
+})();


bin/updateSourceModeration.ts
```diff
@@ -0,0 +1,32 @@
...
+(async () => {
+  const con = await createOrGetConnection();
+  await con.transaction(async (manager) => {
+    console.log('initializing stream');
+
+    const builder = manager
+      .getRepository(SquadSource)
+      .createQueryBuilder('ss')
+      .select('ss.id', 'id')
+      .where('ss."memberPostingRank" = 0');
```
rebelchris, line 20: Do we even need this though?
If the posting rank is 0 nobody but admins/mods can post anyway right?

Did it prove to be an issue somewhere?

bin/updateSourceModeration.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const stream = await builder.stream();
+
+    await runInQueueStream<Result>(stream, async ({ id }: Result) => {
+      await manager
+        .getRepository(SquadSource)
+        .update({ id }, { moderationRequired: true });
+    });
```
capJavert, line 28: @sshanzel can't this be done in a single query?

```sql
update source s set "moderationRequired" = true 
where s.type = 'squad' and s."memberPostingRank" = 0
```

capJavert: if yes we can even just run it as once off on the DB, no need to create a script

bin/updateSourceModeration.ts
sshanzel: > Did it prove to be an issue somewhere?

Just saw this as part of the RFC. Maybe @ilasw can enlighten us with more details on why we need to update existing squads that allows posting for everyone?

bin/updateSourceModeration.ts
sshanzel: It can, like most of our updates. If I remember correctly, we run these scripts to manage the amount of updates per batch.

If we don't have to worry the amount this time, then yeah, we can just execute this as a regular sql script.

capJavert: you can check how many squads it takes, also if you want to update per batch usually we use limit/offset with `insert from select` or date based.

@sshanzel its only 8k rows `select count(1) from source s where s.type = 'squad' and s."memberPostingRank" = 0`

so you can do it as a one time query

capJavert: You can check how many squads it takes, also if you want to update per batch usually we use limit/offset.

@sshanzel its the small amount of rows, see `select count(1) from source s where s.type = 'squad' and s."memberPostingRank" = 0`

so you can do it as a one time query 🚀

sshanzel: We've got 8k rows. Wdyt? Do we run it in batches?

Screenshot 2024-11-11 at 9 42 57 PM

sshanzel:
Turns out, we shouldn't need this 🥳

sshanzel:
Seems like we still need it 😅

capJavert (approved):

sshanzel:
This should be run only once we have shipped everything in prod 👀

rebelchris (approved):

PR #2423, opened 2024-11-11
feat: add date columns for feed_tag

capJavert:
- add createdAt and updatedAt columns
- added updatedAt trigger so updatedAt is correctly set when updating filters (included only blocked column change since that is the only one that can really change)
- added index for filtering over updatedAt in migration script

Since migration is lot more heavy then I thought this will allow us to do a migration in more incremental way where we can set a date when we start it and then incrementally only update rows which changed after the set date.

### DIFF
--- a/src/entity/FeedTag.ts
+++ b/src/entity/FeedTag.ts
@@ -1,4 +1,12 @@
-import { Column, Entity, Index, ManyToOne, PrimaryColumn } from 'typeorm';
+import {
+  Column,
+  CreateDateColumn,
+  Entity,
+  Index,
+  ManyToOne,
+  PrimaryColumn,
+  UpdateDateColumn,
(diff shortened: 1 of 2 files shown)


idoshamun (approved):

idoshamun (changes requested):
oh we do need index for updatedAt i guess for the migration?

capJavert:
@idoshamun `insert into select * where updatedAt > X` will use it, I can remove and try without and then add if its slow?

idoshamun:
@capJavert wdym remove? i don't see any index currently, we need one for `updatedAt`

capJavert:
@idoshamun oh I did not push the commit 😆

idoshamun (approved):

rebelchris (approved):

PR #2426, opened 2024-11-11
refactor: source cdc to update existing post for moderation

sshanzel:
When the Squad was accepting posts but require moderation made an update to not require them anymore, we should have everything pending to be approved.


src/workers/cdc/primary.ts
```diff
@@ -710,6 +711,22 @@ const onSourceChange = async (
...
+      squadBefore.moderationRequired !== squadAfter.moderationRequired &&
+      !squadAfter.moderationRequired
+    ) {
+      await con.getRepository(SourcePostModeration).update(
+        {
+          sourceId: squadAfter.id,
+          status: SourcePostModerationStatus.Pending,
+        },
+        { status: SourcePostModerationStatus.Approved },
+      );
```
capJavert, line 728: Should we add a specific moderationReason so that we know when posts get approved this way, just so we can understand how much it happens?

capJavert (approved):

src/workers/cdc/primary.ts
sshanzel: Not sure where we would use it but maybe we can check with product or @idoshamun

capJavert: I think we should have it just for our sake, so we can have insight on how the post was approved.

capJavert: also because this will not have `moderatedById` so we might think its some kind of issue on our side

sshanzel: Yeah, do you have any suggestions on what should be the key to store in the `reason` column?

capJavert: whatever, some constant value like `moderationRequiredChanged` should do

sshanzel: Cool. Let's do that.

sshanzel: We should maybe change the column name since it is `rejectionReason` which is not applicable to this case anymore.

capJavert: ah its rejection reason, thought it was moderationReason 🤦 ... let's ignore it then it does not fit the schema

sshanzel: We can add it to the `moderatorMessage` if you like.

PR #2427, opened 2024-11-12
feat: email template for moderation notifications

sshanzel:


src/workers/newNotificationV2Mail.ts
```diff
@@ -89,7 +92,7 @@ export const notificationToTemplateId: Record<NotificationType, string> = {
   user_given_top_reader: CioTransactionalMessageTemplateId.UserGivenTopReader,
 };
 
-type TemplateData = Record<string, string | number>;
+type TemplateData = Record<string, unknown>;
```
capJavert, line 95: @sshanzel we should avoid using `!` at almost all costs, I am aware we have some in the code (because I could not convert everything without breaking changes when doing strict-ts.

I updated in 408df5d306e627d985aa364a9995ef6eb365be45 so it does not use `!` for your code changes.

Also the `unknown` here should help with inferring the types from template functions.

capJavert (approved):

src/workers/newNotificationV2Mail.ts
sshanzel: Yeah, I hated using it, but many types were breaking in this part.

PR #2428, opened 2024-11-12
fix: external links submission

sshanzel:
There are two things fixed here, separated by each commit:
- In our CDC pipeline, there can be a case where two of the same links were submitted. Is it not possible to process them at the same time when approved, in this case, the external link would have been created on the first instance, and on the second instance, it would produce an error due to having the same link.
- Second fix was on the submission of post for moderation. The values passed through the mutation would be present depending on the case. We have title, content, and commentary but the post entity only has a title and content.
  - Freeform: must have a title and content. Empty `titleHtml` but `contentHtml` should be defined.
  - Share (from internal id): must have `title` and `titleHtml`. Empty `content` and `contentHtml`.
  - Share (from external link): must have `title` that is going to be used as the title of the then-created Article post, which won't have `titleHtml`. The `content` and `contentHtml` should be present and be used as the `title` of the `Share` post created.
- Also made the tests more explicit.

### DIFF
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -487,6 +487,23 @@ export const processApprovedModeratedPost = async (
 
   if (externalLink) {
     const cleanUrl = standardizeURL(externalLink);
+    const existingPost = await getExistingPost(con, cleanUrl);
+
+    if (existingPost) {
+      const post = await createSharePost({
+        con,
+        args: {
+          sourceId,
+          commentary: content,
+          authorId: createdById,
+          postId: existingPost.id,
+          visible: existingPost.visible,
+        },
+      });
+
+      return { ...moderated, postId: post.id };
+    }
+
     const post = await createExternalLink({
       con,
       args: {
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -1705,6 +1705,10 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       ctx: AuthContext,
       info,
     ): Promise<GQLSourcePostModeration> => {
+      if (![PostType.Share, PostType.Freeform].includes(type)) {
+        throw new ValidationError('Invalid post type!');
+      }
+
       const { con, userId } = ctx;
 
       const sourceMember = await con.getRepository(SourceMember).findOne({
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -1673,8 +1673,6 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       const mentions = await getMentions(con, content, userId, sourceId);
 
       const pendingPost: CreateSourcePostModeration = {
-        title,
-        content,
         sourceId,
         type,
         sharedPostId,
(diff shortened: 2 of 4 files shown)


src/schema/posts.ts
```diff
@@ -1673,21 +1673,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       }
-      if (content) {
-        pendingPost.contentHtml = markdown.render(content, { mentions });
+
+      if (content && type === PostType.Freeform) {
+        pendingPost.title = title;
+        pendingPost.content = content;
+        pendingPost.contentHtml = markdown
+          .render(content, { mentions })
+          ?.trim();
```
sshanzel, line 1760: For some reason, there is an extra new line.

Screenshot 2024-11-12 at 12 47 02 PM

capJavert: Trim removes whitespaces

src/schema/posts.ts
```diff
@@ -1729,21 +1729,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+        if (isExternal) {
+          pendingPost.title = title;
+          pendingPost.content = commentary;
+          pendingPost.contentHtml = commentaryHtml;
+        } else {
+          pendingPost.title = commentary;
+          pendingPost.titleHtml = commentaryHtml;
+        }
       }
```
capJavert, line 1753: I find it hard to follow with all the if else on what is added and I am sure we can easily forget a field or two in all combinations. What happens if someone sends no commentary and content with type article? Will it go through?

src/common/post.ts
```diff
@@ -487,6 +487,23 @@ export const processApprovedModeratedPost = async (
...
+          sourceId,
+          commentary: content,
+          authorId: createdById,
+          postId: existingPost.id,
+          visible: existingPost.visible,
+        },
+      });
+
+      return { ...moderated, postId: post.id };
+    }
```
capJavert, line 505: So this is just in case post already exists and now instead of creating it we add a shared post for approval?

src/schema/posts.ts
sshanzel: Yes.

src/schema/posts.ts
sshanzel: I missed during the review that there were three parameters where there should have been only two. Sending a type article is not possible, but on that case, if you send an empty commentary with `share` type for example, a type we accept, it should proceed as we won't care the actual content.

capJavert: > Sending a type article is not possible

You mean from the frontend or in general its prevented somehow on API side?

sshanzel: In general, through this process. Type article only comes from post scraper and external links, but then that is created within the system not by sending a `article` type.

capJavert: yes but someone can send it directly to API and what happens then?

capJavert: Rather limit what types can be sent here.

sshanzel: Added the checks now. It should not allow types unless defined 🚀

capJavert (approved):

PR #2429, opened 2024-11-12
feat: Create moderation for existing post edit

AmarTrebinjac:
Flow for creating a moderation entry for when user wants to edit their existing post

### DIFF
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -262,11 +262,7 @@ export const createSourcePostModeration = async (
       .getRepository(Post)
       .findOneByOrFail({ id: args.postId });
 
-    const isUnauthorizedEdit =
-      (post.type === PostType.Freeform && args.createdById !== post.authorId) ||
-      (post.type !== PostType.Freeform && args.createdById !== post.scoutId);
-
-    if (isUnauthorizedEdit) {
+    if (args.createdById !== post.authorId) {
       throw new ForbiddenError('Cannot edit posts created by other users');
     }
   }
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -13,7 +13,7 @@ import {
   User,
   WelcomePost,
 } from '../entity';
-import { ValidationError } from 'apollo-server-errors';
+import { ForbiddenError, ValidationError } from 'apollo-server-errors';
 import { isValidHttpUrl, standardizeURL } from './links';
 import { markdown } from './markdown';
 import { generateShortId } from '../ids';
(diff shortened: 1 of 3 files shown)


sshanzel (approved):
Looks amazing! Great work on the tests!

src/common/post.ts
```diff
@@ -250,17 +250,32 @@ export type CreateSourcePostModeration = Omit<
...
 ) => {
-  const newPost = con.getRepository(SourcePostModeration).create({
+  if (args.postId) {
+    const post = await con
+      .getRepository(Post)
+      .findOneByOrFail({ id: args.postId });
+
+    const isUnauthorizedEdit =
+      (post.type === PostType.Freeform && args.createdById !== post.authorId) ||
+      (post.type !== PostType.Freeform && args.createdById !== post.scoutId);
```
capJavert, line 267: Why do we need `post.type` check here?

src/common/post.ts
```diff
@@ -250,17 +250,32 @@ export type CreateSourcePostModeration = Omit<
...
+      throw new ForbiddenError('Cannot edit posts created by other users');
+    }
+  }
+
+  const newModerationEntry = con.getRepository(SourcePostModeration).create({
     ...args,
     status: SourcePostModerationStatus.Pending,
   });
-  return await con.getRepository(SourcePostModeration).save(newPost);
+  return await con.getRepository(SourcePostModeration).save(newModerationEntry);
```
capJavert, line 278: It will create a new row in SourcePostModeration right?

capJavert (approved):
Just few questions, but overall good

src/common/post.ts
AmarTrebinjac: @capJavert This was the reason I asked you about yesterday who is the post creator. I understood you such as that when the post is scraped or written on our site, the "creator" is the author id, and for share/link it is scout id. 

Did I understand that correctly?

src/common/post.ts
AmarTrebinjac: Correct!

src/common/post.ts
sshanzel: We should just check whether the user is the `authorId`. We only allow editing on `freeform` anyway. Also, the scout part is for community picks. Even `share` post can have the right `authorId`.

AmarTrebinjac: Talked with Lee in DMs and he clarified that we do not use scoutId in squad, and both share and freeform use authorId for the "creator".

WIll therefor only check if createdById === authorId

PR #2430, opened 2024-11-12
feat: paddle

rebelchris:
Add paddle integration.

Introduces webhook functionality to listen to paddle events.
We only care about two:
- subscription created (user got either monthly or yearly subscription)
- subscription cancelled (user cancelled a subscription)

We store these as flags, in case we want to eventually know which of the 2 they have.
We then use `isPlus` to determine if one of those is true.

TODO:
Test cases, but we should be able to review already high level.


src/entity/user/User.ts
```diff
@@ -176,6 +180,10 @@ export class User {
   })
   referral?: Promise<User>;
 
+  @Column({ type: 'jsonb', default: {} })
+  @Index('IDX_user_subscription_flags', { synchronize: false })
```
rebelchris, line 184:
```suggestion
```

src/routes/boot.ts
```diff
@@ -469,12 +470,17 @@ const loggedInBoot = async ({
...
         ]),
         providers: [null],
         roles,
         permalink: `${process.env.COMMENTS_PREFIX}/${user.username || user.id}`,
         canSubmitArticle: user.reputation >= submitArticleThreshold,
         isTeamMember,
+        isPlus:
+          user.subscriptionFlags?.monthly ||
+          user.subscriptionFlags?.yearly ||
+          false,
```
capJavert, line 483:
```suggestion
          !!user.subscriptionFlags?.monthly ||
          !!user.subscriptionFlags?.yearly ||
          false,
```

src/routes/webhooks/paddle.ts
```diff
@@ -0,0 +1,105 @@
...
+    return false;
+  }
+  const subscriptionType = data.data?.items.reduce((acc, item) => {
+    if (planTypes[item.price?.id]) {
+      acc = planTypes[item.price?.id];
+    }
+    return acc;
+  }, null);
+  if (!subscriptionType) {
+    logger.error('Subscription type missing in payload');
```
capJavert, line 46:
```suggestion
    logger.error({ data }, 'Subscription type missing in payload');
```

log the data as well for debug? not sure if any info is sensitive?

src/routes/webhooks/paddle.ts
```diff
@@ -0,0 +1,105 @@
...
+                  state: true,
+                });
+                break;
+              case EventName.SubscriptionCanceled:
+                await updateUserSubscription({
+                  data: eventData,
+                  state: false,
+                });
+              default:
+                logger.info(eventData.eventType);
```
capJavert, line 93:
```suggestion
                logger.info({ eventType: eventData.eventType}, 'paddle unknown event type');
```

src/routes/webhooks/paddle.ts
```diff
@@ -0,0 +1,105 @@
...
+  data: SubscriptionCreatedEvent | SubscriptionCanceledEvent | undefined;
+  state: boolean;
+}) => {
+  if (!data) {
+    return;
+  }
+  const con = await createOrGetConnection();
+  const userId = data.data?.customData?.user_id;
+  if (!userId) {
+    logger.error('User ID missing in payload');
```
capJavert, line 36: can we prefix everything with `paddle` or add `{ type: 'paddle' }` so we can easily search through logs?

src/routes/webhooks/paddle.ts
```diff
@@ -0,0 +1,105 @@
...
+    return false;
+  }
+  await con.getRepository(User).update(
+    {
+      id: data.data?.customData?.user_id,
+    },
+    {
+      subscriptionFlags: updateSubscriptionFlags({
+        [subscriptionType]: state,
+      }),
```
capJavert, line 56: Logging to CIO later?

capJavert (approved):
Nothing blocking, few comments to make it a bit easier to track since its a new system.

src/routes/webhooks/paddle.ts
omBratteng: Yeah, think it's a good idea.
I added prefixes to the top reader cron so it's easier to locate logs

omBratteng: (kinda miss the logging in python)

omBratteng:
As general cleanup, maybe we should cleanup references to `premium`? 🙈 
https://github.com/search?q=repo%3Adailydotdev%2Fdaily-api%20premium&type=code

src/graphorm/index.ts
```diff
@@ -17,6 +17,7 @@ import {
   FeatureType,
   SettingsFlagsPublic,
   UserStats,
+  UserSubscriptionFlags,
```
omBratteng, line 20: Missing isPlus on the User so we can show it on profile pages

omBratteng (changes requested):

.infra/Pulumi.adhoc.yaml
```diff
@@ -69,6 +69,8 @@ config:
     otelEnabled: false
     otelExporterOtlpEndpoint: http://otel-collector.local.svc.cluster.local:4318/v1/traces
     otelTracesSampler: always_on
+    paddleApiKey: topsecret
+    paddleWebhookSecret: topsecret
```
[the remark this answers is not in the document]
rebelchris: done

src/routes/webhooks/paddle.ts
rebelchris: Good id, will add prefix

src/routes/webhooks/paddle.ts
rebelchris: Yeah still coming, need to check how I'm going to do it.

src/graphorm/index.ts
omBratteng: Added it

.infra/Pulumi.adhoc.yaml
```diff
@@ -69,6 +69,8 @@ config:
     otelEnabled: false
     otelExporterOtlpEndpoint: http://otel-collector.local.svc.cluster.local:4318/v1/traces
     otelTracesSampler: always_on
+    paddleApiKey: topsecret
```
idoshamun, line 72: I don't see these secrets in prod

src/entity/user/User.ts
```diff
@@ -22,6 +22,12 @@ export type UserFlags = Partial<{
   trustScore: number;
 }>;
 
+export type UserSubscriptionFlags = Partial<{
```
idoshamun, line 25: Since we need to redirect the user to paddle management page, shouldn't we store an id or something like that?

.infra/Pulumi.adhoc.yaml
rebelchris: Yet to add indeed :)

src/entity/user/User.ts
```diff
@@ -22,6 +22,12 @@ export type UserFlags = Partial<{
...
+  monthly: boolean;
+  yearly: boolean;
```
idoshamun, line 27: Instead of two booleans should we consider a string?

src/entity/user/User.ts
rebelchris: If we need that yes, haven't got to the manage link yet (just asked for that on slack)
I mean we can store what we want, but wanted to keep the flags as lean as possible.
Do you think we should just store everything?

src/entity/user/User.ts
rebelchris: Why the string? what does it contain? the subscription_id?

src/routes/boot.ts
```diff
@@ -469,12 +470,17 @@ const loggedInBoot = async ({
...
         ]),
         providers: [null],
         roles,
         permalink: `${process.env.COMMENTS_PREFIX}/${user.username || user.id}`,
         canSubmitArticle: user.reputation >= submitArticleThreshold,
         isTeamMember,
+        isPlus:
+          !!user.subscriptionFlags?.monthly ||
+          !!user.subscriptionFlags?.yearly ||
+          false,
```
idoshamun, line 483: I see this logic in different places, let's encapsulate please

src/routes/webhooks/paddle.ts
```diff
@@ -0,0 +1,112 @@
...
+import { User } from '../../entity';
+import { logger } from '../../logger';
+
+const paddleInstance = new Paddle(process.env.PADDLE_API_KEY, {
+  environment: Environment.sandbox,
+});
+
+const planTypes = {
+  pri_01jcdp5ef4yhv00p43hr2knrdg: 'monthly',
+  pri_01jcdn6enr5ap3ekkddc6fv6tq: 'yearly',
```
idoshamun, line 20: Should we consider the pricing ids as env var to support local env?

src/routes/webhooks/paddle.ts
```diff
@@ -0,0 +1,112 @@
...
+  SubscriptionCanceledEvent,
+  SubscriptionCreatedEvent,
+} from '@paddle/paddle-node-sdk';
+import createOrGetConnection from '../../db';
+import { updateSubscriptionFlags } from '../../common';
+import { User } from '../../entity';
+import { logger } from '../../logger';
+
+const paddleInstance = new Paddle(process.env.PADDLE_API_KEY, {
+  environment: Environment.sandbox,
```
idoshamun, line 15: Just making sure we want sandbox environment

idoshamun (changes requested):
Please see my comments

src/entity/user/User.ts
idoshamun: monthly and yearly as two separate variables is confusing. it's basically a billing cycle. so we can have a variable named cycle as enum

src/entity/user/User.ts
idoshamun: not everything, but since we know we need to send them to the management link. I'd make sure we have everything we need to do that. I'm assuming it would need some sort of subscription id

src/routes/webhooks/paddle.ts
rebelchris: Nope, I'll make this env based as well :)

src/entity/user/User.ts
rebelchris: It's hard for me to determine what we need, just add as I see a need...
I'll add this flag as well, but maybe during my loom Sam can let us know if we miss important info.

We could also consider storing the whole object somewhere else just for reading ops later?

idoshamun: I'm afraid the object may contain personal info that we don't need. let's just add subscription id for now because i think it will be good for debugging and analytics. the rest let's leave it as is

idoshamun (approved):

src/paddle.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const isPlusMember = (cycle: SubscriptionCycles | undefined): boolean =>
+  !!cycle?.length || false;
+
+export const planTypes = remoteConfig.vars.pricingIds || {
+  pri_01jcdp5ef4yhv00p43hr2knrdg: SubscriptionCycles.Monthly,
+  pri_01jcdn6enr5ap3ekkddc6fv6tq: SubscriptionCycles.Yearly,
+  pri_01gsz8x8sawmvhz1pv30nge1ke: SubscriptionCycles.Yearly,
+};
```
rebelchris, line 15:
```suggestion
export const planTypes = remoteConfig.vars.pricingIds || {};
```

PR #2432, opened 2024-11-12
feat: added edit/delete endpoint and tests

ilasw:
- Admin/Mod/Author can delete SourcePostModeration
- Only Author can edit own _non-approved_ SourcePostModeration

### DIFF
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -10,8 +10,6 @@ import {
   Post,
   PostOrigin,
   PostType,
-  SourceMember,
-  SourceType,
   SquadSource,
   User,
   validateCommentary,
@@ -38,7 +36,6 @@ import {
   SourcePostModerationStatus,
 } from '../entity/SourcePostModeration';
 import { mapCloudinaryUrl, uploadPostFile, UploadPreset } from './cloudinary';
-import { canAccessSource, SourcePermissions } from '../schema/sources';
 import { getMentions } from '../schema/comments';
 
 export const defaultImage = {
@@ -559,26 +556,6 @@ export const validateSourcePostModeration = async (
   }
 
   const { con, userId } = ctx;
-
-  const sourceMember = await con.getRepository(SourceMember).findOne({
-    where: { sourceId: sourceId, userId: userId },
-    relations: ['source'],
-  });
-
-  const source = await sourceMember?.source;
-
-  if (!source || source.type !== SourceType.Squad) {
-    throw new ForbiddenError('Access denied!');
-  }
-
-  await canAccessSource(
-    ctx,
-    source,
-    sourceMember,
-    SourcePermissions.Post,
-    sourceId,
-  );
-
   const pendingPost: CreateSourcePostModeration = {
     title: validateCommentary(title),
     postId,
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -543,6 +543,7 @@ export const processApprovedModeratedPost = async (
 export const validateSourcePostModeration = async (
   ctx: AuthContext,
   {
+    postId,
     title,
     content,
     sourceId,
@@ -580,6 +581,7 @@ export const validateSourcePostModeration = async (
 
   const pendingPost: CreateSourcePostModeration = {
     title: validateCommentary(title),
+    postId,
     sourceId,
     type,
     sharedPostId,
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -6,14 +6,18 @@ import {
   createSharePost,
   ExternalLinkPreview,
   FreeformPost,
+  generateTitleHtml,
   Post,
   PostOrigin,
   PostType,
+  SourceMember,
+  SourceType,
   SquadSource,
   User,
+  validateCommentary,
   WelcomePost,
 } from '../entity';
-import { ValidationError } from 'apollo-server-errors';
+import { ForbiddenError, ValidationError } from 'apollo-server-errors';
 import { isValidHttpUrl, standardizeURL } from './links';
 import { markdown } from './markdown';
 import { generateShortId } from '../ids';
@@ -33,7 +37,9 @@ import {
   SourcePostModeration,
   SourcePostModerationStatus,
 } from '../entity/SourcePostModeration';
-import { mapCloudinaryUrl } from './cloudinary';
+import { mapCloudinaryUrl, uploadPostFile, UploadPreset } from './cloudinary';
+import { canAccessSource, SourcePermissions } from '../schema/sources';
+import { getMentions } from '../schema/comments';
 
 export const defaultImage = {
   urls:
@@ -249,7 +255,7 @@ export type CreateSourcePostModeration = Omit<
     'titleHtml' | 'content' | 'type' | 'sharedPostId' | 'createdById'
   > & {
     contentHtml?: string;
-    externalLink?: string;
+    externalLink?: string | null;
   };
 
 export const createSourcePostModeration = async (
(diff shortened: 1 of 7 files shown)


src/schema/posts.ts
```diff
@@ -2353,6 +2409,70 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      if (!isModerator && !isAuthor) {
+        throw new ForbiddenError('Access denied!');
+      }
+
+      await ctx.con.getRepository(SourcePostModeration).delete({ id: postId });
+
+      return { _: true };
+    },
+    editSourcePostModeration: async (
```
sshanzel, line 2434: This should go through the same process of creating. Maybe we can create a generic function that will be used by both mutations.

sshanzel: I can help with this one if you are working on something.

ilasw: Yes thanks, would appreciate if you are free atm ✔️

capJavert: Agree with @sshanzel

sshanzel (approved):
Tests for the creating source post moderation passed without any update (except removal of `commentary`. In that case, we should be good to go 🚀

src/common/post.ts
```diff
@@ -534,3 +539,87 @@ export const processApprovedModeratedPost = async (
...
+    if (!!externalLink) {
+      const cleanContent = validateCommentary(content);
+
+      if (cleanContent) {
+        pendingPost.content = cleanContent;
+        pendingPost.contentHtml = generateTitleHtml(cleanContent, mentions);
+      }
+    } else if (pendingPost.title) {
+      pendingPost.titleHtml = generateTitleHtml(pendingPost.title, mentions);
+    }
```
capJavert, line 604: else only?

src/schema/posts.ts
```diff
@@ -2390,6 +2363,69 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        .getRepository(SourcePostModeration)
+        .findOneByOrFail({ id });
+
+      const isAuthor = moderation.createdById === ctx.userId;
+      const isApproved =
+        moderation.status === SourcePostModerationStatus.Approved;
+
+      if (!isAuthor || isApproved) {
+        throw new ForbiddenError('Access denied!');
+      }
```
capJavert, line 2405: Feels like `isApproved` should be ConflictError and separate?

src/schema/posts.ts
```diff
@@ -2390,6 +2363,69 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      { postId }: { postId: string },
+      ctx: AuthContext,
+    ): Promise<GQLEmptyResponse> => {
+      const moderation = await ctx.con
+        .getRepository(SourcePostModeration)
+        .findOneByOrFail({ id: postId });
+
+      await ensureSourcePermissions(ctx, moderation.sourceId);
+
+      const isModerator = await isPrivilegedMember(ctx, moderation.sourceId);
```
capJavert, line 2377: This reads from Source and SourceMember two times in a row, since you introduced `isPrivilegedMember` all over this maybe just use that one?

src/schema/posts.ts
```diff
@@ -2390,6 +2363,69 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      post: CreateSourcePostModerationArgs & { id: string },
+      ctx: AuthContext,
+      info,
+    ): Promise<SourcePostModeration> => {
+      const { id } = post;
+      const moderation = await ctx.con
+        .getRepository(SourcePostModeration)
+        .findOneByOrFail({ id });
+
+      const isAuthor = moderation.createdById === ctx.userId;
```
capJavert, line 2399: Should we also check source permission access?

src/common/post.ts
```diff
@@ -534,3 +539,87 @@ export const processApprovedModeratedPost = async (
...
+  }
+
+  const { con, userId } = ctx;
+
+  const sourceMember = await con.getRepository(SourceMember).findOne({
+    where: { sourceId: sourceId, userId: userId },
+    relations: ['source'],
+  });
+
+  const source = await sourceMember?.source;
```
capJavert, line 568: Can this be passed from outside, `validateSourcePostModeration` could be easily reused and then we will just be querying the DB all over the place. Let's keep the permissions/source check directly in mutations/queries

capJavert (commented):
Nothing blocking but think we query some data one too many times, if you can check my comments.

src/schema/posts.ts
```diff
@@ -2390,6 +2363,69 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+    editSourcePostModeration: async (
+      _,
+      post: CreateSourcePostModerationArgs & { id: string },
+      ctx: AuthContext,
+      info,
+    ): Promise<SourcePostModeration> => {
+      const { id } = post;
+      const moderation = await ctx.con
+        .getRepository(SourcePostModeration)
+        .findOneByOrFail({ id });
```
rebelchris, line 2397: We can narrow this with a select statement :)

src/schema/posts.ts
```diff
@@ -2390,6 +2363,69 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         ),
       }));
     },
+    deleteSourcePostModeration: async (
+      _,
+      { postId }: { postId: string },
+      ctx: AuthContext,
+    ): Promise<GQLEmptyResponse> => {
+      const moderation = await ctx.con
```
rebelchris, line 2371: Here we can also add select statement

src/common/post.ts
sshanzel: Was thinking about that, but we wouldn't want to process anything if the title is empty.

src/schema/posts.ts
sshanzel: We can do validation error too since the request is not valid anymore.

src/schema/posts.ts
sshanzel: Maybe we shouldn't have to check the moderator since they can reject the request. Let's just make author-only access to simplify.

src/common/post.ts
sshanzel: I've removed some things that should be automatic when we run `ensureSourcePermission`.

src/schema/posts.ts
sshanzel: True. Let's minimize the queried props.

rebelchris (approved):
Nothing more from my side besides Ante's comments

sshanzel:
I introduced a `PostRequest` permission; this is to make it easier to manage whether the user should have the right access for the action and to streamline the validation of permission through the function we use everywhere.

src/schema/sources.ts
```diff
@@ -992,24 +994,27 @@ export const isPrivilegedMember = async (
...
-  if (
-    memberRank >= sourceRoleRank.moderator &&
-    memberRank >= squad.memberPostingRank
-  )
-    return true;
+  if (squad.moderationRequired) {
+    if (memberRank === sourceRoleRank.member) {
+      return permission === SourcePermissions.PostRequest;
+    }
+  }
```
capJavert, line 1015: Is this correct, why `===`?

src/schema/posts.ts
capJavert: also check source permission first before querying the moderation as done in edit?

capJavert (approved):

src/schema/sources.ts
capJavert: asking because condition looks completely different then what it was

AmarTrebinjac: My idea with this condition was to check if user is a moderator or higher (admin), and let them bypass the moderation

src/schema/posts.ts
sshanzel: Yeah, makes sense. Let's do everything after that check.

sshanzel: Ah, for this case, we can't since we need some values from the moderation object.

PR #2434, opened 2024-11-12
feat: read from content preference source

capJavert:
- [x] update tests


rebelchris (approved):

idoshamun (approved):
Once we're done, we can consider simplifying the query, now that everything is in one table

PR #2436, opened 2024-11-12
feat: include feedId to content_preference pk

capJavert:
We need feedId as well as PK, [thread](https://dailydotdev.slack.com/archives/C07ATUPBPJ9/p1731420025841499)

- updated PK and column
- removed `feed*` types (not needed anymore since feedId will make them unique together with their original type)
- will update prod data before merging so constraints can be created

### DIFF
--- a/__tests__/contentPreference.ts
+++ b/__tests__/contentPreference.ts
@@ -103,11 +103,21 @@ describe('query userFollowers', () => {
       }),
     );
 
+    await saveFixtures(
+      con,
+      Feed,
+      usersFixture.map((item) => ({
+        id: `${item.id}-ufq`,
+        userId: `${item.id}-ufq`,
+      })),
+    );
+
     const now = new Date();
 
     await con.getRepository(ContentPreferenceUser).save([
       {
         userId: '2-ufq',
+        feedId: '2-ufq',
         referenceId: '1-ufq',
         referenceUserId: '1-ufq',
         status: ContentPreferenceStatus.Follow,
@@ -115,6 +125,7 @@ describe('query userFollowers', () => {
       },
       {
         userId: '3-ufq',
+        feedId: '3-ufq',
         referenceId: '1-ufq',
         referenceUserId: '1-ufq',
         status: ContentPreferenceStatus.Subscribed,
@@ -122,6 +133,7 @@ describe('query userFollowers', () => {
       },
       {
         userId: '1-ufq',
+        feedId: '1-ufq',
         referenceId: '2-ufq',
         referenceUserId: '2-ufq',
         status: ContentPreferenceStatus.Follow,
@@ -129,6 +141,7 @@ describe('query userFollowers', () => {
       },
       {
         userId: '4-ufq',
+        feedId: '4-ufq',
         referenceId: '1-ufq',
         referenceUserId: '1-ufq',
         status: ContentPreferenceStatus.Follow,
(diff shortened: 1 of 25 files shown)


src/migration/1731424255967-ContentPreferenceReplicaFull.ts
```diff
@@ -0,0 +1,19 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "content_preference" REPLICA IDENTITY FULL`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "content_preference" REPLICA IDENTITY DEFAULT`,
+    );
```
capJavert, line 17: Needed for clickhouse sync

src/common/contentPreference.ts
```diff
@@ -114,6 +112,7 @@ const followUser: FollowEntity = async ({ ctx, id, status }) => {
 
     const contentPreference = repository.create({
       userId: ctx.userId,
+      feedId: ctx.userId,
```
capJavert, line 115: by default feedId is userId, for custom feeds its gonna be actual id of the feed

__tests__/contentPreference.ts
```diff
@@ -103,32 +103,45 @@ describe('query userFollowers', () => {
...
+        userId: `${item.id}-ufq`,
+      })),
+    );
+
     const now = new Date();
 
     await con.getRepository(ContentPreferenceUser).save([
       {
         userId: '2-ufq',
+        feedId: '2-ufq',
```
rebelchris, line 120: How did this test not fail before 😢

rebelchris: But maybe we should add 1 user - 3 feeds with same tags for safety

src/common/contentPreference.ts
rebelchris: Yeah internally userid is just yet another feed, i like it 👍

src/common/contentPreference.ts
```diff
@@ -247,7 +246,7 @@ const unfollowSource: UnFollowEntity = async ({ ctx, id }) => {
 
     await repository.delete({
       userId: ctx.userId,
-      sourceId: id,
```
rebelchris, line 250: Dont you need these still?

rebelchris (approved):
Checked from phone, is the deletion needed, dont we need the sync still else we might delete to globally?

Also maybe add what would previously fail as a test?

src/common/contentPreference.ts
capJavert: no, it was a mistake from before, referenceId is sourceId in this case and also PK so delete is faster

rebelchris: Ah but whats the denominator between source-tag then?

Matbe im not seeing it on the changeset though

capJavert: Ah that, typeorm automatically adds `type` where clause because I am using `ContentPreferenceSource` entity

rebelchris: Ah i see, that answers it 🙏

__tests__/contentPreference.ts
capJavert: Added in 2815f85c2e0863a391f38b3d0f34048a3c75ef8c

rebelchris (approved):

PR #2438, opened 2024-11-13
infra: bump pulumi common to use spot nodes

idoshamun:


idoshamun:
@omBratteng this will move debezium to a spot node. any objections?

omBratteng:
> @omBratteng this will move debezium to a spot node. any objections?

@idoshamun debezium is still running on old nodes because of the disk, I'm in the progress of moving them over to new nodes. Then we can move them to new nodes

idoshamun:
@omBratteng oh so we can't do this update yet. but why is the disk a blocker? it's external to gke

omBratteng (approved):
Should be possible to merge this, the debezium instances have a `nodeSelector` on them, so they'll still be scheduled onto the old nodes

rebelchris (approved):

PR #2443, opened 2024-11-14
fix: approved email not being sent

sshanzel:
Root cause of the failure was that the data being passed to the CDC does not contain the ID returned by the processing.

Also updated the gql schema for unnecessary props.


src/schema/posts.ts
```diff
@@ -268,22 +268,14 @@ export const typeDefs = /* GraphQL */ `
     """
     image: String
     """
-    Url to image if applicable
+    Related source this is posted to
     """
-    imageUrl: String
```
sshanzel, line 273: This is found on the mutation, not on the actual entity.

src/schema/posts.ts
```diff
@@ -268,22 +268,14 @@ export const typeDefs = /* GraphQL */ `
...
-    """
-    Id of the Squad
-    """
-    sourceId: String
```
sshanzel, line 277: These IDs can be found on the related entity. For example, `source.id`.

src/schema/posts.ts
```diff
@@ -1400,19 +1384,26 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   Query: {
     sourcePostModeration: async (
       _,
-      { id, sourceId }: { id: string; sourceId: string },
+      { id }: { id: string },
```
sshanzel, line 1387: The ID of the entity is a unique identifier in itself; we don't need to check for the `sourceId`.

src/schema/posts.ts
```diff
@@ -1411,8 +1399,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         info,
         (builder) => {
           const queryBuilder = builder.queryBuilder.where(
-            `"${builder.alias}"."id" = :id AND ${builder.alias}."sourceId" = :sourceId`,
-            { id, sourceId },
+            `"${builder.alias}"."id" = :id`,
```
sshanzel, line 1402: The id on its own is a unique identifier, no need for any other params. Probably would just make the query worse, although not significant.

ilasw (approved):
Looks good to me 💯

capJavert (approved):

PR #2444, opened 2024-11-14
fix: small type cycle issue

rebelchris:
This should probably fix the deployment, stuck in dep cycle imoport

### DIFF
--- a/src/remoteConfig.ts
+++ b/src/remoteConfig.ts
@@ -60,3 +60,5 @@ class RemoteConfig {
 }
 
 export const remoteConfig = new RemoteConfig();
+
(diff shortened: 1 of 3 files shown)


src/remoteConfig.ts
```diff
@@ -60,3 +60,5 @@ class RemoteConfig {
 }
 
 export const remoteConfig = new RemoteConfig();
+
+export const planTypes = remoteConfig.vars.pricingIds || {};
```
omBratteng, line 64: Don't forget to update import

omBratteng: Nvm, you just did

omBratteng (approved):
🤞

PR #2445, opened 2024-11-14
infra: adjust spot nodes settings

idoshamun:


capJavert (commented):
We are just going 100% on all?

idoshamun:
@capJavert on background processes yes, on services that has api no

.infra/index.ts
```diff
@@ -306,10 +306,6 @@ if (isAdhocEnv) {
         { targetPort: 3000, port: 80, name: 'http' },
         { targetPort: 9464, port: 9464, name: 'metrics' },
       ],
-      spot: {
-        enabled: true,
-        weight: 50,
-      },
```
omBratteng, line 312: I don't see a reason to prevent API from running on spot nodes, it's been running on them since performance sprint, and haven't caused any issues

capJavert (approved):

.infra/index.ts
omBratteng: It shouldn't really affect it, but that's fine.
I really wish it was possible to say "Always run x of y replicas on these nodes"

idoshamun: yap once it is possible we can consider. also this will give us good look if it really changes anything to put it on spot nodes vs not

PR #2446, opened 2024-11-14
fix: remove plantypes to use remoteconfig directly

rebelchris:
Removed the need for function, just use vars inline

### DIFF
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -36,9 +36,9 @@ const updateUserSubscription = async ({
     return false;
   }
   const subscriptionType = data.data?.items.reduce((acc, item) => {
-    const pricingIds = remoteConfig.vars?.pricingIds || [];
-    if (item.price?.id && pricingIds[item.price.id]) {
-      acc = pricingIds[item.price.id];
+    const pricingIds = remoteConfig.vars?.pricingIds;
+    if (item.price?.id && pricingIds?.[item.price.id]) {
+      acc = pricingIds?.[item.price.id] || '';
     }
     return acc;
   }, '');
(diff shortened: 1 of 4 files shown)


src/routes/webhooks/paddle.ts
```diff
@@ -36,8 +36,8 @@ const updateUserSubscription = async ({
     return false;
   }
   const subscriptionType = data.data?.items.reduce((acc, item) => {
-    if (item.price?.id && planTypes()[item.price.id]) {
-      acc = planTypes()[item.price.id];
+    if (item.price?.id && remoteConfig.vars?.pricingIds?.[item.price.id]) {
+      acc = remoteConfig.vars.pricingIds?.[item.price.id] ?? '';
```
rebelchris, line 40: This was just TS complaining

capJavert: its meant to complain yes 😆 so we can fallback by design

src/schema/paddle.ts
```diff
@@ -32,7 +32,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   Query: {
     pricingIds: (): PricingId => {
       return {
-        value: JSON.stringify(planTypes()),
+        value: JSON.stringify(remoteConfig.vars?.pricingIds),
```
capJavert, line 35: since you defined `value` as required do you want to do `|| {}` here so its always string object?

capJavert: `JSON.stringify(undefined)` is `undefined`

capJavert (approved):

src/routes/webhooks/paddle.ts
omBratteng: Could define `remoteConfig.vars?.pricingIds?` as a constant in this parent function

omBratteng: `const pricingIds = remoteConfig.vars?.pricingIds || []` maybe

src/schema/paddle.ts
rebelchris, line 35:
```suggestion
        value: JSON.stringify(remoteConfig.vars?.pricingIds || {}),
```

omBratteng (approved):

PR #2447, opened 2024-11-14
chore: re-introduce public.content_preference

vpol:


.infra/clickhouse-sync.yml
```diff
@@ -71,6 +71,9 @@ max.batch.size: 32000
 metrics.enable: "false"
 cli.port: 7052
 
+# let's try something scary
+single.threaded: "true"
```
omBratteng, line 75: 😂

omBratteng (commented):
Lets give the deployment some more memory, temporarily

vpol:
> Lets give the deployment some more memory, temporarily

I'll update yaml manually if that would be the case.

omBratteng (approved):

.infra/clickhouse-sync.yml
capJavert: why single threaded?

vpol: 👻 🎃

vpol: > why single threaded?

testing performance.

PR #2449, opened 2024-11-14
feat: Add missing properties to SourcePostModeration

AmarTrebinjac:
Added props needed for edit flow


src/schema/posts.ts
```diff
@@ -115,6 +116,8 @@ export interface GQLSourcePostModeration {
   status: SourcePostModerationStatus;
   createdAt: Date;
   updatedAt: Date;
+  source: Source;
```
capJavert, line 119: source would also be optional?

capJavert (approved):

sshanzel:
This should be added here too: https://github.com/dailydotdev/daily-api/pull/2443

sshanzel (approved):

src/schema/posts.ts
sshanzel: Source is always there compared to post, if I'm not mistaken.

src/schema/posts.ts
```diff
@@ -276,6 +279,10 @@ export const typeDefs = /* GraphQL */ `
     """
     sourceId: String
     """
+    Source
+    """
+    source: Source
+    """
```
capJavert, line 285: Then here it should be
```suggestion
    Source
    """
    source: Source!
    """
```

src/schema/posts.ts
capJavert: You are right, then in schema it should be https://github.com/dailydotdev/daily-api/pull/2449#discussion_r1842118045

capJavert: @AmarTrebinjac also you can revert https://github.com/dailydotdev/daily-api/pull/2449/commits/21c0e6ff8dad063bca27dc704a414b58ee3571ef, sorry for the juggle 😆

AmarTrebinjac: @AmarTrebinjac haha, yep. I did a double take and checked, but my local branch is memeing now 💩💩

PR #2451, opened 2024-11-15
feat: read from content_preference source member

capJavert:
AS-681

- read from ContentPreferenceSource instead of SourceMember
- test updates
  - lots of changes, good thing is that none of them really failed due to logic but just difference in test setup (content preference vs source member fixtures) 
  - adjust tests to use ContentPreferenceSource instead of SourceMember
  - some tests also include SourceMember due to backward compatibility (we still write to SourceMember as well)
- triggers left as is until we completely remove SourceMember
- all queries read from ContentPreference but fields return the same source member fields so no changes needed for frontend (see graphorm/index.ts fields changes)
- content_preference table is now used for cdc instead of source_member (debezium config updated)

### DIFF
--- a/__tests__/workers/userUpdatedPlusSubscription.ts
+++ b/__tests__/workers/userUpdatedPlusSubscription.ts
@@ -6,10 +6,12 @@ import { User, Source, Feed } from '../../src/entity';
 import { ghostUser, PubSubSchema } from '../../src/common';
 import { typedWorkers } from '../../src/workers';
 import createOrGetConnection from '../../src/db';
-import { DataSource } from 'typeorm';
+import { DataSource, Not } from 'typeorm';
 import { usersFixture } from '../fixture/user';
 import { SourceMemberRoles } from '../../src/roles';
 import { randomUUID } from 'crypto';
+import { ContentPreferenceSource } from '../../src/entity/contentPreference/ContentPreferenceSource';
+import { ContentPreferenceStatus } from '../../src/entity/contentPreference/types';
 
 const PLUS_MEMBER_SQUAD_ID = '05862288-bace-4723-9218-d30fab6ae96d';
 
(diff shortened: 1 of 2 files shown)


src/workers/cdc/primary.ts
```diff
@@ -835,16 +837,32 @@ const onUserStateChange = async (
 const onSourceMemberChange = async (
   con: DataSource,
   logger: FastifyBaseLogger,
-  data: ChangeMessage<SourceMember>,
+  data: ChangeMessage<ContentPreferenceSource>,
 ) => {
+  const sourceId =
+    data.payload.after!.sourceId || data.payload.before!.sourceId;
+
+  const source = await con
```
idoshamun, line 845: Maybe we should consider storing a flag that says on content preference that this is a squad? It will make fetching the source redundant and the content preference object will be complete

idoshamun (commented):
Massive work @capJavert, way to go!!!

src/common/feedGenerator.ts
```diff
@@ -136,9 +137,18 @@ const getRawFiltersData = async (
...
         .select('"sourceId"')
-        .addSelect("COALESCE((flags->'hideFeedPosts')::boolean, FALSE)", 'hide')
-        .from(SourceMember, 't')
-        .where('"userId" = $2'),
+        .addSelect(
+          "COALESCE((t.flags->'hideFeedPosts')::boolean, FALSE)",
+          'hide',
+        )
+        .from(ContentPreference, 't')
+        .innerJoin(
```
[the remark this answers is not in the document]
capJavert: Same issue as https://github.com/dailydotdev/daily-api/pull/2451#discussion_r1849932721

capJavert (commented):
I added some missed stuff to not show blocked squads i member lists and sidebar and also to avoid sending them notifications.

__tests__/workers/userUpdatedPlusSubscription.ts
```diff
@@ -80,13 +82,16 @@ describe('userUpdatedPlusSubscription', () => {
...
-    expect(sourceMember.role).toEqual(SourceMemberRoles.Member);
-    expect(sourceMember.sourceId).toEqual(PLUS_MEMBER_SQUAD_ID);
-    expect(sourceMember.userId).toEqual(newBase.id);
+    const sourceMember = await con
+      .getRepository(ContentPreferenceSource)
+      .findOneBy({
+        referenceId: PLUS_MEMBER_SQUAD_ID,
+        userId: newBase.id,
+        status: Not(ContentPreferenceStatus.Blocked),
+      });
```
capJavert, line 91: @rebelchris if you can sanity check test changes for plus in this file

capJavert: functional change is only this `src/workers/userUpdatedPlusSubscription.ts`

src/workers/cdc/primary.ts
idoshamun: I think it's a downgrade from what we had and might make it unclear for other devs that it needs to be queried

capJavert: Thinking, should we have gone with a route to have different type for squad `content_preference` then.

idoshamun: Since it's water under the bridge now, let's proceed as is then. We should just be clear about it and pay attention

PR #2452, opened 2024-11-16
fix: streak issue - fetching last recover date with timezone

sshanzel:
This is a 2-part PR series about the issues surrounding the streak reset.

In this PR, this fixes an issue where the user has just restored their streak, but suddenly it was broken again. Referring to the issue mentioned here: https://github.com/dailydotdev/daily/issues/1538#issuecomment-2462438066

The screenshot below tells the user did not read for two days straight, however, on the next screenshot, it can be seen that the user restored the streak that should allow him to continue the streak even when 2 days had passed.
Screenshot 2024-11-16 at 11 06 19 AM

For context, the user lost his streak and restored it as seen below. At 12:03 the email notification about the streak being reset was sent, and immediately after, the restoration happened.
Screenshot 2024-11-16 at 10 54 24 AM

We have a cron job that runs on an internal, and it takes consideration of the current streak and restores performed by the user. However, there was an issue with how we pulled the data about the date of recovering the streak.

In postgresql, setting a timezone on a `timestamp` column (without timezone information column), would either push or pull the time depending on the timezone value. In order to set correctly the timezone, we must convert the value itself to a `timestamptz`, and then set the correct timezone preferred by the user.

This explains the reset that happened, the query to pull all users that should get reset was inaccurate by converting inaccurate value from the recovery date.

Note: to show that it actually fixes the issue, I've introduced a test with the user's configuration. I will [revert the fix temporarily](https://github.com/dailydotdev/daily-api/pull/2452/commits/29442290ead71a2b65094f77cb0dc0a2c3fce88e) to see if the test would fail.


src/cron/updateCurrentStreak.ts
```diff
@@ -16,13 +16,16 @@ const cron: Cron = {
...
         const usersPastStreakTime = await entityManager
           .createQueryBuilder()
-          .select(
-            `us.*, (date_trunc('day', us."lastViewAt" at time zone COALESCE(u.timezone, 'utc'))::date) AS "lastViewAtTz", u.timezone`,
+          .select(`us.*`)
+          .addSelect(
+            `date_trunc('day', us."lastViewAt" at time zone COALESCE(u.timezone, 'utc'))::date`,
+            'lastViewAtTz',
           )
+          .addSelect('u.timezone', 'timezone')
```
sshanzel, line 24: Just broke the 3 columns into their separate `select` statements to make the columns fetched more obvious.

src/cron/updateCurrentStreak.ts
```diff
@@ -16,13 +16,16 @@ const cron: Cron = {
...
           .addSelect('us.currentStreak', 'current')
           .addSelect('u."weekStart"', 'weekStart')
           .addSelect(
-            `(date_trunc('day', usa."lastRecoverAt" at time zone COALESCE(u.timezone, 'utc'))::date) - interval '1 day'`,
+            `(date_trunc('day', usa."lastRecoverAt"::timestamptz at time zone COALESCE(u.timezone, 'utc'))::date) - interval '1 day'`,
```
rebelchris, line 28: Didn't we discuss not doing the timestamptz columns for lots of reasons?
Not sure if we ended up not doing it, but they such a pain...

rebelchris (approved):
Trust this flow you tested, just sad we used timestamptz still, we manually convert anyhow so has no benefit now.

src/cron/updateCurrentStreak.ts
sshanzel: Yes, this column is not `timestamptz`. The problem occurs when you apply the time zone to a column similar to this (just `timestamp`).

Using `at time zone` in a non-timezone column messes up the values rather than adjusting to the right timezone. For us to use the `at time zone` statement, the column must be converted to `timestamptz` first even when the column is just a pure `timestamp`. Some nuances working with timezone in postgresql.

sshanzel:
> Trust this flow you tested, just sad we used timestamptz still, we manually convert anyhow so has no benefit now.

When working with `AT TIME ZONE`, we will have to convert the column first to get the right value. The entity itself is just a `timestamp` and if you add `AT TIME ZONE` with that type of column, postgresql would treat it as if the value is in that time zone already and not adjust the value then converts it to UTC, hence it gets pushed back.

src/cron/updateCurrentStreak.ts
capJavert: Potentially we can extract `date_trunc(...)` and everything else in a function maybe so the query is a bit more readable. Does not need to happen in this PR just a note it is starting to be really hard to follow 😆

capJavert: Also, we are using `AT TIME ZONE` in digest and it correctly transforms there from what I see. But there we convert from `NOW()` so not sure if that is the difference.

capJavert: I guess because `NOW()` includes the timezone offset
image

capJavert (approved):

src/cron/updateCurrentStreak.ts
capJavert: the date_trunc and other things could become a function so we can reuse, since the same thing is used in some other parts outside of cron. Not needed for this PR though.

src/cron/updateCurrentStreak.ts
sshanzel: Yeah, using `now` works as intended since it includes the timezone information. Basically a `timestamptz` type is the value returned from the function.

src/cron/updateCurrentStreak.ts
capJavert: postgresql

sshanzel: https://dailydotdev.atlassian.net/jira/software/projects/WT/boards/11/backlog?selectedIssue=WT-2249

src/cron/updateCurrentStreak.ts
sshanzel: > Potentially we can extract date_trunc(...) and everything else in a function maybe so the query is a bit more readable. Does not need to happen in this PR just a note it is starting to be really hard to follow 😆

Totally agree, especially have it in many places. Ticker created here: https://github.com/dailydotdev/daily-api/pull/2452#discussion_r1849688689

PR #2453, opened 2024-11-17
feat: return geo info in boot

idoshamun:


.infra/index.ts
```diff
@@ -306,6 +306,9 @@ if (isAdhocEnv) {
         { targetPort: 3000, port: 80, name: 'http' },
         { targetPort: 9464, port: 9464, name: 'metrics' },
       ],
+      backendConfig: {
+        customRequestHeaders: ['X-Client-Region:{client_region}'],
+      },
```
capJavert, line 311: Is `x-client-region` blocked from req headers, eg. clients can't inject it from req headers?

src/routes/boot.ts
```diff
@@ -130,6 +136,13 @@ type BootMiddleware = (
   res: FastifyReply,
 ) => Promise<Record<string, unknown>>;
 
+const geoSection = (req: FastifyRequest): BaseBoot['geo'] => {
+  return {
+    region: req.headers['x-client-region'] as string,
+    ip: req.ip,
```
capJavert, line 142: Do we need the ip to be returned publicly? Even with boot being a private endpoint it opens the attack vector which can dox the client's IP.

capJavert (commented):
Nice feature 👌

src/routes/boot.ts
idoshamun: Originally, I planned to provide the IP to paddle for the localization but decided to go with a more privacy-oriented way once I found this geolocation trick. We can remove it and there will be no breaking changes

.infra/index.ts
idoshamun: haven't tested it but i think gcp lb overrides it no matter what. we can try

src/routes/boot.ts
idoshamun: https://github.com/dailydotdev/daily-api/pull/2465

PR #2457, opened 2024-11-18
feat: add plus subscribers to squad

rebelchris:
Adds plus members to the squad automatically once they become subscriber.

AS-736 #done

### DIFF
--- a/src/workers/userUpdatedPlusSubscription.ts
+++ b/src/workers/userUpdatedPlusSubscription.ts
@@ -1,0 +1,44 @@
+import { ghostUser } from '../common';
+import { TypedWorker } from './worker';
+import { addNewSourceMember } from '../schema/sources';
+import { SourceMemberRoles } from '../roles';
+import { User } from '../entity';
+
+const PLUS_MEMBER_SQUAD_ID = '05862288-bace-4723-9218-d30fab6ae96d';
+const worker: TypedWorker<'user-updated'> = {
+  subscription: 'api.user-updated-plus-subscribed',
+  handler: async (message, con, log) => {
+    const {
+      data: { newProfile: user, user: oldUser },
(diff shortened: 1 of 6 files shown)


src/workers/userUpdatedPlusSubscription.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (afterFlags?.cycle === beforeFlags?.cycle) {
+      return;
+    }
+
+    const member = {
+      sourceId: PLUS_MEMBER_SQUAD_ID,
+      userId: user.id,
+      role: SourceMemberRoles.Member,
+    };
+    await addNewSourceMember(con, member);
```
idoshamun, line 39: This can fail if they're already a member right?

src/workers/userUpdatedPlusSubscription.ts
```diff
@@ -0,0 +1,44 @@
+import { ghostUser } from '../common';
+import { TypedWorker } from './worker';
+import { addNewSourceMember } from '../schema/sources';
+import { SourceMemberRoles } from '../roles';
+import { User } from '../entity';
+
+const PLUS_MEMBER_SQUAD_ID = '05862288-bace-4723-9218-d30fab6ae96d';
+const worker: TypedWorker<'user-updated'> = {
+  subscription: 'api.user-updated-plus-subscribed',
+  handler: async (message, con, log) => {
```
idoshamun, line 10: If the user stopped being a subscriber we should remove them from the squad

src/workers/userUpdatedPlusSubscription.ts
rebelchris: Guess it can yes, I can add a check before since it's worker anyway.
But very edge-casey I guess

PR #2460, opened 2024-11-19
fix: post moderation submission email props

sshanzel:
Currently, there is only 1 prop for us to send a name. We need to make it two so the names of the moderator and creator of the post be distinct from one another.

### DIFF
--- a/src/workers/newNotificationV2Mail.ts
+++ b/src/workers/newNotificationV2Mail.ts
@@ -176,7 +176,12 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
     }
 
     const { sharedPostId } = moderation;
-    const [squad, createdBy, sharedPost, moderator] = await Promise.all([
+    const [squad, createdBy, sharedPost, moderator]: [
+      Pick<SquadSource, 'type' | 'name' | 'handle' | 'image'> | null,
+      Pick<User, 'name' | 'reputation' | 'image'> | null,
+      Pick<ArticlePost, 'title' | 'image'> | null,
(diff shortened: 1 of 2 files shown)


omBratteng (approved):

src/workers/newNotificationV2Mail.ts
```diff
@@ -151,21 +151,29 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
   source_post_submitted: async (con, user, notification) => {
     const moderation = await con.getRepository(SourcePostModeration).findOne({
       where: { id: notification.referenceId },
-      select: ['sourceId', 'image', 'title', 'content', 'type', 'sharedPostId'],
+      select: [
```
capJavert, line 154: Can you type the `moderation` with `Pick` so it properly shows what fields are available.

capJavert: That was part of the issue because typeorm select is not applied over findOne type.

capJavert: Also on below `select` queries in Promise.all

capJavert (approved):

src/workers/newNotificationV2Mail.ts
sshanzel: Makes sense. This is very interesting. I might actually check if we can contribute to this part of the type in the package 👀

capJavert: that would be nice, they do offer it as generic with QueryBuilder and `getRawMany` but you still have to `Pick` by hand

PR #2461, opened 2024-11-19
fix: update subscription tracking

rebelchris:
We should only fire the event if subscription cycle actually changed, not any of the other changes.
I also decided to update our flags in this case so we have the right setting stored at all times.

AS-744 #done

### DIFF
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -97,7 +97,7 @@ const getUserId = async ({
   subscriptionId,
   userId,
 }: {
-  subscriptionId?: false | string | null | unknown;
+  subscriptionId?: false | string | null;
   userId?: string | undefined;
 }): Promise<string> => {
   if (userId) {
@@ -122,7 +122,7 @@ const planChanged = async (data: SubscriptionUpdatedEvent) => {
   const customData = data.data?.customData as { user_id: string };
   const userId = await getUserId({
     userId: customData?.user_id,
-    subscriptionId: 'subscriptionId' in data.data && data.data.subscriptionId,
+    subscriptionId: data.data && data.data.id,
   });
   const con = await createOrGetConnection();
   const flags = await con.getRepository(User).findOne({
@@ -154,7 +154,9 @@ const logPaddleAnalyticsEvent = async (
   const currency = data.data?.items?.[0]?.price?.unitPrice?.currencyCode;
   const userId = await getUserId({
     userId: customData?.user_id,
-    subscriptionId: 'subscriptionId' in data.data && data.data.subscriptionId,
+    subscriptionId:
+      ('subscriptionId' in data.data && data.data.subscriptionId) ||
+      data.data.id,
   });
   if (!userId) {
     return;
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -97,7 +97,7 @@ const getUserId = async ({
   subscriptionId,
   userId,
 }: {
-  subscriptionId?: false | string | null;
+  subscriptionId?: false | string | null | unknown;
   userId?: string | undefined;
 }): Promise<string> => {
   if (userId) {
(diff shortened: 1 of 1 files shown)


src/routes/webhooks/paddle.ts
```diff
@@ -203,10 +225,16 @@ export const paddle = async (fastify: FastifyInstance): Promise<void> => {
                 );
                 break;
               case EventName.SubscriptionUpdated:
-                await logPaddleAnalyticsEvent(
-                  eventData,
-                  AnalyticsEventName.ChangeBillingCycle,
-                );
+                if (!(await planChanged(eventData))) {
```
omBratteng, line 228: For readability, would this be better?
```suggestion
				const didPlanChange = await planChanged(eventData);
                if (!didPlanChange) {
```

sshanzel (approved):
Don't have anything blocking 👀

src/routes/webhooks/paddle.ts
rebelchris: Agreed :D

omBratteng (approved):

src/routes/webhooks/paddle.ts
```diff
@@ -93,7 +97,7 @@ const getUserId = async ({
   subscriptionId,
   userId,
 }: {
-  subscriptionId?: false | string | null;
+  subscriptionId?: false | string | null | unknown;
```
capJavert, line 100: Why this? When you add unknown you are effectively saying I don't know the type so the other types are not really needed anymore.

Do we know this type?

src/routes/webhooks/paddle.ts
```diff
@@ -114,6 +118,24 @@ const getUserId = async ({
...
+  const customData = data.data?.customData as { user_id: string };
+  const userId = await getUserId({
+    userId: customData?.user_id,
+    subscriptionId: 'subscriptionId' in data.data && data.data.subscriptionId,
+  });
+  const con = await createOrGetConnection();
+  const flags = await con.getRepository(User).findOne({
+    where: { id: userId },
+    select: ['subscriptionFlags'],
+  });
```
capJavert, line 131: We double query the user, once in `getUserId` then again in the call in this function?

capJavert: If we want to abstract anything maybe better that `getUserId` is just `getUser` and then we get user object either way.

src/routes/webhooks/paddle.ts
rebelchris: From the last addition it just became unknown type in their types.
For the others we do know and didn't need the unknown 😢

src/routes/webhooks/paddle.ts
rebelchris: Yeah was thinking of that as well, just meant more refactors on end type as it's a optional return.

rebelchris: Also the early return only cares about returning the id

src/routes/webhooks/paddle.ts
rebelchris: My bad was wrong field in this type

src/routes/webhooks/paddle.ts
rebelchris: Alternative could be to have a seperate function for this case? (only 1 user now, but maybe future more)
Let me know if you prefer that

capJavert: I would just do it inline in both places but if we use util then I would go with `getUser`. Neither blocking for this PR.

src/routes/webhooks/paddle.ts
```diff
@@ -114,6 +118,24 @@ const getUserId = async ({
   return user.id;
 };
 
+const planChanged = async (data: SubscriptionUpdatedEvent) => {
+  const customData = data.data?.customData as { user_id: string };
+  const userId = await getUserId({
+    userId: customData?.user_id,
+    subscriptionId: data.data && data.data.id,
```
capJavert, line 125:
```suggestion
    subscriptionId: data.data?.id,
```

PR #2462, opened 2024-11-19
feat: slack notification on completed transactions

omBratteng:


src/routes/webhooks/paddle.ts
```diff
@@ -166,6 +166,115 @@ const logPaddleAnalyticsEvent = async (
   ]);
 };
 
+const concatText = (a: string, b: string) => `${a}\n${b}`;
```
capJavert, line 169: Something like this?

```suggestion
const concatText = (a: string, b: string) => [a, b].filter(Boolean).join(`\n`)
```

to avoid concating empty values

src/routes/webhooks/paddle.ts
```diff
@@ -213,6 +322,7 @@ export const paddle = async (fastify: FastifyInstance): Promise<void> => {
                   eventData,
                   AnalyticsEventName.ReceivePayment,
                 );
+                await notifyNewPaddleTransaction(eventData);
```
capJavert, line 325: `Promise.all`?

capJavert (approved):

rebelchris (approved):
Ante's comment makes sense but not blocking

PR #2466, opened 2024-11-19
fix: add experimentation for plus members

rebelchris:
Add experimentation exposure for plus members.

If user is plus it looks like this:

```
  "exp": {
    "f": "{}",
    "e": [],
    "a": {
      "show_error": 1,
      "plus": 1
    }
  },
```


omBratteng (approved):

src/routes/boot.ts
```diff
@@ -530,7 +531,13 @@ const loggedInBoot = async ({
       notifications: { unreadNotificationsCount },
       squads,
       accessToken,
-      exp,
+      exp: {
+        ...exp,
+        a: {
+          ...exp.a,
+          ...(isPlus ? { plus: 1 } : undefined),
```
capJavert, line 538:
```suggestion
          plus: isPlus ? 1 : undefined
```

its json so undefined will disappear

capJavert (approved):

src/routes/boot.ts
```diff
@@ -476,6 +476,10 @@ const loggedInBoot = async ({
     const isTeamMember = exp?.a?.team === 1;
     const isPlus = isPlusMember(user.subscriptionFlags?.cycle);
 
+    if (isPlus) {
+      exp.a.plus = 1;
```
capJavert, line 480: Is `a` object here, if its undefined it will break?

rebelchris: Defaults to empty object I think:
https://github.com/dailydotdev/daily-api/blob/cfd63715cf57b59672611a93e514172711883153/src/routes/boot.ts#L362-L365

capJavert: jep!

capJavert (approved):

idoshamun (approved):

PR #2467, opened 2024-11-19
feat: remove notifications once the post has been approved/rejected/deleted

sshanzel:
Reference: https://dailydotdev.slack.com/archives/C07GFU7DX8X/p1732006742496929

Changes:
- Renamed some files to follow current convention.
- Realized we are missing some tests about what would the actual notification generated look like.
  - Found out we were missing the source avatars.
- Clear the notifications sent when the post has been approved/rejected/deleted.


__tests__/notifications/index.ts
```diff
@@ -1253,6 +1258,158 @@ describe('storeNotificationBundle', () => {
     ]);
   });
 
+  it('should generate source_post_submitted notification', async () => {
```
sshanzel, line 1261: Realized we were missing these 3 tests, hence, added here now along with the fix.

src/notifications/common.ts
```diff
@@ -283,11 +285,17 @@ export const generateUserNotificationUniqueKey = ({
...
-  [SourcePostModerationStatus.Pending]: NotificationType.SourcePostSubmitted,
+export const cleanupSourcePostModerationNotifications = async (
+  con: ConnectionManager,
+  post: ChangeObject<SourcePostModeration>,
+) => {
+  if (!post?.id) {
+    return;
+  }
+
+  await con.getRepository(NotificationV2).delete({
```
capJavert, line 296: I am not sure we want to delete notification, they are usually immutable and can reference other stuff, maybe introduce soft delete cc: @idoshamun ?

idoshamun: let me search for some references. I know for sure that when we delete a post we also hide the notifications. Not sure if we delete or soft delete

idoshamun: So I guess we do delete notifications 😂

capJavert: :triggered:

capJavert (approved):

PR #2468, opened 2024-11-20
chore: reduce logging on lettericon proxy

omBratteng:
Since we added lettericon proxy to API, we've been getting lots of extra logs, making it more difficult to go through production logs because of all the extra spam.

So this sets the log level for the proxy to `warn` instead of `info`


rebelchris (approved):
What does it even complain about?

omBratteng:
Just a bunch of info messages
image

PR #2470, opened 2024-11-20
feat: ad placement in digest

idoshamun:
Add type for posts to allow flexible components and support ads.

The email template already supports it so it's safe to roll out. The existing one will drop the ad, and there's a new template that will present it which we gonna test

### DIFF
--- a/src/common/personalizedDigest.ts
+++ b/src/common/personalizedDigest.ts
@@ -24,6 +24,7 @@ import { DayOfWeek } from './date';
 import { GarmrService } from '../integrations/garmr';
 import { baseFeedConfig } from '../integrations/feed';
 import { FeedConfigName } from '../integrations/feed';
+import { isPlusMember } from '../paddle';
 
 type TemplatePostData = Pick<
   ArticlePost,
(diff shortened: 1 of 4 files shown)


src/common/personalizedDigest.ts
```diff
@@ -129,10 +131,22 @@ const getEmailVariation = async ({
   const dayName = dayEntry ? dayEntry[0] : undefined;
   const userName = user.name?.trim().split(' ')[0] || user.username;
   const userStreak = await user.streak;
+
+  const posts: Record<string, unknown>[] = getPostsTemplateData({
+    posts: postsData,
+    feature,
+  });
+  if (posts.length >= 2 && !isPlusMember(user.subscriptionFlags?.cycle)) {
```
capJavert, line 139: Maybe expose this `2` in digest config so we can experiment?

capJavert (approved):

src/common/personalizedDigest.ts
idoshamun: omg! how come i haven't thought about it :thinkaboutit:

PR #2471, opened 2024-11-20
refactor: user streak to include last view with timezone

sshanzel:
Expose the column that contains the timezone-adjusted column for last view at.

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -669,6 +669,7 @@ export const typeDefs = /* GraphQL */ `
     total: Int
     current: Int
     lastViewAt: DateTime
+    lastViewAtTz: DateTime
     weekStart: Int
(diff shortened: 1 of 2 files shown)


src/schema/users.ts
```diff
@@ -669,6 +669,7 @@ export const typeDefs = /* GraphQL */ `
     total: Int
     current: Int
     lastViewAt: DateTime
+    lastViewAtTz: DateTime
```
capJavert, line 672: Why do we need the both timezoned and utc?

sshanzel: Old one is just for backwards support. The one with timezone is for this PR: https://github.com/dailydotdev/apps/pull/3832

capJavert (approved):

PR #2477, opened 2024-11-21
fix: use shared and content title and image for post types

capJavert:
- no posts in email was due to outdated templates being scheduled on our email platform, it was a temporary issue and it should not happen now that everything scheduled before code change was deploy
- no title is because of shared posts, we can show shared post title
- no image, freeform post without image (we should technically load first image from content)
  - also case for shared post, we would load image of a shared post

Closes https://github.com/dailydotdev/daily/issues/1600

### DIFF
--- a/src/common/markdown.ts
+++ b/src/common/markdown.ts
@@ -159,3 +159,40 @@ export const checkHasMention = (content: string, username: string) => {
     return words.some((word) => word === `@${username}`);
   });
 };
+
+export const findMarkdownTag = ({
+  tokens,
+  tag,
+  depth = 0,
+  maxDepth = 2,
+}: {
+  tokens: Token[];
+  tag: string;
+  depth?: number;
+  maxDepth?: number;
+}): Token | undefined => {
+  if (depth > maxDepth) {
+    return undefined;
+  }
+
+  for (let i = tokens.length - 1; i >= 0; i -= 1) {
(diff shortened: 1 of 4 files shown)


src/common/markdown.ts
```diff
@@ -159,3 +159,40 @@ export const checkHasMention = (content: string, username: string) => {
...
+  maxDepth = 2,
+}: {
+  tokens: Token[];
+  tag: string;
+  depth?: number;
+  maxDepth?: number;
+}): Token | undefined => {
+  if (depth > maxDepth) {
+    return undefined;
+  }
```
capJavert, line 176: since this is recursive (due to markdown structure being tree) I added depth so we don't go too deep and affect performance.

By default its 2 level, which should be more then enough for markdown content we can create in our editor.

src/common/personalizedDigest.ts
```diff
@@ -88,8 +93,13 @@ const getPostsTemplateData = ({
     const summary = post.summary || '';
 
     return {
-      post_title: post.title,
-      post_image: mapCloudinaryUrl(post.image || pickImageUrl(post)),
+      post_title: post.title || post.sharedPostTitle,
```
rebelchris, line 96: I thought I already added this, but maybe it was for another email.

rebelchris (approved):
Cool :)

PR #2481, opened 2024-11-22
fix: reject reason should be required when rejecting a post

sshanzel:
- When rejecting a post, reason should be required.
- When the reason is `Other`, we should require the note from moderator.

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -2342,16 +2342,17 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         throw new ForbiddenError('Access denied!');
       }
 
-      if (status === SourcePostModerationStatus.Rejected && !rejectionReason) {
-        throw new ValidationError('Rejection reason is required');
-      }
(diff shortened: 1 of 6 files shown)


src/schema/posts.ts
```diff
@@ -2342,6 +2342,19 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         throw new ForbiddenError('Access denied!');
       }
 
+      if (status === SourcePostModerationStatus.Rejected) {
+        if (!rejectionReason) {
+          throw new ValidationError('Rejection reason is required');
+        }
+
+        if (
+          rejectionReason?.toLowerCase()?.includes('other') &&
```
capJavert, line 2351: should it not be `=== 'other'?

capJavert (approved):
Just verify this is backwards compatible

src/schema/posts.ts
sshanzel: Yeah, we can use `===` too. Did not remember if it was `other` or `others` from the FE, and just to ensure it with this one. Although I've checked it now and it is just `other`.

PR #2483, opened 2024-11-23
fix: vordr check to work with moderation

omBratteng:


sshanzel (approved):
🚀

src/entity/posts/utils.ts
```diff
@@ -386,30 +386,28 @@ export const createSharePost = async ({
...
-        createdPost.banned = true;
-        createdPost.showOnFeed = false;
+    const vordrStatus = await checkWithVordr(
+      {
+        id: createdPost.id,
+        content: createdPost.title || undefined,
+        type: VordrFilterType.Post,
+      },
+      { con, userId, req: ctx?.req },
+    );
```
capJavert, line 396: What is the goal here, because this is only for shared posts right? 

Would it make sense to move it to approval worker `api.source-post-moderation-approved-notification` so it catches all posts approved after moderation?

sshanzel: I think Ido also wants it to happen pre-submission: https://dailydotdev.slack.com/archives/C013RGVGPN1/p1732728195138469?thread_ts=1732726519.529059&cid=C013RGVGPN1

idoshamun: Exactly. we should silently ignore it

omBratteng: Yeah, it could've been moved to that worker. But it should also happen before it reaches moderation too

sshanzel: Still, I think we can push this PR. If in case the user was vordred after the submission which won't have the flag yet. Edge casey but at least a second level filter.

omBratteng: @sshanzel's PR #2497 handles it so that it blocks it from reaching moderation

capJavert (approved):

PR #2485, opened 2024-11-25
fix: reject reason on notification

sshanzel:


src/notifications/generate.ts
```diff
@@ -136,7 +137,7 @@ export const notificationTitleMap: Record<
   source_post_approved: (ctx: NotificationPostContext) =>
     `Woohoo! Your post has been approved and is now live in ${ctx.source.name}. Check it out!`,
   source_post_rejected: (ctx: NotificationPostModerationContext) =>
-    `Your post in ${ctx.source.name} was not approved for the following reason: ${ctx.post.rejectionReason}. Please review the feedback and consider making changes before resubmitting.`,
+    `Your post in ${ctx.source.name} was not approved for the following reason: ${rejectReason[ctx.post.rejectionReason!]}. Please review the feedback and consider making changes before resubmitting.`,
```
rebelchris, line 140: For Other it will also show the commentary somewhere?

rebelchris (approved):

src/notifications/generate.ts
sshanzel: Yes. Should be part of the description already.

capJavert (approved):

src/notifications/generate.ts
capJavert: sounds good to me

PR #2489, opened 2024-11-27
feat: option to block words

rebelchris:
Re-used content preference so we can go all ways with words as well (block/follow/notify)

For now only block/unblock will be used.
(This way we can also seperate main feed from custom feeds block words :))

### DIFF
--- a/__tests__/contentPreference.ts
+++ b/__tests__/contentPreference.ts
@@ -1822,6 +1822,34 @@ describe('mutation unblock', () => {
         },
       ]);
     });
+
+    it('should unblock', async () => {
+      loggedUser = '1-ublm';
+
+      const res = await client.query(MUTATION, {
+        variables: {
+          id: '2-ublm',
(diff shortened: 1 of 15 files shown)


__tests__/contentPreference.ts
```diff
@@ -1821,6 +1852,48 @@ describe('mutation unblock', () => {
     });
   });
 
+  describe('word', () => {
+    beforeEach(async () => {
+      await saveFixtures(con, Feed, [{ id: '1-ublm', userId: '1-ublm' }]);
+      await con.getRepository(ContentPreferenceKeyword).save([
```
capJavert, line 1858: `ContentPreferenceWord`?

capJavert (approved):

idoshamun (approved):

__tests__/contentPreference.ts
rebelchris: yes 🙈

PR #2491, opened 2024-11-29
feat: functional block words

rebelchris:
Option to block words and resolve them in feed config.

Note:
So far only `blockWords` accepts a comma seperated string, this is due to the ability to block more then one.
We make it unique and onConflict ignore.

### DIFF
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -18,7 +18,7 @@ import {
   NotificationPreferenceSource,
   NotificationPreferenceUser,
 } from '../entity';
-import { ghostUser } from './utils';
+import { ghostUser, uniqueifyArray } from './utils';
 import { randomUUID } from 'crypto';
 import { SourceMemberRoles } from '../roles';
 import { ContentPreferenceKeyword } from '../entity/contentPreference/ContentPreferenceKeyword';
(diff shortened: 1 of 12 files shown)


src/common/contentPreference.ts
```diff
@@ -343,17 +343,39 @@ const blockKeyword: BlockEntity = async ({ ctx, id }) => {
...
-    type: ContentPreferenceType.Word,
-  });
+  const ids = uniqueifyArray(
+    id
+      .split(',')
+      .map((item) => item.trim())
+      .filter(Boolean),
+  );
+
+  if (ids.length > 50) {
```
rebelchris, line 360: Just failsafe so someone can't insert a million :)

src/common/feedGenerator.ts
```diff
@@ -133,6 +133,15 @@ const getRawFiltersData = async (
         .andWhere('"userId" = $2')
         .andWhere(`status = '${ContentPreferenceStatus.Blocked}'`),
     ),
+    rawFilterSelect(con, 'words', (qb) =>
+      qb
+        .select(['"referenceId"'])
+        .from(ContentPreference, 'w')
+        .where('"feedId" = $1')
+        .andWhere(`type = '${ContentPreferenceType.Word}'`)
+        .andWhere(`status = '${ContentPreferenceStatus.Blocked}'`)
```
rebelchris, line 142: For now we only care about blocked, eventually if we want reverse side we can move this to return status :)

src/schema/contentPreference.ts
```diff
@@ -120,6 +120,28 @@ export const typeDefs = /* GraphQL */ `
...
+      entity: ContentPreferenceType!
+      """
+      Paginate after opaque cursor
+      """
+      after: String
+      """
+      Paginate first
+      """
+      first: Int
+    ): ContentPreferenceConnection
```
capJavert, line 144: @auth?

capJavert (approved):

src/schema/contentPreference.ts
rebelchris: Yeah good point, I copied userFollow which is public.
Adding auth

PR #2492, opened 2024-11-29
fix: we should store words lowercased

rebelchris:
We should store blocked words lowercased.
(regardless feed implementation will do lowercasing both sides)

### DIFF
--- a/src/common/contentPreference.ts
+++ b/src/common/contentPreference.ts
@@ -353,7 +353,7 @@ const blockWord: BlockEntity = async ({ ctx, id }) => {
   const ids = uniqueifyArray(
     id
       .split(',')
-      .map((item) => item.trim())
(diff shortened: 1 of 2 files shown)


src/common/contentPreference.ts
```diff
@@ -353,7 +353,7 @@ const blockWord: BlockEntity = async ({ ctx, id }) => {
   const ids = uniqueifyArray(
     id
       .split(',')
-      .map((item) => item.trim())
+      .map((item) => item.trim().toLowerCase())
```
capJavert, line 356: better to do lowercase once on `id`?

capJavert (approved):

src/common/contentPreference.ts
rebelchris: Solved

capJavert (approved):

PR #2494, opened 2024-12-03
feat: following feed API

rebelchris:
Following feed API.

Went for the easy implementation, we should monitor the rawFeedFilter query and maybe rethink.
(can be later I feel like)

Also re-used versionToFeed which isn't 100% optional, but was the faster implementation to re-use lofn feed.

### DIFF
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -36,6 +36,7 @@ import {
 } from '../entity/contentPreference/types';
 import { ContentPreference } from '../entity/contentPreference/ContentPreference';
 import { ContentPreferenceWord } from '../entity/contentPreference/ContentPreferenceWord';
+import { ContentPreferenceUser } from '../entity/contentPreference/ContentPreferenceUser';
 
 export const WATERCOOLER_ID = 'fd062672-63b7-4a10-87bd-96dcd10e9613';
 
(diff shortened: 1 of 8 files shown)


src/integrations/feed/generators.ts
```diff
@@ -123,11 +124,14 @@ export const feedGenerators: Partial<Record<FeedVersion, FeedGenerator>> =
...
-export const versionToFeedGenerator = (version: number): FeedGenerator => {
+export const versionToFeedGenerator = (
+  version: number | FeedVersion,
+  inputOpts?: Options,
+): FeedGenerator => {
   return new FeedGenerator(
     feedClient,
     new FeedLofnConfigGenerator(baseFeedConfig, lofnClient, {
-      ...opts,
+      ...(inputOpts ?? opts),
```
capJavert, line 134: I think we should just instantiate lofn generator in `schema/feed` as we do for other non main feeds, this passing of opts multiple times is complex as is.

src/common/feedGenerator.ts
```diff
@@ -222,6 +231,20 @@ const wordsToFilters = ({
   );
 };
 
+const usersToFilters = ({
```
capJavert, line 234: Do we need this, we can just use result from sql?

src/common/feedGenerator.ts
```diff
@@ -261,10 +297,11 @@ const sourcesToFilters = ({
   );
 
   return {
-    excludeSources: (excludeSources || [])
-      .map((s) => s.sourceId)
+    excludeSources: (blocked || [])
```
capJavert, line 300: blocked will now always be array so we don't need `||`

src/integrations/feed/generators.ts
capJavert: Should not be the issue, just copy paste to schema right?

src/common/feedGenerator.ts
rebelchris: I wanted to do this one, in case we introduce the opposite side of the query at one stage?
Also don't think the one level fn loop will add much right?

capJavert: Yeah, agreed

capJavert (approved):
Looks good, just double check the feed once merge, if we did not accidentally break something with sources refactor.

PR #2496, opened 2024-12-03
feat: add `clickbaitShieldEnabled` flag

omBratteng:


src/entity/Settings.ts
```diff
@@ -129,5 +131,6 @@ export const SETTINGS_DEFAULT = {
     sidebarOtherExpanded: true,
     sidebarResourcesExpanded: true,
     sidebarBookmarksExpanded: true,
+    clickbaitShieldEnabled: true,
```
omBratteng, line 134: Need to check with product if it should be enabled by default or not.

capJavert (approved):

PR #2497, opened 2024-12-03
feat: vordr source post moderation submissions

sshanzel:
- Check vordr when user submits a post for moderation.
- Query for pulling pending requests will filter vordred request.
  - Of course only for moderators, for regular users, you will still see your submission 😈 
- Fixed some copy-pasta issue with the index naming.

### DIFF
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -256,10 +256,17 @@ export type CreateSourcePostModeration = Omit<
     postId?: string;
   };
 
-export const createSourcePostModeration = async (
-  con: DataSource | EntityManager,
-  args: CreateSourcePostModeration,
-) => {
+interface CreateSourcePostModerationProps {
+  con: DataSource | EntityManager;
+  ctx: AuthContext;
+  args: CreateSourcePostModeration;
+}
+
+export const createSourcePostModeration = async ({
+  con,
+  ctx,
+  args,
+}: CreateSourcePostModerationProps) => {
   if (args.postId) {
     const post = await con
       .getRepository(Post)
@@ -274,6 +281,20 @@ export const createSourcePostModeration = async (
     ...args,
     status: SourcePostModerationStatus.Pending,
   });
+
+  const content = `${args.title} ${args.content}`.trim();
+
+  newModerationEntry.flags = {
+    vordr: await checkWithVordr(
+      {
+        id: newModerationEntry.id,
+        type: VordrFilterType.PostModeration,
+        content,
+      },
+      { con, userId: ctx.userId, req: ctx.req },
+    ),
+  };
+
   return await con.getRepository(SourcePostModeration).save(newModerationEntry);
 };
 
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -257,16 +257,16 @@ export type CreateSourcePostModeration = Omit<
   };
 
 interface CreateSourcePostModerationProps {
-  con: DataSource | EntityManager;
   ctx: AuthContext;
   args: CreateSourcePostModeration;
 }
 
 export const createSourcePostModeration = async ({
-  con,
   ctx,
   args,
 }: CreateSourcePostModerationProps) => {
+  const { con } = ctx;
+
   if (args.postId) {
     const post = await con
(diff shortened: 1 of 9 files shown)


src/entity/Comment.ts
```diff
@@ -50,7 +50,7 @@ export class Comment {
   featured: boolean;
 
   @Column({ type: 'jsonb', default: {} })
-  @Index('IDX_user_flags_vordr', { synchronize: false })
```
sshanzel, line 53: Copy-pasta issue. Confirmed from the migration file on what should be the right name. Should not impact anything since we have `synchronize: false`.

src/common/post.ts
```diff
@@ -274,6 +281,20 @@ export const createSourcePostModeration = async (
     ...args,
     status: SourcePostModerationStatus.Pending,
   });
+
+  const content = `${args.title} ${args.content}`.trim();
```
sshanzel, line 285: @omBratteng what do you think with this? Since the user can either add a title/content. And title can also be coming from external link generated by our scraper.

rebelchris (approved):
Nothing blocking from my side, but best wait for Ante/Ole-Martin as well

src/common/post.ts
omBratteng: 👍

omBratteng:
Does this prevent the notification being sent out to the squad admins?

sshanzel:
> Member

Good point. I will check.

sshanzel:
@omBratteng added vordr check before sending pubsub event 🚀

src/migration/1733244072344-SourcePostModerationFlagsVordr.ts
```diff
@@ -0,0 +1,25 @@
...
+  implements MigrationInterface
+{
+  name = 'SourcePostModerationFlagsVordr1733244072344';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "source_post_moderation" ADD "flags" jsonb NOT NULL DEFAULT '{}'`,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_source_post_moderation_flags_vordr" ON post USING HASH (((flags->'vordr')::boolean))`,
```
omBratteng, line 13:
```suggestion
      `CREATE INDEX IF NOT EXISTS "IDX_source_post_moderation_flags_vordr" ON post USING HASH (((flags->'vordr')::boolean))`,C
```

src/migration/1733244072344-SourcePostModerationFlagsVordr.ts
```diff
@@ -0,0 +1,25 @@
...
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DROP INDEX "IDX_source_post_moderation_flags_vordr"`,
```
omBratteng, line 19:
```suggestion
      `DROP INDEX IF EXISTS "IDX_source_post_moderation_flags_vordr"`,
```

omBratteng (changes requested):
Minor changes in the migration

src/schema/posts.ts
```diff
@@ -1764,10 +1767,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       const pendingPost = await validateSourcePostModeration(ctx, props);
-      const moderatedPost = await createSourcePostModeration(
-        ctx.con,
-        pendingPost,
-      );
+      const moderatedPost = await createSourcePostModeration({
+        ctx,
+        con: ctx.con,
+        args: pendingPost,
+      });
```
omBratteng, line 1774: Since we're passing `ctx`, we don't need to pass `ctx.con` as well, since we can access it from `ctx` in the function?

sshanzel: Right. I thought this was also used somewhere else that utilizes a `transaction`. But yeah, basically about that.

omBratteng (approved):

capJavert (approved):

PR #2499, opened 2024-12-04
feat: update BookmarkList entity and CRUD

ilasw:
- Updated entity for `BookmarkList` and `Bookmark`
- Updated CRUD operations for `BookmarkList`
- Updated tests for changed methods

### DIFF
--- a/__tests__/helpers.ts
+++ b/__tests__/helpers.ts
@@ -43,7 +43,7 @@ export class MockContext extends Context {
 
   constructor(
     con: DataSource,
-    userId: string = null,
+    userId: string | null = null,
     premium = false,
     roles = [],
     req?: FastifyRequest,
--- a/src/entity/Bookmark.ts
+++ b/src/entity/Bookmark.ts
@@ -25,7 +25,7 @@ export class Bookmark {
   })
   post: Promise<Post>;
 
-  @ManyToOne(() => BookmarkList, (list) => list.bookmarks, {
+  @ManyToOne(() => BookmarkList, {
     lazy: true,
     onDelete: 'CASCADE',
   })
--- a/src/entity/Bookmark.ts
+++ b/src/entity/Bookmark.ts
@@ -25,9 +25,9 @@ export class Bookmark {
   })
   post: Promise<Post>;
 
-  @ManyToOne(() => BookmarkList, {
+  @ManyToOne(() => BookmarkList, (list) => list.bookmarks, {
     lazy: true,
-    onDelete: 'SET NULL',
+    onDelete: 'CASCADE',
   })
   list: Promise<BookmarkList>;
 
--- a/src/entity/BookmarkList.ts
+++ b/src/entity/BookmarkList.ts
@@ -38,9 +38,8 @@ export class BookmarkList {
   @UpdateDateColumn()
   updatedAt: Date;
 
-  @OneToMany('Bookmark', 'list', {
+  @OneToMany(() => Bookmark, (bookmark) => bookmark.listId, {
     lazy: true,
-    onDelete: 'CASCADE',
   })
   bookmarks: Promise<Bookmark[]>;
 }
--- a/src/entity/BookmarkList.ts
+++ b/src/entity/BookmarkList.ts
@@ -1,4 +1,15 @@
-import { Column, Entity, Index, PrimaryGeneratedColumn } from 'typeorm';
+import {
+  Column,
+  Entity,
+  Index,
+  PrimaryGeneratedColumn,
+  CreateDateColumn,
+  UpdateDateColumn,
+  ManyToOne,
+  OneToMany,
+} from 'typeorm';
+import type { User } from './user';
+import { Bookmark } from './Bookmark';
 
 @Entity()
 export class BookmarkList {
@@ -9,6 +20,27 @@ export class BookmarkList {
   @Index()
   userId: string;
 
+  @ManyToOne('User', {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  user: Promise<User>;
+
   @Column({ type: 'text' })
   name: string;
+
+  @Column({ type: 'text', nullable: true })
+  icon: string;
+
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @UpdateDateColumn()
+  updatedAt: Date;
+
+  @OneToMany('Bookmark', 'list', {
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  bookmarks: Promise<Bookmark[]>;
 }
--- a/src/schema/bookmarks.ts
+++ b/src/schema/bookmarks.ts
@@ -10,7 +10,7 @@ import {
 } from './common';
 import { traceResolvers } from './trace';
 import { AuthContext, BaseContext, Context } from '../Context';
-import { Bookmark, BookmarkList, Post } from '../entity';
+import { Bookmark, BookmarkList, Post, User } from '../entity';
 import {
   base64,
   bookmarksFeedBuilder,
@@ -22,6 +22,8 @@ import {
 import { SelectQueryBuilder } from 'typeorm';
 import { GQLPost } from './posts';
 import { Connection } from 'graphql-relay';
+import { isPlusMember } from '../paddle';
+import { ForbiddenError } from 'apollo-server-errors';
 
 interface GQLAddBookmarkInput {
   postIds: string[];
@@ -30,6 +32,9 @@ interface GQLAddBookmarkInput {
 export interface GQLBookmarkList {
   id: string;
   name: string;
+  icon: string;
+  createdAt: Date;
+  updatedAt: Date;
 }
 
 export const typeDefs = /* GraphQL */ `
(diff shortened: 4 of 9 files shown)


src/entity/BookmarkList.ts
```diff
@@ -9,6 +20,26 @@ export class BookmarkList {
...
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @UpdateDateColumn()
+  updatedAt: Date;
+
+  @OneToMany(() => Bookmark, (bookmark) => bookmark.listId, {
+    lazy: true,
+  })
+  bookmarks: Promise<Bookmark[]>;
```
capJavert, line 44:
```suggestion
  @OneToMany('Bookmark', (bookmark: Bookmark) => bookmark.listId, {
    lazy: true,
  })
  bookmarks: Promise<Bookmark[]>;
```

capJavert: Also if you can adjust existing relations here and in `Bookmark` entity to use this kind of pattern. It helps with circular dependencies.

src/schema/bookmarks.ts
```diff
@@ -325,16 +346,41 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        }
+
+        throw new ForbiddenError(
+          `You have reached the maximum list count (${maxFoldersCount})`,
+        );
+      }
+
+      return ctx.con.getRepository(BookmarkList).save({
         name,
+        icon,
```
capJavert, line 378: Missing code to check if its one of the allowed icons?

src/schema/bookmarks.ts
```diff
@@ -325,16 +346,41 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       ctx: AuthContext,
-    ): Promise<GQLBookmarkList> =>
-      ctx.con.getRepository(BookmarkList).save({
+    ): Promise<GQLBookmarkList> => {
+      const user = await ctx.con.getRepository(User).findOneOrFail({
+        where: { id: ctx.userId },
+        select: ['subscriptionFlags'],
+      });
+      const isPlus = isPlusMember(user.subscriptionFlags?.cycle);
+      const maxFoldersCount = isPlus ? 50 : 1;
```
capJavert, line 358: let's make this two constants in `src/types.ts`

src/entity/Bookmark.ts
```diff
@@ -27,7 +27,7 @@ export class Bookmark {
 
   @ManyToOne(() => BookmarkList, {
     lazy: true,
-    onDelete: 'SET NULL',
+    onDelete: 'CASCADE',
   })
   list: Promise<BookmarkList>;
```
AmarTrebinjac, line 32: I think we always add a seperate property for id as well, so in this case "listId".

But also, maybe it's a bit redundant, I would still prefer this to be called "bookmarkList" rather than "list", but that's perhaps just me.

src/schema/bookmarks.ts
```diff
@@ -325,16 +346,41 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        .getRepository(BookmarkList)
+        .countBy({ userId: ctx.userId });
+
+      if (userFoldersCount >= maxFoldersCount) {
+        if (isPlus) {
+          logger.warn(
+            { listCount: userFoldersCount, userId: ctx.userId },
+            'bookmark folders limit reached',
+          );
+        }
```
capJavert, line 369: Don't think we need this, we will see error below already.

capJavert (commented):
Overall looks good, few comments/checks.

src/entity/BookmarkList.ts
```diff
@@ -9,6 +20,26 @@ export class BookmarkList {
...
+    lazy: true,
+    onDelete: 'CASCADE',
+  })
+  user: Promise<User>;
+
   @Column({ type: 'text' })
   name: string;
+
+  @Column({ type: 'text', nullable: true })
+  icon: string;
```
AmarTrebinjac, line 33: should be icon?: string | null if its nullable

AmarTrebinjac (changes requested):
Good stuff!

Left a couple of minor comments :)

src/entity/BookmarkList.ts
ilasw: Nice catch, updated!

src/entity/Bookmark.ts
ilasw: we already have it on line 17 ✔️

src/schema/bookmarks.ts
ilasw: I found a [nice way to check if is an emoji](https://stackoverflow.com/questions/18862256/how-to-detect-emoji-using-javascript), but we need to change tsconfig target. 
Do you think this could be possibile?

src/entity/BookmarkList.ts
ilasw: Updated ✔️

__tests__/helpers.ts
```diff
@@ -43,7 +43,7 @@ export class MockContext extends Context {
 
   constructor(
     con: DataSource,
-    userId: string = null,
+    userId: string | null = null,
```
sshanzel, line 46: Surprised that the type was not inferred automatically 👀

src/schema/bookmarks.ts
```diff
@@ -385,20 +438,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-      ctx.loader.loadMany<BookmarkList>(
-        BookmarkList,
-        { userId: ctx.userId },
-        info,
-        {
-          order: { name: 'ASC' },
-        },
-      ),
+    bookmarkLists: (_, __, ctx: AuthContext): Promise<GQLBookmarkList[]> =>
+      ctx.con.getRepository(BookmarkList).find({
```
sshanzel, line 442: I think we should use the graphorm function here.

sshanzel (commented):
Looks amazing!

src/schema/bookmarks.ts
rebelchris: Yes, on reads it should always be the case.
Probably actually also can just read from RR here. 

CC @capJavert we talked about this during RFC stages, but missed it as well.

rebelchris (changes requested):
Overall neat, minor query with graphORM query

src/schema/bookmarks.ts
ilasw: Nvm used another way and seems working fine ✅

src/schema/bookmarks.ts
ilasw: Updated ✔️ 

@rebelchris where I can find a sample of using replica? 👀

capJavert: @ilasw look for `queryReadReplica` in code

rebelchris: https://github.com/dailydotdev/daily-api/blob/cc174bcecdad2cac05371d2e1be854337bc5aea5/src/schema/comments.ts#L736

Screenshot 2024-12-05 at 11 39 41

Last prop is readReplica :)

src/entity/Bookmark.ts
```diff
@@ -25,9 +25,9 @@ export class Bookmark {
   })
   post: Promise<Post>;
 
-  @ManyToOne(() => BookmarkList, {
+  @ManyToOne('BookmarkList', {
```
capJavert, line 28: Missing `import type { BookmarkList }` above

src/schema/bookmarks.ts
```diff
@@ -325,16 +347,49 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-      source,
-      { name }: { name: string },
+      _,
+      { name, icon }: Record<'name' | 'icon', string>,
       ctx: AuthContext,
-    ): Promise<GQLBookmarkList> =>
-      ctx.con.getRepository(BookmarkList).save({
+    ): Promise<GQLBookmarkList> => {
+      const onlyOneEmoji = /^(\p{Emoji_Presentation})$/gu;
+      const isValidIcon = !icon || onlyOneEmoji.test(icon);
```
capJavert, line 355: can we move this into helper, also to avoid computing regex each time.

capJavert (commented):
Nice, I think only read replica and we are good to go.

src/entity/Bookmark.ts
ilasw: Added ✔️

src/schema/bookmarks.ts
ilasw: Updated, thanks! ✔️

rebelchris (approved):
Cool nothing blocking from my side.
Nice on the emoji regex :)

src/common/utils.ts
```diff
@@ -176,6 +176,10 @@ export function isNumber(value: string | number): boolean {
   return !isNaN(Number(value.toString()));
 }
 
+export const isOneEmoji = (text: string): boolean => {
+  return /^(\p{Emoji_Presentation})$/gu.test(text);
+};
+
```
capJavert, line 182:
```suggestion
const emojiMatch = /^(\p{Emoji_Presentation})$/gu

export const isOneEmoji = (text: string): boolean => {
  return emojiMatch.test(text);
};

```

capJavert (approved):

PR #2500, opened 2024-12-04
feat: create query to fetch Smart Title / original title of post

omBratteng:
Created new query to fetch the Smart Title or the original title of a post, based on the settings of the user.

If the user is a free user, we allow them to query one time before we return a forbidden error.

Also added a new post to seeds that has alt title added

TODO:
- [x] Add tests

AS-803

### DIFF
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -631,7 +631,7 @@ export const getPostTranslatedTitle = (
   contentLanguage: ContentLanguage,
 ) =>
   post.contentMeta?.translate_title?.translations?.[contentLanguage] ||
-  post.title;
+  (post.title as string);
 
 export const getPostSmartTitle = (
   post: Partial<Pick<Post, 'title' | 'contentMeta'>>,
@@ -639,4 +639,4 @@ export const getPostSmartTitle = (
 ) =>
   post.contentMeta?.alt_title?.translations?.[contentLanguage] ||
   post.contentMeta?.alt_title?.translations?.[ContentLanguage.English] ||
-  post.title;
+  getPostTranslatedTitle(post, contentLanguage);
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -627,16 +627,16 @@ export const findPostImageFromContent = ({
 };
 
 export const getPostTranslatedTitle = (
-  post: Pick<Post, 'title' | 'contentMeta'>,
+  post: Partial<Pick<Post, 'title' | 'contentMeta'>>,
   contentLanguage: ContentLanguage,
 ) =>
   post.contentMeta?.translate_title?.translations?.[contentLanguage] ||
   post.title;
 
 export const getPostSmartTitle = (
-  post: Pick<Post, 'contentMeta'>,
-  contentLanguage: ContentLanguage,
+  post: Partial<Pick<Post, 'title' | 'contentMeta'>>,
+  contentLanguage?: ContentLanguage,
 ) =>
   post.contentMeta?.alt_title?.translations?.[
     contentLanguage ?? ContentLanguage.English
-  ];
+  ] || post.title;
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -635,10 +635,8 @@ export const getPostTranslatedTitle = (
 
 export const getPostSmartTitle = (
   post: Partial<Pick<Post, 'title' | 'contentMeta'>>,
-  contentLanguage?: ContentLanguage,
+  contentLanguage: ContentLanguage,
 ) =>
-  post.contentMeta?.alt_title?.translations?.[
-    contentLanguage ?? ContentLanguage.English
-  ] ||
+  post.contentMeta?.alt_title?.translations?.[contentLanguage] ||
   post.contentMeta?.alt_title?.translations?.[ContentLanguage.English] ||
   post.title;
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -32,6 +32,7 @@ import { downloadJsonFile } from './googleCloud';
 import {
   ContentLanguage,
   type ChangeObject,
+  type I18nRecord,
   type PostCodeSnippetJsonFile,
 } from '../types';
 import { uniqueifyObjectArray } from './utils';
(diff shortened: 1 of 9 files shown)


src/common/post.ts
```diff
@@ -621,3 +625,18 @@ export const findPostImageFromContent = ({
...
+  post.contentMeta?.translate_title?.translations?.[contentLanguage] ||
+  (post.title as string);
+
+export const getPostSmartTitle = (
+  post: Partial<Pick<Post, 'title' | 'contentMeta'>>,
+  contentLanguage: ContentLanguage,
+) =>
+  post.contentMeta?.alt_title?.translations?.[contentLanguage] ||
+  post.contentMeta?.alt_title?.translations?.[ContentLanguage.English] ||
+  getPostTranslatedTitle(post, contentLanguage);
```
omBratteng, line 642: Added future support for when/if we support smart titles in other languages. It first tries specific language, then english, and finally fallbacks to original title.

src/schema/posts.ts
```diff
@@ -1749,6 +1769,57 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         true,
       );
     },
+    fetchSmartTitle: async (
```
rebelchris, line 1772: Isn't it actually enough if this field exist on post graphql, you can just query POST buy only request set field?

src/entity/posts/Post.ts
```diff
@@ -259,7 +272,7 @@ export class Post {
   slug: string;
 
   @Column({ type: 'jsonb', default: {} })
-  contentMeta: object;
+  contentMeta: PostContentMeta;
```
capJavert, line 275: I did not want this typed on purpose since it has a lot more stuff that API does not use so this implies it has only above fields. With `object` it requires manual casting when used which is I think better in this case.

src/entity/user/UserAction.ts
```diff
@@ -10,6 +10,7 @@ export enum UserActionType {
   SquadInvite = 'squad_invite',
   MyFeed = 'my_feed',
   EditWelcomePost = 'edit_welcome_post',
+  FetchedSmartTitle = 'fetched_smart_title',
```
capJavert, line 13:
```suggestion
  FetchedSmartTitle = 'undicied_smart_title',
```

? 😆 🙈

src/paddle.ts
```diff
@@ -1,7 +1,25 @@
...
+export const isUserPlusMember = async (
+  con: DataSource,
+  userId?: string,
+): Promise<boolean> => {
+  const { subscriptionFlags } = await queryReadReplica(con, ({ queryRunner }) =>
+    queryRunner.manager.getRepository(User).findOneOrFail({
+      where: { id: userId },
+      select: ['subscriptionFlags'],
+    }),
+  );
```
capJavert, line 22: use props syntax `{ con, userId }`

src/schema/posts.ts
capJavert: It could be but it would be harder to do this logic inside the field.

src/schema/posts.ts
```diff
@@ -1749,6 +1769,57 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          .findOneBy({
+            userId: ctx.userId,
+            type: UserActionType.FetchedSmartTitle,
+          });
+
+        if (hasUsedFreeTrial) {
+          throw new ForbiddenError(
+            'Free trial has been used! Please upgrade to Plus to continue using this feature.',
+          );
+        } else {
```
capJavert, line 1798: no need to else since throw will kill the flow either way like early return

src/schema/posts.ts
capJavert: don't have preference on one or the other

src/schema/posts.ts
```diff
@@ -1749,6 +1769,57 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      }
+
+      const settings = await queryReadReplica(ctx.con, ({ queryRunner }) =>
+        queryRunner.manager.getRepository(Settings).findOne({
+          where: { userId: ctx.userId },
+          select: ['flags'],
+        }),
+      );
+
+      if (settings?.flags?.clickbaitShieldEnabled ?? true) {
```
capJavert, line 1813: Is it not supposed to be reverse, if its enabled then we return smart title?

src/schema/posts.ts
```diff
@@ -1749,6 +1769,57 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         true,
       );
     },
+    fetchSmartTitle: async (
+      _,
+      { id }: { id: string },
+      ctx: Context,
+    ): Promise<GQLPostSmartTitle> => {
+      const post: Pick<Post, 'title' | 'contentMeta'> = await queryReadReplica(
```
capJavert, line 1777: I think the whole handler can be wrapped inside of this because you always query from RR? It will also avoid multiple connections to the DB?

capJavert (commented):
Nice, nothing really blocking, just few ideas here or there.

src/entity/posts/Post.ts
omBratteng: 👍

src/paddle.ts
omBratteng: I might change this to use the isPlus from context I added in #2503

src/schema/posts.ts
omBratteng: I do one write, but I guess I could wrap it all inside RR and still call write like normal?

src/schema/posts.ts
omBratteng: 👍

src/paddle.ts
capJavert: That would be even better, nice 👌

src/schema/posts.ts
rebelchris: That's fine with me, if it's a speed/DX issue let's keep as is.

My initial concern was how to do the opposite side (already translated -> original)
but think we should include that either way for SEO, so let's keep this query.

src/schema/posts.ts
omBratteng: I realise the name might be a little bit confusing 🙈
We want to fetch the opposite, i.e. when you want to view the original title on a post

src/schema/posts.ts
capJavert: ah I see, no let's leave it as is then

src/entity/posts/Post.ts
omBratteng: Reverted to object and use type casting instead

src/paddle.ts
omBratteng: I'll merge this one first, then in #2503 I can it with isPlus from context

rebelchris (approved):

src/paddle.ts
omBratteng: Changed in 6e69905 in #2503

capJavert (approved):

PR #2501, opened 2024-12-04
feat: Bookmark add and move methods

ilasw:
- On bookmark creation: 
  - auto set last folder used from before (if any); 
  - now returns the just-created item;
- Rename `addBookmarkToList` endpoint to `moveBookmark`

### DIFF
--- a/src/schema/bookmarks.ts
+++ b/src/schema/bookmarks.ts
@@ -75,6 +75,7 @@ export const typeDefs = /* GraphQL */ `
   }
 
   type Bookmark {
+    postId: ID!
     list: BookmarkList
     createdAt: DateTime
     remindAt: DateTime
--- a/src/schema/bookmarks.ts
+++ b/src/schema/bookmarks.ts
@@ -22,9 +22,16 @@ import {
 import { SelectQueryBuilder } from 'typeorm';
 import { GQLPost } from './posts';
 import { Connection } from 'graphql-relay';
+import graphorm from '../graphorm';
 
 interface GQLAddBookmarkInput {
   postIds: string[];
+}
+
+export interface GQLBookmark {
+  createdAt: Date;
+  remindAt: Date;
+  list: GQLBookmarkList;
 }
 
 export interface GQLBookmarkList {
@@ -46,6 +53,7 @@ export const typeDefs = /* GraphQL */ `
   }
 
   type Bookmark {
+    list: BookmarkList
     createdAt: DateTime
     remindAt: DateTime
   }
@@ -84,7 +92,7 @@ export const typeDefs = /* GraphQL */ `
     """
     Add new bookmarks
     """
-    addBookmarks(data: AddBookmarkInput!): EmptyResponse! @auth
+    addBookmarks(data: AddBookmarkInput!): [Bookmark]! @auth
 
     """
     Add or move bookmark to list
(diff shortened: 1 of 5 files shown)


src/schema/bookmarks.ts
```diff
@@ -301,16 +326,14 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-        .onConflict(
-          `("postId", "userId") DO UPDATE SET "listId" = EXCLUDED."listId"`,
-        )
-        .execute();
+      await ctx.con.getRepository(Bookmark).update(
+        {
+          postId: id,
+          userId: ctx.userId,
+        },
+        { listId: listId ?? null },
```
sshanzel, line 334: Good catch. Otherwise, it won't be updated.

sshanzel (approved):
Looks amazing :shipit:

src/schema/bookmarks.ts
```diff
@@ -300,22 +307,39 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        where: { userId: ctx.userId },
+        select: ['listId'],
+      });
+      const lastUsedListId = lastAddedBookmark?.listId ?? null;
+
       const [query, params] = ctx.con
         .createQueryBuilder()
         .select('id', 'postId')
         .addSelect(`'${ctx.userId}'`, 'userId')
+        .addSelect(`:listId`, 'listId')
```
rebelchris, line 322: Why the weird select setting?
```suggestion
        .addSelect('listId')
```

src/schema/bookmarks.ts
```diff
@@ -300,22 +307,39 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
         .getQueryAndParameters();
+
       await ctx.con.query(
-        `insert into bookmark("postId", "userId") ${query} on conflict do nothing`,
+        `insert into bookmark("postId", "userId", "listId") ${query} on conflict do nothing`,
         params,
       );
-      return { _: true };
+
+      return await graphorm.query<GQLBookmark>(ctx, info, (builder) => ({
```
rebelchris, line 333: Why do we need the return data here again?
Just because we don't know where it's going to be. moved to?

rebelchris (approved):

src/schema/bookmarks.ts
capJavert: Agree with Chris, can we just return `listId` if its needed, we should be able to do:
```suggestion
     return {
        lastUsedListId
     }
```

capJavert: The reason why we would want to only return stuff we need or can compute is because of read replica. If we query bookmark after addition we need to query on master which is worse and adds another read to the master db.

ilasw: @capJavert @rebelchris We need the bookmark just added because on frontend we can change folder using toast action button. So we need the list.name and bookmark object for referencing it 👀

src/schema/bookmarks.ts
ilasw: Updated ✔️

src/schema/bookmarks.ts
```diff
@@ -300,22 +307,39 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
         .from(Post, 'post')
         .where('post.id IN (:...postIds)', { postIds: data.postIds })
+        .setParameter('listId', lastUsedListId)
         .getQueryAndParameters();
+
       await ctx.con.query(
-        `insert into bookmark("postId", "userId") ${query} on conflict do nothing`,
+        `insert into bookmark("postId", "userId", "listId") ${query} on conflict do nothing`,
         params,
       );
```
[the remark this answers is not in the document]
ilasw: I've tried to setup a different way but couldn't find an efficient way with typeorm 😞

src/schema/bookmarks.ts
capJavert: you know post ID on the frontend?

src/schema/bookmarks.ts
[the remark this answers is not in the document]
capJavert: @ilasw check a83aaf5832a94f0be8af8239a0dd0634c72fcdc3

I know its not a single query but I don't think we need to be as strict here with insert into and its much more readable and safe.

AmarTrebinjac:
How do we handle moving to the default folders "Read it later" and "Quick saves" ?

AmarTrebinjac (approved):
Looks good! 🚀

src/schema/bookmarks.ts
capJavert: @ilasw we can merge this to unblock but let's look on frontend side and try to avoid querying everything with regards to above comments. Then we should update this so it returns the computed data.

capJavert (approved):

src/schema/bookmarks.ts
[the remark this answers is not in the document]
ilasw: Nice, thanks! 🙌

PR #2502, opened 2024-12-04
feat: return `clickbaitTitleDetected` on posts

omBratteng:
Adds `clickbaitTitleDetected` field to posts

AS-806

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -1068,6 +1068,51 @@ describe('query post', () => {
       'FORBIDDEN',
     );
   });
+
+  describe('clickbaitTitleDetected', () => {
+    const LOCAL_QUERY = /* GraphQL */ `
+      query Post($id: ID!) {
+        post(id: $id) {
+          clickbaitTitleDetected
+        }
+      }
+    `;
+
+    it('should return true if clickbait title probability is above threshold', async () => {
+      await con.getRepository(ArticlePost).update('p1', {
+        contentQuality: { is_clickbait_probability: 1 },
+      });
+
+      const res = await client.query(LOCAL_QUERY, {
+        variables: { id: 'p1' },
+      });
+
+      expect(res.errors).toBeFalsy();
+      expect(res.data.post.clickbaitTitleDetected).toEqual(true);
+    });
+
+    it('should return false if clickbait title probability is below threshold', async () => {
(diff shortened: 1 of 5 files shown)


__tests__/posts.ts
```diff
@@ -1068,6 +1068,51 @@ describe('query post', () => {
...
+  describe('clickbaitTitleDetected', () => {
+    const LOCAL_QUERY = /* GraphQL */ `
+      query Post($id: ID!) {
+        post(id: $id) {
+          clickbaitTitleDetected
+        }
+      }
+    `;
+
+    it('should return true if clickbait title probability is above threshold', async () => {
```
rebelchris, line 1081: Can we maybe add a test where it's more mimicked on floats?
Just to ensure float comparison never breaks down the line?

rebelchris (changes requested):
Minor request on test, but nothing blocking.

src/graphorm/index.ts
```diff
@@ -272,6 +273,17 @@ const obj = new GraphORM({
         select: 'tagsStr',
         transform: (value: string): string[] => value?.split(',') ?? [],
       },
+      clickbaitTitleDetected: {
+        rawSelect: true,
+        select: `"contentQuality"->'is_clickbait_probability'`,
+        transform: (value: string): boolean => {
+          const probability = parseFloat(value || '0');
```
capJavert, line 280: Is JSON value not jsonb float/number value already?

capJavert: I guess just to be safe?

__tests__/posts.ts
capJavert: also since we use parseFload, can we also have test for string being there instead of number in JSON, just to be sure it also works

capJavert (approved):

src/graphorm/index.ts
omBratteng: Just to be safe

__tests__/posts.ts
omBratteng: Changed it to be `'1.99'`

capJavert: it could be two different tests, default which will most probably be float already, and separate one to handle the string case, better to detect what breaks

omBratteng: Added separate test for float

PR #2503, opened 2024-12-04
feat: return smart title on posts if user is plus and have clickbait shield enabled

omBratteng:
TODO:
- [ ] Add tests

AS-807

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -126,7 +126,7 @@ const plugin = async (
             req.premium = payload.premium;
             req.roles = payload.roles;
             req.isTeamMember = payload.isTeamMember;
-            req.isPlus = payload.isPlus;
+            req.isPlus = !!payload.isPlus;
             req.accessToken = {
               token: validValue,
               expiresIn: new Date(payload.exp * 1000),
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -19,6 +19,7 @@ interface AuthPayload {
   premium: boolean;
   roles?: Roles[];
   isTeamMember?: boolean;
+  isPlus?: boolean;
   exp: number;
 }
 
@@ -92,6 +93,7 @@ const plugin = async (
   fastify.decorateRequest('roles');
   fastify.decorateRequest('accessToken');
   fastify.decorateRequest('isTeamMember');
+  fastify.decorateRequest('isPlus');
 
   // Machine-to-machine authentication
   fastify.addHook('preHandler', async (req) => {
@@ -103,6 +105,7 @@ const plugin = async (
       if (req.headers['user-id'] && req.headers['logged-in'] === 'true') {
         req.userId = req.headers['user-id'] as string;
         req.premium = req.headers.premium === 'true';
+        req.isPlus = req.headers['is-plus'] === 'true';
         req.roles =
           ((req.headers['roles'] as string)?.split(',') as Roles[]) ?? [];
       }
@@ -123,6 +126,7 @@ const plugin = async (
             req.premium = payload.premium;
             req.roles = payload.roles;
             req.isTeamMember = payload.isTeamMember;
+            req.isPlus = payload.isPlus;
             req.accessToken = {
               token: validValue,
               expiresIn: new Date(payload.exp * 1000),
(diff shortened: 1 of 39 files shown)


src/graphorm/index.ts
```diff
@@ -247,10 +255,24 @@ const obj = new GraphORM({
...
+            userId: ctx.userId,
+          })
+          .addSelect(
+            `"s"."flags"->'clickbaitShieldEnabled'`,
+            'clickbaitShieldEnabled',
+          );
+      }
+
+      return qb;
+    },
```
omBratteng, line 275: FML tried for so long to figure out a way to select user settings from `createSmartTitleField` function... if only it had accepted async functions.

src/graphorm/index.ts
```diff
@@ -412,9 +439,8 @@ const obj = new GraphORM({
         alias: { field: 'url', type: 'string' },
         transform: (value: string): string => domainOnly(value),
       },
-      title: createI18nField({
+      title: createSmartTitleField({
```
rebelchris, line 442: This will return only as titlte right?
I think we need the addition of `fieldAs` here maybe?

Output should always include original title still as well.

rebelchris (changes requested):
Just one comment, but not sure without test, if the title is still included.
I think it should be output as:

```
title: Blabla
altTitle: clickbaitTitle || translatedTitle
```

This assumption good?

src/auth.ts
```diff
@@ -123,6 +126,7 @@ const plugin = async (
             req.premium = payload.premium;
             req.roles = payload.roles;
             req.isTeamMember = payload.isTeamMember;
+            req.isPlus = payload.isPlus;
```
capJavert, line 129:
```suggestion
            req.isPlus = !!payload.isPlus;
```

and then we can make it non optional in `AuthPayload`?

src/graphorm/index.ts
```diff
@@ -247,10 +255,24 @@ const obj = new GraphORM({
...
+      if (ctx.userId && ctx.isPlus) {
+        qb = qb
+          .innerJoin(Settings, 's', 's."userId" = :userId', {
+            userId: ctx.userId,
+          })
+          .addSelect(
+            `"s"."flags"->'clickbaitShieldEnabled'`,
+            'clickbaitShieldEnabled',
+          );
+      }
```
capJavert, line 272: This will do inner join for all plus user post queries, which could affect performance, I think we should query only when needed or expose it on context.

src/graphorm/index.ts
capJavert: I am not sure as well

omBratteng: We currently only return `title` even when they are translated.
During RFC meeting, we decided on doing it like translation, and replacing it in title.

So now when you have clickbait shield enabled, it will attempt to return the alternative title first, then falls back to translated title, and finally original title.

src/graphorm/index.ts
omBratteng: This was the only way I figured out how to do it in GraphORM 🙈

src/graphorm/index.ts
rebelchris: Ah but then we also said for OG/SEO purposes we should always have the original as well right?
Also when translating back from this it would be easier if it's just there.

My suggestion is we keep title as is, but have `altTitle` as secondary field.
OR title is always rendered title (what you do above), but we add originalTitle somewhere? (and modify only for SEO etc)?

omBratteng: For OG/SEO, it's accessed as anonymous user, so it will fallback to original title

omBratteng: As discussed, this is fine as OG/SEO is fetched as anon user, so it always returns original title.

src/graphorm/index.ts
omBratteng: As discussed, we'll leave it as is for now, and I'll circle back to this later to see if there's any improvements that can be done.

src/auth.ts
omBratteng: Made this non-optional

rebelchris (approved):
Good on my end

rebelchris (approved):

capJavert (approved):

PR #2504, opened 2024-12-06
fix: only return smart title if original title is detected as clickbait

omBratteng:


src/graphorm/index.ts
```diff
@@ -72,6 +72,14 @@ const nullIfNotSameUser = <T>(
   return ctx.userId === user.id ? value : null;
 };
 
+const checkIfTitleIsClickbait = (value?: string): boolean => {
+  const clickbaitProbability = parseFloat(value || '0');
+  const threshold = remoteConfig.vars.clickbaitTitleProbabilityThreshold || 1;
+  console.log('clickbaitProbability', clickbaitProbability);
```
rebelchris, line 78: Testing?

rebelchris, line 78:
```suggestion
```

rebelchris (approved):
Looks good.
Did we set clickbait shield by default on for plus?

src/graphorm/index.ts
omBratteng: Yup

omBratteng:
> Did we set clickbait shield by default on for plus?

Yeah

PR #2505, opened 2024-12-06
feat: move workers queries to read replica 1

capJavert:
Moved workers:
- digest generation
- all workers not needing real time data (upvotes, rep, notification with counts)

Did not move any worker that needs to write or live data as well. Some of them I implemented but in testing realized they would return stale data since RR would be out of date. 

We could move rest of the real time workers if we moved CDC to RR, see [here](https://dailydotdev.slack.com/archives/C03C3JZPRU4/p1733484456135089), but something to be evaluated separate.

### DIFF
--- a/src/common/feedGenerator.ts
+++ b/src/common/feedGenerator.ts
@@ -11,6 +11,7 @@ import {
   ObjectLiteral,
   QueryBuilder,
   SelectQueryBuilder,
+  type EntityManager,
 } from 'typeorm';
 import { Connection, ConnectionArguments } from 'graphql-relay';
 import { IFieldResolver } from '@graphql-tools/utils';
@@ -80,7 +81,7 @@ export const whereKeyword = (
 };
 
 const rawFilterSelect = <T extends ObjectLiteral>(
-  con: DataSource,
+  con: DataSource | EntityManager,
   name: string,
   func: (qb: QueryBuilder<T>) => QueryBuilder<T>,
 ): QueryBuilder<T> =>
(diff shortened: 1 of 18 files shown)


src/common/feedGenerator.ts
```diff
@@ -307,7 +308,7 @@ const sourcesToFilters = ({
 };
 
 export const feedToFilters = async (
-  con: DataSource,
+  con: DataSource | EntityManager,
```
capJavert, line 311: Main change for digest to work, in digest we inject `queryRunner.manager` from `queryReadReplica`.

src/common/personalizedDigest.ts
```diff
@@ -231,11 +232,13 @@ export const getPersonalizedDigestEmailPayload = async ({
   previousSendDate: Date;
   feature: PersonalizedDigestFeatureConfig;
 }): Promise<SendEmailRequestWithTemplate | undefined> => {
-  const feedConfig = await feedToFilters(
-    con,
-    personalizedDigest.userId,
-    personalizedDigest.userId,
-  );
+  const feedConfig = await await queryReadReplica(con, ({ queryRunner }) => {
```
rebelchris, line 235: Yeah here it should be fine, the feed one is different story I guess, this one can be async/outdated even.

rebelchris (approved):
Looks good to me.
Don't see any issues on any of these, but let's monitor after deployment for a bit. 

Nice we have this function 😂

src/common/personalizedDigest.ts
omBratteng, line 235:
```suggestion
  const feedConfig = await queryReadReplica(con, ({ queryRunner }) => {
```

omBratteng (approved):

PR #2506, opened 2024-12-06
refactor: use dataloader for storing user settings per request

omBratteng:
Should reduce the need for querying user settings too many times in post queries

AS-813

### DIFF
--- a/src/dataLoaderService.ts
+++ b/src/dataLoaderService.ts
@@ -1,7 +1,7 @@
 import DataLoader, { BatchLoadFn } from 'dataloader';
 import { Context } from './Context';
 import { getShortUrl } from './common';
-import { SourceMember } from './entity';
+import { Settings, SourceMember } from './entity';
 import { GQLSource } from './schema/sources';
 
 export const defaultCacheKeyFn = <K extends object | string>(key: K) => {
@@ -51,6 +51,15 @@ export class DataLoaderService {
     }
 
     return this.loaders[type] as DataLoader<K, V>;
+  }
+
+  get userSettings() {
+    return this.getLoader<{ userId: string }, Settings>({
+      type: 'userSettings',
+      loadFn: async ({ userId }) =>
+        this.ctx.con.getRepository(Settings).findOneBy({ userId }),
+      cacheKeyFn: ({ userId }) => defaultCacheKeyFn({ userId }),
+    });
   }
 
   get shortUrl() {
--- a/src/dataLoaderService.ts
+++ b/src/dataLoaderService.ts
@@ -56,8 +56,13 @@ export class DataLoaderService {
   get userSettings() {
     return this.getLoader<{ userId: string }, Settings>({
       type: 'userSettings',
-      loadFn: async ({ userId }) =>
-        this.ctx.con.getRepository(Settings).findOneBy({ userId }),
+      loadFn: async ({ userId }) => {
+        if (!userId) {
+          return null;
+        }
+
+        return this.ctx.con.getRepository(Settings).findOneBy({ userId });
+      },
       cacheKeyFn: ({ userId }) => defaultCacheKeyFn({ userId }),
     });
   }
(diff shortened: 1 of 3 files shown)


src/schema/posts.ts
```diff
@@ -1783,60 +1783,58 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       _,
       { id }: { id: string },
       ctx: Context,
-    ): Promise<GQLPostSmartTitle> => {
-      const post: Pick<Post, 'title' | 'contentMeta'> = await queryReadReplica(
-        ctx.con,
-        ({ queryRunner }) =>
-          queryRunner.manager.getRepository(Post).findOneOrFail({
+    ): Promise<GQLPostSmartTitle> =>
+      queryReadReplica(ctx.con, async ({ queryRunner }) => {
```
omBratteng, line 1787: Forgot to this in my previous PR

rebelchris: Only change is RR?

omBratteng: Yeah

omBratteng: Guess looking at diff is difficult because of indentation and prettier doing things :triggered:

rebelchris (approved):
Almost forgot we hade them.
Do wait for Ante to check as well please

src/graphorm/index.ts
```diff
@@ -82,21 +81,24 @@ const checkIfTitleIsClickbait = (value?: string): boolean => {
...
-        clickbaitShieldEnabled?: boolean;
+        i18nTitle: I18nRecord;
+        smartTitle: I18nRecord;
         clickbaitProbability?: string;
         [key: string]: unknown;
       };
 
+      const settings = await ctx.dataLoader.userSettings.load({
+        userId: ctx.userId!,
+      });
```
capJavert, line 94: It would be awesome if we could also reuse this in graphorm, but I guess that would require fiddling more on graphorm side to replace queries there with data loader result. But I think this should be better overall.

src/dataLoaderService.ts
```diff
@@ -53,6 +53,15 @@ export class DataLoaderService {
     return this.loaders[type] as DataLoader<K, V>;
   }
 
+  get userSettings() {
+    return this.getLoader<{ userId: string }, Settings>({
+      type: 'userSettings',
+      loadFn: async ({ userId }) =>
+        this.ctx.con.getRepository(Settings).findOneBy({ userId }),
```
capJavert, line 60: should we return `null` if `userId`is undefined for any reason, just thinking if mistake happens it is going to query all user settings because of typeorm..

capJavert:
```suggestion
        if (!userId) {
            return null
        }

        this.ctx.con.getRepository(Settings).findOneBy({ userId }),
```

capJavert (approved):
Looks good! 🚀

src/dataLoaderService.ts
omBratteng: Returning null if no userId

src/graphorm/index.ts
idoshamun: What would be ideal case in your opinion? graphorm is a translation layer from graphql to pgsql.
not sure how dataloader fits to this picture

capJavert: if we could make hints for reusable entities so that graphorm can maybe fetch those through data loader instead of nested sub query

Another example for source:
- we fetch feed for source X
- in that response we nested query source for each post (it will be the same source X)
- we return the response

With data loader:
- we fetch feed for source X
- we hint to the graphorm that source should be fetched through data loader
- we execute the query in parallel with dataloader requesting source
- we merge and reuse data loader response in place of all source sub queries

It could be a nice caching layer inside graphorm when we have a lot of nested fields and complexity of resolving those rises.

Something like that was on my mind.

PR #2507, opened 2024-12-06
refactor: bookmarks feed query with list params

sshanzel:


__tests__/__snapshots__/bookmarks.ts.snap
```diff
@@ -91,51 +97,15 @@ Object {
 }
 `;
 
-exports[`query bookmarks should return bookmarks from list 1`] = `
```
sshanzel, line 94: Removed some snapshots and converted them to look into actual values.

__tests__/bookmarks.ts
```diff
@@ -544,6 +562,153 @@ describe('query bookmarks', () => {
     expect(res.data).toMatchSnapshot();
   });
 
+  describe('plus user', () => {
+    beforeEach(async () => {
+      loggedUser = '5';
+      isPlus = true;
+      await con.getRepository(Settings).save({ userId: loggedUser });
```
sshanzel, line 569: Basically when testing plus users, users need to have a settings. Otherwise, the query might return values that are not expected :hide_the_pain: - spent some good hours on this.

src/common/feedGenerator.ts
```diff
@@ -670,14 +700,44 @@ export const bookmarksFeedBuilder = (
...
+  const [firstFolderSubQuery, params] = getFirstFolderSubQuery({
+    builder: newBuilder,
+    userId: ctx.userId!,
+  });
+
+  if (ctx.isPlus) {
+    if (listId) {
+      newBuilder = newBuilder.andWhere('bookmark.listId = :listId', { listId });
+    } else {
+      newBuilder = newBuilder.andWhere('bookmark.listId IS NULL');
```
rebelchris, line 719: Doesn't this battle the query bellow actually?

rebelchris (commented):
More of a question.

src/common/feedGenerator.ts
```diff
@@ -654,14 +655,43 @@ export const configuredFeedBuilder = (
...
+  query?: string | null;
+}
+
+interface GetFirstFolderSubQuery {
+  builder: SelectQueryBuilder<Post>;
+  userId: string;
+  alias?: string;
+}
+
+const getFirstFolderSubQuery = ({
```
rebelchris, line 673: This will execute on a loop right?
Maybe also good opportunity for dataLoader? not urgent imo, but nice addition.

capJavert: We can't since this is a query, not a function to fetch data.

capJavert: Also let's move this to `common` since it not directly related to `feedGenerator`

src/common/feedGenerator.ts
```diff
@@ -670,14 +700,44 @@ export const bookmarksFeedBuilder = (
...
+      newBuilder = newBuilder.andWhere('bookmark.listId IS NULL');
+    }
   }
+
+  if (!ctx.isPlus) {
+    if (listId) {
+      // Check if the list id is the first folder.
+      // This is when the user unsubscribed and trying to access locked folders.
+      newBuilder = newBuilder.andWhere(
+        `bookmark."listId" = (${firstFolderSubQuery})`,
```
capJavert, line 728: Can we query that `listId` once on top and then pass it here, feels like sub query will be inefficient on each row? Similar to what Chris said here https://github.com/dailydotdev/daily-api/pull/2507/files#r1875522680

src/common/feedGenerator.ts
```diff
@@ -670,14 +700,44 @@ export const bookmarksFeedBuilder = (
...
+        params,
+      );
+    } else {
+      // Get everything except the first folder
+      // This returns the bookmarked posts from the locked folders but as part of Quick Saves
+      newBuilder = newBuilder.andWhere(
+        `(bookmark."listId" IS NULL OR bookmark."listId" IS DISTINCT FROM (${firstFolderSubQuery}))`,
```
capJavert, line 735: Also here.

capJavert (commented):
Think it would be great if we could replace `firstFolderSubQuery` with static value which is fetched once.

src/common/feedGenerator.ts
sshanzel: Should be not. This one was wrapped with isPlus and the other one is the opposite 🙏

src/common/feedGenerator.ts
sshanzel: Yes, tried to fetch the ID to make it simpler, but it was a sync function, we'll have to always fetch the first folder from the outside if we want, though there can be cases where we will not use it.

src/common/feedGenerator.ts
sshanzel: I'm open to either of the two, feel free to let me know on what you think is best.

src/common/feedGenerator.ts
sshanzel: Updated it to make the conditions simpler 🥳

sshanzel:
Also added the filter to return bookmarks with reminder only.

__tests__/bookmarks.ts
```diff
@@ -568,56 +573,235 @@ describe('mutation moveBookmark', () => {
...
+    });
+
+    it('should return bookmarks without list id only as plus user', async () => {
+      await saveBookmarkFixtures();
+      const list = await con
+        .getRepository(BookmarkList)
+        .save({ userId: loggedUser, name: 'Test' });
+      await con
+        .getRepository(Bookmark)
+        .update({ userId: loggedUser, postId: 'p3' }, { listId: list.id });
```
rebelchris, line 654: Many updates in this file, might cause you headaches.
It might be better to store individual unique ones for this test, but non blocking.

Also noticed you never wait for nocks, might also cause some flakiness down the line (just noting for if it happens)

src/common/feedGenerator.ts
```diff
@@ -671,14 +684,39 @@ export const bookmarksFeedBuilder = (
...
+
   if (unreadOnly) {
     newBuilder = newBuilder.andWhere((subBuilder) =>
       whereUnread(ctx.userId, subBuilder, alias),
     );
   }
-  if (listId && ctx.premium) {
-    newBuilder = newBuilder.andWhere('bookmark.listId = :listId', { listId });
+
+  if (reminderOnly) {
```
rebelchris, line 694: Lot more readable like this 👏

src/schema/bookmarks.ts
```diff
@@ -295,10 +302,24 @@ const applyBookmarkPaging = (
 const searchResolver = feedResolver(
```
rebelchris, line 302: Where does this get used from?

src/schema/compatibility.ts
```diff
@@ -214,15 +215,21 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-    bookmarks: compatFeedResolver((ctx, args, opts, builder, alias) =>
-      bookmarksFeedBuilder(
-        ctx,
-        false,
-        null,
-        builder.orderBy('bookmark.createdAt', 'DESC'),
-        alias,
-      ),
-    ),
+    bookmarks: async (source, args, context: AuthContext, info) => {
```
rebelchris, line 218: What is this even used for? 😅

rebelchris (approved):
Non blocking but some questions.

src/schema/bookmarks.ts
sshanzel: When searching bookmarks I believe.

src/schema/compatibility.ts
sshanzel: Not entirely sure, but I think it is for backwards compat for way waaay back.

src/common/bookmarks.ts
```diff
@@ -0,0 +1,14 @@
+import { AuthContext } from '../Context';
+import { BookmarkList } from '../entity';
+
+export const getFirstFolderId = async (
+  ctx: AuthContext,
+): Promise<string | null | undefined> => {
```
capJavert, line 6:
```suggestion
): Promise<string | undefined> => {
```

capJavert (approved):

PR #2508, opened 2024-12-06
fix: require alt title for clickbaitTitleDetected: true

omBratteng:
Fixes an edge case when we've detected the title to be clickbait, but we do not have an alternative title for it, so that we don't show the clickbait shield in frontend.

### DIFF
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -644,14 +644,20 @@ export const getPostTranslatedTitle = (
     contentLanguage
   ] || (post.title as string);
 
+export const getSmartTitle = (
+  contentLanguage: ContentLanguage,
+  translations?: I18nRecord,
+): string | undefined => {
+  return (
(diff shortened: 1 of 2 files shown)


src/common/post.ts
```diff
@@ -644,14 +644,20 @@ export const getPostTranslatedTitle = (
     contentLanguage
   ] || (post.title as string);
 
+export const getSmartTitle = (
+  contentLanguage: ContentLanguage,
+  translations?: I18nRecord,
+): string | undefined => {
+  return (
+    translations?.[contentLanguage!] ?? translations?.[ContentLanguage.English]
```
capJavert, line 652: why `!`?

capJavert (commented):
Can we add a test for this please?

src/common/post.ts
omBratteng: Leftover from testing something else

rebelchris (approved):

capJavert (approved):

PR #2511, opened 2024-12-09
feat: add admin option to manually protect posts with clickbait shield

omBratteng:
https://github.com/dailydotdev/apps/pull/3947

AS-822

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -100,11 +100,7 @@ import { generateStorageKey, StorageTopic } from '../config';
 import { subDays } from 'date-fns';
 import { ReportReason } from '../entity/common';
 import { reportPost, saveHiddenPost } from '../common/reporting';
-import {
-  MANUAL_CLICKBAIT_VALUE,
-  PostCodeSnippetLanguage,
-  UserVote,
-} from '../types';
+import { PostCodeSnippetLanguage, UserVote } from '../types';
 import { PostCodeSnippet } from '../entity/posts/PostCodeSnippet';
 import {
   SourcePostModeration,
@@ -2233,28 +2229,23 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
           select: ['contentQuality'],
         });
 
-      const is_clickbait_probability = parseFloat(
+      const clickbaitProbability = parseFloat(
         (contentQuality.is_clickbait_probability as unknown as string) || '0.0',
       );
       const clickbaitTitleProbabilityThreshold =
         remoteConfig.vars.clickbaitTitleProbabilityThreshold || 1.0;
 
-      // We +/- 10 as it is a number that is above the natural range from Yggdrasil,
-      // so it will be easy to identify if the value was changed by a user.
-      const newProbability =
-        is_clickbait_probability > clickbaitTitleProbabilityThreshold
-          ? is_clickbait_probability - MANUAL_CLICKBAIT_VALUE
-          : is_clickbait_probability + MANUAL_CLICKBAIT_VALUE;
-
-      const isManual = newProbability > 1.0 || newProbability < 0.0;
+      if ('manual_clickbait_probability' in contentQuality) {
+        delete contentQuality.manual_clickbait_probability;
+      } else {
+        contentQuality.manual_clickbait_probability =
+          clickbaitProbability > clickbaitTitleProbabilityThreshold ? 0 : 1;
+      }
 
       await ctx.con.getRepository(Post).update(
         { id },
         {
-          contentQuality: {
-            is_clickbait_probability: newProbability,
-            manual_clickbait_probability: isManual,
-          },
+          contentQuality,
         },
       );
     },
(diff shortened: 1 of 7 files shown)


rebelchris (approved):

src/schema/posts.ts
```diff
@@ -2215,6 +2221,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+    clickbaitPost: async (_, { id }: { id: string }, ctx: AuthContext) => {
+      const { contentQuality }: Pick<Post, 'contentQuality'> = await ctx.con
+        .getRepository(Post)
+        .findOneOrFail({
+          where: { id },
+          select: ['contentQuality'],
+        });
+
+      const is_clickbait_probability =
+        contentQuality.is_clickbait_probability || 0.0;
```
capJavert, line 2233: parseFloat?

src/schema/posts.ts
```diff
@@ -2215,6 +2221,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        is_clickbait_probability > clickbaitTitleProbabilityThreshold
+          ? is_clickbait_probability - 10.0
+          : is_clickbait_probability + 10.0;
+
+      await ctx.con.getRepository(Post).update(
+        { id },
+        {
+          contentQuality: {
+            is_clickbait_probability: newProbability,
+          },
```
capJavert, line 2249: Check with Denis if Yggdrasil can send this again with post update which will overwrite the set value, we might need to update `postUpdated` to ignore yggdrasil value in there?

src/schema/posts.ts
```diff
@@ -2215,6 +2221,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const is_clickbait_probability =
+        contentQuality.is_clickbait_probability || 0.0;
+      const clickbaitTitleProbabilityThreshold =
+        remoteConfig.vars.clickbaitTitleProbabilityThreshold || 1.0;
+
+      // We +/- 10 as it is a number that is above the natural range from Yggdrasil,
+      // so it will be easy to identify if the value was changed by a user.
+      const newProbability =
+        is_clickbait_probability > clickbaitTitleProbabilityThreshold
+          ? is_clickbait_probability - 10.0
```
capJavert, line 2241: we can make `10.0` a constant in root `types.ts`

capJavert (approved):
Looks good, some edge cases to verify.

src/schema/posts.ts
omBratteng: cc @denisb0

src/schema/posts.ts
omBratteng: Added

src/schema/posts.ts
denisb0: Yes, any values sent with published content, can be sent with update. From Yggdrasil side there is only publish action, no difference between first publish and subsequent update.
It very rarely happens in the course of normal operation, but backfilling the data may be the issues.

omBratteng: I added a flag `manual_clickbait_probability` that worker checks for and if exists, keeps value that is set in database.

rebelchris (approved):

src/schema/posts.ts
```diff
@@ -2215,6 +2225,39 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        });
+
+      const is_clickbait_probability = parseFloat(
+        (contentQuality.is_clickbait_probability as unknown as string) || '0.0',
+      );
+      const clickbaitTitleProbabilityThreshold =
+        remoteConfig.vars.clickbaitTitleProbabilityThreshold || 1.0;
+
+      // We +/- 10 as it is a number that is above the natural range from Yggdrasil,
+      // so it will be easy to identify if the value was changed by a user.
```
capJavert, line 2243: Do we need values outside of `0..1` now that we have `manual_clickbait_probability`?

omBratteng: I guess we could check for either manual_clickbait_probability or is_clickbait_probability, and just leave is_clickbait_probability alone

omBratteng: To set it to be either above or below threshold

capJavert: I don't understand, we can just set to `1` if moderator sets as clickbait?

omBratteng: Yeah, but should also be able to set as not clickbait/ revert to original. 
So if we just keep the `manual_clickbait_probability` for post updated, we don't need to add logic to title field to care about the `manual_clickbait_probability` flag.

capJavert: Could `manual_clickbait_probability` then also just become 0-1 and if its set we use it, if not we use `clickbait_probability` from yggdrasil? Then reverting is just a matter of clearing that one field and then we use whatever yggdrasil set the last time?

omBratteng: Refactored it so that it prefers `manual_clickbait_probability` when set. And mutation will remove flag if it is set, so to revert to original value.

PR #2513, opened 2024-12-10
chore: remove lofn sizes

rebelchris:
We no longer support this from Lofn side.
Should do close to Lofn deployment I think.

### DIFF
--- a/src/integrations/feed/configs.ts
+++ b/src/integrations/feed/configs.ts
@@ -226,19 +226,8 @@ export class FeedLofnConfigGenerator implements FeedConfigGenerator {
           this.feedPreferencesConfigGenerator.generate(ctx, opts),
         ]);
 
-        const configWithoutPagination = lofnConfig.config as Omit<
-          typeof lofnConfig.config,
-          'page_size' | 'total_pages'
-        > & {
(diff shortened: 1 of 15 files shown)


src/integrations/feed/configs.ts
```diff
@@ -226,19 +226,8 @@ export class FeedLofnConfigGenerator implements FeedConfigGenerator {
...
-        const configWithoutPagination = lofnConfig.config as Omit<
-          typeof lofnConfig.config,
-          'page_size' | 'total_pages'
-        > & {
-          page_size?: number;
-          total_pages?: number;
-        };
-
-        delete configWithoutPagination.page_size;
-        delete configWithoutPagination.total_pages;
```
capJavert, line 238: Just to make sure, these don't exist on lofn anymore so we don't have to delete them?

capJavert (approved):

src/integrations/feed/configs.ts
rebelchris: Yeah @pnvasanth  is removing it here:
https://github.com/dailydotdev/lofn/pull/77

capJavert (approved):

PR #2515, opened 2024-12-10
fix: some icons don't pass folder validation

ilasw:
Looks like the regex didn't work for some icons. 
Since we are allowing only a strict amount of emojis changes the check for this subset.


capJavert (approved):

rebelchris (approved):
Good for now :D

AmarTrebinjac (approved):

PR #2516, opened 2024-12-10
feat: query single bookmarks folder by id

sshanzel:


src/schema/bookmarks.ts
```diff
@@ -517,6 +522,16 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
     },
+    bookmarkList: async (
+      _,
+      { id }: { id: string },
+      ctx: AuthContext,
+    ): Promise<GQLBookmarkList> => {
+      return ctx.con.getRepository(BookmarkList).findOneByOrFail({
+        id,
+        userId: ctx.userId,
+      });
```
capJavert, line 533: What do we use this query for, can we use read replica?

capJavert (approved):

AmarTrebinjac (approved):

src/schema/bookmarks.ts
sshanzel: Just for the static props to get the relevant entity based on the query param. Definitely we can use read replica. Let me check some examples.

rebelchris (approved):

src/schema/bookmarks.ts
```diff
@@ -517,6 +523,17 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
       return resolver(source, args, ctx, info);
     },
+    bookmarkList: async (
+      _,
+      { id }: { id: string },
+      ctx: AuthContext,
+    ): Promise<GQLBookmarkList> =>
+      queryReadReplica(ctx.con, async ({ queryRunner }) =>
```
capJavert, line 531: maybe use graphorm better, see https://github.com/dailydotdev/daily-api/pull/2516#discussion_r1877957934

src/schema/bookmarks.ts
sshanzel: Do you have some samples for this, can't seem to make it work with graphorm.

sshanzel: Screenshot 2024-12-10 at 7 57 03 PM

sshanzel: Does this mean I will just pass the `ctx` and not use the `queryRunner` completely?

PR #2517, opened 2024-12-10
feat: cancel subscription when user deletes account

rebelchris:
On deletion of user we should cancel their plus subscription... sad I know 😂

Log:
Screenshot 2024-12-10 at 13 33 28

Cancel on sandbox:
Screenshot 2024-12-10 at 13 33 52

AS-829 #done

### DIFF
--- a/src/common/paddle.ts
+++ b/src/common/paddle.ts
@@ -1,0 +1,38 @@
+import {
+  type Environment,
+  Paddle,
+  type Subscription,
+} from '@paddle/paddle-node-sdk';
+import { logger } from '../logger';
+
+export const paddleInstance = new Paddle(process.env.PADDLE_API_KEY, {
+  environment: process.env.PADDLE_ENVIRONMENT as Environment,
+});
+
+type CancelSubscriptionProps = {
+  subscriptionId: string;
+};
+export const cancelSubscription = ({
+  subscriptionId,
+}: CancelSubscriptionProps): Promise<Subscription> => {
+  logger.info(
+    {
+      type: 'paddle',
+      subscriptionId,
+    },
+    'Subscription cancelled user deletion',
+  );
+  try {
+    return paddleInstance.subscriptions.cancel(subscriptionId, {
+      effectiveFrom: null,
+    });
+  } catch (e) {
+    logger.error(
+      {
+        type: 'paddle',
+        subscriptionId,
+      },
+      'Subscription cancellation failed',
+    );
+  }
+};
--- a/src/directive/user.ts
+++ b/src/directive/user.ts
@@ -17,6 +17,7 @@ import {
   View,
 } from '../entity';
 import { ghostUser } from '../common';
+import { cancelSubscription } from '../common/paddle';
 
 export const deleteUser = async (
   con: DataSource,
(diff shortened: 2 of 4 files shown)


src/common/paddle.ts
```diff
@@ -0,0 +1,38 @@
...
+export const paddleInstance = new Paddle(process.env.PADDLE_API_KEY, {
+  environment: process.env.PADDLE_ENVIRONMENT as Environment,
+});
+
+type CancelSubscriptionProps = {
+  subscriptionId: string;
+};
+export const cancelSubscription = ({
+  subscriptionId,
+}: CancelSubscriptionProps): Promise<Subscription> | undefined => {
```
capJavert, line 17:
```suggestion
export const cancelSubscription = async ({
  subscriptionId,
}: CancelSubscriptionProps): Promise<Subscription | undefined> => {
```

src/common/paddle.ts
```diff
@@ -0,0 +1,38 @@
...
+  logger.info(
+    {
+      type: 'paddle',
+      subscriptionId,
+    },
+    'Subscription cancelled user deletion',
```
capJavert, line 23: It's not really true, we could use this util for other cases, so ideally this log is in the directive where we call `cancelSubscription` on user deletion

src/common/paddle.ts
```diff
@@ -0,0 +1,38 @@
...
+  );
+  try {
+    return paddleInstance.subscriptions.cancel(subscriptionId, {
```
capJavert, line 26:
```suggestion
    return await paddleInstance.subscriptions.cancel(subscriptionId, {
```

try catch will not catch error from promise that is not awaited

src/common/paddle.ts
rebelchris: Good point :)

src/common/paddle.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+export const cancelSubscription = async ({
+  subscriptionId,
+}: CancelSubscriptionProps): Promise<Subscription | undefined> => {
+  try {
+    return await paddleInstance.subscriptions.cancel(subscriptionId, {
+      effectiveFrom: null,
+    });
+  } catch (e) {
+    logger.error(
```
capJavert, line 23: just last one, maybe log original `e.message` as well?

capJavert (approved):

src/directive/user.ts
```diff
@@ -26,6 +28,23 @@ export const deleteUser = async (
...
+        });
+        logger.info(
+          {
+            type: 'paddle',
+            userId,
+            subscriptionId: user.subscriptionFlags.subscriptionId,
+          },
+          'Subscription cancelled user deletion',
+        );
+      }
```
idoshamun, line 47: Ideally it should be before the transaction. I don't want to hold the transaction hostage while we wait for paddle

src/common/paddle.ts
```diff
@@ -0,0 +1,31 @@
...
+  subscriptionId: string;
+};
+export const cancelSubscription = async ({
+  subscriptionId,
+}: CancelSubscriptionProps): Promise<Subscription | undefined> => {
+  try {
+    return await paddleInstance.subscriptions.cancel(subscriptionId, {
+      effectiveFrom: null,
+    });
+  } catch (e) {
```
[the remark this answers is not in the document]
rebelchris: Because user requested deletion, seemed wrong to "not to it" because let's say Paddle is down?
That why I added log so we can always retro do it?

But if you want we can throw, it will just be unclear on frontend why they didn't delete?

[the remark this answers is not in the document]
idoshamun: yes it's better to stop the deletion, it's easy to forget to check these logs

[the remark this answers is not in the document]
rebelchris: Alright, maybe we should then consider (non blocking) to show some toast or something if it fails.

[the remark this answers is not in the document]
rebelchris: I'll throw manually, so we can keep the log anyway for digesting :)

idoshamun (approved):

PR #2519, opened 2024-12-11
feat: limit reputation events per user per day

idoshamun:
Also remove the version property in docker compose as it's legacy

### DIFF
--- a/src/entity/ReputationEvent.ts
+++ b/src/entity/ReputationEvent.ts
@@ -4,7 +4,10 @@ import {
   PrimaryColumn,
   Column,
   BeforeInsert,
+  EntityManager,
+  MoreThan,
 } from 'typeorm';
+import { subDays } from 'date-fns';
 import type { User } from './user';
 
 export enum ReputationReason {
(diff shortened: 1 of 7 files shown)


src/entity/ReputationEvent.ts
```diff
@@ -42,10 +45,21 @@ export const REPUTATION_THRESHOLD = parseInt(
   process.env.REPUTATION_THRESHOLD || '250',
 );
 
-export const canGrantReputation = (
-  grantBy: Pick<User, 'reputation' | 'flags'>,
-): boolean => {
-  return grantBy.reputation >= REPUTATION_THRESHOLD && !grantBy.flags?.vordr;
+const GRANTS_PER_DAY = parseInt(process.env.REPUTATION_GRANTS_PER_DAY || '5');
```
omBratteng, line 48: Remote config maybe?

idoshamun: I don't expect to change it. I was tempted to set it hardcoded to 5 😅

PR #2520, opened 2024-12-11
feat: add az sorting in folders

ilasw:


rebelchris (approved):

src/schema/bookmarks.ts
```diff
@@ -555,12 +555,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-          queryBuilder: builder.queryBuilder.where(
-            `"${builder.alias}"."userId" = :userId`,
-            {
+          queryBuilder: builder.queryBuilder
+            .where(`"${builder.alias}"."userId" = :userId`, {
               userId: ctx.userId,
-            },
-          ),
+            })
+            .addOrderBy(`LOWER("${builder.alias}"."name")`, 'ASC'),
```
capJavert, line 562: we probably need order by index with lower here

capJavert: can you also add test for this

ilasw: Seems that [we cannot add lower to index](https://github.com/typeorm/typeorm/issues/327), but maybe this will not be too heavy as query because we have only few folders per user, not?

I'm going to add the test case still ✔️

capJavert: @ilasw you can add the index manually through migration and say `syncronize: false`, we have some examples already

rebelchris: Yeah we have some indeed:
https://github.com/dailydotdev/daily-api/blob/98947bf22e7ccd227bd5d86110f21744346ab4b3/src/migration/1707983477045-NewIndexes.ts#L16-L18

ilasw: Thanks for suggestion, added ✅

sshanzel (approved):
🚀

src/schema/bookmarks.ts
capJavert: Made it a bit more specific and `@Index` on entity. Good to go!

capJavert (approved):

src/migration/1733918463705-BookmarkListLowerIndex.ts
```diff
@@ -0,0 +1,14 @@
...
+export class BookmarkListLowerIndex1733918463705 implements MigrationInterface {
+    name = 'BookmarkListLowerIndex1733918463705'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(
+        `CREATE INDEX "bookmark_list_idx_lowername_asc" ON "bookmark_list" (LOWER(name) ASC)`,
+      );
+    }
+    public async down(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`DROP INDEX "bookmark_list_idx_lowername_asc"`);
```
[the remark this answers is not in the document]
omBratteng:
```suggestion
      await queryRunner.query(`DROP INDEX IF EXISTS "bookmark_list_idx_lowername_asc"`);
```

PR #2521, opened 2024-12-11
fix: re-issue JWT when user is upgraded to plus

omBratteng:
Fixes edge cases where user has upgrade to plus, but can not use the `fetchSmartTitle` on posts because JWT is outdated.


rebelchris (approved):
Works for me :)

PR #2523, opened 2024-12-11
fix: thread notifications

idoshamun:
Send notifications to mentioned users even if they are part of the thread since we send a reply notification only to the parent commenter. We also respect mute settings now

MI-604 #done

### DIFF
--- a/__tests__/workers/notifications/commentMention.ts
+++ b/__tests__/workers/notifications/commentMention.ts
@@ -6,6 +6,7 @@ import createOrGetConnection from '../../../src/db';
 import {
   Comment,
   CommentMention,
+  NotificationPreferenceComment,
   Post,
   Source,
   User,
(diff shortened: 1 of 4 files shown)


src/workers/notifications/commentReply.ts
```diff
@@ -47,6 +45,7 @@ const worker: NotificationWorker = {
       .findBy({
         referenceId: In([comment.id, parent.id]),
         notificationType: In(commentReplyNotificationTypes),
+        status: NotificationPreferenceStatus.Muted,
```
capJavert, line 48: Hmm curious how we missed this?

__tests__/workers/notifications/commentMention.ts
```diff
@@ -71,4 +91,81 @@ describe('commentMention', () => {
...
+
+    const result = await invokeNotificationWorker(
+      worker,
+      payload as unknown as Record<string, unknown>,
+    );
+
+    expect(result).toBeUndefined();
+  });
+
+  it('should not send notification when the author is mentioned', async () => {
```
capJavert, line 112: we don't send when author is mentioned because he will receive it either was as new comment notification?

capJavert (approved):

__tests__/workers/notifications/commentMention.ts
idoshamun: yes correct

src/workers/notifications/commentReply.ts
idoshamun: you can't really subscribe to comments, so no harm done

PR #2525, opened 2024-12-11
fix: notification being sent on vordred submission

sshanzel:
Also fixed the type to avoid confusion during the CDC part. For notification workers, I'd rather be explicit with the type than make mistakes 🙏

### DIFF
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -1,4 +1,4 @@
-import { PostType, FreeformPost } from '../entity';
+import { PostType, FreeformPost, KeywordFlags } from '../entity';
 import { NotificationBuilder } from './builder';
 import { NotificationIcon } from './icons';
 import {
@@ -131,7 +131,8 @@ export const notificationTitleMap: Record<
     return `New post from <b>${userName}</b>, check it out now!`;
   },
   user_given_top_reader: (ctx: NotificationUserTopReaderContext) => {
-    const keyword = ctx.keyword.flags?.title || ctx.keyword.value;
+    const keyword =
+      (ctx.keyword.flags as KeywordFlags)?.title || ctx.keyword.value;
     return `Great news! You've earned the top reader badge in ${keyword}.`;
   },
   source_post_approved: (ctx: NotificationPostContext) =>
--- a/src/types.ts
+++ b/src/types.ts
@@ -102,7 +102,11 @@ export type ChangeObject<Type> = {
   [Property in keyof Type as Exclude<
     Property,
     Required<Type>[Property] extends IgnoredTypes ? Property : never
-  >]: Required<Type>[Property] extends Date ? number : Type[Property];
+  >]: Required<Type>[Property] extends Date
+    ? number
+    : Required<Type>[Property] extends Record<string | number | symbol, unknown>
+      ? string
+      : Type[Property];
 };
 
 export type ChangeSchema = {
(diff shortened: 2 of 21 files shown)


src/types.ts
```diff
@@ -102,7 +102,11 @@ export type ChangeObject<Type> = {
   [Property in keyof Type as Exclude<
     Property,
     Required<Type>[Property] extends IgnoredTypes ? Property : never
-  >]: Required<Type>[Property] extends Date ? number : Type[Property];
+  >]: Required<Type>[Property] extends Date
+    ? number
+    : Required<Type>[Property] extends Record<string | number | symbol, unknown>
+      ? string
+      : Type[Property];
```
sshanzel, line 109: Diff looked crazy but basically just added one condition as below:
```
: Required<Type>[Property] extends Record<string | number | symbol, unknown>
      ? string
      : Type[Property];
```

That if it is an object, CDC turns it to false.

src/types.ts
```diff
@@ -102,7 +102,11 @@ export type ChangeObject<Type> = {
   [Property in keyof Type as Exclude<
     Property,
     Required<Type>[Property] extends IgnoredTypes ? Property : never
-  >]: Required<Type>[Property] extends Date ? number : Type[Property];
```
sshanzel, line 105: Diff looked crazy but basically just added one condition as below:
```
: Required<Type>[Property] extends Record<string | number | symbol, unknown>
      ? string
      : Type[Property];
```

That if it is an object, CDC turns it to false.

src/workers/cdc/primary.ts
```diff
@@ -736,10 +736,22 @@ const onSourcePostModerationChange = async (
...
+          logger,
+          'api.v1.source-post-moderation-submitted',
+          { post: data.payload.after },
+        );
+    } catch (err) {
+      logger.error(
+        { post: data.payload.after },
+        'Error parsing flags for source post submission',
+      );
+    }
```
capJavert, line 754: I don't think we need catch here `JSON.parse(data.payload.after.flags || '{}');` will always resolve and if not something is really wrong so better that cdc fails and we catch it due to message not being processed

src/types.ts
capJavert: Ah nice, yeah I agree its better that it is casts to string so we know we need to parse

src/notifications/generate.ts
```diff
@@ -131,7 +131,8 @@ export const notificationTitleMap: Record<
     return `New post from <b>${userName}</b>, check it out now!`;
   },
   user_given_top_reader: (ctx: NotificationUserTopReaderContext) => {
-    const keyword = ctx.keyword.flags?.title || ctx.keyword.value;
+    const keyword =
+      (ctx.keyword.flags as KeywordFlags)?.title || ctx.keyword.value;
```
capJavert, line 135: only issue now is that we need to do this on some other places were we use `Reference` because its `ChangeObject` under the hood. But let's leave it for now.

src/workers/cdc/primary.ts
sshanzel: Right. Yeah, we can remove the try-catch and logging.

src/notifications/generate.ts
sshanzel: Yep. Although I think these are the only places, which is why build failed and I had to update them.

capJavert: Some fruits of strict type checking 👌

PR #2527, opened 2024-12-12
feat: add default feed for users

rebelchris:
Add default feed for users and expose id in boot.
Made it relationship to have cascasing, not sure we need anything besides ID?

### DIFF
--- a/src/entity/user/User.ts
+++ b/src/entity/user/User.ts
@@ -3,6 +3,7 @@ import {
   Column,
   Entity,
   Index,
+  JoinColumn,
   ManyToOne,
   OneToMany,
   OneToOne,
@@ -17,6 +18,7 @@ import type { UserStreakAction } from './UserStreakAction';
 import type { UserCompany } from '../UserCompany';
 import type { UserTopReader } from './UserTopReader';
 import type { SubscriptionCycles } from '../../paddle';
+import { Feed } from '../Feed';
 
 export type UserFlags = Partial<{
   vordr: boolean;
@@ -177,6 +179,9 @@ export class User {
   @Column({ type: 'boolean', default: true })
   followNotifications: boolean;
 
+  @Column({ type: 'text', nullable: true })
+  defaultFeedId?: string | null;
+
   @ManyToOne(() => User, {
     lazy: true,
     onDelete: 'SET NULL',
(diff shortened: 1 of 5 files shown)


src/entity/user/User.ts
```diff
@@ -17,6 +18,7 @@ import type { UserStreakAction } from './UserStreakAction';
 import type { UserCompany } from '../UserCompany';
 import type { UserTopReader } from './UserTopReader';
 import type { SubscriptionCycles } from '../../paddle';
+import { Feed } from '../Feed';
```
capJavert, line 21:
```suggestion
import type { Feed } from '../Feed';
```

src/migration/1733987625520-UserDefaultFeed.ts
```diff
@@ -0,0 +1,25 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UserDefaultFeed1733987625520 implements MigrationInterface {
+  name = 'UserDefaultFeed1733987625520';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "user" ADD "defaultFeedId" text`);
+    await queryRunner.query(
+      `ALTER TABLE "user" ADD CONSTRAINT "UQ_ff0f49b797aca629f81cef47610" UNIQUE ("defaultFeedId")`,
```
capJavert, line 9: why unique constraint?

src/routes/boot.ts
```diff
@@ -420,6 +420,7 @@ const getUser = (
       'followingEmail',
       'followNotifications',
       'subscriptionFlags',
+      'defaultFeedId',
```
capJavert, line 423: also add to `UpdateUserInput` so we can set it on frontend?

src/migration/1733987625520-UserDefaultFeed.ts
rebelchris: Ah missed it, just something ORM generated I guess 😅

src/entity/user/User.ts
rebelchris: Really need to fix webstorm auto import type :jenkins:

src/routes/boot.ts
rebelchris: Added, but no specific test yet

capJavert (approved):

omBratteng (approved):

PR #2528, opened 2024-12-12
feat: add custom feed for new plus members

rebelchris:
Automatic create custom feed for plus member.
It's behind remoteConfig so it doesn't work until we launch.

AS-834 #done

### DIFF
--- a/.infra/common.ts
+++ b/.infra/common.ts
@@ -29,7 +29,11 @@ export const workers: Worker[] = [
   },
   {
     topic: 'user-updated',
-    subscription: 'api.user-updated-plus-subscribed',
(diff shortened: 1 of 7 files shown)


.infra/common.ts
```diff
@@ -29,7 +29,11 @@ export const workers: Worker[] = [
   },
   {
     topic: 'user-updated',
-    subscription: 'api.user-updated-plus-subscribed',
+    subscription: 'api.user-updated-plus-subscribed-squad',
```
rebelchris, line 32: Renamed as it was not clear what it did

capJavert (approved):

omBratteng (approved):

PR #2536, opened 2024-12-13
feat: add new flags to Feed entity

omBratteng:
AS-832

### DIFF
--- a/src/schema/feeds.ts
+++ b/src/schema/feeds.ts
@@ -2239,7 +2239,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     },
     createFeed: async (
       _,
-      { name }: { name: string },
+      { name }: FeedFlagsPublic,
       ctx: AuthContext,
     ): Promise<GQLFeed> => {
       validateFeedPayload({ name });
@@ -2295,7 +2295,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     },
     deleteFeed: async (
       _,
-      { feedId }: { feedId: string; name: string },
+      { feedId }: { feedId: string },
       ctx: AuthContext,
     ): Promise<GQLEmptyResponse> => {
       const feedRepo = ctx.con.getRepository(Feed);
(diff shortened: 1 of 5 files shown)


rebelchris (approved):
Think we also need to expose in updateFeedSettings right?

rebelchris (approved):

src/schema/feeds.ts
```diff
@@ -2204,7 +2268,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     },
     updateFeed: async (
       _,
-      { feedId, name }: { feedId: string; name: string },
+      { feedId, name }: { feedId: string } & FeedFlagsPublic,
```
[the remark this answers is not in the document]
omBratteng: I might've misunderstood the task, and just added the typing. But I've added it so that it'll save the flags properly now. Will add tests

src/schema/feeds.ts
```diff
@@ -2195,19 +2289,17 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       return feed;
     },
     updateFeed: async (
       _,
-      { feedId, name }: { feedId: string; name: string },
+      { feedId, ...flags }: { feedId: string } & FeedFlagsPublic,
       ctx: AuthContext,
     ): Promise<GQLFeed> => {
-      validateFeedPayload({ name });
+      validateFeedPayload({ name: flags.name });
```
capJavert, line 2302: Can you add validation of icon inside this function with `isOneValidEmoji`, of course its optional as well.

omBratteng: Added

src/schema/feeds.ts
```diff
@@ -2175,10 +2269,10 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
     },
     createFeed: async (
       _,
-      { name }: { name: string },
+      flags: FeedFlagsPublic,
       ctx: AuthContext,
     ): Promise<GQLFeed> => {
-      validateFeedPayload({ name });
+      validateFeedPayload({ name: flags.name, icon: flags?.icon });
```
capJavert, line 2275: `?.` should not be needed

capJavert (approved):

src/schema/feeds.ts
omBratteng: Removed

PR #2537, opened 2024-12-13
feat: add new type column to Feed entity

omBratteng:
Creates a new generated column `type` on `Feed` entity which is `main` when feed.id equals feed.userId, otherwise it is `custom`.

AS-837

### DIFF
--- a/src/entity/Feed.ts
+++ b/src/entity/Feed.ts
@@ -7,7 +7,10 @@ export type FeedFlags = Partial<{
 
 export type FeedFlagsPublic = Pick<FeedFlags, 'name'>;
 
-export type FeedType = 'main' | 'custom';
+export enum FeedType {
+  Main = 'main',
+  Custom = 'custom',
+}
 
 @Entity()
 @Index('IDX_feed_id_user_id', ['id', 'userId'], { unique: true })
--- a/src/entity/Feed.ts
+++ b/src/entity/Feed.ts
@@ -7,6 +7,8 @@ export type FeedFlags = Partial<{
 
 export type FeedFlagsPublic = Pick<FeedFlags, 'name'>;
 
+export type FeedType = 'main' | 'custom';
+
 @Entity()
 @Index('IDX_feed_id_user_id', ['id', 'userId'], { unique: true })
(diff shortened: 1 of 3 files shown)


src/migration/1734071060646-FeedTypeColumn.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class FeedTypeColumn1734071060646 implements MigrationInterface {
+  name = 'FeedTypeColumn1734071060646'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "feed" ADD "type" text GENERATED ALWAYS AS (CASE WHEN "id" = "userId" THEN 'main' ELSE 'custom' END) STORED NOT NULL`);
+  }
```
capJavert, line 8: I think usually there is insert to `typeorm_metadata` as well, see examples for other generated column migrations

src/entity/Feed.ts
```diff
@@ -35,6 +40,17 @@ export class Feed {
...
+  @Column({
+    type: 'text',
+    update: false,
+    insert: false,
+    nullable: false,
+    unique: false,
+    generatedType: 'STORED',
+    asExpression: `CASE WHEN "id" = "userId" THEN 'main' ELSE 'custom' END`,
+  })
+  type: FeedType;
```
capJavert, line 52: hmm I thought to put it to graphorm, but I guess this works as well 👍

src/migration/1734071060646-FeedTypeColumn.ts
omBratteng: Probably missed it, the auto generated migration is so noisy 🙈

src/entity/Feed.ts
omBratteng: Didn't think of using GraphORM, though think it's nicer to keep it here, so could use it for for SQL stuff.

rebelchris (approved):

capJavert (approved):

PR #2538, opened 2024-12-13
feat: add feedId to content preference follow

omBratteng:
AS-833


src/schema/contentPreference.ts
```diff
@@ -119,6 +123,10 @@ export const typeDefs = /* GraphQL */ `
       Paginate first
       """
       first: Int
+      """
```
rebelchris, line 126: I'm not sure if here it should be a option @capJavert ?
Because technically can only ever be my feed right?

rebelchris (commented):
Just one comment on the userFollowing/userFollowers one

src/schema/contentPreference.ts
capJavert: User following yes, because yo can follow source, tags on different feeds. `userFollowers` no, because that only exists for main feed.

capJavert (approved):

src/common/contentPreference.ts
```diff
@@ -107,7 +113,7 @@ export const cleanContentNotificationPreference = async ({
   });
 };
 
-const followUser: FollowEntity = async ({ ctx, id, status }) => {
+const followUser: FollowEntity = async ({ ctx, id, status, feedId }) => {
```
capJavert, line 116: Can we add a test to make sure additions work for mutations?

src/schema/contentPreference.ts
capJavert: Also if we can add a test for this

src/common/contentPreference.ts
omBratteng: Yeah, I'll add some tests. Was holding off a bit on that to make sure I hadn't completely missed the objective of this task

src/schema/contentPreference.ts
rebelchris: Oh right we query it for the custom feed yeah
I added now feedPlus, but we probably need to merge these both to add tests neatly.

rebelchris (approved):

capJavert (approved):

PR #2543, opened 2024-12-16
feat: advanced settings for custom feeds

capJavert:


src/schema/feeds.ts
```diff
@@ -2233,18 +2238,20 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-      const feed = await feedRepo.findOneBy({ id: feedId });
+      await feedRepo.findOneByOrFail({
+        id: feedId,
+        userId: ctx.userId,
+      });
       const feedAdvSettingsrepo = ctx.con.getRepository(FeedAdvancedSettings);
 
-      if (!feed) {
-        await feedRepo.save({ userId: feedId, id: feedId });
-      }
```
capJavert, line 2246: users have main feed by default so this fallback is no longer needed

src/schema/feeds.ts
```diff
@@ -2233,18 +2238,20 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
       ctx: AuthContext,
     ): Promise<GQLFeedAdvancedSettings[]> => {
-      const feedId = ctx.userId;
+      const feedId = feedIdArg || ctx.userId;
       const feedRepo = ctx.con.getRepository(Feed);
-      const feed = await feedRepo.findOneBy({ id: feedId });
+      await feedRepo.findOneByOrFail({
+        id: feedId,
+        userId: ctx.userId,
+      });
```
capJavert, line 2252: only update settings for user owned feeds

__tests__/feeds.ts
```diff
@@ -2147,6 +2151,63 @@ describe('mutation updateFeedAdvancedSettings', () => {
     });
     expect(res.data).toMatchSnapshot();
   });
+
+  it('should update custom feed advanced settings', async () => {
+    loggedUser = '1';
+
```
rebelchris, line 2157: We should add @feedPlus here right, then this will fail without setting isPlus

src/schema/feeds.ts
```diff
@@ -894,6 +894,11 @@ export const typeDefs = /* GraphQL */ `
     Update user's feed advanced settings
     """
     updateFeedAdvancedSettings(
```
rebelchris, line 896: We need to add feedPlus here.

rebelchris (changes requested):
We need feedPlus right?

src/schema/feeds.ts
capJavert: yes, gonna update after your PR is merged

rebelchris (approved):

PR #2546, opened 2024-12-16
feat: feed thresholds validation

capJavert:


src/entity/Feed.ts
```diff
@@ -11,7 +11,7 @@ export enum FeedOrderBy {
 export type FeedFlags = Partial<{
   name: string;
   orderBy: FeedOrderBy;
-  minDayRange: number;
+  maxDayRange: number;
```
capJavert, line 14: Wrong name, was supposed to be max

rebelchris (approved):

sshanzel (approved):
Are the props not used yet in production? If so then I think this looks amazing! 🚀

capJavert:
@sshanzel yes, all are optional and used for custom feeds when it goes live.

PR #2549, opened 2024-12-17
fix: search also needs a feed id

rebelchris:
We need to add feed id to search as well.
(excluded specific tests)

### DIFF
--- a/src/schema/search.ts
+++ b/src/schema/search.ts
@@ -218,9 +218,19 @@ export const typeDefs = /* GraphQL */ `
       limit: Int = ${defaultSearchLimit}
 
       """
+      Whether to include content preference status in the search
+      """
+      includeContentPreference: Boolean = false
+
+      """
       Version of the search algorithm
       """
       version: Int = 2
+
+      """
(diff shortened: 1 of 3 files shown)


src/schema/search.ts
```diff
@@ -483,6 +519,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
                 })
                 .andWhere('cp."userId" = :cpUserId', {
                   cpUserId: ctx.userId,
+                })
+                .andWhere('cp."feedId" = :cpFeedId', {
```
capJavert, line 523: Not 100%, but can you check that user can't just send someone else feedId and search through it, I think above `userId` condition should make it impossible, but just to be sure. Adding a test for that would be golden!

capJavert: Same for source search

rebelchris: Ah good point, let me add it and validate :)

capJavert (approved):
Tnx for checking and adding tests!

PR #2550, opened 2024-12-17
feat: cio optimization

sshanzel:
CIO optimization by removing/registering users based on their current activity, along with downgrading their digest if we need to.

Once we've agreed on the processing code-wise, I will start working on the rollout plan, as it should basically use the same processing.


.infra/application.properties
```diff
@@ -9,7 +9,7 @@ debezium.source.database.password=%database_pass%
 debezium.source.database.dbname=%database_dbname%
 debezium.source.database.server.name=api
 debezium.source.table.include.list=public.comment,public.user_comment,public.comment_mention,public.source_request,public.post,public.user,public.post_report,public.source_feed,public.settings,public.reputation_event,public.submission,public.user_state,public.notification_v2,public.source_member,public.feature,public.source,public.post_mention,public.content_image,public.comment_report,public.user_post,public.banner,public.post_relation,public.marketing_cta,public.squad_public_request,public.user_streak,public.bookmark,public.user_company,public.source_report,public.user_top_reader,public.source_post_moderation
-debezium.source.column.exclude.list=public.post.tsv,public.post.placeholder,public.source.flags,public.user_top_reader.image
+debezium.source.column.exclude.list=public.post.tsv,public.post.placeholder,public.source.flags,public.user_top_reader.image,public.user.cioRegistered,public.user.acceptedMarketing,public.user.followingEmail,public.user.notificationEmail
```
sshanzel, line 12: We don't need to track changes from any of these columns.

src/cio.ts
```diff
@@ -89,15 +90,26 @@ export async function identifyUserStreak({
...
-  con: DataSource;
-  cio: TrackClient;
-  user: ChangeObject<User>;
-}): Promise<void> {
+export const generateIdentifyObject = async (
+  con: ConnectionManager,
+  user: ChangeObject<User>,
+) => {
+  const { id } = user;
+  const changed = JSON.parse(JSON.stringify(user));
```
sshanzel, line 98: Not sure if we have a util function that can turn regular objects into `ChangeObject` type.

.infra/application.properties
rebelchris: Are you sure though?
For example acceptedMarketing should trigger userUpdated no? to sync them to CIO (when they not in your flow)

src/cio.ts
rebelchris: structuredClone maybe?

src/common/googleCloud.ts
```diff
@@ -29,3 +31,104 @@ export const downloadJsonFile = async <T>({
...
 };
+
+export const userActiveStateQuery = `
+  with d as (
+    select uss.primary_user_id,
+    min(last_app_timestamp) as last_app_timestamp,
+    min(registration_timestamp) as registration_timestamp,
+
+      min(
+        case when period_end between date('2024-12-07' - interval 6*7 day) and '2024-12-07' then '1. active_last_6w'
```
rebelchris, line 42: Why do we have all the hard coded dates here?
Also I kind of expected us to simply read from a export Dave would generate is that not the case?

src/cron/validateActiveUsers.ts
```diff
@@ -0,0 +1,136 @@
...
+
+            await callWithRetryDefault(() =>
+              cioV2.request.post('/users', { batch: data }),
+            );
+
+            await con
+              .getRepository(User)
+              .update({ id: In(ids) }, { cioRegistered: true });
+
+            await setTimeout(20); // wait for a bit to avoid rate limiting
```
rebelchris, line 58: Do we need this when run inside cb?

src/cron/validateActiveUsers.ts
```diff
@@ -0,0 +1,136 @@
...
+              ),
+            );
+
+            await callWithRetryDefault(() =>
+              cioV2.request.post('/users', { batch: data }),
+            );
+
+            await con
+              .getRepository(User)
+              .update({ id: In(ids) }, { cioRegistered: true });
```
rebelchris, line 56: Also edge-case when one user fails on cio but still get's updated here right?
(It's probably fine, just thinking about it now)

src/entity/user/User.ts
```diff
@@ -131,6 +131,9 @@ export class User {
   @Column({ default: false })
   devcardEligible: boolean;
 
+  @Column({ default: true })
```
rebelchris, line 134: For some reason thought we said flag but it's fine.

rebelchris (commented):
Some minor questions at this point

src/common/googleCloud.ts
sshanzel: Modified this one, the static dates were mostly for testing. Applied params to the query to make the fetch dynamic 🙏

src/cron/validateActiveUsers.ts
sshanzel: Ah, good point. In my head when the http fails, this would fail too, but forgot we have a backoff strategy. I should probably add this when the call was successful there.

src/cron/validateActiveUsers.ts
sshanzel: Yes, since this is now the callback that sends the request to CIO. To have some interval in each calls.

sshanzel:
@rebelchris thank you for the quick review! I modified things a bit to have the finalized query, then applied the flow to handle missed dates 🙏

.infra/application.properties
sshanzel: Ah, right. We have these props in CIO, they should definitely be synced. Let me check the worker.

sshanzel: Maybe I should not add these columns, but only when running the initial rollout.

src/cio.ts
sshanzel: Right. Yeah, I think that's better than json-parse-stringify.

src/cio.ts
```diff
@@ -115,24 +126,46 @@ export async function identifyUser({
     }),
   ]);
 
+  const getDate = (value: number | string | Date) => {
+    if (typeof value === 'number') {
+      return debeziumTimeToDate(dup.createdAt);
```
idoshamun, line 131: Feels like a copy paste issue?

```suggestion
      return debeziumTimeToDate(value);
```

Also this seems like a generic function, so I don't think it needs to be here

src/common/googleCloud.ts
```diff
@@ -29,3 +32,129 @@ export const downloadJsonFile = async <T>({
...
+            else '4' end
+      ) as current_state,
+      min(
+        case
+          when period_end is null then '4. never_active'
+          when period_end between date(@previous_date - interval 6*7 day) and @previous_date then '1. active_last_6w'
+          when period_end between date(@previous_date - interval 12*7 day) and date(@previous_date - interval 6*7 + 1 day) then '2. active_7w_12w'
+          when date(u.last_app_timestamp) <  date(@previous_date - interval 12*7 day) then '3. active_12w+'
+          when date(u.registration_timestamp) <  date(@previous_date - interval 12*7 day) then '3. active_12w+'
+          else '4. never_active' end
```
idoshamun, line 73: I think Dave forgot to change the values here to numbers only

src/common/googleCloud.ts
```diff
@@ -29,3 +32,129 @@ export const downloadJsonFile = async <T>({
...
+      ) as previous_state_name,
+      min(
+        case
+            when period_end is null then '4. never_active'
+            when period_end between date(@run_date - interval 6*7 day) and @run_date then '1. active_last_6w'
+            when period_end between date(@run_date - interval 12*7 day) and date(@run_date - interval 6*7 + 1 day) then '2. active_7w_12w'
+            when date(u.last_app_timestamp) <  date(@run_date - interval 12*7 day) then '3. active_12w+'
+            when date(u.registration_timestamp) <  date(@run_date - interval 12*7 day) then '3. active_12w+'
+            else '4. never_active' end
```
idoshamun, line 82: also here

src/common/mailing.ts
```diff
@@ -171,3 +179,134 @@ export const addPrivateSourceJoinParams = ({
...
+      const validReactivateUsers = await con.getRepository(User).find({
+        select: ['id'],
+        where: { id: In(batch), cioRegistered: false },
+      });
+
+      if (validReactivateUsers.length === 0) {
+        return true;
+      }
+
+      await blockingBatchRunner({
```
idoshamun, line 207: Why do we need batch runner inside a batch runner? Let's do one batch size (the smaller among the two) and keep the flow simple

src/common/mailing.ts
```diff
@@ -171,3 +179,134 @@ export const addPrivateSourceJoinParams = ({
...
+        batchLimit: ITEMS_PER_IDENTIFY,
+        data: validReactivateUsers.map(({ id }) => id),
+        runner: async (ids) => {
+          const users = await con
```
idoshamun, line 211: Why do we fetch again the users? we fetched them earlier

src/common/googleCloud.ts
sshanzel: We have that, way above, he duplicated the lines to have both.

src/common/mailing.ts
```diff
@@ -171,3 +179,134 @@ export const addPrivateSourceJoinParams = ({
...
+      const validInactiveUsers = await con.getRepository(User).find({
+        select: ['id'],
+        where: { id: In(batch), cioRegistered: true },
+      });
+
+      if (validInactiveUsers.length === 0) {
+        return true;
+      }
+
+      await blockingBatchRunner({
```
idoshamun, line 252: Same comment about batching

src/common/mailing.ts
```diff
@@ -171,3 +179,134 @@ export const addPrivateSourceJoinParams = ({
...
+          const data = ids.map((id) => ({
+            action: 'destroy',
+            type: 'person',
+            identifiers: { id },
+          }));
+
+          await callWithRetryDefault({
+            callback: () => cioV2.request.post('/users', { batch: data }),
+            onSuccess: async () => {
+              await con.getRepository(User).update(
```
idoshamun, line 265: We also need to remove the digest entry right?

src/common/mailing.ts
sshanzel: The other fetch is from BQ data. This fetch is internally querying the users from API DB. This saves us time whether we should really process the user or not. For example, this cron failed midway while some users were already processed, fetching here would allow us to avoid running those users again.

src/common/mailing.ts
sshanzel: Right. Good catch. Let me add the removal 👀

src/common/mailing.ts
```diff
@@ -171,3 +179,134 @@ export const addPrivateSourceJoinParams = ({
...
+          await setTimeout(20); // wait for a bit to avoid rate limiting
+        },
+      });
+    },
+  });
+
+  // inactive for 6 weeks: downgrade from daily to weekly digest
+  await blockingBatchRunner({
+    data: downgradeUsers,
+    runner: async (current) => {
```
idoshamun, line 289: We can squeeze everything to one query here. We don't need two. It will make it heavier and slower given the current scale

src/cron/index.ts
```diff
@@ -13,6 +13,7 @@ import updateHighlightedViews from './updateHighlightedViews';
 import hourlyNotifications from './hourlyNotifications';
 import updateCurrentStreak from './updateCurrentStreak';
 import syncSubscriptionWithCIO from './syncSubscriptionWithCIO';
+// import validateActiveUsers from './validateActiveUsers';
```
idoshamun, line 16: You don't have to comment here. It will not be triggered anyway since it's commented on pulumi side

src/cron/validateActiveUsers.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    const lastRunDate = new Date(lastSuccessfulDate);
+    const difference = getAbsoluteDifferenceInDays(lastRunDate, processingDate);
+
+    if (difference === 0) {
+      return;
+    }
+
+    for (let i = 1; i <= difference; i++) {
```
idoshamun, line 39: Neat solution!

src/entity/user/User.ts
idoshamun: We decided a column so we can easily ignore cdc for this

src/common/googleCloud.ts
idoshamun: Are you sure? why is it needed? Please double check

.infra/application.properties
rebelchris: I think it does, didn't we fix that issue a while back by finally updating debezium?

src/common/googleCloud.ts
rebelchris: So this query is the once-off one?
Or we always gonna full query all the time?

I sort of assumed we would just hit some results already?

src/common/mailing.ts
idoshamun: I agree with this notion but still there's no need to query two times.
Here at first: https://github.com/dailydotdev/daily-api/pull/2550/files/3e718033b199746e7a98eead96c2ab640e7fc0cc#diff-a4ce5390febde3cbc254fdf2d25dec35b0144e25fa3df809578cca6d638a7037R198
And now again in the line of the comment.

Also please select only the fields you need throughout the PR. User object is a big one

src/entity/user/User.ts
rebelchris: Ah either way is easy, but it's ok :)

.infra/application.properties
idoshamun: It's not an issue just misunderstanding of what this config does from our side. But it's better to check with adhoc infra any change in debezium props

src/common/googleCloud.ts
sshanzel: Yes, this is what we're going to use on a day-to-day basis. If I set the date today, we will get the correct values. Hence, we won't have this enabled until initial rollout.

src/common/googleCloud.ts
sshanzel: Yes, we can just remove these extra columns. Maybe I will just then add a comment up above to define the values. Or the enum itself is the definition.

src/common/googleCloud.ts
```diff
@@ -29,3 +32,129 @@ export const downloadJsonFile = async <T>({
...
+  )
+  select *
+  from d
+  where current_state != previous_state
+  and previous_state != '4'
+`;
+
+export const getUserActiveStateQuery = (runDate: Date): Query => {
+  const run_date = runDate.toISOString().split('T')[0];
+  const previous_date = subDays(runDate, 1).toISOString().split('T')[0];
```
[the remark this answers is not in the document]
sshanzel: We are not changing the timezone, we are just getting the date part only.

src/common/mailing.ts
sshanzel: Yeah, we can do that. This is just so we will do the fetch only once.

src/common/mailing.ts
sshanzel: Ah, yes, this must be a mistake, we don't need this extra fetch.

As for the columns, from the looks of how we pass the values to CIO, we send the whole user object. Unless we explicitly define all the columns to receive by CIO, we can't specify it here as well.

src/common/googleCloud.ts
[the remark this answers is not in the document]
idoshamun: oh right 🤦

src/cio.ts
```diff
@@ -1,18 +1,15 @@
...
-  camelCaseToSnakeCase,
-  CioUnsubscribeTopic,
-  debeziumTimeToDate,
-  getFirstName,
-  getShortGenericInviteLink,
-} from './common';
+import { camelCaseToSnakeCase, debeziumTimeToDate } from './common/utils';
+import { CioUnsubscribeTopic, getFirstName } from './common/mailing';
+import { getShortGenericInviteLink } from './common/links';
 import type { UserCompany } from './entity/UserCompany';
```
sshanzel, line 13: I'm getting some weird issues that the imports are not defined. Had to separate each to make them work.

src/cio.ts
sshanzel: Let me move this to our utils file 👀

__tests__/cron/validateActiveUsers.ts
```diff
@@ -0,0 +1,249 @@
...
+          'cio_subscription_preferences.topics.topic_7': true,
+          'cio_subscription_preferences.topics.topic_8': true,
+          'cio_subscription_preferences.topics.topic_9': true,
+          created_at: 1656427727,
+          first_name: 'Ido',
+          name: 'Ido',
+          permalink: 'http://localhost:5002/idoshamun',
+          referral_link: 'http://localhost:5002/join?cid=generic&userid=1',
+          updated_at: undefined,
+          username: 'idoshamun',
```
sshanzel, line 227: Notice: these are the only attributes we are now sending to CIO for reactivation. Feel free to comment if it is missing any other properties.

bin/syncCioBasedOnActivity.ts
```diff
@@ -0,0 +1,62 @@
...
+    left join analytics.user_state_sparse as uss on uss.primary_user_id = u.primary_user_id
+      and uss.period_end between '2018-01-01' and @run_date
+      and uss.period = 'daily'
+      and uss.app_active_state = 'active'
+      and uss.registration_state = 'registered'
+    where u.registration_timestamp is not null
+    and date(u.registration_timestamp) < @run_date
+    group by 1
+  )
+  select COUNT(*)
```
sshanzel, line 38: Amount of users we are going to process.

I've added a condition where we would only return `!= 1` since every user is already registered. Hence, we don't need the reactivation process.

Screenshot 2024-12-20 at 1 10 50 PM

```suggestion
  select *```

File: .infra/application.properties [hunk 21]
```diff
@@ -28,4 +28,5 @@ debezium.transforms.ReadOperationFilter.condition=!(valueSchema.field('op') && v
 debezium.transforms.PostsFilter.type=io.debezium.transforms.Filter
 debezium.transforms.PostsFilter.language=jsr223.groovy
 debezium.transforms.PostsFilter.condition=!(valueSchema.field('op') && value.op == 'u' && value.source.table == 'post' && value.before.views != value.after.views)
+# debezium.transforms.PostsFilter.condition=!(valueSchema.field('op') && value.op == 'u' && value.source.table == 'post' && value.before.views != value.after.views) && !(valueSchema.field('op') && value.op == 'u' && value.source.table == 'user' && value.before.cioRegistered != value.after.cioRegistered && !value.after.cioRegistered)
```
idoshamun, line 31: Reactivation is still one-by-one based on this filter right?

src/common/googleCloud.ts
```diff
@@ -29,3 +32,119 @@ export const downloadJsonFile = async <T>({
...
+
+export enum UserActiveState {
+  Active = '1',
+  InactiveSince6wAgo = '2',
+  InactiveSince12wAgo = '3',
+  NeverActive = '4',
+}
+
+export const userActiveStateQuery = `
+  with d as (
```
idoshamun, line 44: I have to double check the query with Dave. I feel there's an edge case here

src/common/mailing.ts
```diff
@@ -171,3 +178,130 @@ export const addPrivateSourceJoinParams = ({
...
+        action: 'destroy',
+        type: 'person',
+        identifiers: { id },
+      }));
+
+      await callWithRetryDefault({
+        callback: () => cioV2.request.post('/users', { batch: data }),
+        onSuccess: async () => {
+          const ids = users.map(({ id }) => id);
+          await Promise.all([
```
idoshamun, line 260: Please use transaction

src/cron/validateActiveUsers.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  const lastRunDate = new Date(lastSuccessfulDate);
+  const difference = getAbsoluteDifferenceInDays(lastRunDate, runDate);
+
+  if (difference === 0) {
+    return;
+  }
+
+  for (let i = 1; i <= difference || i > 100; i++) {
```
idoshamun, line 40: why `i > 100`?

sshanzel: This just means it will run max of 100 days, in case we were going for this route on the initial rollout. We can remove the condition now.

.infra/application.properties
sshanzel: Oh, we are running it also in batches to avoid the rate-limiting as the users can reach up to thousands.

src/cron/validateActiveUsers.ts
idoshamun: so you need `i < 100` for what you want, right?

idoshamun (commented):
we can't commit the initial rollout csv yet. we need to do it only when we're ready

sshanzel:
> we can't commit the initial rollout csv yet. we need to do it only when we're ready

Yes, this was just a sample. Just 1k records to test it.

sshanzel:
I will add more details in the rollout plan confluence page on how to proceed.

src/cron/validateActiveUsers.ts
sshanzel: Yep, should've been that 🤦 

Anyway, I've now based the initial rollout on CSV files. We should not need this anymore.

.infra/application.properties
sshanzel: Ah, yes for that part. This is in part of the preparation, so we can merge this PR once done, and when the initial rollout is about to start, we will just uncomment this line. I will mention this in the confluence page.

idoshamun: but we need this to be uncommented for production as well, not just initial rollout. Don't we? We don't rely on CDC anymore

idoshamun (approved):
Looks good but let's not merge until we meet to do the initial rollout

rebelchris (approved):
Overall looks good to me, just monitor very closely I guess.

PR #2552, opened 2024-12-17
fix: user following query without feedId

capJavert:


src/schema/contentPreference.ts
```diff
@@ -321,11 +321,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         await getFeedByIdentifiersOrFail({
           con: ctx.con,
           feedIdOrSlug: args.feedId,
-          userId: ctx.userId,
+          userId: args.userId,
         });
       }
 
-      const feedId = args.feedId || ctx.userId;
+      const feedId = args.feedId || args.userId;
```
capJavert, line 328: this should never be auth user id, its a userId of requested user

omBratteng: 🙈

omBratteng (approved):

idoshamun (approved):

src/schema/contentPreference.ts
omBratteng: I blame my copy-pasta

capJavert: usual suspect 😆

PR #2554, opened 2024-12-18
feat: rate limiting based on reputation

idoshamun:
Rate limit also affects our power users and I'd like to avoid that. Hence, introducing reputation based rate limiting. It will allow us to set stricter limits to new accounts with low reputation to deal with the spam

### DIFF
--- a/src/remoteConfig.ts
+++ b/src/remoteConfig.ts
@@ -43,9 +43,7 @@ class RemoteConfig {
 
   get vars(): Partial<RemoteConfigValue> {
     if (!process.env.GROWTHBOOK_API_CONFIG_CLIENT_KEY) {
-      return {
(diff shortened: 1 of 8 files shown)


src/remoteConfig.ts
```diff
@@ -40,7 +43,9 @@ class RemoteConfig {
 
   get vars(): Partial<RemoteConfigValue> {
     if (!process.env.GROWTHBOOK_API_CONFIG_CLIENT_KEY) {
-      return {};
+      return {
+        rateLimitReputationThreshold: 10,
+      };
```
capJavert, line 48: We mock for tests in `__setup__`

idoshamun: fixed :)

rebelchris (approved):

PR #2557, opened 2024-12-19
feat: limit custom feed view from date

capJavert:
AS-874

also not removing default custom feed if not plus

### DIFF
--- a/.infra/Pulumi.adhoc.yaml
+++ b/.infra/Pulumi.adhoc.yaml
@@ -84,7 +84,6 @@ config:
     typeormUsername: postgres
     tz: UTC
     urlPrefix: http://api.local.com
-  api:image: api-image:tilt-3c540e5fe45b2a14
   api:k8s:
     namespace: local
   api:temporal:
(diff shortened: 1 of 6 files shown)


rebelchris (approved):

.infra/Pulumi.adhoc.yaml
```diff
@@ -84,6 +84,7 @@ config:
     typeormUsername: postgres
     tz: UTC
     urlPrefix: http://api.local.com
+  api:image: api-image:tilt-3c540e5fe45b2a14
```
sshanzel, line 87: Accidentally pushed I think.

src/routes/boot.ts
```diff
@@ -510,6 +510,7 @@ const loggedInBoot = async ({
         language: user.language || undefined,
         image: mapCloudinaryUrl(user.image),
         cover: mapCloudinaryUrl(user.cover),
+        defaultFeedId: isPlus ? user.defaultFeedId : null,
```
sshanzel, line 513: Should we return the my-feed id (userId) or basically wouldn't really matter?

sshanzel (approved):
Just the accidental push, other than that it looks amazing 🚀

src/routes/boot.ts
capJavert: null is default when you don't have anything set, frontend expects it as well

.infra/Pulumi.adhoc.yaml
capJavert: yep, good catch, removed!

PR #2571, opened 2025-01-03
fix: id of prompts

omBratteng:
already changed in prod


rebelchris (approved):
Cool, just apply to prod as well :)

omBratteng:
> Cool, just apply to prod as well :)

already done

PR #2574, opened 2025-01-08
fix: cio url and digest update

sshanzel:


src/common/mailing.ts
```diff
@@ -248,39 +248,39 @@ export const syncSubscriptionsWithActiveState = async ({
       }
 
       const data = users.map(({ id }) => ({
-        action: 'destroy',
+        action: 'delete',
```
rebelchris, line 251: Just double check you don't call action elsewhere wrongly.

rebelchris (approved):

src/common/mailing.ts
sshanzel: So far we only implemented two areas, one for identify and one for delete. Should be as what is intended 🚀

idoshamun (approved):

PR #2576, opened 2025-01-08
chore: enable cron job

sshanzel:


__tests__/cron/validateActiveUsers.ts
```diff
@@ -34,7 +34,7 @@ describe('validateActiveUsers', () => {
     await saveFixtures(con, User, usersFixture);
   });
 
-  it('should NOT be registered yet', () => {
```
sshanzel, line 37: This was just a wrong description, but the check below should be correct.

idoshamun (approved):

PR #2577, opened 2025-01-08
feat: Enable reporting users

AmarTrebinjac:


src/migration/1736350167538-UserReport.ts
```diff
@@ -0,0 +1,20 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class UserReport1736350167538 implements MigrationInterface {
+    name = 'UserReport1736350167538'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE TABLE "user_report" ("reportedUserId" text NOT NULL, "userId" character varying(36) NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "reason" character varying(36) NOT NULL, "note" text, CONSTRAINT "PK_f804f389ba2f3ac0313080a1a82" PRIMARY KEY ("reportedUserId", "userId"))`);
+        await queryRunner.query(`CREATE INDEX "IDX_user_report_reported_user_id" ON "user_report" ("reportedUserId") `);
+        await queryRunner.query(`CREATE INDEX "IDX_user_report_user_id" ON "user_report" ("userId") `);
```
omBratteng, line 9:
```suggestion
        await queryRunner.query(`CREATE TABLE IF NOT EXISTS "user_report" ("reportedUserId" text NOT NULL, "userId" character varying(36) NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "reason" character varying(36) NOT NULL, "note" text, CONSTRAINT "PK_f804f389ba2f3ac0313080a1a82" PRIMARY KEY ("reportedUserId", "userId"))`);
        await queryRunner.query(`CREATE INDEX IF NOT EXISTS "IDX_user_report_reported_user_id" ON "user_report" ("reportedUserId") `);
        await queryRunner.query(`CREATE INDEX IF NOT EXISTS "IDX_user_report_user_id" ON "user_report" ("userId") `);
```

I know it's a brand new table, so chances that these are already existing is low, but just a best practice safety measure imho

src/migration/1736350167538-UserReport.ts
```diff
@@ -0,0 +1,20 @@
...
+        await queryRunner.query(`ALTER TABLE "user_report" ADD CONSTRAINT "FK_cfc9cf9a552e98d6a634377496f" FOREIGN KEY ("userId") REFERENCES "user"("id") ON DELETE CASCADE ON UPDATE NO ACTION`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "user_report" DROP CONSTRAINT "FK_cfc9cf9a552e98d6a634377496f"`);
+        await queryRunner.query(`DROP INDEX "public"."IDX_user_report_user_id"`);
+        await queryRunner.query(`DROP INDEX "public"."IDX_user_report_reported_user_id"`);
+        await queryRunner.query(`DROP TABLE "user_report"`);
```
omBratteng, line 17:
```suggestion
        await queryRunner.query(`DROP INDEX IF EXISTS "public"."IDX_user_report_user_id"`);
        await queryRunner.query(`DROP INDEX IF EXISTS "public"."IDX_user_report_reported_user_id"`);
        await queryRunner.query(`DROP TABLE IF EXISTS "user_report"`);
```

AmarTrebinjac:
@omBratteng I removed the composite key because I realized a user should be able to report another user multiple times, but I added your suggestions to the new migration ✅

src/common/slack.ts
```diff
@@ -261,6 +261,37 @@ export const notifyCommentReport = async (
   });
 };
 
+export const notifyReportUser = async (
```
sshanzel, line 264: Looks to be almost exactly the same as others. I wonder if we can create a generic function for this. Not related to this PR, just noticed.

sshanzel (approved):
Don't have anything to add! 🚀

ilasw:
Tested integration with frontend but think that we need to double check reason keys, I've found that `HATE_SPEECH` do not exist on server, maybe some more are missing!

AmarTrebinjac:
> Tested integration with frontend but think that we need to double check reason keys, I've found that `HATE_SPEECH` do not exist on server, maybe some more are missing!

Correct key for hate speech is `HATEFUL`

AmarTrebinjac:
> Tested integration with frontend but think that we need to double check reason keys, I've found that `HATE_SPEECH` do not exist on server, maybe some more are missing!

@ilasw Correct key for hate speech is `HATEFUL`

AmarTrebinjac:
> Tested integration with frontend but think that we need to double check reason keys, I've found that `HATE_SPEECH` do not exist on server, maybe some more are missing!

@ilasw There's an existing key for "HATEFUL", I think we can reuse that one instead

omBratteng (approved):

PR #2579, opened 2025-01-09
feat: boot to include continent based from country code

sshanzel:
As discussed, continent is not supported out of the box so we keep a list of country code to continent.


src/routes/boot.ts
```diff
@@ -136,8 +138,11 @@ type BootMiddleware = (
 ) => Promise<Record<string, unknown>>;
 
 const geoSection = (req: FastifyRequest): BaseBoot['geo'] => {
+  const region = req.headers['x-client-region'] as string;
```
capJavert, line 141: Usually geo services provide continent code as well though one of the headers, did you check if one is available so we don't have to use/maintain above mapping?

sshanzel: Yeah, Ido checked and mentioned that it is not part of the headers unfortunately. Yeah we can do either continent code or full name.

capJavert: Maybe code is better since `region` is a code as well?

capJavert (approved):

src/routes/boot.ts
sshanzel: I agree, better consistency. Should be updated now 🚢

PR #2581, opened 2025-01-09
feat: filter blocked users comments

ilasw:
- Added utility function `whereNotUserBlocked` for managing both comments and threads at once
- Added tests

> If blocked user comments you still receive notifications, but this is another task on board


src/common/contentPreference.ts
```diff
@@ -516,3 +518,27 @@ export const unblockEntity = async ({
...
+  }>,
+) => {
+  const query = qb
+    .subQuery()
+    .from(ContentPreference, 'pref')
+    .where({
+      status: ContentPreferenceStatus.Blocked,
+      type: ContentPreferenceType.User,
+      userId,
+      feedId: feedId ?? userId,
```
AmarTrebinjac, line 539: There should be an additional "where" that always checks that it's not blocked on the main feed, because the main feed is a global block. So it should filter out users blocked from main feed + the custom feed, if provided

AmarTrebinjac (changes requested):
Nice work! 🤩

Only left one comment about global block

src/common/contentPreference.ts
ilasw: Updated ✔️

rebelchris (approved):
Need to monitor indexes maybe :)

AmarTrebinjac (approved):

src/common/contentPreference.ts
```diff
@@ -516,3 +518,28 @@ export const unblockEntity = async ({
...
+    .subQuery()
+    .from(ContentPreference, 'pref')
+    .where({
+      status: ContentPreferenceStatus.Blocked,
+      type: ContentPreferenceType.User,
+      userId,
+      feedId: In(feedIds),
+    })
+    .andWhere(`"pref"."referenceId" = ${qb.alias}."userId"`)
+    .getQuery();
```
capJavert, line 543: content preference is a big table, can you show examples of query plans for this to make sure we have indexes at least, on top of my head not sure if there is a more efficient way to query this.

src/common/contentPreference.ts
```diff
@@ -516,3 +518,28 @@ export const unblockEntity = async ({
...
+  qb: QueryBuilder,
+  {
+    userId,
+    feedId,
+  }: Partial<{
+    userId: string;
+    feedId: string;
+  }>,
+) => {
+  const feedIds = feedId ? [feedId, userId] : [userId];
```
capJavert, line 532:
```suggestion
    userId: string;
    feedId?: string;
  }>,
) => {
  const feedIds = feedId ? [feedId, userId] : [userId];
```

if feedId can be optional this should be added in the function's type

capJavert, line 532: userId can be optional as well no, so basically if nothing was passed this query would query all the blocked user rows, we need to guard from function signature and inside the query

ilasw: `userId` is mandatory, i fixed this! ✔️

src/common/contentPreference.ts
ilasw: We need always to search for status/userId/feedId/referenceId, maybe we can remove type but I don't thing is to much impactful. We don't have indexes that involve all those field at the moment. Should I add one? 🤔

capJavert: we do have `idx_content_preferences_feedid_type_userid_status` and the rest is PK, as said can you check the query plan to see how it looks and if we need to add anything?

ilasw: Seems that we don't have any use of indexes with this query, I've tried to add `.useIndex()` but didn't work. I'm going to test if adding an index for all the fields changes something.

Pasted Graphic

capJavert: did you do it locally or in prod, locally psql will almost never use indexes because there is not enough data?

basically best way is to test on metabase or read replica, we can take a look on monday together if you need help

capJavert (approved):
Let's try and monitor DB performance when we merge.

PR #2582, opened 2025-01-09
feat: Filter out blocked users from CommentFeed

AmarTrebinjac:


rebelchris (approved):

capJavert (approved):
Good one, but need to check/sort out the base PR

PR #2583, opened 2025-01-09
feat: Unfollow user on global block

AmarTrebinjac:


src/schema/contentPreference.ts
```diff
@@ -509,6 +509,14 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
 
       const feedId = feedIdArg || ctx.userId;
       await blockEntity({ ctx, id, entity, feedId });
+      if (feedId === ctx.userId && entity === ContentPreferenceType.User) {
```
rebelchris, line 512: Wait why do we need to do this specifically for this entity type?

capJavert: yeah, on block the content preference will already change to blocked, its the same row

AmarTrebinjac: @rebelchris This task is specifically to remove the blocked user from the blocker's "following" list, and to not make the blocker appear on the blocked user's "followers" list

rebelchris: But then it's not a "block" action?
Maybe I'm not understanding the task you working on, which one is this specifically that needs this?

AmarTrebinjac: I just noticed that this feature is already in. I didn't see where it was implemented initially and when I tested on the frontend we don't invalidate cache, so the follower/following count didn't update, so I thought we needed to add this.

My bad

PR #2585, opened 2025-01-10
fix: return original post title of content language is disabled

omBratteng:
Return original post title of content language is disabled (header is missing)

AS-930

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -99,7 +99,7 @@ const createSmartTitleField = ({ field }: { field: string }): GraphORMField => {
       });
 
       const i18nValue =
-        ctx.contentLanguage && typedParent.i18nTitle?.[ctx.contentLanguage];
+        ctx.contentLanguage ? typedParent.i18nTitle?.[ctx.contentLanguage] : undefined
 
       const altValue = getSmartTitle(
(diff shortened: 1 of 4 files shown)


rebelchris (approved):

rebelchris:
Generic, we'll modify raw query to unset everyone to null right? (that's now english)

src/graphorm/index.ts
```diff
@@ -98,7 +98,9 @@ const createSmartTitleField = ({ field }: { field: string }): GraphORMField => {
         userId: ctx.userId!,
       });
 
-      const i18nValue = typedParent.i18nTitle?.[ctx.contentLanguage];
+      const i18nValue =
+        ctx.contentLanguage && typedParent.i18nTitle?.[ctx.contentLanguage];
+
```
capJavert, line 103:
```suggestion
      const i18nValue =
        ctx.contentLanguage ? typedParent.i18nTitle?.[ctx.contentLanguage] : undefined

```

just to make it more specific

capJavert (approved):

omBratteng:
> Generic, we'll modify raw query to unset everyone to null right? (that's now english)

Yeah, will run a query directly to update the ones with English and set it to null

PR #2594, opened 2025-01-14
fix: blocked users no longer show in followers

capJavert:


AmarTrebinjac:
Thanks for handling this. Forgot to look at it after our chat yesterday!

I tried testing it out, but it doesn't seem to work properly ? 🤔  I pulled this branch and `MI-655-report-and-block-user` on the frontend. "Amar Google" is blocked by current user, but still showing up in the list
image

capJavert:
@AmarTrebinjac this is not handling blocking for current user but fixes not showing user's that blocked the user you are viewing. Its a bug that we discovered yesterday indeed.

ilasw (approved):
Looks good! 👍

PR #2595, opened 2025-01-14
feat: add filter for blocked following/followed users

ilasw:


AmarTrebinjac (approved):
Looks good!

Did a user test as well, seems to work as it should

src/common/contentPreference.ts
```diff
@@ -524,9 +524,11 @@ export const whereNotUserBlocked = (
   {
     userId,
     feedId,
+    columnName = 'userId',
```
capJavert, line 527: Lets make this union userId | referenceId

capJavert (approved):

src/common/contentPreference.ts
capJavert: type it as `'userId' | 'referenceId'` instead of `string`

rebelchris (approved):

PR #2596, opened 2025-01-14
feat: reading streak fixes 1

capJavert:
- replace lastViewAtTz (we always return utc for frontend, left it for older clients)
- replace reading history date to also be UTC
- fix cio streak tests and convert to real local dates for display
- no changes to streak increment logic, just standardizing output for frontend

AS-914


src/workers/userStreakUpdatedCio.ts
```diff
@@ -21,20 +21,49 @@ const worker: TypedWorker<'api.v1.user-streak-updated'> = {
...
+        dateOfWeek,
+        user.timezone || DEFAULT_TIMEZONE,
+        'iiiiii',
+      );
       acc.push({
-        day: days[d.getDay()],
-        read: readHistoryDates.includes(d.toISOString().split('T')[0]),
+        day,
+        read: readHistoryDates.includes(dStamp),
       });
```
capJavert, line 67: From what I see in CIO and digest emails when we show reading streak we are expecting local days per timezone, but currently for timezones not close to UTC the day would basically be wrong as it was on frontend. 

Since I am not sure CIO snippets have the ability to manipulate timezones we continue to do that only know it should be correct for all timezones.

rebelchris: Top of head the CIO snippet actually takes user timezone and offsets to UTC should be ok there.
But this is top of head.

capJavert: Yeah it does not, but new logic should be alright. I'll test by moving myself to another TZ.

rebelchris (approved):
Can't be worse then it is now :) 
Just keep an eye out error from users

PR #2598, opened 2025-01-14
feat: Filter out notifications from blocked users

AmarTrebinjac:
This should cover the following workers:

- articleNewCommentCommentedComment (what the hell is this name 😂)
- articleNewCommentPostCommented
- commentMention
- commentReply
- postAdded
- postAddedUserNotification
- postMention

I think these are all the workers that are triggered by another user which are not specifically meant for admins/moderators. Let me know if I missed any!


src/notifications/index.ts
```diff
@@ -212,9 +217,52 @@ export async function generateAndStoreNotificationsV2(
...
+        where: {
+          feedId: In(chunk),
+          referenceId: ctx.initiatorId!,
+          status: ContentPreferenceStatus.Blocked,
+          type: ContentPreferenceType.User,
+        },
+      }),
+    );
+
+    const blockedUsersChunks = await Promise.all(blockedUsersPromises);
```
capJavert, line 245: If we split in chunks and then again do Promise.all it will run all queries in the same time again having a larger impact. I would do a loop here and chunk fetch one by one.

src/notifications/index.ts
```diff
@@ -212,9 +217,52 @@ export async function generateAndStoreNotificationsV2(
...
+    const blockedUsers = blockedUsersChunks.flat();
+
+    const blockedUserIds = new Set(blockedUsers.map((pref) => pref.userId));
```
capJavert, line 248: `flatMap` or reduce to do all in one iteration?

src/workers/notifications/squadMemberJoined.ts
```diff
@@ -75,6 +75,7 @@ const worker: NotificationWorker = {
           post,
           source,
           doneBy,
+          initiatorId: doneBy.id,
```
capJavert, line 78: Even though admin blocked the user I think they should still see the notification if they are admin of the squad.

src/notifications/index.ts
capJavert: It will be slower but thats something we accepted with this approach.

src/workers/notifications/utils.ts
```diff
@@ -80,6 +80,7 @@ export const buildPostContext = async (
       post,
       source: await post.source,
       sharedPost,
+      initiatorId: post.authorId,
```
capJavert, line 83: Adding it here will affect some other workers/notifications, for example moderators might not get notifications from blocked users, can you check?

src/notifications/index.ts
```diff
@@ -212,9 +217,52 @@ export async function generateAndStoreNotificationsV2(
...
+    const filteredUserIds = ctx.userIds.filter((id) => !blockedUserIds.has(id));
+
+    if (filteredUserIds.length === 0) {
+      continue;
+    }
```
capJavert, line 253: would we not `filteredArgs.push(arg)` if no filters are to be applied?

AmarTrebinjac: @capJavert Maybe the variable names are a bit poorly named. FilteredUserIds contains the ids that **should** receive notifications. If there are none in the array, then no point in creating the notification.

I renamed the variables to make it more clear!

capJavert: 👍

src/notifications/index.ts
AmarTrebinjac: Done. Made some simplifications to the syntax as well :)

capJavert (approved):

src/notifications/index.ts
```diff
@@ -212,9 +217,55 @@ export async function generateAndStoreNotificationsV2(
...
+  for (const arg of args) {
+    const { type, ctx } = arg;
+    if (!ctx.initiatorId) {
+      filteredArgs.push(arg);
+      continue;
+    }
+
+    const userIdChunks: string[][] = [];
+    for (let i = 0; i < ctx.userIds.length; i += 500) {
+      userIdChunks.push(ctx.userIds.slice(i, i + 500));
```
sshanzel, line 231: Max size of parameters we can pass to postgresql is 60k. Just if you guys want to increase the batching, although I have nothing against with 500 units.

sshanzel (approved):
Don't have anything blocking. Looks amazing 🚀

src/notifications/index.ts
capJavert: It is also the fact that it is a large query on a large table, but good to note about exact max 👍

PR #2602, opened 2025-01-17
feat: add translation worker

omBratteng:
Adds new worker that listens for post translations from Kvasir.

AS-923

### DIFF
--- a/.infra/common.ts
+++ b/.infra/common.ts
@@ -321,6 +321,10 @@ export const workers: Worker[] = [
     topic: 'api.v1.user-top-reader',
     subscription: 'api.user-top-reader-added',
   },
+  {
+    topic: 'kvasir.post-translated',
+    subscription: 'api.post-translated'
+  }
 ];
 
 export const personalizedDigestWorkers: Worker[] = [
--- a/.infra/common.ts
+++ b/.infra/common.ts
@@ -322,7 +322,7 @@ export const workers: Worker[] = [
     subscription: 'api.user-top-reader-added',
   },
   {
-    topic: 'kvasir.post-translated',
+    topic: 'kvasir.v1.post-translated',
     subscription: 'api.post-translated'
   }
 ];
--- a/__tests__/workers/postTranslated.ts
+++ b/__tests__/workers/postTranslated.ts
@@ -20,7 +20,7 @@ beforeEach(async () => {
   await saveFixtures(con, Post, postsFixture);
 });
 
-describe('vordrPostPrevented', () => {
+describe('postTranslated', () => {
   it('should be registered', () => {
     const registeredWorker = typedWorkers.find(
       (item) => item.subscription === worker.subscription,
--- a/__tests__/workers/postTranslated.ts
+++ b/__tests__/workers/postTranslated.ts
@@ -1,0 +1,121 @@
+import { DataSource } from 'typeorm';
+import createOrGetConnection from '../../src/db';
+import { Post, Source } from '../../src/entity';
+import { postTranslated as worker } from '../../src/workers/postTranslated';
+import { postsFixture } from '../fixture/post';
+import { typedWorkers } from '../../src/workers';
+import { expectSuccessfulTypedBackground, saveFixtures } from '../helpers';
+import { sourcesFixture } from '../fixture';
+import { ContentLanguage } from '../../src/types';
+
+let con: DataSource;
+
+beforeAll(async () => {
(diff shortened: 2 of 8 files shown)


.infra/common.ts
```diff
@@ -321,6 +321,10 @@ export const workers: Worker[] = [
     topic: 'api.v1.user-top-reader',
     subscription: 'api.user-top-reader-added',
   },
+  {
+    topic: 'kvasir.post-translated',
```
capJavert, line 325: do we want to do `v1`, just gives us some flexibility in the future if anything breaking changes

__tests__/workers/postTranslated.ts
```diff
@@ -0,0 +1,121 @@
...
+  con = await createOrGetConnection();
+});
+
+beforeEach(async () => {
+  jest.resetAllMocks();
+  await saveFixtures(con, Source, sourcesFixture);
+  await saveFixtures(con, Post, postsFixture);
+});
+
+describe('vordrPostPrevented', () => {
```
capJavert, line 23: vordr?

src/workers/postTranslated.ts
```diff
@@ -0,0 +1,30 @@
...
+import { validLanguages } from '../types';
+import { logger } from '../logger';
+
+export const postTranslated: TypedWorker<'kvasir.post-translated'> = {
+  subscription: 'api.post-translated',
+  handler: async (message, con) => {
+    const { id, translations, language } = message.data;
+
+    if (!validLanguages.includes(language)) {
+      logger.error({ id, language }, '[postTranslated]: Invalid language');
```
capJavert, line 12: would we not throw so then we can update languages and it auto reprocesses?

src/workers/postTranslated.ts
```diff
@@ -0,0 +1,30 @@
...
+      return;
+    }
+
+    const post = await con.getRepository(Post).findOneByOrFail({ id });
+
+    const existingTranslation = post.translation?.[language];
+    post.translation[language] = {
+      ...existingTranslation,
+      ...translations,
+    };
```
capJavert, line 22: should we not be able to use json `newFlags || flags` to merge on psql side?

__tests__/workers/postTranslated.ts
omBratteng: 🙈 
```suggestion
describe('postTranslated', () => {
```

src/workers/postTranslated.ts
omBratteng: Could lead to a massive backlog of unprocessed languages. The same check happens in Kvasir too, so in API it's more of a safeguard.

src/workers/postTranslated.ts
omBratteng: Yeah, guess that's possible too

src/workers/postTranslated.ts
capJavert: Yeah thats true of course. How would we add a support for new language for example, what thats updated when?

omBratteng: I guess we maybe could store it in Growthbook, then Kvasir can also read from that?

.infra/common.ts
omBratteng: Yeah, good idea.

omBratteng: https://github.com/dailydotdev/streams/pull/43
```suggestion
    topic: 'kvasir.v1.post-translated',
```

src/workers/postTranslated.ts
capJavert: Not a bad idea yes, because now we would have to update on both places?

omBratteng: Yeah, need to update both places. And technically front-end too so it's a dropdown option.

src/workers/postTranslated.ts
```diff
@@ -0,0 +1,44 @@
...
+      await con
+        .getRepository(Post)
+        .createQueryBuilder()
+        .update(Post)
+        .set({
+          translation: () => `jsonb_set(
+          COALESCE(translation, '{}'::jsonb),
+          '{${language}}',
+          '${JSON.stringify(translations)}'::jsonb,
+          true
```
capJavert, line 27: why `COALESCE` here?

omBratteng: I guess `COALESCE` can be dropped, we have `default={}`.

capJavert (approved):

PR #2603, opened 2025-01-17
feat: add gift support in webhook

ilasw:
- Add gift logic in paddle webhook
- Add tests for plus gifting


src/entity/user/User.ts
```diff
@@ -29,6 +29,8 @@ export type UserSubscriptionFlags = Partial<{
   cycle: SubscriptionCycles;
   createdAt: Date;
   subscriptionId: string;
+  gifterId: string;
+  giftExpirationDate: Date;
```
capJavert, line 33: optional?

```suggestion
  gifterId?: string;
  giftExpirationDate?: Date;
```

src/routes/webhooks/paddle.ts
```diff
@@ -74,6 +82,39 @@ const updateUserSubscription = async ({
     );
     return false;
   }
+
+  const { gifter_id: gifterId, duration } = customData;
+  const durationTime = duration && parseInt(duration);
```
capJavert, line 87: is this part set as custom data on frontend side?

src/routes/webhooks/paddle.ts
```diff
@@ -74,6 +82,39 @@ const updateUserSubscription = async ({
...
+
+    const targetUser = await con.getRepository(User).findOne({
+      select: ['subscriptionFlags'],
+      where: { id: userId },
+    });
+    if (isPlusMember(targetUser?.subscriptionFlags?.cycle)) {
+      logger.error({ type: 'paddle', data }, 'User is already a Plus member');
+      return false;
+    }
+  }
```
capJavert, line 116: This means that even though user is already member the gifter payed so not sure if we need to do anything there money wise?

__tests__/paddle.ts
```diff
@@ -0,0 +1,174 @@
...
+let con: DataSource;
+
+beforeAll(async () => {
+  con = await createOrGetConnection();
+  app = await appFunc();
+  return app.ready();
+});
+
+beforeEach(async () => {
+  await saveFixtures(con, User, [...usersFixture, ...plusUsersFixture]);
```
capJavert, line 25: can you also add prefixes as I mentioned in cron PR?

src/routes/webhooks/paddle.ts
```diff
@@ -74,6 +82,39 @@ const updateUserSubscription = async ({
     );
     return false;
   }
+
+  const { gifter_id: gifterId, duration } = customData;
```
rebelchris, line 86: Shouldn't we extract duration from the actual message?
So it will say start date or something like that of purchase and API we can just do + 1 year?

src/routes/webhooks/paddle.ts
```diff
@@ -74,6 +82,39 @@ const updateUserSubscription = async ({
...
+  const durationTime = duration && parseInt(duration);
+  const isGift = !!durationTime && gifterId;
+  if (isGift) {
+    if (userId === gifterId) {
+      logger.error({ type: 'paddle', data }, 'User and gifter are the same');
+      return false;
+    }
```
rebelchris, line 93: We don't do any protection on the frontend for this?
If not, we should :)

src/routes/webhooks/paddle.ts
```diff
@@ -74,6 +82,39 @@ const updateUserSubscription = async ({
...
+
+    const gifterUser = await con
+      .getRepository(User)
+      .findOneBy({ id: gifterId });
+    if (!gifterUser) {
+      logger.error({ type: 'paddle', data }, 'Gifter user not found');
+      return false;
+    }
```
rebelchris, line 106: Might be good to add alerts for some of these (CC @capJavert )

rebelchris (changes requested):
Overall looks good, some minor questions really.

src/routes/webhooks/paddle.ts
capJavert: yeah agree, we can do it once this is deployed 👌

src/routes/webhooks/paddle.ts
capJavert: I was worried if we pass to frontend user can inject it to whatever so might need to be set fixed on the API https://github.com/dailydotdev/daily-api/pull/2603#discussion_r1922274898

src/routes/webhooks/paddle.ts
ilasw: Not yet, but we need to add it during the next tasks

capJavert: in that case it can be mended by malicious user so we should have it coded on the API side only and not be dynamic and user injectable

src/routes/webhooks/paddle.ts
ilasw: Yes we should, but is better to put a check even on server, not?

src/routes/webhooks/paddle.ts
ilasw: Ok, since currently we plan only for yearly plan gift I can hardcode it here ✔️  @rebelchris @capJavert

src/routes/webhooks/paddle.ts
rebelchris: Yeah, just have to make sure we probably throw error someone goes and looks at as it should never happen hopefully.

src/routes/webhooks/paddle.ts
ilasw: User shouldn't be able to gift to a plus user, this is a really edge case imo 👀

__tests__/paddle.ts
ilasw: Added ✔️

src/routes/webhooks/paddle.ts
ilasw: Updated ✔️

capJavert (approved):

src/routes/webhooks/paddle.ts
capJavert: @ilasw can you make note to setup this alerts after the feature goes live?

ilasw: Yes ✔️ 
Where do we setup those alerts? 👀

capJavert: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1234567185/Alerting+and+incident+management

src/routes/webhooks/paddle.ts
sshanzel: I agree with Ante's concern. Should we maybe record this somewhere? Or the logging would enough? Maybe we need to send a Slack message or something.

__tests__/setup.ts
```diff
@@ -26,6 +26,7 @@ jest.mock('../src/remoteConfig', () => ({
       vordrIps: ['192.0.2.0/24', '198.51.100.0/24', '203.0.113.0/24'],
       ignoredWorkEmailDomains: ['igored.com', 'ignored.org'],
       rateLimitReputationThreshold: 1,
+      pricingIds: { pricingGift: 'yearly' },
```
rebelchris, line 29: PlusPriceType.Yearly :) but not urgent

src/paddle.ts
```diff
@@ -3,5 +3,7 @@ export enum SubscriptionCycles {
   Yearly = 'yearly',
 }
 
+export const subscriptionGiftDuration = 31557600000;
```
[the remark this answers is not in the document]
rebelchris: Can we just re-use this one?
https://github.com/dailydotdev/apps/blob/a680fefe4c27d68cdc38bc954fe14851bf0ac8f6/packages/shared/src/lib/dateFormat.ts#L15

rebelchris (approved):
Non blocking comments, nice work

__tests__/setup.ts
ilasw: Tried to use `SubscriptionCycles` but didn't work due to load time, reverted and maybe add later on

PR #2604, opened 2025-01-18
feat: clear expired gifted plus subscription

ilasw:
- add new cron for cleaning expired gifted plus subscription;
- add test case


src/cron/cleanGiftedPlus.ts
```diff
@@ -0,0 +1,25 @@
+import { Cron } from './cron';
+import { User } from '../entity';
+
+const cron: Cron = {
+  name: 'clean-gifted-plus',
+  handler: async (con, logger) => {
+    logger.info('cleaning gifted plus membership...');
```
capJavert, line 7: no need to log this, we will see cron ran from the jobs panel, but if you want to leave it you can change it to `debug`

src/cron/cleanGiftedPlus.ts
```diff
@@ -0,0 +1,25 @@
...
+    const timeThreshold = new Date();
+
+    const { affected } = await con
+      .getRepository(User)
+      .createQueryBuilder('user')
+      .update()
+      .set({ subscriptionFlags: {} })
+      .where('"user"."subscriptionFlags" ->> :expire  < :time', {
+        expire: 'giftExpirationDate',
```
capJavert, line 16:
```suggestion
      .where('"user"."subscriptionFlags"->>'giftExpirationDate' < :time', {
```

capJavert: this part can be static right?

__tests__/cron/cleanGiftedPlus.ts
```diff
@@ -0,0 +1,70 @@
...
+  const yesterdayDate = subDays(new Date(), 1);
+  const tomorrowDate = addDays(new Date(), 1);
+
+  // 5 total users:
+  // - 1 is gifted and expired;
+  // - 2 is gifted and NOT expired;
+  // - 3 & 4 are not plus.
+  // - 5 is classic yearly plus;
+  await con.getRepository(User).update(
+    { id: '1' },
```
capJavert, line 30: can you insert users with prefix eg. `1-clgp` for this test spec, this is so that they don't overlap with other tests and become flaky? 

Check here for example https://github.com/dailydotdev/daily-api/commit/2666bca4c84875e8e12d25c3da3c30ba51b750f0#diff-6dda666a738903e7b6ff809e0a3fb7dff8a41b86a053d1c1d278f15df92e274dR31-R42

src/entity/user/User.ts
```diff
@@ -29,6 +29,8 @@ export type UserSubscriptionFlags = Partial<{
   cycle: SubscriptionCycles;
   createdAt: Date;
   subscriptionId: string;
+  gifterId: string;
+  giftExpirationDate: Date;
```
capJavert, line 33:
```suggestion
  gifterId?: string;
  giftExpirationDate?: Date;
```

optional no?

src/cron/index.ts
```diff
@@ -37,4 +38,5 @@ export const crons: Cron[] = [
   updateSourcePublicThreshold,
   calculateTopReaders,
   validateActiveUsers,
+  cleanGiftedPlus,
```
capJavert, line 41: You have to setup cron in `.infra/crons`

__tests__/cron/cleanGiftedPlus.ts
```diff
@@ -0,0 +1,70 @@
...
+
+beforeAll(async () => {
+  con = await createOrGetConnection();
+});
+
+beforeEach(async () => {
+  await saveFixtures(con, User, [...usersFixture, ...plusUsersFixture]);
+});
+
+it('should remove expired gifted plus subscription', async () => {
```
capJavert, line 20: can you use `describe` to group it, easier to add multiple tests after?

and also add `should be registered` test (check other crons for example)

capJavert (changes requested):
Few things missed, overall looks good 👍

src/cron/cleanGiftedPlus.ts
rebelchris: We decided not to check if user didn't buy plus already?
Secondary thing comes to mind, a gifted/plus user can't be re-gifted before ends from FE?

rebelchris (changes requested):
Nothing besides Ante's comments

src/cron/cleanGiftedPlus.ts
ilasw: You cannot gift to a already-plus user, so no re-gift is possible :(

src/cron/index.ts
ilasw: Thanks! 🙏

src/cron/cleanGiftedPlus.ts
ilasw: Updated to `debug`

src/cron/cleanGiftedPlus.ts
ilasw: > We decided not to check if user didn't buy plus already?
If user buy plus `giftExpirationDate` is cleaned so is not possible to delete "standard" subscription 👀

__tests__/cron/cleanGiftedPlus.ts
ilasw: Updated ✔️

src/entity/user/User.ts
ilasw: Updated ✔️

src/cron/index.ts
ilasw: Updated ✔️

__tests__/cron/cleanGiftedPlus.ts
ilasw: Updated ✔️

src/cron/cleanGiftedPlus.ts
```diff
@@ -0,0 +1,24 @@
...
+  handler: async (con, logger) => {
+    logger.debug('cleaning gifted plus membership...');
+    const timeThreshold = new Date();
+
+    const { affected } = await con
+      .getRepository(User)
+      .createQueryBuilder('user')
+      .update()
+      .set({ subscriptionFlags: {} })
+      .where('"user"."subscriptionFlags" ->> \'giftExpirationDate\'  < :time', {
```
capJavert, line 15:
```suggestion
      .where(`"user"."subscriptionFlags"->>'giftExpirationDate'  < :time`, {
```

.infra/crons.ts
```diff
@@ -114,4 +114,8 @@ export const crons: Cron[] = [
     name: 'calculate-top-readers',
     schedule: '0 2 1 * *',
   },
+  {
+    name: 'clean-gifted-plus',
+    schedule: '0 1 * * *',
```
capJavert, line 119:
```suggestion
    schedule: '45 2 * * *',
```

lets push it a bit so it does not overlap with other crons (just load wise)

its also good to not put it on even hour because usually other stuff happens in even hour

capJavert (approved):
Minor suggestions, but good to go after that.

rebelchris (approved):
Nothing besides Ante's comment.

PR #2605, opened 2025-01-18
feat: add notification inital structure for gifted plus

ilasw:


src/workers/notifications/userGiftedPlusNotification.ts
```diff
@@ -0,0 +1,70 @@
...
+    send_to_unsubscribed: true,
+    identifiers: { id: ctx.gifter.id },
+    transactional_message_id:
+      CioTransactionalMessageTemplateId.UserSentPlusGift,
+    to: ctx.gifter.email,
+    message_data,
+  });
+}
+
+const worker = generateTypedNotificationWorker<'user-updated'>({
```
rebelchris, line 26: Probably weird one, but this should also receive when you blocked this user? 😅

rebelchris (approved):
Minor question on blocked users, otherwise nothing from my side.

src/workers/notifications/userGiftedPlusNotification.ts
ilasw: Yes, I think that this case is not covered here. Should I mute notification for blocked users? This will result in a very odd case 🤔

capJavert: Technically I think blocked users should not show in the gift modal right?

capJavert: > Maybe is better to disable users that blocked you, from the select dropdown on frontend, if we want to handle this case. Wdyt?

Yes

src/workers/notifications/userGiftedPlusNotification.ts
```diff
@@ -0,0 +1,70 @@
...
+      (user.subscriptionFlags as string) || '{}',
+    );
+    const afterSubscriptionFlags: Partial<UserSubscriptionFlags> = JSON.parse(
+      (recipient.subscriptionFlags as string) || '{}',
+    );
+
+    if (
+      isPlusMember(beforeSubscriptionFlags?.cycle) ||
+      !isGiftedPlus(afterSubscriptionFlags) ||
+      !afterSubscriptionFlags?.gifterId
```
capJavert, line 41: Should we log warning for the edge case where user is somehow plus member but he was gifted, it should not happen but still feels wrong to just ignore user's gift.

src/workers/newNotificationV2Mail.ts
```diff
@@ -856,6 +857,24 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
...
+    const { subscriptionFlags } = await con.getRepository(User).findOneOrFail({
+      where: {
+        id: user.id,
+      },
+      select: ['subscriptionFlags'],
+    });
+    const gifter = await con.getRepository(User).findOneOrFail({
+      where: {
+        id: subscriptionFlags?.gifterId,
+      },
```
capJavert, line 870: Watch out, make sure `gifterId` is not `undefined` or `null`, because this will return random user from the DB then. Probably good to throw in that case.

capJavert (changes requested):

src/workers/notifications/userGiftedPlusNotification.ts
ilasw: Yes, adding a `logger.warn`

src/workers/notifications/userGiftedPlusNotification.ts
```diff
@@ -0,0 +1,70 @@
+import { NotificationType } from '../../notifications/common';
+import { generateTypedNotificationWorker } from './worker';
+import { NotificationGiftPlusContext } from '../../notifications';
+import { SquadSource, User, UserSubscriptionFlags } from '../../entity';
+import { isGiftedPlus, isPlusMember } from '../../paddle';
+import { queryReadReplica } from '../../common/queryReadReplica';
+import { PLUS_MEMBER_SQUAD_ID } from '../userUpdatedPlusSubscriptionSquad';
+import { CioTransactionalMessageTemplateId, sendEmail } from '../../common';
+
+async function sendNotificationEmailToGifter(ctx: NotificationGiftPlusContext) {
```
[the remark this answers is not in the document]
ilasw: Added ✔️

src/workers/notifications/userGiftedPlusNotification.ts
ilasw: Added ✔️

__tests__/workers/notifications.ts
```diff
@@ -2433,3 +2437,78 @@ describe('user post added', () => {
...
+        type: SourceType.Squad,
+      },
+    ]);
+  });
+
+  it('should early return for currently non plus user', async () => {
+    const worker = await import(
+      '../../src/workers/notifications/userGiftedPlusNotification'
+    );
+    const actual = await invokeNotificationWorker(worker.default, {
```
capJavert, line 2476: Can you add this tests in a separate file as we did for moderation workers, we want to deprecate this giant notifications worker test file in the future and stop adding to it.

src/workers/newNotificationV2Mail.ts
```diff
@@ -856,6 +857,29 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
...
+      select: ['subscriptionFlags'],
+    });
+
+    if (!subscriptionFlags?.gifterId) {
+      throw new Error('Gifter user not found');
+    }
+
+    const gifter = await con.getRepository(User).findOneOrFail({
+      where: {
+        id: subscriptionFlags?.gifterId,
```
capJavert, line 874:
```suggestion
        id: subscriptionFlags.gifterId,
```

src/workers/newNotificationV2Mail.ts
```diff
@@ -856,6 +857,29 @@ const notificationToTemplateData: Record<NotificationType, TemplateDataFunc> = {
...
   },
+  user_gifted_plus: async (con, user) => {
+    const { subscriptionFlags } = await con.getRepository(User).findOneOrFail({
+      where: {
+        id: user.id,
+      },
+      select: ['subscriptionFlags'],
+    });
+
+    if (!subscriptionFlags?.gifterId) {
```
capJavert, line 868:
```suggestion
    if (!subscriptionFlags.gifterId) {
```

capJavert (approved):
Almost there, can you just move the tests to separate file please. Approving so feel free to merge after.

src/workers/notifications/userGiftedPlusNotification.ts
rebelchris: That's a great suggestion let's go with that.

src/workers/newNotificationV2Mail.ts
ilasw: Could not remove this, `subscriptionFlags` may be undefined here but I can remove the prev one ✔️

src/workers/newNotificationV2Mail.ts
ilasw: Done ✔️

src/workers/newNotificationV2Mail.ts
capJavert: ah right, cool 👍

__tests__/workers/notifications/userGiftedPlusNotification.ts
```diff
@@ -0,0 +1,94 @@
...
+      type: SourceType.Squad,
+    },
+  ]);
+});
+
+const WORKER_RELATIVE_PATH =
+  '../../../src/workers/notifications/userGiftedPlusNotification';
+
+describe('plus subscription gift', () => {
+  it('should early return for currently non plus user', async () => {
```
capJavert, line 58: @ilasw no need to back adjust it, but we always want to have `should be registered` test for each worker, this ensures worker is setup correctly. Keep it in mind for the future.

PR #2609, opened 2025-01-20
fix: posts public/private field assignment

capJavert:


src/workers/postUpdated.ts
```diff
@@ -568,8 +568,14 @@ const getSourcePrivacy = async ({
...
-    return source?.private;
+
+    // if source is not found this will return undefined
+    // and DB will default to false, which means post will be public
+    return !!source?.private;
   } catch (err) {
+    // in case of an error it will return undefined
+    // and DB will default to false, which means post will be public
+
     logger.error({ data, err }, 'failed find source for post');
```
capJavert, line 579: In case of an error we want to set post to private: false?

src/workers/postUpdated.ts
```diff
@@ -568,8 +568,14 @@ const getSourcePrivacy = async ({
     }
 
     const source = await query.getOne();
-    return source?.private;
+
+    // if source is not found this will return undefined
+    // and DB will default to false, which means post will be public
+    return !!source?.private;
```
capJavert, line 574: In case we can't find the source what would we do? Not sure when this would happen but maybe explicitly also return private: false

or change above to be `getOneOrFail` to fallback to catch case below

idoshamun: if there's no source we should ignore the message and log it just in case

src/workers/postUpdated.ts
```diff
@@ -715,7 +720,11 @@ const worker: Worker = {
           await fixData({
             logger,
             entityManager,
-            data,
+            data: {
+              ...data,
+              // pass resolved post id or fallback to original data
+              post_id: postId || data.post_id,
+            },
```
capJavert, line 727: If message without `post_id` and `source_id` arrives it would fail to resolve source because source is resolved over one of those filelds

this fixes the case where we resolve postId from yggdrasil but do not pass it in `fixData`

src/workers/postUpdated.ts
```diff
@@ -567,10 +567,15 @@ const getSourcePrivacy = async ({
...
+    const source = await query.getOneOrFail();
+
+    return source.private;
   } catch (err) {
-    logger.error({ data, err }, 'failed find source for post');
+    const sourcePostError = new Error('failed to find source for post');
+
+    logger.error({ data, err }, sourcePostError.message);
+
+    throw sourcePostError;
```
capJavert, line 578: We now throw if source is not found.

Only place where `source_id` is not sent from yggdrasil is on freeform (due to existing on API) but we have that case covered.

Will monitor of course after deploy for this errors.

__tests__/workers/postUpdated.ts
```diff
@@ -557,6 +558,58 @@ it('save a post as private if source is private', async () => {
...
+  const post = await con.getRepository(Post).findOne({
+    where: {
+      id: 'f99a445f-e2fb-48e8-959c-e02a17f5e816',
+    },
+  });
+  expect(post).toBeNull();
+});
+
+it('should save post if source_id is not passed but post exists', async () => {
+  const newPostId = await generateShortId();
```
capJavert, line 577: test to support https://github.com/dailydotdev/daily-api/pull/2609/files#r1926068941

src/workers/postUpdated.ts
capJavert: done, will monitor after deploy for any errors on processing

idoshamun (approved):

sshanzel (approved):
🚀

rebelchris (approved):

PR #2610, opened 2025-01-22
fix: don't show shared post deleted on feed

rebelchris:
We decided not to show shared post that have deleted reference on feed anymore.
Luckily `showOnFeed` handles all feed types for us.

To retro fix I suggest we simply run a manual query. (Probably very few hits anyway)

### DIFF
--- a/__tests__/workers/postDeletedSharedPostCleanup.ts
+++ b/__tests__/workers/postDeletedSharedPostCleanup.ts
@@ -5,6 +5,7 @@ import createOrGetConnection from '../../src/db';
 import { Post, SharePost, Source } from '../../src/entity';
 import { postsFixture, sharedPostsFixture } from '../fixture/post';
 import { sourcesFixture } from '../fixture';
+import { typedWorkers, workers } from '../../src/workers';
 
 let con: DataSource;
 beforeEach(async () => {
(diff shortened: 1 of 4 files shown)


src/workers/postDeletedSharedPostCleanup.ts
```diff
@@ -0,0 +1,55 @@
...
+const worker: Worker = {
+  subscription: 'api.post-deleted-shared-post-cleanup',
+  handler: async (message, con, logger): Promise<void> => {
+    const data: Data = messageToJson(message);
+    const { post } = data;
+
+    const databasePost = await con
+      .getRepository(Post)
+      .findOneBy({ id: post?.id });
+    console.log('del: ', databasePost?.deleted);
```
capJavert, line 19: cleanup

capJavert (approved):

__tests__/workers/postDeletedSharedPostCleanup.ts
```diff
@@ -0,0 +1,49 @@
...
+      return {
+        ...item,
+        id: `pdspc-${item.id}`,
+        sharedPostId: `pdspc-p1`,
+      };
+    }),
+  );
+});
+
+it('should set shared post to not show on feed if post gets deleted', async () => {
```
capJavert, line 38: can we use `describe` and add `should be registered` test

omBratteng (approved):

PR #2611, opened 2025-01-22
feat: expose translation field on Post

omBratteng:
Expose `translation` field on `Post`. It returns an object containing keys of what translations are available. Currently we only return `title`, but it can be extended in the future.

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -855,7 +855,7 @@ describe('translation field', () => {
   it('should return false for fields when content-language header is not set', async () => {
     const res = await client.query(QUERY);
     expect(res.data.post.translation).toEqual({
-      title: false,
+      title: null,
     });
   });
 
@@ -873,7 +873,7 @@ describe('translation field', () => {
       },
     });
     expect(res.data.post.translation).toEqual({
-      title: false,
+      title: null,
     });
   });
 
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -823,6 +823,61 @@ describe('type field', () => {
   });
 });
 
+describe('translation field', () => {
+  const QUERY = /* GraphQL */ `
+    {
+      post(id: "p1") {
+        translation {
+          title
+        }
+      }
+    }
+  `;
+
+  it('should return false for fields when content-language header is not set', async () => {
+    const res = await client.query(QUERY);
+    expect(res.data.post.translation).toEqual({
+      title: false,
+    });
+  });
+
+  it('should return false for fields when translation does not exist', async () => {
+    await con.getRepository(ArticlePost).update('p1', {
+      translation: {
+        es: {
+          title: 'Hola',
+        },
+      },
+    });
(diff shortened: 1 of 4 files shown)


__tests__/posts.ts
```diff
@@ -823,6 +823,61 @@ describe('type field', () => {
...
+
+  it('should return false for fields when content-language header is not set', async () => {
+    const res = await client.query(QUERY);
+    expect(res.data.post.translation).toEqual({
+      title: false,
+    });
+  });
+
+  it('should return false for fields when translation does not exist', async () => {
+    await con.getRepository(ArticlePost).update('p1', {
```
capJavert, line 845: can you use unique post id eg. `p1-tf`, so it does not overlap with other tests

src/graphorm/index.ts
```diff
@@ -490,6 +492,24 @@ const obj = new GraphORM({
       title: createSmartTitleField({
         field: 'title',
       }),
+      translation: {
+        jsonType: true,
+        transform: (
+          translations: Partial<Record<ContentLanguage, PostTranslation>>,
+          ctx: Context,
+        ): Record<keyof PostTranslation, boolean> => {
+          const translation = ctx.contentLanguage
```
capJavert, line 501: can we early return `{}` if translation is not available, think it will be cleaner to not worry about optional in reduce?

capJavert (approved):

src/graphorm/index.ts
capJavert: I just thought early return `{}` before reduce if `translations[ctx.contentLanguage]` is undefined

```js
const translation = translations[ctx.contentLanguage]

if (!translation) {
  return {}
}

// ... reduce
```

capJavert: because technically each field is not required so if its not there we don't have to return `false`

capJavert: but if you want to return `false` then I would prefer just leaving it as is

omBratteng: I can do as you suggest. I see GraphQL converts it to what I need anyways :lolsob:

src/graphorm/index.ts
```diff
@@ -490,6 +492,26 @@ const obj = new GraphORM({
...
+        jsonType: true,
+        transform: (
+          translations: Partial<Record<ContentLanguage, PostTranslation>>,
+          ctx: Context,
+        ): Partial<Record<keyof PostTranslation, boolean>> => {
+          if (!ctx.contentLanguage || !translations[ctx.contentLanguage]) {
+            return {};
+          }
+
+          const translation = translations[ctx.contentLanguage]!;
```
capJavert, line 505: Better to avoid `!`, typescript will automatically know `translation` is defined after early return:
```suggestion
          const translation = ctx.contentLanguage
            ? translations[ctx.contentLanguage]
            : undefined;

          if (!translation) {
            return {};
          }

```

capJavert (approved):

PR #2612, opened 2025-01-23
feat: return translated title for team members

omBratteng:
An intermediate change to start returning the translated title for team members. Also added remote config if we want to test it for all users.

AS-924


src/graphorm/index.ts
```diff
@@ -85,24 +85,34 @@ const checkIfTitleIsClickbait = (value?: string): boolean => {
...
       const settings = await ctx.dataLoader.userSettings.load({
-        userId: ctx.userId!,
+        userId: ctx.userId,
       });
 
       const i18nValue = ctx.contentLanguage
-        ? typedParent.i18nTitle?.[ctx.contentLanguage]
+        ? shouldUseNewTranslation(ctx)
+          ? typedParent.translation?.[ctx.contentLanguage]?.title
+          : typedParent.i18nTitle?.[ctx.contentLanguage]
```
capJavert, line 115: :hidethepain:

capJavert (approved):

src/graphorm/index.ts
omBratteng: Heh.. don't worry, it's only temporary. It'll be nice and clean ™️ once rolled out

rebelchris (approved):

PR #2617, opened 2025-01-27
feat: multi squad moderation

AmarTrebinjac:
Enables a moderator to view moderation items from multiple squads at once, as well as bulk approve for multiple squads at once.

### DIFF
--- a/__tests__/notifications/index.ts
+++ b/__tests__/notifications/index.ts
@@ -1294,7 +1294,7 @@ describe('storeNotificationBundle', () => {
     );
     expect(actual.notification.description).toBeFalsy();
     expect(actual.notification.targetUrl).toEqual(
-      'http://localhost:5002/squads/a/moderate',
+      'http://localhost:5002/squads/moderate',
     );
     expect(actual.avatars).toEqual([
       {
@@ -1351,7 +1351,7 @@ describe('storeNotificationBundle', () => {
     );
     expect(actual.notification.description).toEqual('Lacks value.');
     expect(actual.notification.targetUrl).toEqual(
-      'http://localhost:5002/squads/a/moderate',
+      'http://localhost:5002/squads/moderate?handle=a',
     );
     expect(actual.avatars).toEqual([
       {
--- a/src/notifications/generate.ts
+++ b/src/notifications/generate.ts
@@ -180,7 +180,7 @@ export const generateNotificationMap: Record<
       .avatarSource(ctx.source)
       .avatarUser(ctx.user)
       .referencePostModeration(ctx.post)
-      .targetSourceModeration(ctx.source),
+      .targetSourceModeration(),
   community_picks_failed: (builder, ctx: NotificationSubmissionContext) =>
     builder.systemNotification().referenceSubmission(ctx.submission),
   community_picks_succeeded: (builder, ctx: NotificationPostContext) =>
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -42,6 +42,11 @@ import {
   validateSourcePostModeration,
   getPostTranslatedTitle,
   getPostSmartTitle,
+  getModerationItemsAsAdminForSource,
+  getModerationItemsByUserForSource,
+  type GQLSourcePostModeration,
+  type SourcePostModerationArgs,
+  getAllModerationItemsAsAdmin,
 } from '../common';
 import {
   ArticlePost,
@@ -70,7 +75,6 @@ import {
   UserAction,
   Settings,
   type PostTranslation,
-  SourceMember,
 } from '../entity';
 import { GQLEmptyResponse, offsetPageGenerator } from './common';
 import {
@@ -93,7 +97,7 @@ import {
   queryPaginatedByDate,
 } from '../common/datePageGenerator';
 import { GraphQLResolveInfo } from 'graphql';
-import { Roles, SourceMemberRoles } from '../roles';
+import { Roles } from '../roles';
 import { markdown, saveMentions } from '../common/markdown';
 // @ts-expect-error - no types
 import { FileUpload } from 'graphql-upload/GraphQLUpload';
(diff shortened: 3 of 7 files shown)


src/schema/posts.ts
```diff
@@ -819,7 +821,7 @@ export const typeDefs = /* GraphQL */ `
       """
       Id of the source
       """
-      sourceId: ID!
+      sourceId: ID
```
rebelchris, line 824: In what case do you not send it?

src/schema/posts.ts
```diff
@@ -1464,12 +1462,55 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-      const isModerator = await isPrivilegedMember(ctx, args.sourceId);
       const page = sourcePostModerationPageGenerator.connArgsToPage(args);
-
       const statuses = Array.from(new Set(args.status));
+      const { userId } = ctx;
+
+      if (args?.sourceId) {
+        const isModerator = await isPrivilegedMember(ctx, args.sourceId);
+        if (isModerator) {
+          return graphorm.queryPaginated(
```
rebelchris, line 1472: What exactly is this for?
Not full following this addition here?

rebelchris (changes requested):
Minor questions

src/schema/posts.ts
AmarTrebinjac: If sourceId is provided, we continue as before. If not, we fetch all the items the current user is allowed to moderate [for the shared moderation page](https://www.figma.com/design/nmfWPS7x3kzLUvYMkBx2kW/daily.dev---Dev-Mode?node-id=23251-26609&p=f&t=uBgm9bI30aFsmhBv-0)

capJavert: Thinking maybe we want a new mutation for multi moderation, its a bit hard to follow what is going on and I would bet we again missed some branch of the logic?

AmarTrebinjac: It's all made from the same page on the frontend. Rather than creating multiple endpoints and having to deal with that on the frontend as well, how about maybe abstracting the logic instead?

src/schema/posts.ts
capJavert: Also curious?

AmarTrebinjac: I answered it [here](https://github.com/dailydotdev/daily-api/pull/2617#discussion_r1931761308)

capJavert: I guess when you want to ask for all moderations.

src/schema/sources.ts
```diff
@@ -982,6 +982,35 @@ export const canAccessSource = async (
...
+    select: ['sourceId'],
+  });
+
+  if (!posts.length) {
+    return false;
+  }
+
+  const sourceIds = Array.from(new Set(posts.map((p) => p.sourceId)));
+
+  const memberships = await ctx.con.getRepository(SourceMember).findBy({
```
capJavert, line 1000: you need only `role` column here, we can add to `select`

src/schema/posts.ts
```diff
@@ -1529,11 +1582,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
           ),
         (builder) => {
           builder.queryBuilder
-            .where(`"${builder.alias}"."sourceId" = :sourceId`, {
-              sourceId: args.sourceId,
-            })
-            .andWhere(`"${builder.alias}"."createdById" = :userId`, {
-              userId,
+            .where(`"${builder.alias}"."sourceId" IN (:...sourceIds)`, {
```
capJavert, line 1585: I think you should be able to do this in a single query with join over source member and then filter by role? No need to first query `sourceIds` above.

src/schema/posts.ts
capJavert: I think main issue is 3 big queries with `queryPaginated`, they share some params but its hard to see what and where so if those can be extracted it would help.

__tests__/notifications/index.ts
```diff
@@ -1351,7 +1351,7 @@ describe('storeNotificationBundle', () => {
     );
     expect(actual.notification.description).toEqual('Lacks value.');
     expect(actual.notification.targetUrl).toEqual(
-      'http://localhost:5002/squads/a/moderate',
```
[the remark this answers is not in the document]
sshanzel: Yeah, we might have to add it.

src/notifications/generate.ts
```diff
@@ -183,7 +183,7 @@ export const generateNotificationMap: Record<
       .avatarSource(ctx.source)
       .avatarUser(ctx.user)
       .referencePostModeration(ctx.post)
-      .targetSourceModeration(ctx.source),
+      .targetSourceModeration(),
```
sshanzel, line 186: I think we can just inline this. We have no other place that uses this `targetSourceModeration` which we can also completely remove now.

```suggestion
      .targetUrl(`${process.env.COMMENTS_PREFIX}/squads/moderate`),
```

src/schema/posts.ts
sshanzel: Personally, I'd also go for a separate query if it would make things much simpler.

src/notifications/generate.ts
AmarTrebinjac: We do actually. We also use it here:

```
  source_post_rejected: (builder, ctx: NotificationPostModerationContext) =>
    builder
      .icon(NotificationIcon.Bell)
      .referencePostModeration(ctx.post)
      .targetSourceModeration(ctx.source)
      .description(ctx.post.moderatorMessage ?? '')
      .avatarSource(ctx.source),

Comment (sshanzel) [reply src/notifications/generate.ts:186]: Oh, nice catch. Ignore me with this one then.

**Comment by AmarTrebinjac** [comment]
@capJavert I've made some changes, take a look.

I managed to solve the count stuff on the frontend without needing to update boot  :)

File: src/schema/posts.ts [hunk 7]
```diff
@@ -1244,10 +1225,6 @@ export const typeDefs = /* GraphQL */ `
       """
       status: String
       """
-      Source to moderate the post in
-      """
-      sourceId: ID!
-      """
```
capJavert, line 1250: This is breaking change so we can't really merge before frontend?

__tests__/notifications/index.ts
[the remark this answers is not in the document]
capJavert: We did this?

src/schema/posts.ts
AmarTrebinjac: @capJavert  But you can't get to this page through the extension, though? 🤔 

The plan is to merge both frontend and backend at the same time

AmarTrebinjac: @capJavert This is not an extension page, so should be fine

The plan is to merge both frontend and backend at the same time

capJavert: Ok

> The plan is to merge both frontend and backend at the same time


Yeah this is not ideal because if we have to revert it will take more time, but let's keep it in mind for the next time, we should always be backward compatible, thats why sometimes its better to add separate query.

All in all, I recommend we merge API first and then frontend (since that one is faster to deploy).

__tests__/notifications/index.ts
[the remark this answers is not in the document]
capJavert: @AmarTrebinjac just a reminder if you did not add it already 🙏 if yes ignore

capJavert (approved):

PR #2618, opened 2025-01-27
feat: clear title translation when post title is changed

omBratteng:
When a post title is changed, we want to clear out the translation, so that we can fetch translation for the new title.

### DIFF
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -11,7 +11,12 @@ import {
 import { User } from '../user';
 import { PostKeyword } from '../PostKeyword';
 import { ArticlePost } from './ArticlePost';
-import { Post, PostOrigin } from './Post';
+import {
+  Post,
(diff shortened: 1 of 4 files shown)


src/entity/posts/utils.ts
```diff
@@ -603,3 +608,26 @@ export const normalizeCollectionPostSources = async ({
...
+
+  await con
+    .getRepository(Post)
+    .createQueryBuilder()
+    .update(Post)
+    .set({
+      translation: () => `(
+SELECT jsonb_object_agg(key, value - '${field}')
+FROM jsonb_each(translation)
+)`,
```
omBratteng, line 629: Because the field structure is
```json
[language]: {
  [field]: string
}
```
we need to iterate over the column and only remove the required keys, so that we don't accidentally wipe out other keys we might have in the future.

capJavert (approved):

PR #2619, opened 2025-01-27
fix: escape post translated JSON

omBratteng:
During the demo, I noticed that it didn't store translations where the translated title contained `'`, because it wasn't getting escaped properly 🙈

### DIFF
--- a/src/workers/postTranslated.ts
+++ b/src/workers/postTranslated.ts
@@ -22,10 +22,14 @@ export const postTranslated: TypedWorker<'kvasir.v1.post-translated'> = {
         .set({
           translation: () => `jsonb_set(
           translation,
-          '{${language}}',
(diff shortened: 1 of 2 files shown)


src/workers/postTranslated.ts
```diff
@@ -22,11 +22,15 @@ export const postTranslated: TypedWorker<'kvasir.v1.post-translated'> = {
...
           translation,
-          '{${language}}',
-          '${JSON.stringify(translations)}'::jsonb,
+          :language,
+          :translations::jsonb,
           true
         )`,
         })
+        .setParameters({
+          language: [language],
```
omBratteng, line 31: Just put language in escape as well for good measure

capJavert: yeah, always better then string concat

capJavert (approved):

PR #2623, opened 2025-01-27
chore: temporarily disable `language` column on `user`

omBratteng:
Needed so that I can run the migration to replace `en` with `null`.


capJavert (approved):

idoshamun (changes requested):
I don't think this will do what you expect it to do. it will just hide it from the payload but still trigger on change afaik

capJavert:
I always forget what this column means, it is poorly named :hidethepain:

omBratteng:
We have `debezium.source.skip.messages.without.change=true`, so since we ignore user language field, it technically should skip sending the message.

omBratteng:
Yeah, testing locally, it doesn't trigger a change

omBratteng:
So yeah, @idoshamun it technically hides it from the payload. But with the `debezium.source.skip.messages.without.change` setting, there's no change, so Debezium skips sending a message.

omBratteng:
https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-property-skip-messages-without-change

We might've had issues where it didn't work properly if we didn't have the proper replica identity
> This property is applied only when the REPLICA IDENTITY of the table is set to FULL.

idoshamun:
Oh good to know!

idoshamun (approved):

PR #2624, opened 2025-01-27
refactor: alerts entity to manage plus modal

sshanzel:
- user gets a subscription gift -> we set `flags.showPlusGift` to true;
- on boot return `shouldShowGiftPlus` based on flag;
- Clear user flag on gift expire.


src/cron/cleanGiftedPlus.ts
```diff
@@ -11,7 +11,10 @@ const cron: Cron = {
       .getRepository(User)
       .createQueryBuilder('user')
       .update()
-      .set({ subscriptionFlags: {} })
+      .set({
+        subscriptionFlags: {},
+        flags: () => `"user"."flags"::jsonb - 'showPlusGift'`,
```
capJavert, line 16: Let's just set it to `false` to be more specific with `updateFlagsStatement`

capJavert (commented):
Looks good

src/entity/user/User.ts
```diff
@@ -23,6 +23,7 @@ import type { Feed } from '../Feed';
 export type UserFlags = Partial<{
   vordr: boolean;
   trustScore: number;
+  showPlusGift: boolean;
 }>;
```
capJavert, line 27: We should create new type `UserFlagsPublic` which will pick only `showPlusGift`

src/schema/users.ts
```diff
@@ -142,6 +142,7 @@ export interface GQLUpdateUserInput {
   followingEmail?: boolean;
   followNotifications?: boolean;
   defaultFeedId?: string;
+  showPlusGift?: boolean;
```
capJavert, line 145: Let's accept `flags` and make it the same type as `UserFlagsPublic`

src/schema/users.ts
```diff
@@ -275,6 +276,16 @@ export const typeDefs = /* GraphQL */ `
     company: Company
   }
 
+  """
+  User flags
+  """
+  type UserFlags {
```
capJavert, line 282: Also let's name it `UserFlagsPublic`

src/schema/users.ts
```diff
@@ -439,6 +450,11 @@ export const typeDefs = /* GraphQL */ `
     Returns the latest top reader badge for the user
     """
     topReader: UserTopReader
+
+    """
+    Flags for the user
+    """
+    flags: UserFlags
```
capJavert, line 457: You don't need this really for the feature since you will look for the value in boot? But if you want to expose this you have to adjust `graphorm/index`, look for `flags` definition for `Settings` entity

src/schema/users.ts
```diff
@@ -565,6 +581,10 @@ export const typeDefs = /* GraphQL */ `
     Default feed id for the user
     """
     defaultFeedId: String
+    """
+    Flags for the user
+    """
+    showPlusGift: Boolean
```
capJavert, line 587:
```suggestion
    """
    Flags for the user
    """
    flags: UserFlagsPublic
```

src/schema/users.ts
```diff
@@ -1905,6 +1925,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       try {
         const updatedUser = { ...user, ...data, image: avatar };
         updatedUser.email = updatedUser.email?.toLowerCase();
+
+        if ('showPlusGift' in data) {
+          updatedUser.flags.showPlusGift = data.showPlusGift;
+        }
```
capJavert, line 1931: With above changes to accept flags we can update flags with `updateFlagsStatement` directly

capJavert (commented):
Looks good, added some notes to standardize how we deal with flags on other entities like `Settings`, basically we want to accept/return `flags` as object.

rebelchris (approved):
Nothing blocking besides Ante's comments so approving already.

src/schema/users.ts
ilasw: Could not use updateFlagsStatement because in save it is not possible, but I'm updating the whole flags merging with old ones

src/schema/users.ts
```diff
@@ -1905,6 +1921,11 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       try {
         const updatedUser = { ...user, ...data, image: avatar };
         updatedUser.email = updatedUser.email?.toLowerCase();
+
+        if ('flags' in data) {
+          updatedUser.flags = { ...user.flags, ...data.flags };
+        }
```
capJavert: Difference with this and `updateFlagsStatement` is that if update to user happens between fetching the flags from the db and saving those flags would be lost:

1. repo.findOneBy(user)
2. some other logic updates user flags in the db
3. we merge the flags from 1. with input
4. repo.save
5. flags from 2. are overwritten

There is an exact pattern in `updateUserSettings` which can be reused, I pushed it in 1ff17ae3e7951fb940afbc3868fb17fa2d101589 for sanity check

capJavert (approved):

PR #2627, opened 2025-01-29
fix: new feature for shared post

rebelchris:
New flow that will always share 1 up only.

If 1 up is a shared post with commentary we re-share that.
If 1 up is a shared post with no commentary we share it's parent.
if 1 up is some other post we just share that.

### DIFF
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -354,8 +354,7 @@ interface CreateSharePostArgs {
 }
 
 export const determineSharedPostId = (post: Post | SharePost): string => {
-  if (post.type === PostType.Share) {
-    if (!post.title) {
+  if (post.type === PostType.Share && !post.title) {
       return post instanceof SharePost ? post.sharedPostId : post.id;
     }
   }
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -14,8 +14,9 @@ import { ArticlePost } from './ArticlePost';
 import {
   Post,
   PostOrigin,
-  translateablePostFields,
+  PostType,
   type TranslateablePostField,
+  translateablePostFields,
 } from './Post';
 import { MAX_COMMENTARY_LENGTH, SharePost } from './SharePost';
 import { ForbiddenError, ValidationError } from 'apollo-server-errors';
@@ -351,6 +352,16 @@ interface CreateSharePostArgs {
   ctx?: AuthContext;
   args: SharePostArgs;
 }
+
+export const determineSharedPostId = (post: Post | SharePost): string => {
+  if (post.type === PostType.Share) {
+    if (!post.title) {
+      return post instanceof SharePost ? post.sharedPostId : post.id;
+    }
+  }
+
+  return post.id;
+};
 
 export const createSharePost = async ({
   con,
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -355,7 +355,9 @@ interface CreateSharePostArgs {
 
 export const determineSharedPostId = (post: Post | SharePost): string => {
   if (post.type === PostType.Share && !post.title) {
-    return post instanceof SharePost ? post.sharedPostId : post.id;
+    if (post instanceof SharePost) {
+      return post.sharedPostId;
+    }
   }
 
   return post.id;
--- a/src/entity/posts/utils.ts
+++ b/src/entity/posts/utils.ts
@@ -355,8 +355,7 @@ interface CreateSharePostArgs {
 
 export const determineSharedPostId = (post: Post | SharePost): string => {
   if (post.type === PostType.Share && !post.title) {
-      return post instanceof SharePost ? post.sharedPostId : post.id;
-    }
+    return post instanceof SharePost ? post.sharedPostId : post.id;
   }
 
   return post.id;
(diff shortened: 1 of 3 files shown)


src/schema/posts.ts
```diff
@@ -2344,19 +2345,21 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       ctx: AuthContext,
       info,
     ): Promise<GQLPost> => {
-      await Promise.all([
+      const [post]: [post: Post] = await Promise.all([
```
omBratteng, line 2348: I feel we need a few more `post` here :joy:

src/entity/posts/utils.ts
```diff
@@ -352,6 +353,16 @@ interface CreateSharePostArgs {
   args: SharePostArgs;
 }
 
+export const determineSharedPostId = (post: Post | SharePost): string => {
+  if (post.type === PostType.Share) {
+    if (!post.title) {
```
omBratteng, line 358: Could this be in one if statement, so we don't need to nest if in if?
```suggestion
  if (post.type === PostType.Share && !post.title) {
```

rebelchris: Ah yeah when I wrote it, I felt there was another condition...
There wasn't 😅

src/entity/posts/utils.ts
```diff
@@ -352,6 +353,15 @@ interface CreateSharePostArgs {
   args: SharePostArgs;
 }
 
+export const determineSharedPostId = (post: Post | SharePost): string => {
+  if (post.type === PostType.Share && !post.title) {
+      return post instanceof SharePost ? post.sharedPostId : post.id;
+    }
+  }
```
omBratteng, line 360:
```suggestion
  if (post.type === PostType.Share && !post.title) {
    return post instanceof SharePost ? post.sharedPostId : post.id;
  }
```
should maybe remove the dangling `}` too 😬

src/schema/posts.ts
```diff
@@ -2344,19 +2345,22 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       ctx: AuthContext,
       info,
     ): Promise<GQLPost> => {
-      await Promise.all([
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const [post, __, ___] = await Promise.all([
```
idoshamun, line 2349: Should do it?

```suggestion
      const [post] = await Promise.all([
```

src/schema/posts.ts
```diff
@@ -2344,19 +2345,22 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
         ctx.con.getRepository(Post).findOneByOrFail({ id }),
```
idoshamun, line 2350: Should we select only the necessary fields here?

src/schema/posts.ts
rebelchris: I can, just need different eslint ignore then for not using all promises 😅

src/schema/posts.ts
rebelchris: we can, didn't want to touch it as no changes needed.

rebelchris: Oh we actually can't because it queries Post, but can also satisfy SharedPost we can't query those fields.

src/entity/posts/utils.ts
```diff
@@ -352,6 +353,14 @@ interface CreateSharePostArgs {
   args: SharePostArgs;
 }
 
+export const determineSharedPostId = (post: Post | SharePost): string => {
+  if (post.type === PostType.Share && !post.title) {
+    return post instanceof SharePost ? post.sharedPostId : post.id;
```
capJavert, line 358:
```suggestion
    if (post instanceof SharePost) {
        return post.sharedPostId
    }
```

this way we can add more conditions later below because it will fall-through to default `post.id`

src/schema/posts.ts
```diff
@@ -2344,19 +2345,21 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       ctx: AuthContext,
       info,
     ): Promise<GQLPost> => {
-      await Promise.all([
-        ctx.con.getRepository(Post).findOneByOrFail({ id }),
+      const [post] = await Promise.all([
+        ctx.con.getRepository(Post).findOneOrFail({ where: { id } }),
```
capJavert, line 2349: why even this change, is it not the same thing `findOneByOrFail` or `findOneOrFail`?

src/schema/posts.ts
idoshamun: If you use findOneOrFail instead By it should allow that. We should avoid selecting all fields. I know it wasn't you but since you're changing it already, let's adjust please

capJavert (approved):

src/schema/posts.ts
rebelchris: It's not the by that's the issue.
It's the fact we take `Post` repo, but we need a field from the child entity (`SharedPost`)

However can't make this SharedPost as it needs to satisfy all types.
Hope this makes sense?

src/schema/posts.ts
rebelchris: It was on the try above for select, but figured it won't work due to different entities.
https://github.com/dailydotdev/daily-api/pull/2627#discussion_r1935177097

src/schema/posts.ts
rebelchris: Done :)

PR #2628, opened 2025-01-29
fix: expose `post` `language`

omBratteng:
Required to do some real time translation on webapp side

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -669,6 +669,11 @@ export const typeDefs = /* GraphQL */ `
     List of available translations for the post
     """
     translation: PostTranslation
+
+    """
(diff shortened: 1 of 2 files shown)


src/schema/posts.ts
```diff
@@ -669,6 +669,11 @@ export const typeDefs = /* GraphQL */ `
     List of available translations for the post
     """
     translation: PostTranslation
+
+    """
+    Language of the post
+    """
+    language: String
```
capJavert, line 676: quick test (or update existing one) that it actually returns?

omBratteng: Added

capJavert (approved):

PR #2632, opened 2025-01-30
feat: add query for get gifter data 

ilasw:
Added new endpoint for enable user fetching gifter data from FE


src/schema/users.ts
```diff
@@ -1873,6 +1879,25 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         true,
       );
     },
+    plusGifterUser: async (_, __, ctx: AuthContext): Promise<GQLUser> => {
```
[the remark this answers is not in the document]
ilasw: We don't have the gifterId on FE because flags are not available, so added this endpoint for getting the user only if you are the gifter plus owner 👀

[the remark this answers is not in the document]
capJavert: Hmm, for what do you need gifter id, can you point me to frontend code?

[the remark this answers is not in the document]
ilasw: Yes, this is the PR: https://github.com/dailydotdev/apps/pull/4132
@capJavert

src/schema/users.ts
```diff
@@ -1873,6 +1879,25 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        });
+
+      if (!isGiftedPlus(subscriptionFlags)) {
+        throw new ForbiddenError('User is not a gifted plus user');
+      }
+
+      return await ctx.con.getRepository(User).findOneOrFail({
+        where: { id: subscriptionFlags.gifterId },
+        select: ['id', 'username', 'name', 'image'],
+      });
```
capJavert, line 1899: This should use graphorm queryOneOrFail

src/schema/users.ts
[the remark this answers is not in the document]
capJavert: ah, ok let's do it like this for now 👍

src/paddle.ts
```diff
@@ -12,7 +12,8 @@ export const isPlusMember = (cycle: SubscriptionCycles | undefined): boolean =>
   !!cycle?.length || false;
 
 export const isGiftedPlus = (
-  subscriptionFlags: UserSubscriptionFlags,
```
capJavert, line 15: let's keep it required if that is ok, you can early return in the query if `subscriptionFlags` is undefined?

capJavert (approved):

capJavert (changes requested):
Good to go, just few comments to adjust how we do queries

src/paddle.ts
ilasw: Done ✔️

src/schema/users.ts
```diff
@@ -1873,6 +1879,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      return await graphorm.queryOneOrFail<GQLUser>(
+        ctx,
+        info,
+        (builder) => {
+          builder.queryBuilder = builder.queryBuilder
+            .andWhere(`${builder.alias}."id" = :userId`, {
+              userId: subscriptionFlags.gifterId,
+            })
+            .select(['id', 'username', 'name', 'image']);
```
sshanzel, line 1904: We don't have to limit the properties here, in this case, we will do it through the actual GraphQL query. Automatically, defined props are the only ones to get pulled.

__tests__/users.ts
```diff
@@ -5994,3 +5996,85 @@ describe('query topReaderBadge', () => {
...
+    await testQueryErrorCode(
+      client,
+      {
+        query: QUERY,
+      },
+      'FORBIDDEN',
+    );
+  });
+
+  it('should thwrow an error if the user is plus but has no gifter', async () => {
```
sshanzel, line 6036:
```suggestion
  it('should throw an error if the user is plus but has no gifter', async () => {
```

__tests__/users.ts
```diff
@@ -5994,3 +5996,85 @@ describe('query topReaderBadge', () => {
...
+    await testQueryErrorCode(
+      client,
+      {
+        query: QUERY,
+      },
+      'UNAUTHENTICATED',
+    );
+  });
+
+  it('should thwrow an error if the user is not plus', async () => {
```
sshanzel, line 6025:
```suggestion
  it('should throw an error if the user is not plus', async () => {
```

sshanzel (approved):
Looks amazing to me!

src/routes/boot.ts
```diff
@@ -519,6 +519,7 @@ const loggedInBoot = async ({
         image: mapCloudinaryUrl(user.image),
         cover: mapCloudinaryUrl(user.cover),
         defaultFeedId: isPlus ? user.defaultFeedId : null,
+        shouldShowGiftPlus: Boolean(user?.flags?.showPlusGift),
```
capJavert, line 522: I think we should expose it as flags: { showPlusGift } and then when you use update user mutation on frontend it will automatically invalidate the object as well.

src/schema/users.ts
capJavert: +1

src/schema/users.ts
```diff
@@ -1873,6 +1879,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         true,
       );
     },
+    plusGifterUser: async (_, __, ctx: AuthContext, info): Promise<GQLUser> => {
+      const { subscriptionFlags } = await ctx.con
+        .getRepository(User)
```
capJavert, line 1884: We can also query this from read replica

src/paddle.ts
capJavert: Why are gifterId and cycle still required, they can technically be undefined and then `isGiftedPlus` would return false no?

src/routes/boot.ts
ilasw: Updated ✔️

src/schema/users.ts
ilasw: Updated ✔️

src/paddle.ts
ilasw: They are not required in params, but if the guard function pass the return an `UserSubscriptionFlags` object that have mandatory `gifterId` and `cycle`. 

This is helping to simplify the code where this function is used since now ts knows that we have already checked for those properties!

rebelchris (approved):

src/paddle.ts
capJavert: ahh nice in fact, I missed the `is` part 🚀

capJavert (approved):

AmarTrebinjac (approved):

PR #2636, opened 2025-01-31
fix: store `alt_title` in `translation`

omBratteng:
I first tried to use SQL functions, but then encountered escape issue with `'`, so went with js object route
```sql
jsonb_set(
    translation,
    '{en}',
    COALESCE(translation->'en', '{}'::jsonb) || jsonb_build_object('smartTitle', '${smartTitle}'),
    true
  )
```

AS-963

### DIFF
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -32,7 +32,11 @@ import { SubmissionFailErrorKeys, SubmissionFailErrorMessage } from '../errors';
 import { generateShortId } from '../ids';
 import { FastifyBaseLogger } from 'fastify';
 import { EntityManager } from 'typeorm';
-import { parseDate, updateFlagsStatement } from '../common';
+import {
+  addSmartTitleToTranslationStatement,
+  parseDate,
+  updateFlagsStatement,
+} from '../common';
 import { markdown } from '../common/markdown';
 import { counters } from '../telemetry';
 import { I18nRecord } from '../types';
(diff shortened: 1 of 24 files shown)


src/workers/postUpdated.ts
```diff
@@ -479,6 +488,15 @@ const updatePost = async ({
     data.contentQuality.manual_clickbait_probability =
       databasePost.contentQuality.manual_clickbait_probability;
   }
+  if (smartTitle) {
+    if (Object.keys(databasePost.translation).length !== 0) {
+      data.translation = databasePost.translation;
+      data.translation.en = data.translation.en || {};
+      data.translation.en.smartTitle = smartTitle;
```
capJavert, line 495: non blocking but can't we just spread this, feels like it would be easier?

it would also remove an option of needing if-else

capJavert (approved):

src/workers/postUpdated.ts
omBratteng: Think it'd look something like this. Haven't tested or validated that it works.

```js
data.translation = {
  ...databasePost.translation,
  en: {
    ...databasePost.translation.en,
    smartTitle,
  },
};
```

capJavert: yep

rebelchris:
@capJavert You ok with this one, or you want any changes?

capJavert:
@rebelchris yes, just gonna test it a bit more and then merge

capJavert (approved):

src/workers/postUpdated.ts
capJavert: 👌

rebelchris (approved):

PR #2637, opened 2025-01-31
feat: add smartTitle as a translatable field

omBratteng:
Updates the query to not override already existing translations for that language.

AS-968

### DIFF
--- a/src/workers/postTranslated.ts
+++ b/src/workers/postTranslated.ts
@@ -22,14 +22,13 @@ export const postTranslated: TypedWorker<'kvasir.v1.post-translated'> = {
         .set({
           translation: () => /*sql*/ `jsonb_set(
           translation,
-          :languages,
+          ARRAY[(:language)],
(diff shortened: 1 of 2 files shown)


src/workers/postTranslated.ts
```diff
@@ -15,24 +15,26 @@ export const postTranslated: TypedWorker<'kvasir.v1.post-translated'> = {
...
-          :translations::jsonb,
+          :languages,
+          COALESCE(translation->:language, '{}'::jsonb) || :translations::jsonb,
           true
         )`,
         })
         .setParameters({
-          language: [language],
+          language,
+          languages: [language],
```
omBratteng, line 32: Couldn't figure out how to get `:translaion` to be what I wanted. So easy fix is just creating a new param that is the array of translation.

omBratteng: Nvm, fixed it in 514f97b13ee793d6c3a59986ee3f4850be18719c

capJavert (approved):

PR #2638, opened 2025-02-03
chore: adjust when we log warn for gift notification

capJavert:
Split condition because `warn` was firing for each user subscription purchase thus spamming a bit too much

We now:
- early return when user is not gifted, basic case where user bought subscription for themself
- warn if user was gifter but was already gifter id is missing


capJavert:
@ilasw for sanity check, logic is the same it was just spamming warns too much.

rebelchris (approved):

PR #2639, opened 2025-02-04
feat: appstore support in redirects

idoshamun:


src/routes/redirects.ts
```diff
@@ -112,10 +123,17 @@ const redirectToMobile =
...
     await sendRedirectAnalytics(con, req, res, '/mobile');
-    if (isAndroid({ req })) {
+
+    const ua = uaParser(req.headers['user-agent']);
+    const os = ua.os.name?.toLowerCase();
+    if (os?.includes('android')) {
       return redirectToAndroid({ req, res });
     }
 
+    if (os?.includes('ios')) {
```
capJavert, line 133: does ios also includes ipad, not sure how the lib returns names?

capJavert (approved):

src/routes/redirects.ts
idoshamun: yes it should

PR #2643, opened 2025-02-06
feat: translate smart title during fetch

capJavert:
- also fallbacks `title` graphorm field when smart title is returned since it used same utils


omBratteng (approved):

rebelchris (approved):
This part looks ok to me.

PR #2644, opened 2025-02-06
fix: add vordr check to common moderation query

AmarTrebinjac:
Added vordr check to the common moderation fetch logic, so that it's the same as for just a single squad


rebelchris (approved):

sshanzel (approved):
We should apply the same on the query (if any) we use for fetching the amount of pending posts.

PR #2648, opened 2025-02-08
fix: paddle transaction management for gifting

sshanzel:
One-time payments are not considered subscriptions in Paddle's dictionary, so we must handle the event for gifting within the TransactionCompleted. Events are as below:

Screenshot 2025-02-08 at 12 38 33 PM

Test cases still check out.

### DIFF
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -24,7 +24,11 @@ import {
 import { JsonContains } from 'typeorm';
 import { paddleInstance } from '../../common/paddle';
 import { addMilliseconds } from 'date-fns';
-import { isPlusMember, subscriptionGiftDuration } from '../../paddle';
+import {
+  isPlusMember,
+  plusGiftDuration,
+  SubscriptionCycles,
+} from '../../paddle';
 
 const extractSubscriptionType = (
   items:
@@ -86,33 +90,6 @@ export const updateUserSubscription = async ({
     return false;
   }
 
-  const { gifter_id: gifterId } = customData;
-  const duration = subscriptionGiftDuration;
-  const isGift = !!gifterId;
-  if (isGift) {
-    if (userId === gifterId) {
-      logger.error({ type: 'paddle', data }, 'User and gifter are the same');
-      return false;
-    }
-
-    const gifterUser = await con
-      .getRepository(User)
-      .findOneBy({ id: gifterId });
-    if (!gifterUser) {
-      logger.error({ type: 'paddle', data }, 'Gifter user not found');
-      return false;
-    }
-
-    const targetUser = await con.getRepository(User).findOne({
-      select: ['subscriptionFlags'],
-      where: { id: userId },
-    });
-    if (isPlusMember(targetUser?.subscriptionFlags?.cycle)) {
-      logger.error({ type: 'paddle', data }, 'User is already a Plus member');
-      return false;
-    }
-  }
-
   await con.getRepository(User).update(
     {
       id: userId,
@@ -122,18 +99,6 @@ export const updateUserSubscription = async ({
         cycle: state ? subscriptionType : null,
         createdAt: state ? data.data?.startedAt : null,
         subscriptionId: state ? data.data?.id : null,
-        ...(isGift && {
-          gifterId,
-          giftExpirationDate: addMilliseconds(
-            new Date(),
-            duration,
-          ).toISOString(),
-        }),
-      }),
-      ...(isGift && {
-        flags: updateFlagsStatement({
-          showPlusGift: isGift,
-        }),
       }),
     },
   );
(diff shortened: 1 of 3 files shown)


src/routes/webhooks/paddle.ts
```diff
@@ -86,33 +90,6 @@ export const updateUserSubscription = async ({
...
-    const targetUser = await con.getRepository(User).findOne({
-      select: ['subscriptionFlags'],
-      where: { id: userId },
-    });
-    if (isPlusMember(targetUser?.subscriptionFlags?.cycle)) {
-      logger.error({ type: 'paddle', data }, 'User is already a Plus member');
-      return false;
-    }
-  }
-
```
sshanzel, line 115: Moved to its own function to lessen the noise here.

src/routes/webhooks/paddle.ts
```diff
@@ -405,6 +370,66 @@ const notifyNewPaddleTransaction = async ({
...
 };
 
+export const processGiftedPayment = async (
+  data: TransactionCompletedEvent['data'],
+) => {
+  const con = await createOrGetConnection();
+  const { gifter_id, user_id } = data.customData as PaddleCustomData;
+
+  if (user_id === gifter_id) {
+    logger.error({ type: 'paddle' }, 'User and gifter are the same');
```
capJavert, line 380: maybe log user_id?

src/routes/webhooks/paddle.ts
```diff
@@ -405,6 +370,66 @@ const notifyNewPaddleTransaction = async ({
...
+    return;
+  }
+
+  const gifterUser = await con.getRepository(User).findOneBy({ id: gifter_id });
+
+  if (!gifterUser) {
+    logger.error({ type: 'paddle' }, 'Gifter user not found');
```
capJavert, line 387: log gifter_id

src/routes/webhooks/paddle.ts
```diff
@@ -405,6 +370,66 @@ const notifyNewPaddleTransaction = async ({
...
+};
+
+export const processTransactionCompleted = async (
+  result: TransactionCompletedEvent,
+) => {
+  const { gifter_id } = (result?.data?.customData ?? {}) as PaddleCustomData;
+
+  if (gifter_id) {
+    await processGiftedPayment(result.data);
+  }
```
capJavert, line 428: lgtm, but its a bit hard to see what you added since everything in code moved, what was the actual change?

src/routes/webhooks/paddle.ts
capJavert: I actually see we used to send `data` so probably want to return that

capJavert (commented):
Just some questions

src/routes/webhooks/paddle.ts
sshanzel: Yeah, most things were just moved, and this is the only new thing.

capJavert: but what inside `processGiftedPayment` is new?

src/routes/webhooks/paddle.ts
sshanzel: Applied it on all three places.

src/routes/webhooks/paddle.ts
sshanzel: Since we are now passing the `data` property, it should contain the `gifter_id` as well.

src/routes/webhooks/paddle.ts
sshanzel: I mean this block, the content of `processGiftedPayment` was just moved from the deleted lines above.

src/routes/webhooks/paddle.ts
```diff
@@ -462,7 +487,7 @@ export const paddle = async (fastify: FastifyInstance): Promise<void> => {
                     eventData,
                     AnalyticsEventName.ReceivePayment,
                   ),
-                  notifyNewPaddleTransaction(eventData),
+                  processTransactionCompleted(eventData),
```
capJavert, line 490:
```suggestion
                  await processTransactionCompleted(eventData),
```

src/routes/webhooks/paddle.ts
sshanzel: https://github.com/dailydotdev/daily-api/pull/2648#discussion_r1947480532

This basically. Also to make it easier to review since the whole function was just moved, and this is the new thing that might be harder to review if I had kept them in the same place.

src/routes/webhooks/paddle.ts
```diff
@@ -122,18 +99,6 @@ export const updateUserSubscription = async ({
         cycle: state ? subscriptionType : null,
         createdAt: state ? data.data?.startedAt : null,
         subscriptionId: state ? data.data?.id : null,
-        ...(isGift && {
```
capJavert, line 125: Since code was moved from here should we ignore gifter in `SubscriptionCreated` or is it not sent at all for OTP?

src/routes/webhooks/paddle.ts
```diff
@@ -405,6 +370,66 @@ const notifyNewPaddleTransaction = async ({
...
+        ).toISOString(),
+      }),
+      flags: updateFlagsStatement({ showPlusGift: true }),
+    },
+  );
+};
+
+export const processTransactionCompleted = async (
+  result: TransactionCompletedEvent,
+) => {
```
capJavert, line 423: Let's make it accept object `{ data }` or whatever, easier to expand later

src/routes/webhooks/paddle.ts
```diff
@@ -405,6 +370,66 @@ const notifyNewPaddleTransaction = async ({
   await webhooks.transactions.send({ blocks });
 };
 
+export const processGiftedPayment = async (
```
capJavert, line 373: same as https://github.com/dailydotdev/daily-api/pull/2648/files#r1948641153

capJavert (commented):
Looked at the code more, few more questions 🙏

src/routes/webhooks/paddle.ts
capJavert: 🤝

src/routes/webhooks/paddle.ts
capJavert: 🤝

capJavert (approved):
Approving, good to go after comments are resolved.

src/routes/webhooks/paddle.ts
sshanzel: The whole object was needed originally, and didn't want to make any change outside the gifting, but yeah, I can update the other functions too.

capJavert: you can still do the whole object, just wanted to move to object syntax

capJavert:
```ts
export const processTransactionCompleted = async ({
  result: TransactionCompletedEvent,
}) => Promise

rebelchris (approved):
Nothing besides Ante's stuff, but it's kind of hard to see as so much is swapped in diff.

src/routes/webhooks/paddle.ts
sshanzel: Ah, cool, yeah, we can do that.

PR #2654, opened 2025-02-12
feat: add pricepreviews endpoint

AmarTrebinjac:
Added this endpoint so that we can get the prices for the Plus extension modal, as Paddle is being blocked from loading third party scripts.


src/routes/pricePreviews.ts
```diff
@@ -0,0 +1,23 @@
...
+export default async function (fastify: FastifyInstance): Promise<void> {
+  fastify.get('/', async (req, res) => {
+    const region = req.headers['x-client-region'] as CountryCode | undefined;
+    try {
+      const pricePreview = await paddleInstance?.pricingPreview.preview({
+        items: Object.keys(remoteConfig.vars.pricingIds!).map((priceId) => ({
+          priceId,
+          quantity: 1,
+        })),
+        address: region ? { countryCode: region } : undefined,
```
AmarTrebinjac, line 15: This doesn't seem to be working properly. I thought it was returning USD to begin with because local doesn't set geo, but I tried putting in "NO" manually, and I still got USD back.

From looking at the Frontend, this is the way we get the correct currency back from Paddle, or am I missing something here?

src/routes/pricePreviews.ts
```diff
@@ -0,0 +1,23 @@
+import type { FastifyInstance } from 'fastify';
+import { paddleInstance } from '../common/paddle';
+import type { CountryCode } from '@paddle/paddle-node-sdk';
+import { remoteConfig } from '../remoteConfig';
+
+export default async function (fastify: FastifyInstance): Promise<void> {
+  fastify.get('/', async (req, res) => {
```
capJavert, line 7: let's move this to gql query, you can add it to `user`, or if you want you can add new paddle schema

src/remoteConfig.ts
```diff
@@ -44,7 +44,16 @@ class RemoteConfig {
...
-      return {};
+      return {
+        postRateLimit: 10000,
+        commentRateLimit: 10000,
+        rateLimitReputationThreshold: 0,
+        pricingIds: {
+          pri_01jjbwd5j7k0nm45k8e07yfmwr: SubscriptionCycles.Yearly,
+          pri_01jcdp5ef4yhv00p43hr2knrdg: SubscriptionCycles.Monthly,
+          pri_01jcdn6enr5ap3ekkddc6fv6tq: SubscriptionCycles.Yearly,
+        },
```
capJavert, line 55: accidental change?

AmarTrebinjac: @capJavert Accidental how? Wrong pricing ids, or the commit itself? 😅

capJavert: We don't usually put this mocked values in code itself since it would mean API would fallback to them if remote config was not configured.

src/routes/pricePreviews.ts
capJavert: We prefer gql due to ability to use fields, transformations and also just validating input/output.

capJavert: Also gql schema is easily typed across frontend/backend so we don't have to guess if something changes.

capJavert: We use simple endpoints mostly for internal stuff, but those also moved to protobuf which offers similar schema validation across services.

AmarTrebinjac: I see.

But unless you feel strongly about it, I'd like to just keep it like it is now. This endpoint only serves one purpose, and it's getting the price previews specifically for the extension, since we can't load 3rd party scripts. All other cases we just get them from the paddle SDK in frontend.

capJavert: Let's do the gql call please, really no point in having it like this when we have better tools to do it.

capJavert: If you have any issues with gql implementation I am here to help.

sshanzel: Yeah, I think using GQL as our primary means to make API requests makes maintenance much easier. And we have more tools there in case we need something like directives, the graphorm magic, etc. - expanding the endpoint will be easier.

AmarTrebinjac: @capJavert I've moved it over to GraphQL now 😃

src/schema/paddle.ts
```diff
@@ -0,0 +1,100 @@
...
+    """
+    currencyCode: String!
+    """
+    Price previews
+    """
+    items: [Price!]!
+  }
+
+  extend type Query {
+    pricePreviews: PricePreviews!
```
capJavert, line 58: Let's make it `@auth`, don't think anonymous users will need the price info anytime soon.

src/schema/paddle.ts
```diff
@@ -0,0 +1,100 @@
...
+  unknown,
+  BaseContext
+>({
+  Query: {
+    pricePreviews: async (_, __, ctx: BaseContext) => {
+      const region = ctx.req.headers['x-client-region'] as
+        | CountryCode
+        | undefined;
+      const pricePreview = await paddleInstance?.pricingPreview.preview({
+        items: Object.keys(remoteConfig.vars.pricingIds!).map((priceId) => ({
```
capJavert, line 77:
```suggestion
        items: Object.keys(remoteConfig.vars.pricingIds || {}).map((priceId) => ({
```

All remote config keys must be treated as optional, we should fallback to `{}` or throw error if pricing ids is not set.

src/schema/paddle.ts
capJavert: And add a test for it.

capJavert (commented):
Some fallback missing and tests, other then that good to go.

capJavert (approved):

src/schema/paddle.ts
```diff
@@ -0,0 +1,102 @@
...
+  label: string;
+}
+
+export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
+  unknown,
+  BaseContext
+>({
+  Query: {
+    pricePreviews: async (_, __, ctx: BaseContext) => {
+      const region = ctx?.req.headers['x-client-region'] as
```
capJavert, line 73: btw. this could now be just a param in query, no need to use headers

__tests__/paddle.ts
```diff
@@ -105,6 +157,22 @@ const getTransactionData = (customData: PaddleCustomData) =>
...
 
+describe('pricing preview', () => {
+  it('should return pricing preview data', async () => {
+    loggedUser = 'whp-1';
+    const mockPreview = jest.fn().mockResolvedValue(getPricingPreviewData());
+    jest
+      .spyOn(paddleInstance.pricingPreview, 'preview')
+      .mockImplementation(mockPreview);
+
+    const result = await resolvers.Query.pricePreviews(null, {}, state.context);
```
rebelchris, line 168: Why this approach, I'm not against it actually...
Maybe we should add the regular one though as it can test for auth and errors.

rebelchris (commented):
Would maybe just make the tests based on normal graphQL layer and add test for fail+auth

__tests__/paddle.ts
capJavert: yeah, @AmarTrebinjac can you do the regular tests as we do usually, through the gql client calls?

capJavert: I kinda missed you used this weird flow

AmarTrebinjac: @rebelchris @capJavert updated to call via query! ✅

capJavert (approved):
Looks good from my end, only header thingy left.

PR #2655, opened 2025-02-12
feat: expose `titleHtml` in translation

omBratteng:
Adds `titleHtml` as a translatable field, and adds `translate="no"` to the mention links so that it does not translate the usernames


src/workers/postTranslated.ts
```diff
@@ -15,6 +16,26 @@ export const postTranslated: TypedWorker<'kvasir.v1.post-translated'> = {
...
+      const post = await con.getRepository(Post).findOneByOrFail({ id });
+
+      if (post.type === PostType.Share && translations.title) {
+        const mentions = await getMentions(
+          con,
+          translations.title,
+          post.authorId || '',
+        );
+
+        translations.titleHtml = generateTitleHtml(
```
capJavert, line 28: question is will LLM translated response preserve mentions, I think it might also translate them, for example `@rock` would become `@kamen` in Croatian

omBratteng: It might 🙈

capJavert: It is small amount of posts, so still most of the posts would get translated, which in the end is what we want I think.

omBratteng:
@capJavert refactored to just expose titleHtml as translation

src/common/markdown.ts
```diff
@@ -26,7 +26,7 @@ export const markdown: MarkdownIt = MarkdownIt({
 export const getMentionLink = ({ id, username }: MarkdownMention): string => {
   const href = getUserProfileUrl(username || ghostUser.id);
 
-  return `<a href="${href}" data-mention-id="${id}" data-mention-username="${username}">@${username}</a>`;
+  return `<a href="${href}" data-mention-id="${id}" data-mention-username="${username}" translate="no">@${username}</a>`;
```
capJavert, line 29: this will be detected on kvasir?

capJavert (approved):

rebelchris (approved):
Nice one

sshanzel (approved):

src/common/markdown.ts
omBratteng: [Google Translate detects it](https://cloud.google.com/translate/troubleshooting#:~:text=How%20do%20I%20tell%20Cloud%20Translation%20%2D%20Advanced%20to%20not%20translate%20something%3F)

PR #2659, opened 2025-02-13
fix: suggestions should keep same logic as posts

rebelchris:
I saw the suggestion one didn't filter out deleted/private posts.
It should as they not accesible for the masses.

Then it's also inline with the actual post page.


capJavert (approved):
Test would also be good, but nothing blocking 🚀

PR #2663, opened 2025-02-17
fix: add `emailConfirmed` to zombie cleanup cron

omBratteng:


src/cron/cleanZombieUsers.ts
```diff
@@ -8,10 +8,23 @@ const cron: Cron = {
...
+        {
+          infoConfirmed: false,
+        },
+        {
+          emailConfirmed: false,
+        },
+      ])
+      .andWhere({
+        createdAt: LessThan(timeThreshold),
+      });
```
omBratteng, line 25:
```sql
DELETE FROM
  "public"."user"
WHERE
  (
    "infoConfirmed" = FALSE
    OR "emailConfirmed" = FALSE
  )
  AND "createdAt" < NOW() - INTERVAL '1 day'
```

omBratteng: `infoConfirmed` will always be true for users signing up via email.

And `emailConfirmed` will always be true for users signing up via SSO.

So this will now catch zombies that never complete signup via email

capJavert: now that we have multi column match, might need index on this, if for any reason cron starts to be slow

omBratteng: Currently no indexes on `infoConfirmed`, but can add for that and `emailConfirmed`.

Though currently the query takes ~750ms. When I add one index per column, it goes down to ~120ms

capJavert: yeah, don't think is needed right now, but just something to look for

omBratteng: Added single index. Duration is the same as two indexes, but according to chatgippity, a single index is potentially better.

capJavert (approved):

PR #2664, opened 2025-02-19
fix: proper fix search filter

rebelchris:
Add filter for private=false.
Restore limit as we won't need it with this fix.

### DIFF
--- a/src/integrations/meilisearch.ts
+++ b/src/integrations/meilisearch.ts
@@ -10,12 +10,14 @@ type Meili = {
   post_id: string;
   title?: string;
 };
+
 interface MeiliResponse {
   hits: Meili[];
   limit: number;
   offset: number;
   estimatedTotalHits: number;
 }
+
 export type MeiliPagination = {
   limit: number;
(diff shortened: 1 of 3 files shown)


omBratteng (approved):
I assume removing limit is intended

src/integrations/meilisearch.ts
```diff
@@ -37,7 +39,7 @@ export const searchMeili = async (
     headers,
   };
   const res = await retryFetchParse<MeiliResponse>(
-    `${meiliOrigin}indexes/${meiliIndex}/search?${params}`,
+    `${meiliOrigin}indexes/${meiliIndex}/search?${params}&filter=private=false`,
```
capJavert, line 42: when we start to add another filter we should move to `POST` endpoint

capJavert (approved):

src/integrations/meilisearch.ts
rebelchris: Let's keep it in mind :)

PR #2665, opened 2025-02-20
fix: add provider to SubscriptionFlags

omBratteng:


src/migration/1740042768350-UserSubscriptionProvider.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class UserSubscriptionProvider1740042768350 implements MigrationInterface {
+  name = 'UserSubscriptionProvider1740042768350'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`UPDATE "user"
+      SET "subscriptionFlags" = jsonb_set("subscriptionFlags", '{provider}', '"paddle"')
+      WHERE "subscriptionFlags" <> '{}'::jsonb;
+    `);
```
capJavert, line 10: Let's run this as a one of, don't think we need a migration

src/routes/webhooks/paddle.ts
```diff
@@ -99,6 +99,7 @@ export const updateUserSubscription = async ({
         cycle: state ? subscriptionType : null,
         createdAt: state ? data.data?.startedAt : null,
         subscriptionId: state ? data.data?.id : null,
+        provider: state ? SubscriptionProvider.Paddle : null,
```
capJavert, line 102: missing this in `processGiftedPayment`

capJavert: also can we add this field to tests to make sure its set

src/migration/1740042768350-UserSubscriptionProvider.ts
omBratteng: Reverted. Will run manually

PR #2670, opened 2025-02-21
feat: award user

capJavert:
- [x] tests

Lot of TODOs but trying to get it out of the door for initial stuff, implementation will change with Njord side as well.

### DIFF
--- a/src/common/njord.ts
+++ b/src/common/njord.ts
@@ -58,7 +58,7 @@ export const transferCores = async ({
       productId: product.id,
       senderId,
       value: product.value,
-      fee: remoteConfig.vars.fees?.transfer,
+      fee: remoteConfig.vars.fees?.transfer || 0,
       request: ctx.requestMeta,
       // TODO feat/transactions add note
     });
--- a/src/common/njord.ts
+++ b/src/common/njord.ts
@@ -1,0 +1,97 @@
+import { createClient } from '@connectrpc/connect';
+import { createGrpcTransport } from '@connectrpc/connect-node';
+import {
+  Credits,
+  Currency,
+  EntityType,
+  TransferType,
+} from '@dailydotdev/schema';
+import type { AuthContext } from '../Context';
+import { UserTransaction } from '../entity/user/UserTransaction';
+import { Product } from '../entity/Product';
+import { remoteConfig } from '../remoteConfig';
+import { dispatchWhoami } from '../kratos';
+
+const transport = createGrpcTransport({
+  baseUrl: process.env.NJORD_ORIGIN,
+  httpVersion: '2',
+});
+
+const njordClient = createClient<typeof Credits>(Credits, transport);
+
+export type TransferProps = {
+  ctx: AuthContext;
+  receiverId: string;
+  productId: string;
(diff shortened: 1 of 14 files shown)


src/common/njord.ts
```diff
@@ -0,0 +1,104 @@
...
+  productId,
+}: TransferProps): Promise<UserTransaction> => {
+  if (!ctx.userId) {
+    throw new ForbiddenError('Auth is required');
+  }
+
+  // TODO feat/transactions check if user is team member, remove check when prod is ready
+  if (!ctx.isTeamMember && isProd) {
+    throw new ForbiddenError('Not allowed for you yet');
+  }
```
capJavert, line 44: Closed it only for team members in prod

src/entity/user/UserTransaction.ts
```diff
@@ -0,0 +1,77 @@
...
+export type UserTransactionFlags = Partial<{
+  note: string;
+}>;
+
+export type UserTransactionFlagsPublic = Pick<UserTransactionFlags, 'note'>;
+
+export type UserTransactionRequest = RequestMeta;
+
+@Entity()
+export class UserTransaction {
```
rebelchris, line 23: Do we actually want to store this?
Is this what you last discussed with Viktor/Denis?

src/entity/user/UserTransaction.ts
```diff
@@ -0,0 +1,77 @@
...
+    nullable: true,
+  })
+  @JoinColumn({ name: 'senderId' })
+  sender: Promise<User>;
+
+  @Column()
+  value: number;
+
+  @Column()
+  fee: number;
```
rebelchris, line 70: Who will calculate this now?

rebelchris (changes requested):
Mainly question around that userTransaction status

src/entity/user/UserTransaction.ts
capJavert: Yes

src/entity/user/UserTransaction.ts
capJavert: This will be a flat fee from remote config at the time of making a transaction, so no calculations, just a way of knowing what was sent to njord

src/entity/user/UserTransaction.ts
rebelchris: we'll use this id for idempotency?

src/entity/user/UserTransaction.ts
rebelchris: Ah ok, so remote config is in charge of fees then?

src/entity/user/UserTransaction.ts
capJavert: yes

src/entity/user/UserTransaction.ts
capJavert: it is in charge of getting fixed fee percentage and sending it to njord, everything else is on njord side eg. how much credits will be deducated, given etc..

rebelchris (approved):
Nothing blocking

PR #2671, opened 2025-02-24
fix: add better log tracking

rebelchris:
We seemed to always trigger this message regardless of the fail reason.
(might have been added after logs tbh)

So now we only log the user missing when that's the error.
Other cases likely mean user just not eligible for streak.


omBratteng (approved):

capJavert (commented):
tnx for adjusting!

capJavert (approved):

PR #2672, opened 2025-02-24
feat: add count endpoint for pending posts

ilasw:
```
Pending posts count increases in production when on a single squad moderation page.
```

Currently we fetch the entire pending posts array to display the sidebar counter but we are not considering the pagination.
When a user navigates to the detail page, the query updates. Since both the page and sidebar share the same query key, the count gets updated in both places.

I've added a new endpoint that fetches only the count, eliminating pagination issue and reducing the network call overload

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -46,7 +46,8 @@ import {
   type GQLSourcePostModeration,
   type SourcePostModerationArgs,
   getAllModerationItemsAsAdmin,
-  getTranslationRecord, whereUserIsSquadModerator,
+  getTranslationRecord,
+  whereUserIsSquadModerator,
 } from '../common';
 import {
   ArticlePost,
@@ -75,7 +76,8 @@ import {
   UserAction,
   Settings,
   type PostTranslation,
-  determineSharedPostId, SquadSource,
+  determineSharedPostId,
+  SquadSource,
 } from '../entity';
 import { GQLEmptyResponse, offsetPageGenerator } from './common';
 import {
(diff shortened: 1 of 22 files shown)


src/schema/posts.ts
```diff
@@ -813,6 +815,16 @@ export const typeDefs = /* GraphQL */ `
       first: Int
     ): SourcePostModerationConnection! @auth
 
+    """
+    Get count of pending post for all sources
```
AmarTrebinjac, line 819: The "all" is a bit misleading

src/schema/posts.ts
```diff
@@ -1452,6 +1464,37 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      _,
+      args: { sourceId?: SquadSource['id'] },
+      ctx: AuthContext,
+    ): Promise<number> => {
+      const { sourceId } = args;
+      const isAdmin = !!sourceId && (await isPrivilegedMember(ctx, sourceId));
+      return queryReadReplica(ctx.con, ({ queryRunner }) => {
+        const manager = queryRunner.manager;
+        const query = manager
+          .getRepository(SourcePostModeration)
```
AmarTrebinjac, line 1477: We need to filter vordr posts

ilasw: Added ✔️

capJavert:
At first I also don't see a need for another query, but I have to check what we have, I'll review a bit later since I have some credits system stuff to solve first.

ilasw:
@AmarTrebinjac @capJavert out pagination system currently do not support a "totalCount" field nor is easy to add it without a big effort in refactor/test the codebase. This is why I've moved to this kind of solution. Let me know wdyt please

capJavert:
@ilasw I think this is the issue with query keys and it should be handled on frontend by including `status` in query key of `useSquadPendingPosts`. This is why you see that arbitrary 15 because it loads and shows rejected reviews. 

As for this not taking pagination into account we were fine with that before so I think it is out of scope of this task to change that part.

ilasw:
@capJavert Even after adding `status` to the query key and filtering only for `pending` items, we’re still having some issues. For example: on the Squad page, the counter comes from squad.moderationPostCount and isn't limited to 15. We could set a higher limit when fetching data from all squads, but this would result in a very large response for those moderating only a few active squads. What do you think?

image

capJavert:
I understand, but I don't want to introduce it since it is not relevant to the task, task is about fixing changing of count in sidebar randomly, and showing counts either way is always hard and does not scale well. I doubt that a lot of users have 15 or more reviews active at any time, we can do `15+` label in sidebar if value is 15, and then user can always click on it and see all reviews that are there.

ilasw:
Ok np, I'll close this PR then and mention this on the fe one ✔️

PR #2673, opened 2025-02-26
feat: add `plusProvider` to boot

omBratteng:
Exposes the plus provider in boot. Need to use it in webapp to determine what plus to show in app

### DIFF
--- a/__tests__/boot.ts
+++ b/__tests__/boot.ts
@@ -129,6 +129,7 @@ const LOGGED_IN_BODY = {
     mastodon: null,
     language: undefined,
     isPlus: false,
+    plusProvider: null,
     defaultFeedId: null,
     flags: {
       showPlusGift: false,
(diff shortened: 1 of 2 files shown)


__tests__/boot.ts
```diff
@@ -129,6 +129,7 @@ const LOGGED_IN_BODY = {
     mastodon: null,
     language: undefined,
     isPlus: false,
+    plusProvider: null,
```
capJavert, line 132: Can you add a test that it is also returned for logged user boot?

omBratteng: 👍

__tests__/boot.ts
```diff
@@ -550,6 +552,53 @@ describe('logged in boot', () => {
       .expect(200);
     expect(res.body.user.defaultFeedId).toBeNull();
   });
+
+  describe('plusProvider', () => {
+    it('should return the correct plus provider when not plus', async () => {
+      // await con.getRepository(User).save({
+      //   ...usersFixture[0],
+      //   plusProvider: 'stripe',
+      // });
```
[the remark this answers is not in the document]
omBratteng:
```suggestion
```

PR #2675, opened 2025-02-27
feat: get user balance

capJavert:
https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1435172886/RFC-108+-+Credits+backend#User-credits

TODO:

- [x]  boot tests
- [x] garmr integration https://github.com/dailydotdev/daily-api/pull/2679


src/redis.ts
```diff
@@ -60,7 +60,10 @@ export const setRedisObjectWithExpiry = (
   key: string,
   value: RedisObject,
   seconds: number,
-) => ioRedisPool.execute((client) => client.set(key, value, 'EX', seconds));
+) =>
+  ioRedisPool.execute((client) =>
+    client.set(key, value, 'EX', Math.floor(seconds)),
```
capJavert, line 65: We had this multiple times, so I think it is in fact better to fix in this level. Redis accepts only int (seconds) as expires.

src/common/utils.ts
```diff
@@ -228,3 +229,18 @@ export const unwrapArray = <T>(
 
   return arrayOrValue;
 };
+
+export const parseBigInt = (value: bigint): number => {
+  try {
+    return Number(value);
```
capJavert, line 235: On njord side balance is `bigint`

src/common/njord.ts
```diff
@@ -102,3 +110,91 @@ export const transferCores = async ({
...
+          amount: 0,
+        };
+      }
+
+      throw originalError;
+    }
+  },
+);
+
+export const updateBalanceCache = createAuthProtectedFn(
```
capJavert, line 163: will be used on transfer error to instantly update balance of user

src/common/njord.ts
```diff
@@ -102,3 +110,91 @@ export const transferCores = async ({
...
+    const redisKey = getBalanceRedisKey({ ctx });
+
+    const currentTsSec = Date.now() / 1000;
+    const expiresAt = currentTsSec + coresBalanceExpirationSeconds;
+
+    await setRedisObjectWithExpiry(redisKey, JSON.stringify(value), expiresAt);
+  },
+);
+
+export const expireBalanceCache = createAuthProtectedFn(
```
capJavert, line 174: will be used in CDC to invalidate balance cache

rebelchris (approved):
Looks good to me

PR #2676, opened 2025-02-28
refactor: settings entity to store sort comments by

sshanzel:
Settings property to contain the preference of user on how to sort the comments section.


__tests__/settings.ts
```diff
@@ -239,6 +241,19 @@ describe('mutation updateUserSettings', () => {
...
+    loggedUser = '1';
+
+    const repo = con.getRepository(Settings);
+    const settings = await repo.save(repo.create({ userId: '1' }));
+    expect(settings.sortCommentsBy).toEqual('oldest');
+
+    const res = await client.mutate(MUTATION, {
+      variables: { data: { sortCommentsBy: SortCommentsBy.Newest } },
+    });
+    expect(res.data.updateUserSettings.sortCommentsBy).toEqual('newest');
```
omBratteng, line 254: Test that the settings in the database matches as well?

sshanzel: Updated 🚀

src/migration/1740757299661-SettingsCommentsAlgo.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class SettingsCommentsAlgo1740757299661 implements MigrationInterface {
+  name = 'SettingsCommentsAlgo1740757299661';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "settings" ADD "sortCommentsBy" character varying NOT NULL DEFAULT 'oldest'`,
```
capJavert, line 8:
```suggestion
      `ALTER TABLE "settings" ADD "sortCommentsBy" text NOT NULL DEFAULT 'oldest'`,
```

not sure why it added varying, text should be fine

src/schema/settings.ts
```diff
@@ -80,6 +81,8 @@ export const typeDefs = /* GraphQL */ `
     lastPrompt: String
   }
 
+  ${toGQLEnum(SortCommentsBy, 'SortCommentsBy')}
```
capJavert, line 84: Where do we then actually sort by it?

capJavert (commented):
Not sure if WIP still, but few comments

src/schema/settings.ts
sshanzel: On the actual comments request, it is being sort when you pass the right parameter.

sshanzel: The feature already exists. The sort is just not persisted at the moment.

capJavert: I was not thinking about the enum specifically (I was just commenting on this part of code). 

I was thinking where do you use this new setting in the comment query, that part is missing?

capJavert: Or do you just add it in this PR and you are gonna hook it up in separate PR?

sshanzel: Ah cool. This line was a mistake, it was already imported from the `comments.ts`. So the latest changes removed this part since everything is already integrated. We are just saving it now.

sshanzel: > I was thinking where do you use this new setting in the comment query, that part is missing?

It already exists. The query to fetch comments accepts this parameter and sorts accordingly.

capJavert: Ah so you are going to send it from frontend directly, ok

capJavert (approved):

src/entity/Settings.ts
```diff
@@ -44,6 +45,8 @@ export type SettingsFlagsPublic = Pick<
   | 'lastPrompt'
 >;
 
+export { SortCommentsBy };
```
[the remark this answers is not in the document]
sshanzel: Ah. Was just debugging some things. Removed now 🚀

PR #2677, opened 2025-02-28
feat: add `appAccountToken` to `UserSubscriptionFlags`

omBratteng:
Turns out StoreKit / AppStore is a bit more annoying to work with, I can't set the userId on the transaction, but I can use an `appAccountToken` which must be an UUID.

This is what we will use to associate the App Store transactions with the user (hopefully).

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -177,6 +177,25 @@ export const excludeProperties = <T, K extends keyof T>(
   return clone;
 };
 
+export const includeProperties = <T, K extends keyof T>(
+  obj: T,
+  properties: K[],
+): Pick<T, K> => {
+  if (!obj) {
+    return obj;
+  }
+
+  const clone = structuredClone(obj);
+  const keys = Object.keys(clone) as K[];
+  keys.forEach((key) => {
+    if (!properties.includes(key)) {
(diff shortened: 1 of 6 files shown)


rebelchris (approved):
Looks ok to me

src/routes/boot.ts
```diff
@@ -515,14 +533,19 @@ const loggedInBoot = async ({
...
         defaultFeedId: isPlus ? user.defaultFeedId : null,
         flags: {
           showPlusGift: Boolean(user?.flags?.showPlusGift),
         },
+        subscriptionFlags: {
+          ...includeProperties(user.subscriptionFlags!, [
+            'provider',
+            'appAccountToken',
+          ]),
+        },
```
capJavert, line 548: Let's just make it static, it is faster and easier to see what is going on, `excludeProperties` above is something legacy so we never touched it. 
```suggestion
        subscriptionFlags: {
          provider: user.subscriptionFlags?.provider,
          appAccountToken: user.subscriptionFlags?.appAccountToken,
        },
```

omBratteng: I created a new function that's the opposite of exclude, so that it just includes the properties.
But I'll make it static :)

PR #2680, opened 2025-03-03
fix: streak reminder timezone calculation

capJavert:
Streak reminder checks if user read today and skips sending. Issue was that if user has non UTC timezone their read day could actually be different then current day in worker (UTC) so now we check if last view date is the same in user's timezone. 

[Thread](https://dailydotdev.slack.com/archives/C01L0QXQJNB/p1740639959953599)


__tests__/common/timezone.ts
```diff
@@ -39,3 +43,37 @@ describe('validateValidTimeZone tests', () => {
     });
   });
 });
+
+describe('isSameDayInTimezone tests', () => {
```
capJavert, line 47: We are using this same util on frontend already

idoshamun (approved):

rebelchris (approved):
Oh nice find!

PR #2683, opened 2025-03-03
feat: add price preview support for new property structure

ilasw:
Solves https://github.com/dailydotdev/daily/issues/1771  and https://github.com/dailydotdev/daily/issues/1772


src/common/paddle.ts
```diff
@@ -31,3 +32,21 @@ export const cancelSubscription = async ({
     throw e;
   }
 };
+
+export const removeNonNumber = (value: string): string =>
+  value.replace(/,(\d{2})$/, '.$1').replace(/[^\d.]/g, '');
```
rebelchris, line 37: Kind of annoying we have to do this on API and FE, can't we not only do it in 1 place maybe?
(Or maybe we removed FE one but I missed it?)

src/schema/paddle.ts
```diff
@@ -82,15 +127,47 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const items = pricePreview?.details?.lineItems.map((item) => {
+        const isOneOff = !item.price?.billingCycle?.interval;
+        const isYearly = item.price?.billingCycle?.interval === 'year';
+        const duration =
+          isOneOff || isYearly
+            ? SubscriptionCycles.Yearly
+            : SubscriptionCycles.Monthly;
+        const priceAmount = getPriceFromPaddleItem(item);
+        const months = duration === SubscriptionCycles.Yearly ? 12 : 1;
+        const monthlyPrice = Number(
```
rebelchris, line 139: This one as well I guess?

src/schema/paddle.ts
```diff
@@ -82,15 +127,47 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          (priceAmount / months).toString().match(/^-?\d+(?:\.\d{0,2})?/)?.[0],
+        );
+        const currencyCode = pricePreview?.currencyCode;
+        const currencySymbol = item.formattedTotals.total.replace(
+          /\d|\.|\s|,/g,
+          '',
+        );
+        const customData = item.price.customData as GQLCustomData;
+        const priceFormatter = new Intl.NumberFormat('en-US', {
```
rebelchris, line 148: Is this stable? I vaguely recall it not always being stable

rebelchris (changes requested):
More of a question, did we remove the FE duplication now?

src/common/paddle.ts
ilasw: Not removing any client logic for now, since @idoshamun said that we need to focus more on a quick short term fix for now, not on a long term refactor.

I'm open for implementing a long term solution instead, but we need the team to agree for it.

src/schema/paddle.ts
ilasw: As now I've just replicated the client logic since we need to pass the values directly into the client. If we don't want to have this logic here I can simplify returning the items and moving the transformation in our client side.

src/common/paddle.ts
ilasw: From slack thread:
> I would love to keep the transformations in client because we change it very often and apply a large amount of experiment in plus flows 

Maybe is better until we reach a stable point to keep the transformations in client, I've personally changed the price object shape about 3-4 times as today 👀

src/schema/paddle.ts
ilasw: This should be stable, we have it since node v6 👀

src/common/paddle.ts
idoshamun: For now I'd leave this concern aside please, there's a production issue that needs to get fixed. I'll open this thread separately

rebelchris: I'm just really out of the loop why we need 2, but don't really have time to check both sides now.

rebelchris: Also:
> because we change it very often
But now it's also here? 😂

Maybe I'm just confused now?

AmarTrebinjac: The issue explained shortly:

* Extension can't use paddle SDK
* We made this endpoint to mirror the price previews for the extension
* Shortly after we made some bigger changes to the frontend part and the "price" property, but didn't update it here. It went from being a string to being an object.
* Extension now only shows currency, but not amount:
image


We need a short term fix to solve this issue, which is what Luca is doing here, but in the long term, I think most of us agree we need to adjust a little on how we handle Paddle stuff.

ilasw: I was referring to frequently changing the structure of prices due to new product requirements—almost every sprint. I moved these changes to the BE since we already duplicated the old client logic (I’m just updating it, not introducing entirely new logic).

Let me know if this makes things clearer, or if you prefer, we can quickly sync.

capJavert (approved):
I don't have anything much to add, trusting you guys to fix it as best as we can. Then we can work out a better way of handling this. 

I agree with @rebelchris we should just move it to the API, and if this query needs to support experimentation support we should add it through additional parameter, loading user data in GB server side and so on. 

Having dynamic prices and supporting experimentation has become product requirement so I think we should invest time to support it.

rebelchris (approved):

AmarTrebinjac (approved):
Checkout both backend and frontend, and it seems to be working correctly 🙂  
Screenshot 2025-03-04 at 11 29 44

PR #2685, opened 2025-03-04
fix: skip caching garm response

capJavert:
- skip caching garm response (when breaker is open since it already locally returns), this will allow for service to return instantly and fill cache with real response after error
- handle balance account missing to avoid triggering garm, this error is expected for all new users and handled as zero balance


capJavert:
Merging so I can close the garm on prod

PR #2686, opened 2025-03-05
feat: award post

capJavert:
https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1435172886/RFC-108+-+Credits+backend#Award-post-and-comment

- awardPost mutation
- save award under UserPost and connect to transaction
- aggregate `awards` under Post

- [ ] Tests

### DIFF
--- a/src/entity/posts/Post.ts
+++ b/src/entity/posts/Post.ts
@@ -288,4 +288,8 @@ export class Post {
     },
   )
   public codeSnippets: Promise<PostCodeSnippet[]>;
+
+  @Column({ type: 'integer', default: 0 })
+  @Index('IDX_post_awards')
+  awards: number;
 }
--- a/src/entity/user/UserPost.ts
+++ b/src/entity/user/UserPost.ts
@@ -10,9 +10,11 @@ import {
 import type { Post } from '../posts';
 import type { User } from './User';
 import { UserVote } from '../../types';
+import type { UserTransaction } from './UserTransaction';
 
 export type UserPostFlags = Partial<{
   feedbackDismiss: boolean;
+  awardId: string;
 }>;
 
 export type UserPostFlagsPublic = Pick<UserPostFlags, 'feedbackDismiss'>;
@@ -57,4 +59,14 @@ export class UserPost {
     onDelete: 'CASCADE',
   })
   user: Promise<User>;
+
+  @Column({ type: 'uuid', nullable: true })
+  @Index({ unique: true })
+  awardTransactionId: string;
+
+  @ManyToOne('UserTransaction', {
+    lazy: true,
+    onDelete: 'SET NULL',
+  })
+  awardTransaction: Promise<UserTransaction>;
 }
--- a/src/migration/1741187989804-AwardPostTriggers.ts
+++ b/src/migration/1741187989804-AwardPostTriggers.ts
@@ -1,0 +1,95 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AwardPostTriggers1741187989804 implements MigrationInterface {
+  name = 'AwardPostTriggers1741187989804';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "post" ADD "awards" integer NOT NULL DEFAULT '0'`,
+    );
+
+    await queryRunner.query(`
+      CREATE OR REPLACE FUNCTION user_post_award_insert_trigger_function()
+      RETURNS TRIGGER AS $$
+      BEGIN
+          IF NEW."awardTransactionId" IS NOT NULL THEN
+              UPDATE post SET awards = awards + 1 WHERE id = NEW."postId";
+          END IF;
+
+          RETURN NEW;
+      END;
+      $$ LANGUAGE plpgsql;
+    `);
+    await queryRunner.query(`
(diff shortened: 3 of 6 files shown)


capJavert:
Opening for review while I work on frontend integration.

src/schema/njord.ts
```diff
@@ -70,6 +97,87 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          postId: post.id,
+          userId: ctx.userId,
+          awardTransactionId: transaction.id,
+          flags: {
+            awardId: transaction.productId,
+          },
+        })
+        .onConflict(
+          `("postId", "userId") DO UPDATE SET "awardTransactionId" = EXCLUDED."awardTransactionId", "flags" = user_post.flags || EXCLUDED."flags"`,
+        )
```
capJavert, line 178: Since UserPost can exist due to upvotes/downvotes on conflict we upsert the award columns

src/schema/njord.ts
```diff
@@ -70,6 +97,87 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        note,
+      });
+
+      if (!transaction.productId) {
+        throw new Error('Product missing from transaction');
+      }
+
+      await ctx.con
+        .getRepository(UserPost)
+        .createQueryBuilder()
```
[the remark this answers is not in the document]
capJavert: Hmm actually, I can fix it by wrapping everything in an outside transaction and then do a njord call at the end. That will make sure nothing can go wrong after njord call.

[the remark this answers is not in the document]
idoshamun: Shouldn't we call njord and update the API when we get the pubsub message? Similar to CDC in a way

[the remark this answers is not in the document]
capJavert: We get it all here because transfer completes sync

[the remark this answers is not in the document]
capJavert: I solved it now in one transaction, everything in the DB happens before njord, if we do CDC here it means we will have to do whole async flow when user awards post which UX wise is not worth it and will be slower.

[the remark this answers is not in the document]
capJavert: - we create transaction id
- fill up transaction
- send all the parameters in njord
- if it succeeds we commit transaction
- if njord fails we revert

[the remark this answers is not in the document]
idoshamun: flow you mentioned sounds perfect! i didn't review the code just what i understood from your comment before. sorry about that 🙈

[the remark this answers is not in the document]
capJavert: yeah, sorry for pinging too early, if you do have time to review it is all ready now for this iteration

src/entity/user/UserPost.ts
```diff
@@ -57,4 +59,14 @@ export class UserPost {
     onDelete: 'CASCADE',
   })
   user: Promise<User>;
+
```
idoshamun, line 62: Shouldn't we save here the amount of the award?

src/entity/posts/Post.ts
```diff
@@ -288,4 +288,8 @@ export class Post {
     },
   )
   public codeSnippets: Promise<PostCodeSnippet[]>;
+
+  @Column({ type: 'integer', default: 0 })
+  @Index('IDX_post_awards')
+  awards: number;
```
idoshamun, line 294: Don't we need to store also the total amount? I don't remember what we show on the frontend

src/migration/1741187989804-AwardPostTriggers.ts
```diff
@@ -0,0 +1,95 @@
...
+    await queryRunner.query(
+      `ALTER TABLE "post" ADD "awards" integer NOT NULL DEFAULT '0'`,
+    );
+
+    await queryRunner.query(`
+      CREATE OR REPLACE FUNCTION user_post_award_insert_trigger_function()
+      RETURNS TRIGGER AS $$
+      BEGIN
+          IF NEW."awardTransactionId" IS NOT NULL THEN
+              UPDATE post SET awards = awards + 1 WHERE id = NEW."postId";
```
idoshamun, line 16: Same question about the count vs the sum as before

src/schema/njord.ts
```diff
@@ -44,33 +106,143 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
   Mutation: {
     awardUser: async (
       _,
-      { productId, receiverId, note }: AwardInput,
+      props: UserAwardInput,
       ctx: AuthContext,
-    ) => {
-      const product = await queryReadReplica<Pick<Product, 'type'>>(
+    ): Promise<TransactionCreated> => {
```
idoshamun, line 111: The function feels very long, should we consider splitting to smaller steps?

idoshamun (approved):

src/entity/user/UserPost.ts
capJavert: it is saved under `UserTransaction`

src/entity/posts/Post.ts
capJavert: If this is what you meant? https://dailydotdev.slack.com/archives/C07VA1FJTDK/p1741273583885919

idoshamun: yes, perfect

rebelchris (approved):
Looks good to me

PR #2688, opened 2025-03-07
feat: products query

capJavert:
I will add some products to the prod DB manually and also update real one once all assets are ready from design.

Products are returned sorted by value ASC.


rebelchris (approved):
looks good

PR #2690, opened 2025-03-10
feat: award comment

capJavert:
- add support for comment award
- add columns and triggers same as for posts
- expose new fields on comment entities

### DIFF
--- a/src/common/njord.ts
+++ b/src/common/njord.ts
@@ -29,6 +29,11 @@ import { remoteConfig } from '../remoteConfig';
 import { queryReadReplica } from './queryReadReplica';
 import { UserPost } from '../entity/user/UserPost';
 import { Post } from '../entity/posts/Post';
+import { Comment } from '../entity';
+import { UserComment } from '../entity/user/UserComment';
+import { saveComment } from '../schema/comments';
+import { generateShortId } from '../ids';
+import { checkWithVordr, VordrFilterType } from './vordr';
 
 const transport = createGrpcTransport({
   baseUrl: process.env.NJORD_ORIGIN,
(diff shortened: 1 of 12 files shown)


src/common/njord.ts
```diff
@@ -447,3 +452,131 @@ export const awardPost = async (
...
+        ctx,
+        entityManager,
+        productId: product.id,
+        receiverId: comment.userId,
+        note,
+      });
+
+      if (!transaction.productId) {
+        throw new Error('Product missing from transaction');
+      }
```
rebelchris, line 517: In what case does this happen?

src/entity/user/UserComment.ts
```diff
@@ -10,15 +10,20 @@ import {
...
+  awardId: string;
 }>;
 
 @Entity()
 @Index(['commentId', 'userId'], { unique: true })
 @Index(['userId', 'vote', 'votedAt'])
 export class UserComment {
+  get awarded(): boolean {
+    return !!this.awardTransactionId;
+  }
```
rebelchris, line 25: Just check that this works, had some issues in the past with it due to typeorm layer on top.

rebelchris (approved):
Non blocking comments

src/common/njord.ts
capJavert: None currently, but I made `productId` optional so just types being safe, I will look to add NOT NULL constraint when we finish the project depending on the needs

src/entity/user/UserComment.ts
capJavert: yep, it is used in graphorm, it could be this or in graphorm do `{ ...entity, awarded: !!entity.awardTransactionId }`

PR #2694, opened 2025-03-12
feat: iOS Plus

omBratteng:


src/plusSubscription.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const con = await createOrGetConnection();
+
+  // Convert all keys in data to null when status is expired
+  const subscriptionFlags = Object.fromEntries(
+    Object.entries(clone).map(([key, value]) => [
+      key,
+      status === UserSubscriptionStatus.Expired ? null : value,
+    ]),
+  );
```
capJavert, line 37: This seems a bit magical, and error prone if something is added without knowing it will be removed, can we specifically set each key to null?

src/common/user.ts
```diff
@@ -13,34 +12,45 @@ import {
...
       where: { id: userId },
     });
-    if (user?.subscriptionFlags?.subscriptionId) {
+    if (subscriptionFlags?.subscriptionId) {
+      if (
+        subscriptionFlags?.provider === SubscriptionProvider.AppleStoreKit &&
+        subscriptionFlags?.status === UserSubscriptionStatus.Active
+      ) {
+        throw new Error(
+          'Apple subscriptions are not supported for user deletion',
```
capJavert, line 42: This will then block user deletion?

capJavert: How are we handling it then on product side?

capJavert: can we also add a test for this?

src/entity/user/User.ts
```diff
@@ -33,17 +33,32 @@ export enum SubscriptionProvider {
...
   createdAt: Date;
-  subscriptionId: string;
-  gifterId?: string;
-  giftExpirationDate?: Date;
   provider: SubscriptionProvider;
-
-  // StoreKit flags
-  appAccountToken?: string; // StoreKit app account token (UUID)
-}>;
+  status: UserSubscriptionStatus;
```
capJavert, line 58: paddle side won't have status?

src/routes/webhooks/apple.ts
```diff
@@ -57,6 +72,62 @@ interface AppleNotificationRequest {
...
+        return UserSubscriptionStatus.Cancelled;
+      }
+    case NotificationTypeV2.GRACE_PERIOD_EXPIRED:
+      return UserSubscriptionStatus.Cancelled;
+    case NotificationTypeV2.REFUND:
+    case NotificationTypeV2.EXPIRED:
+    case NotificationTypeV2.REVOKE: // We don't support Family Sharing, but to be on the safe side
+      return UserSubscriptionStatus.Expired;
+    default:
+      return null;
```
capJavert, line 127: does null mean we don't know how to handle this notificationType or sometimes it should be `null`?

src/routes/webhooks/paddle.ts
```diff
@@ -100,6 +104,7 @@ export const updateUserSubscription = async ({
         createdAt: state ? data.data?.startedAt : null,
         subscriptionId: state ? data.data?.id : null,
         provider: state ? SubscriptionProvider.Paddle : null,
+        status: state ? UserSubscriptionStatus.Active : null,
```
capJavert, line 107: Ah I see now paddle will also have active status, what about old subs?

capJavert (commented):
Mainly some questions and interop with paddle side

src/routes/webhooks/paddle.ts
omBratteng: I'll add it to existing ones

src/common/user.ts
omBratteng: Yeah, it will block user deletion. Forgot to add test.
On product side I'll disable the delete button

omBratteng: With a message to tell them to cancel the subscription

src/routes/webhooks/apple.ts
omBratteng: It means it's a notification type we don't know how to handle, and we'll log and throw an error

src/plusSubscription.ts
omBratteng: Got to enjoy a bit of magic from time to time. I just didn't really like having to do the ternary check for each item like it's done in the paddle version of this.

src/routes/webhooks/apple.ts
capJavert: ah I see you throw after evaluation, maybe better to throw from here directly? Then this method can also just have `UserSubscriptionStatus` as return type which is cleaner?

omBratteng: Yeah, can do that.

src/common/user.ts
capJavert: Nice, so handled on frontend/app side?

omBratteng: Not yet, but I'll add it

omBratteng: Tests have been added.

src/plusSubscription.ts
capJavert: I would rather do it that way because at least its typed then, or since this is sub flags specifically, maybe just set new object with `{ status }` if that is the only thing that needs to be left after sub expires?

omBratteng: Yeah, we could just set the status field and remove all other fields

omBratteng: Refactored

src/routes/webhooks/apple.ts
```diff
@@ -57,6 +72,66 @@ interface AppleNotificationRequest {
...
+};
+
+const renewalInfoToSubscriptionFlags = (
+  data: JWSRenewalInfoDecodedPayload,
+): UserSubscriptionFlags => {
+  const cycle =
+    productIdToCycle[data.autoRenewProductId as keyof typeof productIdToCycle];
+
+  if (isNullOrUndefined(cycle)) {
+    logger.error({ data }, '[Apple StoreKit]: Invalid auto renew product ID');
```
capJavert, line 88: btw. for paddle we send `{ data, type: 'paddle' }` easier to filter paddle specific logs instead from log strings, so maybe instead of `[Apple StoreKit]` we can also send `{ data, type: 'storekit }`?

omBratteng: 👍

omBratteng: Replaced

src/routes/webhooks/apple.ts
```diff
@@ -98,20 +173,72 @@ export const apple = async (fastify: FastifyInstance): Promise<void> => {
...
         );
+
+        const subscriptionStatus = getSubscriptionStatus(
+          notification.notificationType,
+          notification.subtype,
+        );
+
+        const subscriptionFlags = renewalInfoToSubscriptionFlags(renewalInfo);
+
+        await updateStoreKitUserSubscription({
```
capJavert, line 225: One thing that comes to mind, can we catch if sub already exists on paddle for some reason and throw error in that case so we are aware of it?

capJavert: Same thing would be helpful in inverse case where someone has apple sub but then tries to purchase paddle

capJavert: just so we don't have overwriting subs

omBratteng: Yeah, I've been thinking about that. Not 100% sure what the best course of action is. I think it's better to let StoreKit overwrite, as we can't issue refunds on behalf of the user, but we can on Paddle.

But of course log when it happens, on both sides.

capJavert: I would say throw, and handle on case by case basis manually, we have restrictions on front side so technically it should not be visible, but let's raise it on the channel to see what Ido/product think?

omBratteng: It now throws, and I'll add some log alerts so we get notified on the cases. Yeah, shouldn't happen, but who knows 🙈

capJavert (approved):

PR #2701, opened 2025-03-14
feat: purchase cores

capJavert:
[RFC section](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1435172886/RFC-108+-+Credits+backend#Paddle)

### DIFF
--- a/src/common/paddle.ts
+++ b/src/common/paddle.ts
@@ -2,9 +2,18 @@ import {
   type Environment,
   Paddle,
   type Subscription,
+  type TransactionCompletedEvent,
+  type TransactionCreatedEvent,
+  TransactionPaymentFailedEvent,
+  type TransactionReadyEvent,
+  type TransactionUpdatedEvent,
 } from '@paddle/paddle-node-sdk';
 import { logger } from '../logger';
 import { PricingPreviewLineItem } from '@paddle/paddle-node-sdk/dist/types/entities/pricing-preview';
+import { remoteConfig } from '../remoteConfig';
+import { z } from 'zod';
+import { UserTransaction } from '../entity/user/UserTransaction';
+import type { DataSource, EntityManager } from 'typeorm';
 
 export const paddleInstance = new Paddle(process.env.PADDLE_API_KEY, {
   environment: process.env.PADDLE_ENVIRONMENT as Environment,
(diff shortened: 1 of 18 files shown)


src/common/paddle.ts
```diff
@@ -54,3 +63,125 @@ export const getPriceFromPaddleItem = (
...
+  Plus = 'plus',
+  Core = 'core',
+}
+
+export const getProductPurchaseType = ({
+  id,
+}: {
+  id: string;
+}): ProductPurchaseType => {
+  if (!remoteConfig.vars.coreProductId) {
```
idoshamun, line 77: Learning from the mistakes of Plus, is it possible to use Paddle API somehow to detect if it was a core purchase?
Maintaining product ids on GB is something we often forget to do which breaks the flow, especially with experimentation. So ideally if we can have something more bullet proof it would be great

src/common/paddle.ts
```diff
@@ -54,3 +63,125 @@ export const getPriceFromPaddleItem = (
...
+
+  switch (id) {
+    case remoteConfig.vars.coreProductId:
+      return ProductPurchaseType.Core;
+    default:
+      return ProductPurchaseType.Plus;
+  }
+};
+
+const paddleNotificationCustomDataSchema = z.object(
```
idoshamun, line 89: Should we consider moving the schemas to a separate file?

src/routes/webhooks/paddle.ts
```diff
@@ -460,18 +476,325 @@ export const processGiftedPayment = async ({
...
+
+  return true;
+};
+
 export const processTransactionCompleted = async ({
   event,
 }: {
   event: TransactionCompletedEvent;
 }) => {
+  await notifyNewPaddleTransaction({ event });
```
idoshamun, line 515: Let's update this function to differentiate between plus and cores so we will know from slack what the user purchased

src/routes/webhooks/paddle.ts
```diff
@@ -460,18 +476,325 @@ export const processGiftedPayment = async ({
...
+      throw new Error('Transaction value changed after success');
+    }
+  }
+
+  const payload = con.getRepository(UserTransaction).create({
+    processor: UserTransactionProcessor.Paddle,
+    id: transaction?.id,
+    receiverId: data.customData.user_id,
+    status: nextStatus,
+    productId: null, // no product user is buying cores directly
```
rebelchris, line 584: We don't want this as fixed something for future?

rebelchris (approved):
Love to see Zod in action.
Besides Ido's comments nothing major on my side.

src/routes/webhooks/paddle.ts
capJavert: I though of adding generic `Product` for cores, but since Product entities are something purchased with cores it made more sense to leave `null`, wdyt?

src/routes/webhooks/paddle.ts
capJavert: Yep, gonna update

src/common/paddle.ts
capJavert: Yes, but I also expect a lot more usage in the future so maybe each `common` can have its own schemas like this. I'll think about it 👍

src/routes/webhooks/paddle.ts
rebelchris: Yeah let's leave null then for now

src/common/paddle.ts
capJavert: Yep, good think paddle prefixes them so at least you can see that you passed a wrong one eg. `pro_`, `pri_`..

src/routes/webhooks/paddle.ts
capJavert: Added 0375c52b8755b4e4ffa2caf3ad8608fa8e848001

capJavert:
Missing some tests, I am gonna add them as I work on other stuff, want to merge this so we can have purchasing flow on preview.

src/dataLoaderService.ts
```diff
@@ -125,4 +126,23 @@ export class DataLoaderService {
...
+  get userBalance() {
+    return this.getLoader<
+      {
+        userId: string;
+      },
+      GetBalanceResult
+    >({
+      type: 'userBalance',
+      loadFn: async ({ userId }) => {
+        return getBalance({
```
idoshamun, line 139: Instead of "hacking" context, shouldn't we just provide `userId` as a parameter?

src/routes/webhooks/paddle.ts
```diff
@@ -418,6 +418,111 @@ const notifyNewPaddleTransaction = async ({
   await webhooks.transactions.send({ blocks });
 };
 
+const notifyNewPaddleCoresTransaction = async ({
```
idoshamun, line 421: Can't wait for the first one to fire!! 🤩

idoshamun (approved):
Just check my comment, not blocking

src/routes/webhooks/paddle.ts
capJavert: 🚀 soon!

src/dataLoaderService.ts
capJavert: It was due to auth protection, but I do agree it can be simplified, since I use it like that all over the place, refactored to just `{ userId }` 98f6bdec62641a30fd029de516b8a8ea894b166d

PR #2705, opened 2025-03-18
feat: log iOS purchases with USD

omBratteng:


src/common/utils.ts
```diff
@@ -278,5 +278,5 @@ export const getCurrencySymbol = ({
     .trim();
 };
 
-export const concatText = (a?: string, b?: string) =>
```
omBratteng, line 281: Forgot to add this to #2704 before merging

src/routes/webhooks/apple.ts
```diff
@@ -182,10 +183,21 @@ export const logAppleAnalyticsEvent = async (
...
   const extra = {
     cycle,
+    cost: currencyInUSD,
+    currency: 'USD',
     localCost: cost ? cost / 100 : undefined,
     localCurrency: data?.currency,
+    payout: {
+      total: currencyInUSD * 100,
+      grandTotal: currencyInUSD * 100,
+    },
```
omBratteng, line 200: In cents

src/cron/fetchExchangeRates.ts
```diff
@@ -0,0 +1,16 @@
+import { getOpenExchangeRates } from '../integrations/openExchangeRates';
+import { logger } from '../logger';
+import type { Cron } from './cron';
+
+export const fetchExchangeRates: Cron = {
+  name: 'fetch-exchange-rates',
+  handler: async () => {
+    logger.info('Fetching exchange rates');
+    try {
+      await getOpenExchangeRates();
```
capJavert, line 10: Why do we need the cron, just calling a function should be fine, and then we cache it to redis until expiration?

capJavert: All following calls will be fast.

src/common/utils.ts
```diff
@@ -278,5 +278,5 @@ export const getCurrencySymbol = ({
...
-  [a, b].filter(Boolean).join(`\n`);
+export const concatTextToNewline = (...args: Array<string | undefined>) =>
```
capJavert, line 281: now we are in the rabbit whole, but could have also accept separator 😆 

{
   strings: [...]
   separator: '\n'
}

src/integrations/openExchangeRates.ts
```diff
@@ -0,0 +1,71 @@
...
+export const getOpenExchangeRates = async (): Promise<void> => {
+  if (!env.OPEN_EXCHANGE_RATES_APP_ID) {
+    throw new Error('OPEN_EXCHANGE_RATES_APP_ID is not set');
+  }
+
+  try {
+    const params = new URLSearchParams({
+      app_id: env.OPEN_EXCHANGE_RATES_APP_ID,
+    });
+    const data = await retryFetchParse<OpenExchangeRates>(
```
capJavert, line 43: Could connect Garmr, it has retry support and we can also fallback and trigger proper errors with it if this fails for any reason as 3party service.

capJavert (commented):
Mostly don't see why cron is needed, this could be similar to `getBalance` where we fetch and cache on demand.

src/cron/fetchExchangeRates.ts
omBratteng: Because we only have 1000 API calls per month, so it'll update the exchange rates once per hour

src/common/utils.ts
omBratteng: I renamed it to `concatTextToNewline` so it only concats to newline

src/cron/fetchExchangeRates.ts
capJavert: But same effect would be done if we just set expiration in redis to 1 hour, it would be even less calls since if not needed it would not fire as cron would do.

src/common/utils.ts
capJavert: I am just against this kind of helpers widespread use, they don't add anything else and often need to be expanded when new use arises randomly. It is what it is I guess.

src/cron/fetchExchangeRates.ts
omBratteng: Refactored and set TTL to 1 hour and it fetches in the webhook pre-handler if it doesn't exist in redis

idoshamun (approved):

src/integrations/openExchangeRates.ts
```diff
@@ -0,0 +1,80 @@
...
+      StorageKey.OpenExchangeRates,
+      data.rates,
+      REDIS_EXPIRATION,
+    );
+
+    return data.rates;
+  } catch (_err) {
+    const err = _err as Error;
+    logger.error({ err }, 'Error fetching open exchange rates');
+    throw err;
```
capJavert, line 54: This will throw and break webhook processing if service is unavailable?

capJavert (approved):

src/integrations/openExchangeRates.ts
omBratteng: Yeah, I'll remove the throw and return empty object instead. It webhook should still pass if service is unavailable

src/integrations/openExchangeRates.ts
rebelchris: I think it might be a good idea as well.
Just another failsafe

rebelchris (approved):
Nothing blocking from my side, but check Ante's comment

src/integrations/openExchangeRates.ts
omBratteng: Added it

PR #2709, opened 2025-03-20
fix: log `notification` in sandbox endpoint

omBratteng:
- Add the `notification` payload to the error log when user does not have access to sandbox
- Add the `environment` to all logs too

### DIFF
--- a/src/routes/webhooks/apple.ts
+++ b/src/routes/webhooks/apple.ts
@@ -277,7 +277,7 @@ const handleNotifcationRequest = async (
       !remoteConfig.vars.approvedStoreKitSandboxUsers?.includes(user.id)
     ) {
       logger.error(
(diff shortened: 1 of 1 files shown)


src/routes/webhooks/apple.ts
```diff
@@ -233,7 +237,11 @@ const handleNotifcationRequest = async (
     // Don't proceed any further if it's a test notification
     if (notification.notificationType === NotificationTypeV2.TEST) {
       logger.info(
-        { notification, provider: SubscriptionProvider.AppleStoreKit },
+        {
+          environment,
```
capJavert, line 241: We should start using child loggers in pino, easy to forget adding stuff https://getpino.io/#/docs/api?id=base-object, well at some point we will..

omBratteng: I miss python and its logger 😬

PR #2710, opened 2025-03-21
feat: add credits for subscription

rebelchris:
Add cores to people who complete purchases for plus subscription.
The 1 core value is stores in remoteConfig so we can easily update it.

### DIFF
--- a/src/remoteConfig.ts
+++ b/src/remoteConfig.ts
@@ -22,6 +22,7 @@ type RemoteConfigValue = {
   enableBalance: boolean;
   approvedStoreKitSandboxUsers: string[];
   coreProductId: string;
+  coreModifier: number;
 };
 
 class RemoteConfig {
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -33,10 +33,12 @@ import {
 } from '../../integrations/analytics';
 import { JsonContains, type DataSource, type EntityManager } from 'typeorm';
 import {
+  coreProductCustomDataSchema,
   getPaddleTransactionData,
   getTransactionForProviderId,
   isCoreTransaction,
   paddleInstance,
+  paddleTransactionSchema,
 } from '../../common/paddle';
 import { addMilliseconds } from 'date-fns';
 import {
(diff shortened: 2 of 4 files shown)


src/routes/webhooks/paddle.ts
```diff
@@ -654,9 +696,12 @@ export const processTransactionCompleted = async ({
   }
 
   const { gifter_id } = (event?.data?.customData ?? {}) as PaddleCustomData;
-
   if (gifter_id) {
     await processGiftedPayment({ event });
+  } else {
```
rebelchris, line 701: Used the else here, otherwise it would also apply on gifters?

src/routes/webhooks/paddle.ts
```diff
@@ -523,6 +523,57 @@ const notifyNewPaddleCoresTransaction = async ({
   await webhooks.transactions.send({ blocks });
 };
 
+const assignSubscriptionCores = async ({
+  event: { data },
+}: {
+  event: TransactionCompletedEvent;
+}) => {
+  const con = await createOrGetConnection();
+  const { user_id } = data.customData as PaddleCustomData;
```
capJavert, line 532: you can use schema I use for cores probably to get proper type

capJavert: no need to validate optionals then below separately

src/routes/webhooks/paddle.ts
```diff
@@ -523,6 +523,57 @@ const notifyNewPaddleCoresTransaction = async ({
...
+        processor: UserTransactionProcessor.Paddle,
+        receiverId: user_id,
+        productId: null, // no product user is buying cores directly
+        senderId: null, // no sender, user is buying cores
+        status: UserTransactionStatus.Success,
+        value: coresForPayment,
+        fee: 0, // no fee when buying cores
+        request: {},
+        flags: {
+          providerId: data.id,
```
capJavert, line 567: let's add a note `Plus subscription`

src/routes/webhooks/paddle.ts
capJavert: I mean probably it should apply, let's check with product

src/remoteConfig.ts
```diff
@@ -22,6 +22,7 @@ type RemoteConfigValue = {
   enableBalance: boolean;
   approvedStoreKitSandboxUsers: string[];
   coreProductId: string;
+  coreModifier: number;
```
capJavert, line 25: let's just add if coreModifier is undefined we don't run `assignSubscriptionCores`

capJavert (commented):
Good work, nothing blocking, just some comments to verify.

src/routes/webhooks/paddle.ts
```diff
@@ -523,6 +523,57 @@ const notifyNewPaddleCoresTransaction = async ({
...
+  if (!user_id) {
+    logger.error(
+      { provider: SubscriptionProvider.Paddle, data },
+      'User ID missing in payload',
+    );
+    return;
+  }
+
+  const modifier = remoteConfig.vars?.coreModifier || 0.001;
+  const total = parseInt(data.details?.totals?.grandTotal || '0');
```
capJavert, line 543: Let's check with product if they want tax or no tax here?

PR #2712, opened 2025-03-21
feat: limit cores access

capJavert:
- check region on registration and deduce cores access
- return new field from boot so we can use it to check access on frontend
- rules from remote config (first rule applies)

TODO
- [x] tests for awarding authors

RFC section: https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1435172886/RFC-108+-+Credits+backend#Limiting-credits-access-to-some-users

Changes on hemidall https://github.com/dailydotdev/heimdall/pull/121


src/remoteConfig.ts
```diff
@@ -22,6 +23,10 @@ type RemoteConfigValue = {
   enableBalance: boolean;
   approvedStoreKitSandboxUsers: string[];
   coreProductId: string;
+  coresRoleRules: {
+    regions: string[];
+    role: CoresRole;
+  }[];
```
capJavert, line 29: We can specify multiple rules to block which regions are restricted and how, first found rule applies.

src/common/njord.ts
```diff
@@ -424,13 +425,27 @@ const canAward = async ({
...
   }
+
+  if (
+    (await checkCoresAccess({
+      ctx,
+      userId: receiverId,
+      requiredRole: CoresRole.Creator,
+    })) === false
+  ) {
+    throw new ForbiddenError('You can not award this user');
```
idoshamun, line 447: Is this user facing?

src/dataLoaderService.ts
```diff
@@ -143,4 +145,22 @@ export class DataLoaderService {
...
+      type: 'user',
+      loadFn: async ({ userId }) => {
+        if (!userId) {
+          return null;
+        }
+
+        return queryReadReplica(this.ctx.con, async ({ queryRunner }) => {
+          return queryRunner.manager
+            .getRepository(User)
+            .findOneBy({ id: userId });
```
idoshamun, line 160: should we load specific fields? i feel like querying the readme and some other long columns can be wasteful

src/common/njord.ts
capJavert: Yeah error should propagate, but on frontend side we can show whatever, why do you ask?

src/routes/webhooks/paddle.ts
```diff
@@ -636,6 +638,19 @@ export const processTransactionCompleted = async ({
           event,
         });
 
+        const user = await entityManager.getRepository(User).findOneByOrFail({
```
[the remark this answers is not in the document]
capJavert: done 2a9b6c07248f26d25bb95342ba7b7307687ae31a

src/dataLoaderService.ts
capJavert: done 2a9b6c07248f26d25bb95342ba7b7307687ae31a

src/common/njord.ts
idoshamun: just because of the copy. if we're not showing this copy, then it's fine

idoshamun (approved):

src/common/njord.ts
capJavert: I will note down to verify copy with product on frontend side

PR #2716, opened 2025-03-26
feat: cache cores price preview

capJavert:
- paddle is rate limiting us so for now cache in redis

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -30,6 +30,7 @@ export enum StorageTopic {
   CIO = 'customer_io',
   Streak = 'streak',
   Njord = 'njord',
+  Paddle = 'paddle',
 }
 
 export enum StorageKey {
@@ -38,6 +39,7 @@ export enum StorageKey {
   Reset = 'reset',
   CoresBalance = 'cores_balance',
   OpenExchangeRates = 'open_exchange_rates',
+  PricingPreview = 'pricing_preview',
 }
 
 export const generateStorageKey = (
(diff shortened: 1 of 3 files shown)


src/schema/paddle.ts
```diff
@@ -148,13 +152,37 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
-      const pricePreview = await paddleInstance?.pricingPreview.preview({
-        items: Object.keys(featureValue).map((priceId) => ({
-          priceId,
-          quantity: 1,
-        })),
-        address: region ? { countryCode: region as CountryCode } : undefined,
-      });
+      const hmac = createHmac('sha1', process.env.SLACK_SIGNING_SECRET);
+      hmac.update(Object.keys(featureValue).toString());
+      const pricesHash = hmac.digest().toString('hex');
```
capJavert, line 157: We hash the prices so we don't have to cache paddle cores per user id and all pricing ids

src/config.ts
```diff
@@ -38,6 +39,8 @@ export enum StorageKey {
   Reset = 'reset',
   CoresBalance = 'cores_balance',
   OpenExchangeRates = 'open_exchange_rates',
+  PricingPreview = 'pricing_preview',
```
sshanzel, line 42:
```suggestion
  PricingPreviewPlus = 'pricing_preview_plus',
```

sshanzel (approved):
🚀

src/config.ts
capJavert: ba7cc443ffe02617bd4ef9916de82f1f6abfc1da done ✅

PR #2719, opened 2025-03-31
feat: transaction error handling

capJavert:
- handle njord errors and record the status in `UserTransaction` row
- revert any entities for awards when njord fails
- return balance of user on transfer errors
- handle error messages

TODO:
- [x] Save errors in transaction flags
- [x] Tests

### DIFF
--- a/src/common/njord.ts
+++ b/src/common/njord.ts
@@ -48,6 +48,8 @@ import { generateShortId } from '../ids';
 import { checkWithVordr, VordrFilterType } from './vordr';
 import { CoresRole } from '../types';
 import { GraphQLError } from 'graphql';
+import { randomUUID } from 'node:crypto';
+import { logger } from '../logger';
 
 const transport = createGrpcTransport({
   baseUrl: process.env.NJORD_ORIGIN,
@@ -86,10 +88,12 @@ export const createTransaction = createAuthProtectedFn(
   async ({
     ctx,
     entityManager,
+    id,
     productId,
     receiverId,
     note,
   }: TransactionProps & {
+    id?: string;
     entityManager: EntityManager;
   }): Promise<UserTransaction> => {
     const { userId: senderId } = ctx;
@@ -101,6 +105,7 @@ export const createTransaction = createAuthProtectedFn(
     const userTransaction = entityManager
       .getRepository(UserTransaction)
       .create({
+        id: id || randomUUID(),
         processor: UserTransactionProcessor.Njord,
         receiverId,
         status: UserTransactionStatus.Success,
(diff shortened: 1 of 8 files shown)


src/common/utils.ts
```diff
@@ -20,7 +20,6 @@ export const deletedPost = {
   id: '404',
 };
 
-// TODO feat/transactions not created yet on db side
 export const systemUser = {
```
capJavert, line 23: created system user https://app.daily.dev/system

src/common/njord.ts
```diff
@@ -225,22 +244,22 @@ export const purchaseCores = async ({
...
-    }
+  if (transaction.productId) {
+    throw new Error('Purchase cores transaction can not have product');
+  }
 
-    const njordClient = getNjordClient();
+  const njordClient = getNjordClient();
 
-    const { results } = await njordClient.transfer({
+  const response = await garmNjordService.execute(async () => {
```
idoshamun, line 261: Why don't we add garmr as part of njord client like other clients we have?

src/common/njord.ts
```diff
@@ -709,12 +763,34 @@ export const awardComment = async (
...
+      try {
+        const transfer = await transferCores({
+          ctx,
+          transaction,
+        });
 
-      return { transaction, transfer };
+        return { transaction, transfer };
+      } catch (error) {
+        if (error instanceof TransferError) {
```
idoshamun, line 774: Same question about db availability here

src/common/njord.ts
```diff
@@ -709,12 +763,34 @@ export const awardComment = async (
...
+          if (newComment) {
+            await entityManager.getRepository(Comment).delete({
+              id: newComment.id,
+            });
+          }
+
+          await entityManager.getRepository(UserComment).delete({
+            awardTransactionId: transaction.id,
+          });
```
idoshamun, line 783: Should we use transaction for both?

idoshamun (commented):
Overall looks good but left some comments to make sure we aren't missing anything. I always get confused with the distributed nature of the cores and how we make it consistent

src/common/njord.ts
```diff
@@ -581,12 +617,28 @@ export const awardPost = async (
...
+      try {
+        const transfer = await transferCores({
+          ctx,
+          transaction,
+        });
+
+        return { transaction, transfer };
+      } catch (error) {
+        if (error instanceof TransferError) {
+          await entityManager.getRepository(UserPost).delete({
```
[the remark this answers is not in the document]
capJavert: it would not even start since transaction would not be able to start

src/common/njord.ts
capJavert: it is in transaction

src/common/njord.ts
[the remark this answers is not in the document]
idoshamun: i know it's very edge casey but it could go offline midway through. that's what i mean.
so we managed to create db transaction, initiated njord transaction and now on error the db isn't available

src/common/njord.ts
idoshamun: yes i missed it 🤦 
i always assume the `entityManager` syntax is out of transaction for some reaosn

src/common/njord.ts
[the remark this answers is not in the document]
capJavert: If db is not available on error it would be the same as if db is unavailable on success and we can't commit transaction, it would fail with njord side being succesful.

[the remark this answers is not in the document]
capJavert: In above specific case everything would be alright because:
- njord errored, no balance was passed to user
- transaction would not be commited on api side
- only issue would be that we don't have any log of it because db was not available

We would have an error in log though so we can investigate why it happened.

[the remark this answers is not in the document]
capJavert: In success case where njord suceeded and then db is not available to commit transaction there would be issue where balance was moved from one user to other but user would not see the award represented in the UI.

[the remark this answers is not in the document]
idoshamun: yes please and maybe we should add to the backlog to listen to async events as well. since sync/async should be idempotent. in best case scenario, on sync we update the app transaction and async see that the transaction is up to date. in worst case, async updates a transaction that was stale due to db wasn't available or any other reason

idoshamun (approved):

src/common/njord.ts
```diff
@@ -452,276 +481,446 @@ export const awardUser = async (
...
+    logger.error(
+      {
+        context: 'award',
+        err: error,
+        props,
+        userId: ctx.userId,
+        transactionId: newTransactionId,
+      },
+      'Award error',
+    );
```
capJavert, line 688: In case of critical error we log all the data for mutation, current user and transaction id. That way even if transaction is not commited on API side after njord is finished we can still deduce from which njord transfer error occured based on `transactionId` and `idempotencyKey` in njord.

src/common/njord.ts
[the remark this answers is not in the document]
capJavert: I added logging so we can manually correlate for now. https://github.com/dailydotdev/daily-api/pull/2719/files#r2022623540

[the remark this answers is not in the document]
capJavert: We do listen for async events but we don't do anything with them currently.

src/common/njord.ts
```diff
@@ -146,27 +162,30 @@ export const transferCores = createAuthProtectedFn(
...
+    if (!transaction.senderId) {
+      throw new Error('No sender id');
+    }
 
-      const { results } = await njordClient.transfer({
+    const senderId = transaction.senderId;
+    const receiverId = transaction.receiverId;
+
+    const response = await garmNjordService.execute(async () => {
+      const response = await njordClient.transfer({
```
omBratteng, line 177: Couldn't we just return this directly?

capJavert: We want to await because if you just return `execute` will not catch promise rejection.

idoshamun (approved):

omBratteng (approved):

PR #2720, opened 2025-03-31
refactor: pricing ids removal from remote config

sshanzel:
- Removed the pricing ids.
- Used the billing cycle from the event data.
- Refactored the analytics function to use a function to get the `extra` values. Was a bit hard to follow which one was going to have value and which ones would not.

### DIFF
--- a/src/remoteConfig.ts
+++ b/src/remoteConfig.ts
@@ -1,7 +1,6 @@
 import { GrowthBook } from '@growthbook/growthbook';
 import { logger } from './logger';
 import { isProd } from './common/utils';
-import { SubscriptionCycles } from './paddle';
 import type { CoresRole } from './types';
 
 type RemoteConfigValue = {
@@ -10,7 +9,6 @@ type RemoteConfigValue = {
   vordrIps: string[];
   blockedCountries: string[];
   ignoredWorkEmailDomains: string[];
-  pricingIds: Record<string, SubscriptionCycles>;
   origins: string[];
   clickbaitTitleProbabilityThreshold: number;
   plusCustomFeed: boolean;
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -3,15 +3,14 @@ import {
   EventName,
   TransactionCompletedEvent,
   TransactionCreatedEvent,
-  TransactionItemNotification,
   type EventEntity,
   type SubscriptionCanceledEvent,
   type SubscriptionCreatedEvent,
-  type SubscriptionItemNotification,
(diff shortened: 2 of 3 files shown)


src/remoteConfig.ts
```diff
@@ -10,7 +9,6 @@ type RemoteConfigValue = {
   vordrIps: string[];
   blockedCountries: string[];
   ignoredWorkEmailDomains: string[];
-  pricingIds: Record<string, SubscriptionCycles>;
```
sshanzel, line 13: Webhook is no longer dependent on the defined pricing IDs from the remote config. Everything should function accordingly without needing to update anything.

src/routes/webhooks/paddle.ts
```diff
@@ -192,68 +180,93 @@ const planChanged = async (data: SubscriptionUpdatedEvent) => {
...
+
+  return {
+    cost: cost ? parseInt(cost) / 100 : undefined,
+    currency,
+    payment,
+    localCost: localCost ? parseInt(localCost) / 100 : undefined,
+    localCurrency,
+    payout,
+  };
+};
```
sshanzel, line 236: Basically took the deleted block of codes below to be moved here. Due to the following concerns:
- Typescript issue.
- A lot of places were using `as Type` which can be done just once if as displayed above.
- It was a bit hard to follow which properties are expected to have values and which ones are not.

src/routes/webhooks/paddle.ts
```diff
@@ -192,68 +180,93 @@ const planChanged = async (data: SubscriptionUpdatedEvent) => {
...
+  payout: TransactionPayoutTotalsNotification | null;
+}
+
+const getAnalyticsExtra = (
+  data: (
+    | SubscriptionUpdatedEvent
+    | SubscriptionCanceledEvent
+    | TransactionCompletedEvent
+  )['data'],
+): Partial<AnalyticsExtra> => {
```
capJavert, line 203: Would prefer each of the `data` types is split in its own function (even if code is duplicated a bit) but let's leave it as is since you did not really touched the logic just copied. If we add more events here we will refactor.

capJavert (approved):
Looks good, just keep an eye out when deployed if subs are processing alright.

omBratteng (approved):

src/routes/webhooks/paddle.ts
sshanzel: Can't agree more on that. I was on the verge of actually doing it. But yeah, I'll try to keep the scope small for now and do more maybe once we get more time 🙏

PR #2722, opened 2025-04-02
fix: add comment for note award on post

rebelchris:
Added comment when note is provided on awarding post

### DIFF
--- a/src/common/njord.ts
+++ b/src/common/njord.ts
@@ -672,23 +672,6 @@ export const awardPost = async (
           };
 
           await saveComment(entityManager, newComment, post.sourceId);
-          await entityManager
-            .getRepository(UserComment)
-            .createQueryBuilder()
-            .insert()
-            .into(UserComment)
-            .values({
-              commentId: newComment.id,
-              userId: ctx.userId,
-              awardTransactionId: transaction.id,
-              flags: {
-                awardId: transaction.productId,
-              },
-            })
-            .onConflict(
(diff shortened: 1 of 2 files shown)


src/common/njord.ts
```diff
@@ -644,6 +644,53 @@ export const awardPost = async (
           )
           .execute();
 
+        let newComment: Comment | undefined;
```
rebelchris, line 647: @capJavert Sort of duplication, but it needed the different order compared to the one on comments :jenkins-triggered:

src/common/njord.ts
```diff
@@ -644,6 +644,53 @@ export const awardPost = async (
...
+                type: VordrFilterType.Comment,
+                content: newComment.content,
+              },
+              ctx,
+            ),
+          };
+
+          await saveComment(entityManager, newComment, post.sourceId);
+          await entityManager
+            .getRepository(UserComment)
```
capJavert, line 676: I don't think we need UserComment here because it would imply user awarded their own comment. We already award post through `UserPost`.

src/common/njord.ts
capJavert: All good

capJavert (approved):
Need to test on frontend, but looks good.

src/common/njord.ts
rebelchris: Oh wasn't sure what UserComment was again, thought just the link between comment-user
Why we need it on comments then?

omBratteng (approved):

PR #2725, opened 2025-04-02
feat: experiment variant entity and pricing preview queries

sshanzel:
- Introduced a new entity Experiment Variant.
- New query for fetching metadata (useful for iOS).
- New query for fetching paddle pricing preview.
- I did not touch the existing pricing preview query to avoid any issues, but the code written on how to consolidate the data was based from how it was processed there.
  - We will deprecate it once everything is deployed and we get the right amount of adoption in extension.


src/common/paddle/index.ts
```diff
@@ -9,14 +9,14 @@ import {
   type TransactionReadyEvent,
   type TransactionUpdatedEvent,
 } from '@paddle/paddle-node-sdk';
-import { logger } from '../logger';
+import { logger } from '../../logger';
```
sshanzel, line 12: Just moved this file to a folder. I feel like this file will be super cluttered if I put the newly introduced block of codes here.

src/common/paddle/pricing.ts
```diff
@@ -0,0 +1,125 @@
...
+  );
+  const currencySymbol = item.formattedTotals.total.replace(/\d|\.|\s|,/g, '');
+  const priceFormatter = new Intl.NumberFormat('en-US', {
+    minimumFractionDigits: 2,
+  });
+
+  return {
+    amount: monthlyPrice,
+    formatted: `${currencySymbol}${priceFormatter.format(monthlyPrice)}`,
+  };
```
sshanzel, line 124: These formattings are from the original pricing preview query.

src/common/paddle/pricing.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  try {
+    return JSON.parse(experiment.value);
+  } catch (error) {
+    throw new Error('Invalid experiment JSON value');
+  }
+};
+
+export const getPlusPricePreview = async (ctx: AuthContext, ids: string[]) => {
```
sshanzel, line 69: Caching was based on how it originally was made from the original query. I just added the `ids` from the parameters to be part of the redis key.

src/schema/paddle.ts
```diff
@@ -231,6 +359,61 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
       };
     },
+    plusPricingMetadata: async (
+      _,
+      { variant = DEFAULT_PLUS_METADATA }: PlusMetadataArgs,
+      ctx: AuthContext,
+    ): Promise<PlusPricingMetadata[]> =>
+      await getPlusPricingMetadata(ctx.con, variant),
+    plusPricingPreview: async (_, __, ctx): Promise<PlusPricingPreview[]> => {
+      const gb = getUserGrowthBookInstace(ctx.userId);
```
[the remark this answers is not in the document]
sshanzel: I will modify this typo before merging; to avoid reviewing this a bit difficult due to unrelated changes.

capJavert (changes requested):
Accidentally pressed approved

src/common/paddle/pricing.ts
rebelchris: Shouldn't we also simply return the value if it's already monthly to avoid doing weird things?

src/schema/paddle.ts
```diff
@@ -231,6 +359,61 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
+          metadata: meta,
+          productId: item.price.productId,
+          price: {
+            monthly,
+            amount: baseAmount,
+            formatted: item.formattedTotals.total,
+          },
+          currency: {
+            code: preview.currencyCode,
+            symbol: item.formattedTotals.total.replace(/\d|\.|\s|,/g, ''),
```
rebelchris, line 408: Consider making it a named regex to know what it does :)

rebelchris (approved):
Nice, besides Ante's comment nothing blocking from my side.

src/entity/ExperimentVariant.ts
```diff
@@ -0,0 +1,24 @@
...
+  @Column({ type: 'text', default: null })
+  value: string;
+}
+
+export const getExperimentVariant = (
+  con: ConnectionManager,
+  feature: string,
+  variant: string,
+) =>
+  con.getRepository(ExperimentVariant).findOne({ where: { feature, variant } });
```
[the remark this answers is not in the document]
omBratteng: Also maybe add an index for feature variant?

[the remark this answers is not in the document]
omBratteng: Or might already do that since both are primary columns?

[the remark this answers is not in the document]
capJavert: yep, PK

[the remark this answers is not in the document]
sshanzel: Yes, they are both PKs, so need I assume.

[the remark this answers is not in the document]
sshanzel: I kinda doubt the dep cycle since if you need this entity, you will have to import this file at the end of the day. We just replaced it with the function above.

src/common/paddle/pricing.ts
sshanzel: Right. Yeah, we can early return that, we can just return `null` since we have the original value from the other props already. Wdyt? @rebelchris

src/entity/ExperimentVariant.ts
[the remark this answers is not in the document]
capJavert: Issue is that entity files are imported in migrations, and if any file that has side effects (for example redis) then all of those files are also loaded in migrations

src/common/paddle/pricing.ts
rebelchris: Yeah sounds good :) (just safer I think)

src/entity/ExperimentVariant.ts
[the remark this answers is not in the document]
capJavert: so prefer moving helpers to common or some other file which can be safely imported

src/schema/paddle.ts
```diff
@@ -231,6 +359,61 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
+          },
+          duration,
+          trialPeriod,
+        } as PlusPricingPreview;
+      });
+
+      return consolidated.filter((item) => !!item);
```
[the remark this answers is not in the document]
sshanzel: I'm getting a TS issue for some reason.

Screenshot 2025-04-04 at 7 08 29 PM

src/common/paddle/pricing.ts
sshanzel: @capJavert pretty hard to think of cases that are not covered yet by this, but definitely open to suggestions.

@rebelchris updated the usage. We don't have to check if the cycle is `year` anymore. Using this assumes you want monthly and from the outside, I applied a ternary to check if it is monthly already, we don't need to run the function.

src/schema/paddle.ts
sshanzel: Moved! 🚀

__tests__/paddle.ts
```diff
@@ -266,6 +273,271 @@ describe('pricing preview', () => {
   });
 });
 
+describe('plus pricing metadata', () => {
+  const QUERY = `
```
omBratteng, line 277:
```suggestion
  const QUERY = /* GraphQL */ `
```
For syntax

__tests__/paddle.ts
```diff
@@ -266,6 +273,271 @@ describe('pricing preview', () => {
...
+    const result = await client.query(QUERY, {
+      variables: { variant: 'non_existent' },
+    });
+    expect(result.errors).toBeTruthy();
+    expect(result.errors?.[0].message).toContain('Entity not found');
+  });
+});
+
+describe('plus pricing preview', () => {
+  const QUERY = `
```
omBratteng, line 379:
```suggestion
  const QUERY = /* GraphQL */ `
```
For syntax

src/schema/paddle.ts
```diff
@@ -231,6 +359,71 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
+    ): Promise<PlusPricingMetadata[]> =>
+      await getPlusPricingMetadata(ctx.con, variant),
+    plusPricingPreview: async (_, __, ctx): Promise<PlusPricingPreview[]> => {
+      const user = await ctx.con.getRepository(User).findOneOrFail({
+        where: { id: ctx.userId },
+        select: { createdAt: true },
+      });
+      const gb = getUserGrowthBookInstance(ctx.userId, {
+        subscribeToChanges: false,
+        attributes: { registrationDate: user.createdAt.toISOString() },
```
sshanzel, line 375: @capJavert @idoshamun should we add `AllocationClient` here? This feature is never going to be called from the FE, so they might not get allocated if we experiment on the values.

capJavert: If that is the case I think yes, good catch

rebelchris (approved):

src/schema/paddle.ts
sshanzel: Added! 🚀

src/common/paddle/index.ts
omBratteng: Maybe it can still be called `paddle.ts`?

src/schema/paddle.ts
```diff
@@ -231,6 +362,73 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
         items,
       };
     },
+    plusPricingMetadata: async (
+      _,
+      { variant = DEFAULT_PLUS_METADATA }: PlusMetadataArgs,
+      ctx: AuthContext,
+    ): Promise<PlusPricingMetadata[]> =>
+      await getPlusPricingMetadata(ctx.con, variant),
```
omBratteng, line 370:
```suggestion
      getPlusPricingMetadata(ctx.con, variant),
```

Since we're just returning it directly, we don't need the `await` here.

sshanzel: Kinda helps with error messages normally from what I've experienced, but yeah, we can definitely just remove this if the default error message would be helpful already.

src/schema/paddle.ts
capJavert: don't see its moved?

capJavert (approved):
Pricing endpoints and calcs still look a bit messy but at this point I am not sure what to add. At least its server side now so we can fix and adjust quickly.

src/schema/paddle.ts
```diff
@@ -118,6 +130,121 @@ export const typeDefs = /* GraphQL */ `
   extend type Query {
     pricePreviews: PricePreviews! @auth
     corePricePreviews: PricePreviews! @auth
+    plusPricingMetadata(variant: String): [PlusPricingMetadata!]! @auth
```
capJavert, line 133: Please make sure old endpoints works as is after deployment.

sshanzel: Yeah, we didn't touch anything existing so we can easily deprecate it when the time comes.

src/schema/paddle.ts
sshanzel: Right. I moved the other one that used the same regex. Should be update now 🙏

PR #2733, opened 2025-04-04
fix: only show products with success status

capJavert:


__tests__/njord.ts
```diff
@@ -1731,13 +1731,13 @@ describe('query userProductSummary', () => {
         id: '3f73343b-b1ae-45fc-bc3c-ded01d149596',
         name: 'Award 1',
         image: 'https://daily.dev/award.jpg',
-        count: 3,
+        count: 2,
```
capJavert, line 1734: I actually added cases in data for error but forgot to apply in the query

omBratteng (approved):

PR #2736, opened 2025-04-04
feat: save transaction fees value after transfer

capJavert:
On transfer save value after fees to db so we can show receiver balance/summary correctly

### DIFF
--- a/src/migration/1743776947033-UserTransactionValueFee.ts
+++ b/src/migration/1743776947033-UserTransactionValueFee.ts
@@ -1,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UserTransactionValueFee1743776947033
+  implements MigrationInterface
+{
+  name = 'UserTransactionValueFee1743776947033';
+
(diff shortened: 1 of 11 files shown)


src/migration/1743776947033-UserTransactionValueFee.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UserTransactionValueFee1743776947033
+  implements MigrationInterface
+{
+  name = 'UserTransactionValueFee1743776947033';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "user_transaction" ADD COLUMN IF NOT EXISTS "valueIncFees" integer NOT NULL`,
```
capJavert, line 10: Will manually do alter so I can set existing values appropriately. Mostly to avoid doing migration in two steps:
1. Column with default value 0
2. Update values
3. Then remove default from column

rebelchris (approved):

PR #2737, opened 2025-04-07
feat: check cores role

capJavert:
Query will be called from frontend to check user's region and set correct `coresRole` based on it. It is a once off based on user action. 

Frontend call https://github.com/dailydotdev/apps/pull/4353

### DIFF
--- a/src/entity/user/utils.ts
+++ b/src/entity/user/utils.ts
@@ -37,6 +37,8 @@ import { User } from './User';
 import { Feed } from '../Feed';
 import { remoteConfig } from '../../remoteConfig';
 import { getUserCoresRole } from '../../common/user';
+import { insertOrIgnoreAction } from '../../schema/actions';
+import { UserActionType } from './UserAction';
 
 export type AddUserData = Pick<
   User,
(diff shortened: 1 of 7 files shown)


src/entity/user/utils.ts
```diff
@@ -256,6 +258,12 @@ const safeInsertUser = async (
       const newUserId = newUser.identifiers[0].id;
       const feedId = newUserId;
 
+      await insertOrIgnoreAction(
+        entityManager,
+        newUserId,
+        UserActionType.CheckedCoresRole,
+      );
```
capJavert, line 265: User's who register with region already have it resolved so we don't fulfil user action immediately.

src/schema/users.ts
```diff
@@ -1935,6 +1946,63 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        },
+      });
+
+      if (coresRoleAction) {
+        const coresRoleUser = await coresRoleAction.user;
+
+        return {
+          coresRole: coresRoleUser.coresRole,
+        };
+      }
```
capJavert, line 1980: If user action exists just return the current cores role.

omBratteng (approved):

src/schema/users.ts
```diff
@@ -1935,6 +1946,63 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+            coresRole: true,
+          },
+        },
+        relations: {
+          user: true,
+        },
+      });
+
+      if (coresRoleAction) {
+        const coresRoleUser = await coresRoleAction.user;
```
idoshamun, line 1975: This fetches the full user object. Let's select only cores role

idoshamun (approved):
One optimization but overall looks good!

PR #2738, opened 2025-04-07
feat: add email for user bought cores

omBratteng:
Send user email when they buy cores.

AS-1050


src/workers/transactions/userBoughtCores.ts
```diff
@@ -0,0 +1,54 @@
...
+    const user = await transaction.receiver;
+
+    const coreAmount = Intl.NumberFormat('en-US').format(transaction.value);
+
+    await sendEmail({
+      send_to_unsubscribed: true,
+      transactional_message_id:
+        CioTransactionalMessageTemplateId.UserBoughtCores,
+      message_data: {
+        core_amount: `+${coreAmount}`,
```
capJavert, line 46:
```suggestion
        core_amount: transaction.value
```

src/workers/transactions/userBoughtCores.ts
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+
+    if (!purchaseProcessors.includes(transaction.processor)) {
+      return;
+    }
+
+    if (transaction.status !== UserTransactionStatus.Success) {
+      return;
+    }
```
capJavert, line 35: One thing I am worried is that technically transaction could be updated again after success status and then we will send the email again?

capJavert: As a quick solution we could set a flag for this.

src/workers/transactions/userBoughtCores.ts
capJavert: Ah my bad, the diff fulled me

src/workers/transactions/userBoughtCores.ts
omBratteng: Yeah, I think I should add it. Also need it for regular awards too

capJavert: But for regular awards it will be a notification so it will dedup from that?

omBratteng: You're right, it dedupes, so no need for flag on awards

omBratteng: Added now

src/workers/transactions/userBoughtCores.ts
```diff
@@ -0,0 +1,68 @@
...
+        id: user.id,
+      },
+      to: user.email,
+    });
+
+    await con.getRepository(UserTransaction).update(transaction.id, {
+      flags: updateFlagsStatement({
+        emailSent: true,
+      }),
+    });
```
capJavert, line 66: I think for safety we will have to do:
- set `emailSent: true` before `sendEmail`
- then await `sendEmail`
- catch and on error set `emailSent: false`

check `dedupedSend` in digest
other

capJavert (approved):
One last thing for dedup, other then that nice work 🚀

rebelchris (approved):

src/workers/transactions/userBoughtCores.ts
omBratteng: Wrapped in try...catch

capJavert (approved):

PR #2742, opened 2025-04-07
fix: transaction upsert on multiple updates

capJavert:
On multiple updates to transaction there can be edge case on writes so in that case also upsert over `flags->providerId`.

### DIFF
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -748,16 +748,29 @@ export const updateUserTransaction = async ({
   });
 
   if (!transaction) {
-    const newTransaction = await con
+    const insertResult = await con
       .getRepository(UserTransaction)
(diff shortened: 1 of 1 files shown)


src/routes/webhooks/paddle.ts
```diff
@@ -748,23 +748,37 @@ export const updateUserTransaction = async ({
...
+
+    payload.id = insertResult.raw[0].id;
+
+    return payload;
   } else {
     await con.getRepository(UserTransaction).update(
       { id: transaction.id },
       {
         value: itemData.price.customData.cores,
+        valueIncFees: itemData.price.customData.cores,
```
capJavert, line 773: Also fixed issue where I did not update `valueIncFees` if user selected different cores pack

omBratteng (approved):

PR #2743, opened 2025-04-08
fix: handle unsubscribes from award email and push notifications

omBratteng:
Forgot to add this logic in previous PR 🙈


capJavert (approved):

capJavert:
Let's keep an eye on any strange notifications, but I think we good.

PR #2745, opened 2025-04-08
fix: only publish if transaction status is success

omBratteng:


src/workers/transactions/userBoughtCores.ts
```diff
@@ -71,6 +75,11 @@ export const userBoughtCores: TypedWorker<'api.v1.user-transaction'> = {
         },
         to: user.email,
       });
+
+      logger.info(
+        { transactionId: transaction.id },
+        'Email sent for user bought cores',
+      );
```
capJavert, line 82: accidenal?

capJavert (approved):
Let's try

src/workers/transactions/userBoughtCores.ts
omBratteng: no, wanted to log it so that I can monitor production logs if it sends duplicates for same transaction

capJavert: yeah fine for above duplicates, but this one would log each time?

omBratteng: The above one could be removed, because it returns early and doesn't send email.
This is after email is sent, for cases where there's a race condition

PR #2741, opened 2025-04-07
chore: paddle dev variables

sshanzel:
We now store the metadata in a DB table which is based from a feature key.

Rather than updating the table value every time or having seeds, we can just hard-code these dev values.

This also removes caching from dev env to avoid any confusion during development in case you need to change any value.

PR #2751, opened 2025-04-08
fix: id maps property can be nullable

sshanzel:
As discussed from this PR, the id for iOS can be nullable: https://github.com/dailydotdev/daily-api/pull/2741#discussion_r2032691102

We can probably say the same for Paddle, if we intend to promote something similar to gifting in iOS but implemented differently.

### DIFF
--- a/src/schema/paddle.ts
+++ b/src/schema/paddle.ts
@@ -156,11 +156,11 @@ export const typeDefs = /* GraphQL */ `
     """
     Paddle platform ID
     """
-    paddle: String!
+    paddle: String
     """
     iOS platform ID
     """
-    ios: String!
+    ios: String
   }
 
   """


omBratteng (approved):

src/schema/paddle.ts
```diff
@@ -156,11 +156,11 @@ export const typeDefs = /* GraphQL */ `
     """
     Paddle platform ID
     """
-    paddle: String!
+    paddle: String
```
capJavert, line 159: We have to be careful now because now there can be a case where neither ID is defined, which is error prone.

sshanzel: No worries, because if the ID is empty, then that is not going to be part of the request to the pricing preview endpoint. We will only request items when they are both present at the metadata and the response itself.

PR #2757, opened 2025-04-09
fix: slack cores message send

capJavert:


src/routes/webhooks/paddle.ts
```diff
@@ -814,7 +814,7 @@ export const updateUserTransaction = async ({
       ...transaction,
       value: itemData.price.customData.cores,
       valueIncFees: itemData.price.customData.cores,
-      status: transaction.status || nextStatus,
+      status: nextStatus ?? transaction.status,
```
capJavert, line 817: `0 || true` is `true`

capJavert: fixed some other cases in logging

PR #2762, opened 2025-04-11
feat: new private endpoint for Kvasir

omBratteng:
Creates a new endpoint that Kvasir will call to fetch posts, and it will only return the posts that the user has access to.

Kvasir calls the endpoint by emulating being the user, where it forwards the users auth token as the da3 cookie in the request.

https://github.com/dailydotdev/kvasir/pull/15

### Jira ticket
AS-1002

### DIFF
--- a/src/routes/private/kvasir.ts
+++ b/src/routes/private/kvasir.ts
@@ -7,7 +7,14 @@ import { SourceMemberRoles } from '../../roles';
 import { z } from 'zod';
 
 const postsSchema = z.object({
-  postIds: z.array(z.string()).min(1),
+  postIds: z
+    .array(z.string())
+    .min(1, {
+      message: 'No posts to translate',
+    })
+    .max(100, {
+      message: 'Too many posts to translate',
+    }),
 });
 
 export const kvasir = async (fastify: FastifyInstance): Promise<void> => {
--- a/src/routes/private/kvasir.ts
+++ b/src/routes/private/kvasir.ts
@@ -7,7 +7,7 @@ import { SourceMemberRoles } from '../../roles';
 import { z } from 'zod';
 
 const postsSchema = z.object({
-  postIds: z.array(z.string()),
+  postIds: z.array(z.string()).min(1),
 });
 
 export const kvasir = async (fastify: FastifyInstance): Promise<void> => {
--- a/src/routes/private/kvasir.ts
+++ b/src/routes/private/kvasir.ts
@@ -19,7 +19,7 @@ export const kvasir = async (fastify: FastifyInstance): Promise<void> => {
 
   fastify.post<{
     Body: {
-      postIds?: Array<Post['id']>;
+      postIds: Array<Post['id']>;
     };
   }>('/posts', async (request, response): Promise<Array<Post>> => {
     const con = await createOrGetConnection();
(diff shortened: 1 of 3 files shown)


rebelchris (approved):

src/routes/private/kvasir.ts
```diff
@@ -0,0 +1,63 @@
...
+        .leftJoin(
+          SourceMember,
+          'sm',
+          'sm.sourceId = source.id AND sm.userId = :userId',
+          { userId: request.userId },
+        )
+        .where('post.id IN (:...postIds)', { postIds })
+        .andWhere('post.deleted = false')
+        .andWhere('post.visible = true')
+        .andWhere(
```
capJavert, line 52: same query that we used in Kvasir before?

src/routes/private/kvasir.ts
```diff
@@ -0,0 +1,63 @@
...
+          new Brackets((qb) => {
+            qb.where('source.private = false').orWhere(
+              'sm.userId IS NOT NULL AND sm.role != :blockedRole',
+              { blockedRole: SourceMemberRoles.Blocked },
+            );
+          }),
+        )
+        .getMany();
```
capJavert, line 60: I would like to add some kind of limit, eg. 100, so we don't allow thousands of posts in event this gets exposed for any reason

src/routes/private/kvasir.ts
capJavert: Meaning does it take care of this? https://dailydotdev.slack.com/archives/C07A7QQ27TM/p1739274066461929

src/routes/private/kvasir.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Post, Source, SourceMember } from '../../entity';
+import createOrGetConnection from '../../db';
+import { queryReadReplica } from '../../common/queryReadReplica';
+import { Brackets } from 'typeorm';
+import { SourceMemberRoles } from '../../roles';
+import { z } from 'zod';
+
+const postsSchema = z.object({
+  postIds: z.array(z.string()).min(1),
+});
```
capJavert, line 11: nice 👌

src/routes/private/kvasir.ts
capJavert: we can also just limit length on zod schema as well

capJavert (commented):
Just some questions

src/routes/private/kvasir.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const kvasir = async (fastify: FastifyInstance): Promise<void> => {
+  fastify.addHook('preHandler', async (req, res) => {
+    if (!req.userId) {
+      return res.code(401).send({ error: 'Unauthorized' });
+    }
```
capJavert, line 17: Do you also need:
```
   if (!req.service) {
    }
```

just so it can be accessed only internally

omBratteng: It already should only be accessed internally, as it's on the /p/ prefix, which is a separate deployment.

On Kvasir side, I just forward the users own auth token, but as cookie instead

capJavert: yeah but just for safety sake, if we have it in other private endpoints let's add here. 

seen the user token part, thats cool 👌

omBratteng: Added

capJavert (approved):

PR #2765, opened 2025-04-15
feat: skadi digest support

rebelchris:
Added support for Skadi powered ads in digest.
The endpoint is not configured yet, but this should allow us to test via retool once it is.

For now it's a specific template (see todo), we need to check if we want to support 75+ maybe.

### DIFF
--- a/src/common/personalizedDigest.ts
+++ b/src/common/personalizedDigest.ts
@@ -149,7 +149,7 @@ const getEmailAd = async ({
   const digestAd = ad.value.digest;
 
   return {
-    type: ad.type,
+    type: 'dynamic_ad',
     ...digestAd,
   };
 };
--- a/src/common/personalizedDigest.ts
+++ b/src/common/personalizedDigest.ts
@@ -123,13 +123,17 @@ const getPostsTemplateData = ({
   });
 };
 
+type CIOSkadiAd = {
+  type: string;
+} & SkadiAd;
+
 const getEmailAd = async ({
   user,
   feature,
 }: {
   user: User;
   feature: PersonalizedDigestFeatureConfig;
-}): Promise<SkadiAd | null> => {
+}): Promise<CIOSkadiAd | null> => {
   // TODO: Temporary hardcode 75 check
   if (
     isPlusMember(user.subscriptionFlags?.cycle) ||
@@ -164,7 +168,7 @@ const getEmailVariation = async ({
   userStreak?: UserStreak;
   feature: PersonalizedDigestFeatureConfig;
   currentDate: Date;
-  adProps: DynamicAd | null;
+  adProps: CIOSkadiAd | null;
 }): Promise<
   Pick<SendEmailRequestWithTemplate, 'to' | 'message_data' | 'identifiers'>
 > => {
--- a/src/common/personalizedDigest.ts
+++ b/src/common/personalizedDigest.ts
@@ -30,8 +30,7 @@ import { isPlusMember } from '../paddle';
 import { mapCloudinaryUrl } from './cloudinary';
 import { queryReadReplica } from './queryReadReplica';
 import { counters } from '../telemetry/metrics';
-import { SkadiAd, skadiClient } from '../integrations/skadi';
-import { skadiPersonalizedDigestClient } from '../integrations/skadi/clients';
+import { SkadiAd, skadiPersonalizedDigestClient } from '../integrations/skadi';
 
 type TemplatePostData = Pick<
   ArticlePost,
--- a/src/common/personalizedDigest.ts
+++ b/src/common/personalizedDigest.ts
@@ -31,6 +31,7 @@ import { mapCloudinaryUrl } from './cloudinary';
 import { queryReadReplica } from './queryReadReplica';
 import { counters } from '../telemetry/metrics';
 import { SkadiAd, skadiClient } from '../integrations/skadi';
+import { skadiPersonalizedDigestClient } from '../integrations/skadi/clients';
 
 type TemplatePostData = Pick<
   ArticlePost,
@@ -142,7 +143,7 @@ const getEmailAd = async ({
     return null;
   }
 
-  const ad = await skadiClient.getAd('default_digest', {
+  const ad = await skadiPersonalizedDigestClient.getAd('default_digest', {
     USERID: user.id,
   });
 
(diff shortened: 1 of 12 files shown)


src/common/personalizedDigest.ts
```diff
@@ -341,19 +384,26 @@ export const getPersonalizedDigestEmailPayload = async ({
...
     currentDate,
+    adProps,
   });
 
   return {
     ...baseNotificationEmailData,
     send_at: Math.floor(emailSendDate.getTime() / 1000),
     transactional_message_id: feature.templateId,
     ...variationProps,
+    ...(adProps && adProps),
```
rebelchris, line 406:
```suggestion

```

src/common/personalizedDigest.ts
```diff
@@ -122,19 +123,60 @@ const getPostsTemplateData = ({
...
+  }
+
+  const ad = await skadiClient.getAd('default_digest', {
+    USERID: user.id,
+  });
+
+  return {
+    type: ad.type,
+    title: ad.value.digest.title,
+    link: ad.value.digest.link,
```
omBratteng, line 157: Is `link` correct? The fallback ad uses `post_link` there

rebelchris: They different types of ads, this one is new, so new naming :P

src/common/personalizedDigest.ts
```diff
@@ -122,19 +123,52 @@ const getPostsTemplateData = ({
...
+  }
+
+  const ad = await skadiClient.getAd('default_digest', {
+    USERID: user.id,
+  });
+
+  const digestAd = ad.value.digest;
+
+  return {
+    type: ad.type,
```
idoshamun, line 152: Where is this type coming from?

rebelchris: Passed from Skadi response

```
// response
{
    "type": "DIGEST",
    "value": {
        "digest": {
            "title": "DigestTitle",
            "link": "https://api.daily.dev/v2/c?id=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJPYmplY3QiOnsiaSI6IjYzMmIxMGE2LWQ4MmMtNDI4MS04ZGY3LTVjYmFjZjE2YzI3YyIsInUiOiJEaWdlc3RMaW5rIiwidCI6MTc0NDI4NTk3MH19.83OFFXO_5IdL9P8HGKPv4CsipF_Xhh_7k0o1e_AX8-Y",
            "image": "DigestImage",
            "company_name": "DigestCompanyName",
            "company_logo": "DigestCompanyLogo",
            "call_to_action": "DigestCallToAction"
        }
    },
    "pixels": [
        "https://api.daily.dev/v2/p?id=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJPYmplY3QiOnsiaSI6IjYzMmIxMGE2LWQ4MmMtNDI4MS04ZGY3LTVjYmFjZjE2YzI3YyIsImFpIjoiYmVjMmVjYWItN2MxMS00MzRhLTg4YjItMGEzMzNkMzdjMGI4IiwiYyI6ImVjYjJkM2YwLWFiOTctNDUyNS1hNWEyLWVlM2Y4ZmQwMTkyNCIsImEiOiJkM2FkYjQxZS0zNjRlLTQwNGMtOTgzMi1mZWIwZDZiNTg3NzYiLCJjciI6ImI1MmY5YzE5LWI1YmMtNDU4My04ODRjLWI5NzQwZWY5YzBiMiIsImJtIjoib2xkbG9naWMiLCJiaWQiOiIwLjEiLCJ1IjoiNGYzZmI1OTUtOGUwNS00YzViLWIzYjItMjYwZTcyMjEyZDZkIiwidCI6MTc0NDI4NTk3MH19.6wlCBr5e3fSzTDDwQArNr_QW6NrJWDt29jyTwfTxQHU"
    ]
}
```

idoshamun (approved):
once we validate that it works, we will update the condition to anything >= 75. should be better as you suggested

src/integrations/skadi/clients.ts
```diff
@@ -0,0 +1,60 @@
...
+  breakerOpts: {
+    halfOpenAfter: 5 * 1000,
+    threshold: 0.1,
+    duration: 10 * 1000,
+  },
+});
+
+export const skadiClient = new SkadiClient(process.env.SKADI_ORIGIN!, {
+  garmr: garmrSkadiService,
+});
```
capJavert, line 60: I think we should instantiate separate skadi client for personalized digest as we do with `personalizedDigestFeedClient`

capJavert: key options being:
- minimumRps (because workers can have lower rps due to processing)
- limits
- retryOpts

we can also add `events`

src/common/personalizedDigest.ts
```diff
@@ -149,21 +183,22 @@ const getEmailVariation = async ({
...
-      ),
-    });
+  if (posts.length >= feature.adIndex) {
+    if (adProps) {
+      posts.splice(feature.adIndex, 0, adProps);
+    } else if (!isPlusMember(user.subscriptionFlags?.cycle)) {
+      posts.splice(feature.adIndex, 0, {
+        // type: 'ad_image',
+        // link: `https://email.buysellads.net/?k=CW7DE23N&c=${user.id}`,
+        // image: `https://email.buysellads.net/?k=CW7DE23N&i=${user.id}`,
```
capJavert, line 193: Do we need this still?

capJavert (approved):
Missing some tests, but good if you want to test first 👍 

Minor comments to check

src/integrations/skadi/clients.ts
rebelchris: But isn't this a new Skadi client?
It's only used for personal digest.

rebelchris: Ah ok makes sense just a naming thing :)

src/integrations/skadi/clients.ts
```diff
@@ -0,0 +1,71 @@
...
+    duration: 10 * 1000,
+    minimumRps: 1,
+  },
+  limits: {
+    maxRequests: 150,
+    queuedRequests: 100,
+  },
+  retryOpts: {
+    maxAttempts: 0,
+  },
```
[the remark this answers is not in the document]
rebelchris: Added, can you just check that logic is ok?
You don't need to know which event happened?

capJavert (approved):

PR #2769, opened 2025-04-16
fix: funnel return 500 when userId is read

ilasw:
Add optional chaining to userId and session checks in funnel data retrieval

| Cloud | Vercel |
|--------|--------|
| image | image |

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -769,7 +769,7 @@ const resolveDynamicFunnelId = (userId: string): string => {
 
 // Fetches the funnel data from Freyja
 const getFunnel = async (req: FastifyRequest, res: FastifyReply) => {
-  const userId = req?.userId || req?.trackingId;
+  const userId = req.userId || req.trackingId;
   const query = req.query as { id?: string; v?: string };
   const sessionId = req.cookies[cookies.funnel.key];
 
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -769,7 +769,7 @@ const resolveDynamicFunnelId = (userId: string): string => {
 
 // Fetches the funnel data from Freyja
 const getFunnel = async (req: FastifyRequest, res: FastifyReply) => {
-  const userId = req.userId || req.trackingId;
+  const userId = req?.userId || req?.trackingId;
   const query = req.query as { id?: string; v?: string };
   const sessionId = req.cookies[cookies.funnel.key];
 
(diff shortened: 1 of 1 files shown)


src/routes/boot.ts
```diff
@@ -769,7 +769,7 @@ const resolveDynamicFunnelId = (userId: string): string => {
 
 // Fetches the funnel data from Freyja
 const getFunnel = async (req: FastifyRequest, res: FastifyReply) => {
-  const userId = req.userId || req.trackingId;
+  const userId = req?.userId || req?.trackingId;
   const query = req.query as { id?: string; v?: string };
```
sshanzel, line 773: This line would then fail I think.

src/routes/boot.ts
```diff
@@ -769,7 +769,7 @@ const resolveDynamicFunnelId = (userId: string): string => {
 
 // Fetches the funnel data from Freyja
 const getFunnel = async (req: FastifyRequest, res: FastifyReply) => {
-  const userId = req.userId || req.trackingId;
+  const userId = req?.userId || req?.trackingId;
```
capJavert, line 772: when is req undefined?

ilasw: Should never been undefined imo, but I'm trying to make every occurrence where we are picking userId safe.  I can revert it if needed. 

[Some context about this PR](https://dailydotdev.slack.com/archives/C08J2D9NKPU/p1744795560847689?thread_ts=1744726922.675339&cid=C08J2D9NKPU)

idoshamun: even if it could happen we need to find the root cause. it seems like the issue is elsewhere

ilasw: IMO the issue is [here](https://github.com/dailydotdev/daily-api/pull/2769/files#diff-53d88b8a44fefbe970e714bd8cd0ed3245db56a6645de0d7f174a2a26fd524d9R783) and not in the req, but added those check to be sure 👀

idoshamun: ah you're right. please remove this optional chaining then and keep the other one

ilasw: @idoshamun Done ✔️ 

Let's try only with this change, then if still not working I'll investigate more on the req @capJavert @sshanzel

idoshamun (approved):

lee-dailydev123 (approved):
🚀

lee-dailydev123:
Oh. I'm using my dummy account 😂

src/routes/boot.ts
capJavert: looks like in prod build indeed the line number `512:98` correlates to the changed line:
image

capJavert: so crash is on that part

capJavert (approved):

PR #2772, opened 2025-04-16
fix: protect paddle webhook endpoint by IP in production

omBratteng:
Protects the Paddle webhook endpoint with IP validation in production.
Also logs the payload in the error log if there's a generic paddle error.


rebelchris (approved):

src/routes/webhooks/paddle.ts
```diff
@@ -1030,6 +1032,12 @@ export const processTransactionUpdated = async ({
 
 export const paddle = async (fastify: FastifyInstance): Promise<void> => {
   fastify.register(async (fastify: FastifyInstance): Promise<void> => {
+    fastify.addHook('onRequest', async (request, res) => {
+      if (isProd && !remoteConfig.vars.paddleIps?.includes(request.ip)) {
```
capJavert, line 1036: if its empty lets skip check, so we have an option to pass everything through if needed

capJavert (approved):

PR #2775, opened 2025-04-17
feat: njord request signing

capJavert:
Njord PR that enables it: https://github.com/dailydotdev/njord/pull/54


idoshamun (approved):
Nice implementation! Something to consider, non-blocking, is to add support for request signing as a generic thing and then apply it to njord. But this may be premature

src/common/njord.ts
```diff
@@ -70,6 +79,34 @@ export const getNjordClient = (clientTransport = transport) => {
...
+};
+
+export const createNjordAuth = async (
+  payload: Message,
+): Promise<Pick<CallOptions, 'headers'>> => {
+  const authHeaders = new Headers();
+
+  const { token } = await signJwt<NjordJwtPayload>(
+    {
+      client_id: 'api',
```
omBratteng, line 94: extract API as a const somewhere?

omBratteng (approved):

capJavert:
Yeah I wanted to add generic request signing util but did not manage to sort out the types of the wrapper method as I wanted to. Will keep in mind as usage increases.

PR #2776, opened 2025-04-18
feat: soft delete user

capJavert:
- preserve deleted user ids
- preserve deleted user transactions under ghost
- check for deleted user ids collision on user creation
- add new handler for deleted user collision in `safeInsertUser`
- also apply delete user logic to zombie user cron


src/entity/user/DeletedUser.ts
```diff
@@ -0,0 +1,10 @@
+import { CreateDateColumn, Entity, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class DeletedUser {
```
capJavert, line 4: Opted to create new entity which we can check on user creation, instead of adding `deleted` column to `User` entity which we would then need to check on all other places.

src/migration/1744972478032-SpecialUserDeleteTrigger.ts
```diff
@@ -0,0 +1,42 @@
...
+  implements MigrationInterface
+{
+  name = 'SpecialUserDeleteTrigger1744972478032';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DROP RULE prototect_ghostuser_deletion on "user";`,
+    );
+    await queryRunner.query(`
+      CREATE OR REPLACE FUNCTION prevent_special_user_delete()
```
capJavert, line 13: Instead of `RULE` create trigger, because `RULE` was blocking usage of `RETURNING` in zombie cron.

omBratteng (commented):
What happens in the user_transaction table? Does the rows get deleted?

capJavert:
yes, as other user content, njord keeps the log though, although @idoshamun wdyt, should we move user_transaction items to ghost user instead of cascading when user is deleted?

src/migration/1744972478032-SpecialUserDeleteTrigger.ts
```diff
@@ -0,0 +1,42 @@
...
+      RETURNS trigger AS $$
+      BEGIN
+        IF OLD.id IN ('404', 'system') THEN
```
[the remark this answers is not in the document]
capJavert: yes

omBratteng:
I think we should update it to ghost user, so the receiver still can see awards they've been given.

capJavert:
@omBratteng yeah I agree, see 9402bb6131c658b9b62a291860ae733285839b44

src/common/user.ts
```diff
@@ -24,7 +24,11 @@ import { ForbiddenError } from 'apollo-server-errors';
 import { logger } from '../logger';
 import { CoresRole } from '../types';
 import { remoteConfig } from '../remoteConfig';
+<<<<<<< Updated upstream
 import { DeletedUser } from '../entity/user/DeletedUser';
+=======
+import { UserTransaction } from '../entity/user/UserTransaction';
+>>>>>>> Stashed changes
```
omBratteng, line 31: 😬

capJavert: image

capJavert: see 96ff01afd90486bfc199ec3f907b0eee9335993d

omBratteng (approved):

src/common/user.ts
```diff
@@ -87,7 +89,26 @@ export const deleteUser = async (
       await entityManager
         .getRepository(Post)
         .update({ scoutId: userId }, { scoutId: null });
+      await entityManager.getRepository(UserTransaction).update(
```
idoshamun, line 92: Can we somehow keep the original id? maybe in a flag or anything that can help us track if needed the original user?

src/cron/cleanZombieUsers.ts
```diff
@@ -22,9 +23,26 @@ const cron: Cron = {
...
       .andWhere({
         createdAt: LessThan(timeThreshold),
-      });
+      })
+      .returning(['id']);
+
+    const { affected, raw } = await query.execute();
+
+    if (Array.isArray(raw) && raw.length > 0) {
+      await con
```
idoshamun, line 32: Isn't it a better approach to use a trigger on delete user?

src/entity/user/DeletedUser.ts
idoshamun: I agree. Using soft delete on user table would have forced us to be creative when the same user tries to register with the same email after deleting a previous account for example

src/common/user.ts
capJavert: I was thinking original Ids would be kept on njord side, we can trace through transaction id and idempotencyKey

src/cron/cleanZombieUsers.ts
capJavert: hmm actually yes

capJavert:
@idoshamun ready for review again 🙏

src/common/user.ts
idoshamun: sounds good

PR #2780, opened 2025-04-21
refactor: pricing preview to support cores

sshanzel:
- Refactor the newly introduced query to allow receiving a type that will determine which keys to utilize to fetch from.
- Refactored the same query to have reusable functions and be used for cores.
- Update the way we parse the pricing amount of the item based on Ido's implementation from the FE. I think it is more stable than dividing the value by 10.
  - I applied it to all types:
    - Base price.
    - Monthly (if applicable).
    - Daily (if applicable).
  - This is so the way we parse the value is going to be consistent across three types. I think we wouldn't want to show a different format when it comes to base price vs monthly/daily.

Note: I changed the graphql query's definition but good thing since none uses it yet. Just made it more general as it now is applicable both to plus and cores.


capJavert (approved):
Nothing specific to add, can we confirm with frontend implementation that this matches all cores requirements?

sshanzel:
> Nothing specific to add, can we confirm with frontend implementation that this matches all cores requirements?

I will do a final confirmation before merging. And if it does not yet satisfy the requirements yet, I will raise another PR to point it here 🙏

sshanzel:
Looking about right locally!

Screenshot 2025-04-23 at 7 53 28 PM

PR #2783, opened 2025-04-24
feat: cores iap

capJavert:
- support for IAP and one time purchases
- refactor apple webhooks file to common to split and reuse common logic
- will work on adding mores tests for IAP flow in next PR after some sandbox testing e2e if needed

### DIFF
--- a/src/common/apple/purchase.ts
+++ b/src/common/apple/purchase.ts
@@ -22,6 +22,8 @@ import {
 } from '../utils';
 import type { Block, KnownBlock } from '@slack/web-api';
 import { webhooks } from '../slack';
+import { checkUserCoresAccess } from '../user';
+import { CoresRole } from '../../types';
 
 export const isCorePurchaseApple = ({
   transactionInfo,
@@ -43,7 +45,7 @@ export const notifyNewStoreKitPurchase = async ({
 }: {
   data: JWSTransactionDecodedPayload;
   transaction: UserTransaction;
-  user: User;
+  user: Pick<User, 'id' | 'subscriptionFlags' | 'coresRole'>;
   currencyInUSD: number;
 }) => {
   if (isTest) {
@@ -143,7 +145,7 @@ export const handleCoresPurchase = async ({
   user,
 }: {
   transactionInfo: JWSTransactionDecodedPayload;
-  user: User;
+  user: Pick<User, 'id' | 'subscriptionFlags' | 'coresRole'>;
   environment: Environment;
   notification: ResponseBodyV2DecodedPayload;
 }): Promise<UserTransaction> => {
@@ -153,6 +155,15 @@ export const handleCoresPurchase = async ({
 
   if (!transactionInfo.productId) {
     throw new Error('Missing productId in transactionInfo');
+  }
+
+  if (
+    checkUserCoresAccess({
+      user,
+      requiredRole: CoresRole.User,
+    }) === false
+  ) {
+    throw new Error('User does not have access to cores purchase');
   }
 
   const con = await createOrGetConnection();
--- a/src/common/apple/purchase.ts
+++ b/src/common/apple/purchase.ts
@@ -4,7 +4,11 @@ import type {
   ResponseBodyV2DecodedPayload,
 } from '@apple/app-store-server-library';
 import type { User } from '../../entity/user/User';
-import { getAppleTransactionType } from './utils';
+import {
+  getAnalyticsEventFromAppleNotification,
+  getAppleTransactionType,
+  logAppleAnalyticsEvent,
+} from './utils';
 import { AppleTransactionType } from './types';
 import {
   UserTransaction,
@@ -24,6 +28,7 @@ import type { Block, KnownBlock } from '@slack/web-api';
 import { webhooks } from '../slack';
 import { checkUserCoresAccess } from '../user';
 import { CoresRole } from '../../types';
+import { convertCurrencyToUSD } from '../../integrations/openExchangeRates';
 
 export const isCorePurchaseApple = ({
   transactionInfo,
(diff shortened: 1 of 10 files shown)


src/common/apple/purchase.ts
```diff
@@ -0,0 +1,236 @@
...
+import {
+  concatTextToNewline,
+  isProd,
+  isTest,
+  updateFlagsStatement,
+} from '../utils';
+import type { Block, KnownBlock } from '@slack/web-api';
+import { webhooks } from '../slack';
+
+export const isCorePurchaseApple = ({
```
capJavert, line 26: New file for Cores related and IAP

src/common/apple/subscription.ts
```diff
@@ -0,0 +1,245 @@
...
+import {
+  getAnalyticsEventFromAppleNotification,
+  logAppleAnalyticsEvent,
+} from './utils';
+import { productIdToCycle } from './types';
+import { concatTextToNewline, isTest } from '../utils';
+import type { Block, KnownBlock } from '@slack/web-api';
+import { webhooks } from '../slack';
+
+const getSubscriptionStatus = (
```
capJavert, line 28: Mostly code from webhooks file, for reuse

src/common/apple/types.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const certificatesToLoad = isTest
+  ? ['__tests__/fixture/testCA.der']
+  : [
+      '/usr/local/share/ca-certificates/AppleIncRootCertificate.cer',
+      '/usr/local/share/ca-certificates/AppleRootCA-G2.cer',
+      '/usr/local/share/ca-certificates/AppleRootCA-G3.cer',
+    ];
+
+const getVerifierEnvironment = (): Environment => {
```
capJavert, line 14: Mostly code from webhooks file, for reuse

src/common/apple/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+} from '../../integrations/analytics';
+import {
+  AppleTransactionType,
+  certificatesToLoad,
+  productIdToCycle,
+} from './types';
+import { readFile } from 'fs/promises';
+import { isTest } from '../utils';
+
+export const verifyAndDecodeAppleSignedData = async ({
```
capJavert, line 25: New method for decoding both signed data

src/routes/webhooks/apple.ts
```diff
@@ -250,30 +69,23 @@ const handleNotifcationRequest = async (
       return response.status(200).send({ received: true });
     }
 
-    // Check if the event is a subscription event
-    // NOTE: When adding support for purchasing cores, we must remove this check as it's not a subscription event
-    if (isNullOrUndefined(notification?.data?.signedRenewalInfo)) {
```
capJavert, line 255: Moved check down only if subscription transaction type

src/common/apple/purchase.ts
```diff
@@ -0,0 +1,247 @@
...
+    throw new Error('Missing productId in transactionInfo');
+  }
+
+  if (
+    checkUserCoresAccess({
+      user,
+      requiredRole: CoresRole.User,
+    }) === false
+  ) {
+    throw new Error('User does not have access to cores purchase');
```
idoshamun, line 166: What are we going to do if the user paid and isn't eligible?

src/common/apple/purchase.ts
```diff
@@ -0,0 +1,247 @@
...
+          text: concatTextToNewline('*Currency (local):*', data.currency),
+        },
+      ],
+    },
+  ];
+
+  await webhooks.transactions.send({ blocks });
+};
+
+export const handleCoresPurchase = async ({
```
idoshamun, line 143: Once we approve everything this should be mostly the same as the paddle cores?
I assume the same internal logic, just different trigger?

src/common/apple/purchase.ts
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        throw error;
+      }
+    }
+
+    return userTransaction;
+  });
+
+  if (transaction.status === UserTransactionStatus.Success) {
```
idoshamun, line 237: I think we should also log an event here just like we do with paddle, or does it happen on the ios part?

src/common/apple/purchase.ts
capJavert: they are blocked on the client side as well, so its mostly another check, we do it on paddle side as well

we would handle it case by case, do you want to handle somehow differently?

src/common/apple/purchase.ts
capJavert: yeah, missed calling the function, adding

src/common/apple/purchase.ts
capJavert: it is not the same since payloads differ, but `purchaseCores` is reused

src/common/apple/purchase.ts
capJavert: https://github.com/dailydotdev/daily-api/pull/2783/commits/c57a4b3f5e14b56e922f70c271d180b3d82ab893 added

src/common/apple/purchase.ts
idoshamun: but there's also much logic even here that can be reused, am i missing anything?
if we define a common type with all the info from apple/paddle and have one function that checks access, create transactions, etc. it should be better, isn't it? again hard to review this specifically as it's going to change once we go production

idoshamun (approved):
Nothing blocking. Looks good

src/common/apple/purchase.ts
capJavert: Not really, transaction creation is different on apple and paddle, also detection of product type. I imagine that any new provider will also have it different, so I think its actually nicer for now to keep the logics separate. 

What could be done is to create a generic provider interface which would standardize what we expect for each provider, but that requires refactoring how apple and paddle webhooks work now (which is quite different) so it would be separate work from this PR.

capJavert: I'll try to add the spec for this in a task, and link back so you can see.

idoshamun: Ok makes sense

omBratteng (approved):

PR #2785, opened 2025-04-28
refactor: pricing preview to add locale option

sshanzel:
Add a `locale` param for requesting the pricing preview.


capJavert (approved):

src/schema/paddle.ts
```diff
@@ -131,7 +131,7 @@ export const typeDefs = /* GraphQL */ `
     pricePreviews: PricePreviews! @auth
     corePricePreviews: PricePreviews! @auth
     pricingMetadata(type: PricingType): [ProductPricingMetadata!]! @auth
-    pricingPreview(type: PricingType): [ProductPricingPreview!]! @auth
```
sshanzel, line 134: @capJavert also removed the `@auth` directive since we don't require auth for requesting the prices anymore. For example, on Funnel.

PR #2787, opened 2025-04-28
fix: try using tags in graphorm queries to look at query insights

omBratteng:
Adds a SQL comment to the GraphORM queries with the operation name to attempt at getting better query insights in CloudSQL

https://cloud.google.com/sql/docs/postgres/using-query-insights#adding-tags-to-sql-queries


src/graphorm/graphorm.ts
```diff
@@ -571,6 +572,9 @@ export class GraphORM {
     let res: any[];
 
     try {
+      builder.queryBuilder.comment(
+        `action='${(ctx?.req?.body as GraphqlPayload)?.operationName}'`,
+      );
```
capJavert, line 577: Check if operation name exists before calling `comment`?

capJavert (approved):

PR #2789, opened 2025-04-30
feat: skip njord if test discount

capJavert:
Since we are using test discounts sometimes for e2e purchase testing we don't want to actually log those in njord balance sheet.

### DIFF
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -682,11 +682,25 @@ export const processTransactionCompleted = async ({
         throw new Error('User does not have access to cores purchase');
       }
 
+      // skip njord if transaction has test discount
       const shouldSkipNjord =
         !!transactionData.discountId &&
         !!remoteConfig.vars.paddleTestDiscountIds?.includes(
           transactionData.discountId,
         );
+
+      if (shouldSkipNjord) {
+        await entityManager.getRepository(UserTransaction).update(
+          {
+            id: userTransaction.id,
+          },
+          {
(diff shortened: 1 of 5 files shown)


src/routes/webhooks/paddle.ts
```diff
@@ -682,41 +682,49 @@ export const processTransactionCompleted = async ({
...
+        !!transactionData.discountId &&
+        !!remoteConfig.vars.paddleTestDiscountIds?.includes(
+          transactionData.discountId,
+        );
+
+      if (!shouldSkipNjord) {
+        try {
+          await purchaseCores({
             transaction: userTransaction,
           });
```
capJavert, line 695: Just skip njord call, so Cores are not assigned, everything else works so we can e2e test.

capJavert: @idoshamun thing better then messing with separate njord system account for this test case

idoshamun: So the checkout should work but we will not see our balance gets updated?

capJavert: yes

capJavert: if team members need balance we can manually assign as needed

idoshamun: ok, let's see if someone complains 😂

idoshamun (approved):

src/routes/webhooks/paddle.ts
capJavert: I'll let people know

PR #2790, opened 2025-05-02
fix: deleting an account with gifted subscription

sshanzel:
From Paddle's side of things, there is no subscription ID for gifted Plus membership. Hence we get an error that the subscription is not found and the process causes to throw an error.

### DIFF
--- a/src/common/user.ts
+++ b/src/common/user.ts
@@ -47,16 +47,20 @@ export const deleteUser = async (
       }
 
       if (subscriptionFlags?.provider === SubscriptionProvider.Paddle) {
-        await cancelSubscription({
-          subscriptionId: subscriptionFlags.subscriptionId,
-        });
+        const isGifted = !!subscriptionFlags.giftExpirationDate;
+        // gifted subscription is a one-time payment hence not considered subscription in Paddle's terms
+        if (!isGifted) {
+          await cancelSubscription({
+            subscriptionId: subscriptionFlags.subscriptionId,
+          });
+        }
         logger.info(
           {
             provider: SubscriptionProvider.Paddle,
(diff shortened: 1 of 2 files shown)


src/common/user.ts
```diff
@@ -47,16 +47,20 @@ export const deleteUser = async (
       }
 
       if (subscriptionFlags?.provider === SubscriptionProvider.Paddle) {
-        await cancelSubscription({
-          subscriptionId: subscriptionFlags.subscriptionId,
-        });
+        const isGifted = !!subscriptionFlags.giftExpirationDate;
+        // gifted subscription is a one-time payment hence not considered subscription in Paddle's terms
+        if (!isGifted) {
+          await cancelSubscription({
```
capJavert, line 53: Can we add a test for this case and validate `cancelSubscription` is not called?

src/common/user.ts
```diff
@@ -47,16 +47,20 @@ export const deleteUser = async (
...
+          });
+        }
         logger.info(
           {
             provider: SubscriptionProvider.Paddle,
             userId,
             subscriptionId: subscriptionFlags.subscriptionId,
           },
-          'Subscription cancelled user deletion',
+          `Subscription${isGifted ? ' (gifted)' : ''} cancelled user deletion`,
```
capJavert, line 63: Instead of modifying script you can pass `isGifted` as property to log, more easily queryable.

sshanzel: Oh, right. Yeah, that is better.

src/common/user.ts
sshanzel: Good point!

PR #2801, opened 2025-05-07
fix: product seeds flags

capJavert:
Was saved as string to flags in jsonb column, with this change it actually saves it as json.


omBratteng (approved):

sshanzel (approved):
Do we not have any existing data on this yet for prod?

capJavert:
We do, this was just mistake in seeds.

PR #2802, opened 2025-05-08
feat: list awards

capJavert:
- list awards
- get awards total (separate query but on frontend its queries as a single one)

Frontend: https://github.com/dailydotdev/apps/pull/4483

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -810,7 +810,7 @@ export const typeDefs = /* GraphQL */ `
     node: UserPost!
 
     """
-    Used in \`before\` and \`after\` args
+    Used in before and after args
     """
     cursor: String!
   }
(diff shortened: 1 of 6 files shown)


src/schema/posts.ts
```diff
@@ -803,6 +806,28 @@ export const typeDefs = /* GraphQL */ `
     COLLECTION
   }
 
+  type UserPostEdge {
+    node: UserPost!
+
+    """
+    Used in \`before\` and \`after\` args
```
omBratteng, line 813: Do we need the \\`?

capJavert: just copy pasted from other params

capJavert: probably not

capJavert: just because of same ticks are used inside the docs, removed them but it does not really make any difference

idoshamun (approved):

PR #2804, opened 2025-05-12
feat: add claimable item entity

AmarTrebinjac:
Logic for adding claimable item to an user that subscribes without being registered.


src/entity/ClaimableItem.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { User } from './user';
+
+export enum ClaimableItemTypes {
+  Plus = 'plus',
+}
+
+@Entity()
+export class ClaimableItem {
+  @PrimaryColumn()
+  transactionId: string;
```
AmarTrebinjac, line 11: gonna rename this to "id"

src/entity/ClaimableItem.ts
```diff
@@ -0,0 +1,34 @@
+import { Column, Entity, Index, ManyToOne, PrimaryColumn } from 'typeorm';
+import type { User } from './user';
+
+export enum ClaimableItemTypes {
+  Plus = 'plus',
+}
+
+@Entity()
+export class ClaimableItem {
+  @PrimaryColumn()
```
capJavert, line 10: Let's use UUID

src/entity/ClaimableItem.ts
```diff
@@ -0,0 +1,34 @@
...
+  @Index('IDX_claimable_item_email')
+  email: string;
+
+  @Column({ nullable: false, default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ nullable: true })
+  claimedAt?: Date;
+
+  @Column({ nullable: false })
```
capJavert, line 23: I think `nullable` is `false` by default

capJavert (commented):
Let's not merge this separate but with https://github.com/dailydotdev/daily-api/pull/2805, because we are sure to find some adjustments to entity when real implementation is done and want to avoid additional migrations.

src/entity/ClaimableItem.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+@Entity()
+export class ClaimableItem {
+  @PrimaryColumn()
+  id: string;
+
+  @Column({ type: 'text', nullable: false })
+  @Index('IDX_claimable_item_email')
+  email: string;
```
capJavert, line 15: Should we make `email` unique, because technically user can purchase multiple subscriptions as anonymous and then what happens when they claim?

AmarTrebinjac: @capJavert In the RFC I laid out the idea that this table can be used for more than just funnels. For example it can be used to buy a subscription as a gift for someone who hasn't registered as an incentive to join. 
I think its better to let it be more flexible in this case.

As to your question. I'm having difficulties imagining where this scenario wouldn't be an accident, in which case, we'd need to manually sort it out for them anyway? 🤔

src/entity/ClaimableItem.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  @Column({ nullable: false, default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ nullable: true })
+  claimedAt?: Date;
+
+  @Column({ nullable: false })
+  type: string;
```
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert I am sure I was told to make it a string and let us do the validation last time I created an enum for an Entity 😅 

I am okay with either way, but I'm not sure what our guidelines are on this 🤔

AmarTrebinjac:
> Let's not merge this separate but with #2805, because we might find some adjustments to entity when real implementation is done and want to avoid additional migrations.

@capJavert 
We can let this be the intermediate branch :)

src/entity/ClaimableItem.ts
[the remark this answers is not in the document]
capJavert: As I said, just make it enum in typescript, in db its a string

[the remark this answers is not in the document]
capJavert: Look at `Product.type` for example

src/entity/ClaimableItem.ts
AmarTrebinjac: @capJavert In that case, I don't think it will be necessary. We already have the ID in the flags, so we can just make it a unique UUID.

idoshamun (approved):

src/entity/ClaimableItem.ts
capJavert: yes, if you throw from webhook it should trigger existing one already

src/entity/ClaimableItem.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  @Column({ type: 'text' })
+  @Index('IDX_claimable_item_email')
+  email: string;
+
+  @Column({ default: () => 'now()' })
+  createdAt: Date;
+
+  @Column({ nullable: true })
+  claimedAt?: Date;
```
capJavert, line 27: We might need join index for email/claimedAt due to query for existing items. But not blocking at the start.

capJavert: non blocking

src/entity/user/utils.ts
```diff
@@ -368,6 +370,41 @@ export const addNewUser = async (
...
+        cio,
+        email: body.email,
+        claimedSub: true,
+      });
+    }
+  } catch (err) {
+    req.log.error(
+      { err, userId: body.id },
+      'Error adding claimable items to user',
+    );
```
capJavert, line 404: We shoud leave this to throw, the outer catch will handle it and log properly so we see alert

capJavert (approved):

PR #2805, opened 2025-05-13
feat: paddle webhook handles sub from anonymous

AmarTrebinjac:
Added the logic to the paddle webhook to handle the case where someone subscribes without having registered an account with us yet.

### DIFF
--- a/src/routes/webhooks/paddle.ts
+++ b/src/routes/webhooks/paddle.ts
@@ -54,6 +54,7 @@ import { checkUserCoresAccess } from '../../common/user';
 import { CoresRole } from '../../types';
 import { TransferError } from '../../errors';
 import { remoteConfig } from '../../remoteConfig';
+import { ClaimableItem, ClaimableItemTypes } from '../../entity/ClaimableItem';
 
 export interface PaddleCustomData {
   user_id?: string;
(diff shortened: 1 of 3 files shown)


src/routes/webhooks/paddle.ts
```diff
@@ -103,54 +104,75 @@ export const updateUserSubscription = async ({
...
-    );
-    return false;
-  }
-
-  if (user.subscriptionFlags?.provider === SubscriptionProvider.AppleStoreKit) {
-    logger.error(
-      {
-        user,
-        data: event,
+  if (userId === 'anonymous') {
```
sshanzel, line 107: What does the ID have to be `anonymous`? It can either be empty or if it does, we can just search the DB which will return not found. Normally not a fan of strongly typed strings that are used across systems.

capJavert: Yeah, basically if there is no `customData.user_id` we treat as anonymous purchase and we should create claimable item row.

capJavert: Let's create a helper in `src/common/paddle/index.ts` for handling anonymous purchase and move the logic there, it will be a bit more readable.

src/routes/webhooks/paddle.ts
```diff
@@ -103,54 +104,75 @@ export const updateUserSubscription = async ({
...
+    const [customer, subscription] = await Promise.all([
+      paddleInstance.customers.get(data.customerId),
+      paddleInstance.subscriptions.get(data.id),
```
capJavert, line 110: Don't you have this from subscription event, I don't think we need to fetch?

src/routes/webhooks/paddle.ts
```diff
@@ -103,54 +104,75 @@ export const updateUserSubscription = async ({
...
+    ]);
+    await con.getRepository(ClaimableItem).insert({
```
capJavert, line 112: Might have to handle if claimable item already exists for email https://github.com/dailydotdev/daily-api/pull/2804/files#r2086337335

src/routes/webhooks/paddle.ts
AmarTrebinjac: You are absolutely right! Don't know how I missed that 😅

idoshamun (changes requested):
I don't see where we handle the cancelation of an unclaimed subscription

AmarTrebinjac:
> I don't see where we handle the cancelation of an unclaimed subscription
@idoshamun 

There already is logic for handling cancelled subscriptions from before.

If you mean that we need to check if the subscription is expired in some way before we give access to the user when he registers, then I think that should be done when we are applying the subscription to the user. Maybe at this point we can just set the status to `null` and when claiming the item, we can check if its active/inactive. 

What do you think?

src/routes/webhooks/paddle.ts
capJavert: Reminder to handle this case in regards to https://github.com/dailydotdev/daily-api/pull/2804#discussion_r2088527809

src/common/paddle/index.ts
```diff
@@ -190,3 +195,36 @@ export const getTransactionForProviderId = async ({
...
+  con: DataSource | EntityManager,
+  data: SubscriptionNotification | SubscriptionCreatedNotification,
+) => {
+  const customer = await paddleInstance.customers.get(data.customerId);
+
+  await con.getRepository(ClaimableItem).insert({
+    type: ClaimableItemTypes.Plus,
+    email: customer.email,
+    flags: {
+      cycle: extractSubscriptionCycle(data.items),
```
capJavert, line 223: What if cycle is undefined for any reason, we should handle it like we do for regular subscription case

src/common/paddle/index.ts
```diff
@@ -190,3 +195,36 @@ export const getTransactionForProviderId = async ({
...
+      createdAt: data?.startedAt,
+      subscriptionId: data?.id,
```
capJavert, line 225: Also this, in what cases can it be undefined?

capJavert (commented):
Almost there, some comments to check and two blockers:
- handling duplicate subs by email
- cancellation as Ido mentioned

AmarTrebinjac:
> Almost there, some comments to check and two blockers:
> 
> * handling duplicate subs by email


@capJavert I'm not really sure what we can do here other than maybe send a message/alert on slack that this is happening? I think you suggested earlier making the email field unique, but by the time they hit the claimable items table they've already subbed and paid, so I think it's better to insert them.

Open to suggestions, though!

capJavert:
By duplicates it is the case where a person buys two subscriptions with the same email, basically a guard for a case where they forget about it (no matter how much times was in between). They can even be registered and they buy anonymously after some time. 

I know if this happens it is kinda user's fault because they should be aware what they are buying but basically I would like us to at least know about those cases so we can address them and check if someone raises the support ticket.

src/common/paddle/index.ts
AmarTrebinjac: I moved the logic for checking the cycle earlier in the hook, so it handles both cases!

AmarTrebinjac:
> I don't see where we handle the cancelation of an unclaimed subscription

@idoshamun Handling it in [this PR](https://github.com/dailydotdev/daily-api/pull/2810/files) since the cio stuff was already being added there :)

idoshamun (approved):

sshanzel (approved):
Just wait for Ante's approval as the codeowner, but looks good to me.

src/common/paddle/index.ts
capJavert: @AmarTrebinjac did you not push it, or can you point me where that handling is?

src/common/paddle/index.ts
```diff
@@ -190,3 +195,49 @@ export const getTransactionForProviderId = async ({
...
+      provider: SubscriptionProvider.Paddle,
+      status: UserSubscriptionStatus.Active,
+    },
+  });
+
+  if (existingEntries.length > 0) {
+    throw new Error(
+      `User ${customer.email} already has a claimable subscription`,
+    );
+  }
```
capJavert, line 242: Should we move this above the `insert`, because now it will insert another row and then throw error.

capJavert (approved):

PR #2810, opened 2025-05-14
feat: add claimable item to user

AmarTrebinjac:
Add claimable item to user upon creation, as well as identifies a cio customer by email.


src/entity/user/utils.ts
```diff
@@ -368,6 +369,39 @@ export const addNewUser = async (
...
+  con: DataSource,
+  body: AddUserData,
+  req: FastifyRequest,
+) => {
+  try {
+    const claimableItems = await con
+      .getRepository(ClaimableItem)
+      .findBy({ email: body.email });
+    if (claimableItems.length > 0) {
+      const subscription = claimableItems.find(
```
AmarTrebinjac, line 382: Right now the only thing we're using this for is adding a sub bought from funnels, so this should be okay. We will need to change this to a loop when we add more options, though.

sshanzel:
> canceling through the dashboard

Did you have the proper setup for the webhook so that your system will receive the message from cancelling in the dashboard?

src/entity/ClaimableItem.ts
```diff
@@ -35,6 +35,6 @@ export class ClaimableItem {
   @Column({ nullable: true })
   claimedById?: string;
 
-  @ManyToOne('User', { lazy: true })
+  @ManyToOne('User', { lazy: true, onDelete: 'SET NULL' })
```
sshanzel, line 38: Should we delete the entity here too, when the user deletes their account?

src/common/paddle/index.ts
```diff
@@ -240,4 +250,30 @@ export const insertClaimableItem = async (
...
+
+  await identifyAnonymousFunnelSubscription({
+    cio,
+    email: customer.email,
+    claimedSub: false,
+  });
+};
+
+export const dropClaimableItem = async (
+  con: DataSource | EntityManager,
```
sshanzel, line 262: I think we have a common type for this.
```suggestion
  con: ConnectionManager,
```

src/common/paddle/index.ts
```diff
@@ -240,4 +250,30 @@ export const insertClaimableItem = async (
...
+  data: SubscriptionNotification | SubscriptionCreatedNotification,
```
sshanzel, line 263:
```suggestion
  data: SubscriptionNotification,
```

src/entity/user/utils.ts
sshanzel: Why not make this part of the query conditions?

sshanzel (commented):
Few things but nothing major.

src/entity/ClaimableItem.ts
idoshamun: yes we should

idoshamun (approved):
Just fix @sshanzel's comment and make cascade deletion

AmarTrebinjac:
> > canceling through the dashboard
> 
> Did you have the proper setup for the webhook so that your system will receive the message from cancelling in the dashboard?

@sshanzel omg, you are absolutely right! I hadn't hooked off that event in the paddle dashboard when I made this 🤦🏼

sshanzel (approved):
🚀

src/common/paddle/index.ts
```diff
@@ -237,17 +239,29 @@ export const insertClaimableItem = async (
...
+  const currentSubscription = existingEntries.find(
+    (entry) => entry.flags.subscriptionId === data.id,
+  );
+
+  const flags = {
+    cycle: state ? extractSubscriptionCycle(data.items) : null,
+    createdAt: state ? data.startedAt : null,
+    subscriptionId: state ? data.id : null,
+    provider: state ? SubscriptionProvider.Paddle : null,
+    status: state ? UserSubscriptionStatus.Active : null,
```
sshanzel, line 251: Just a random thought. I think we don't even have to run updates on this entity. This table is more of a record for when user is trying to claim their subscription. Cancelling it shouldn't matter here since the entity has served its purpose already.

src/entity/ClaimableItem.ts
```diff
@@ -30,11 +43,11 @@ export class ClaimableItem {
...
 
   @Column({ type: 'jsonb', default: {} })
-  flags = {};
+  flags: ClaimableItemFlags;
 
   @Column({ nullable: true })
   claimedById?: string;
 
-  @ManyToOne('User', { lazy: true })
+  @ManyToOne('User', { lazy: true, onDelete: 'SET NULL' })
```
sshanzel, line 51:
```suggestion
  @ManyToOne('User', { lazy: true, onDelete: 'CASCADE' })
```

src/common/paddle/index.ts
capJavert: Technically user could change subscription type in paddle while still not registered, and then it would claim different sub?

src/entity/user/utils.ts
```diff
@@ -368,6 +370,39 @@ export const addNewUser = async (
...
+      .findOneBy({ email: body.email, claimedById: IsNull() });
+
+    if (subscription) {
+      await con.getRepository(ClaimableItem).update(subscription.id, {
+        claimedById: body.id,
+        claimedAt: new Date(),
+      });
+      await con.getRepository(User).update(body.id, {
+        subscriptionFlags: subscription.flags,
+      });
```
capJavert, line 390: Let's do it in a transaaction

AmarTrebinjac: @capJavert

src/common/paddle/index.ts
sshanzel: Makes sense. Then yeah, let's keep the update all throughout.

src/common/paddle/index.ts
```diff
@@ -223,21 +239,59 @@ export const insertClaimableItem = async (
...
+  data: SubscriptionNotification,
+) => {
+  const customer = await paddleInstance.customers.get(data.customerId);
+
+  await con.getRepository(ClaimableItem).delete({
+    email: customer.email,
+    claimedAt: IsNull(),
+    flags: JsonContains({
+      subscriptionId: data.id,
+    }),
```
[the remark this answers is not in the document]
capJavert: non blocking

capJavert (approved):

PR #2812, opened 2025-05-14
refactor: create helpers to make funnel boot per use case

idoshamun:
As we are about to swap onboarding to use funnels, I thought it can be a good idea to put the building blocks to help with that. To make sure we follow all the best practices.

I also fixed an issue that was recovering a session even though it shouldn't

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -1,4 +1,9 @@
-import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';
+import {
+  FastifyInstance,
+  FastifyReply,
+  FastifyRequest,
+  RouteHandlerMethod,
+} from 'fastify';
 import createOrGetConnection from '../db';
 import { DataSource, Not, QueryRunner } from 'typeorm';
 import { clearAuthentication, dispatchWhoami } from '../kratos';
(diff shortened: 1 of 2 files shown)


src/routes/boot.ts
```diff
@@ -780,14 +789,18 @@ const getFunnel = async (req: FastifyRequest, res: FastifyReply) => {
   // If the session id is set, we should use it to fetch the funnel data
   if (sessionId) {
     const sessionFunnel = await freyjaClient.getSession(sessionId);
-    if (sessionFunnel?.session?.userId === userId) {
+    if (
+      sessionFunnel?.session?.userId === userId &&
+      (!query?.id || sessionFunnel.funnel.id === query.id) &&
+      (!query.v || sessionFunnel.funnel.version === parseInt(query.v))
+    ) {
```
capJavert, line 796: this looks like hell, can restructure it a bit? For example split to multiple if statements that return undefined

capJavert (approved):
One comment about the session resolving, if we can restructure it a bit.

src/routes/boot.ts
idoshamun: will do 🫡 
i'll add comments as well

capJavert: 👌 awesome!

capJavert (approved):

PR #2815, opened 2025-05-19
feat: geo resolving with maxmind

capJavert:
- it will be used for coresRoles resolving at first and later also in boot
- setup db download and copy to app
- add initialization of reader
- add test endpoint (to validate everything works before plugging into features)


src/routes/private.ts
```diff
@@ -188,4 +189,8 @@ export default async function (fastify: FastifyInstance): Promise<void> {
     routes: rpc,
     prefix: '/rpc',
   });
+
+  fastify.get('/test-ip', (req, res) => {
+    return res.status(200).send(getGeo({ ip: req.ip }));
+  });
```
capJavert, line 195: Test endpoint so I can validated implementation

.circleci/config.yml
```diff
@@ -16,6 +16,12 @@ jobs:
           command: |
             sudo npm install -g corepack@0.31.0
             pnpm install
+      - run:
+          name: Download GeoIP
+          command: wget "https://download.maxmind.com/app/geoip_download?edition_id=GeoIP2-City&license_key=$GEOIP_LICENSE_KEY&suffix=tar.gz" -O geoip.tar.gz
+      - run:
+          name: Untar
+          command: tar -xvzf geoip.tar.gz && mv */GeoIP2-City.mmdb geodb/
```
omBratteng, line 24: How often to these files update?

idoshamun: I think weekly

capJavert: less often then we deploy

omBratteng: I could deploy their geoipupdate as a daemonset on all our nodes, that keeps the db up to date. And then API (and any other service) just loads from k8s host

capJavert: :hidethepain: reverting fetching code

capJavert: 2859117ae9d25b59ed6e15086ece597c6c9d1ff1

src/routes/private.ts
capJavert: Yep I tested them already locally

omBratteng (approved):

PR #2816, opened 2025-05-19
feat: get previews by ids

sshanzel:
An endpoint that fetches the pricing details through IDs.

The existing ones we had for fetching by ids are going to be deprecated soon because we have different ways of processing the formatting of value on our new functions and we didn't touch old ones to avoid breaking old versions.


AmarTrebinjac (dismissed):
Looks good, but I think we should add alerts if there are no items returned, as this should never be the intended cause, and can cause losses like we experienced the other day.

sshanzel:
@capJavert thoughts on logging with 0 returned elements?

src/schema/paddle.ts
```diff
@@ -429,6 +463,24 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
 
       return consolidated.filter(Boolean) as BasePricingPreview[];
     },
+    pricingPreviewByIds: async (
```
capJavert, line 466: Where is this going to be used?

capJavert:
If we expect that all valid requests should return something I say we can trigger alert yes.

src/schema/paddle.ts
sshanzel: For the Funnel since they have a separate pricing IDs as compared to what is displayed from the Plus page.

capJavert: We can't use the method we have and then funnel on its side looks for pricing ids it needs? I am not really sure how it works but we are again starting to have one to many queries for pricing 😆

src/schema/paddle.ts
```diff
@@ -429,6 +465,34 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
+        duration: getPricingDuration(item),
+        trialPeriod: item.price.trialPeriod,
+        currency: {
+          code: preview.currencyCode,
+          symbol: removeNumbers(item.formattedTotals.total),
+        },
+      }));
+
+      if (!pricing?.length) {
+        logger.error(`Pricing preview by ids ${ids} returned no items`); // to send an alert
```
capJavert, line 490: Please use `{ err }, message` syntax, and tag with some property which we can then add to critical alert:

```suggestion
        const error = new ValidationError('pricing returned no items')
        logger.error({ err: error, ids  }, error.message); 
```

src/schema/paddle.ts
```diff
@@ -429,6 +465,34 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
       return consolidated.filter(Boolean) as BasePricingPreview[];
     },
+    pricingPreviewByIds: async (
+      _,
+      { ids, locale }: PaddlePricingPreviewByIdsArgs,
+      ctx,
+    ): Promise<BasePricingWithoutMetadata[]> => {
+      if (!ids?.length) {
+        throw new ValidationError('No ids provided');
+      }
```
capJavert, line 475: Not needed since `ids` are required in schema

src/schema/paddle.ts
capJavert: and then throw error

src/schema/paddle.ts
sshanzel: It can be an empty array though.

capJavert: diff marked it wrong I mean 
```suggestion
      if (!ids.length) {
        throw new ValidationError('No ids provided');
      }
```

src/schema/paddle.ts
omBratteng:
```suggestion
        const err = new ValidationError('pricing returned no items')
        logger.error({ err, ids  }, error.message); 
```

src/schema/paddle.ts
```diff
@@ -429,6 +465,35 @@ export const resolvers: IResolvers<unknown, AuthContext> = traceResolvers<
...
+        price: getProductPrice(item, locale),
+        duration: getPricingDuration(item),
+        trialPeriod: item.price.trialPeriod,
+        currency: {
+          code: preview.currencyCode,
+          symbol: removeNumbers(item.formattedTotals.total),
+        },
+      }));
+
+      if (!pricing?.length) {
```
capJavert, line 489:
```suggestion
      if (!pricing.length) {
```

capJavert (approved):

PR #2819, opened 2025-05-20
fix: rounding up should not be applied

sshanzel:
When the value has an example value of 7.49916 - we should not round up to 7.50. We should stay at 7.49.

### DIFF
--- a/__tests__/common/paddle/pricing.ts
+++ b/__tests__/common/paddle/pricing.ts
@@ -122,8 +122,8 @@ describe('getPrice', () => {
       divideBy: 30, // Convert monthly to daily price
     });
     expect(result).toEqual({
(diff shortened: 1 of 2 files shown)


__tests__/common/paddle/pricing.ts
```diff
@@ -122,8 +122,8 @@ describe('getPrice', () => {
       divideBy: 30, // Convert monthly to daily price
     });
     expect(result).toEqual({
-      amount: 49.97,
-      formatted: '49,97 kr',
+      amount: 49.96,
+      formatted: '49,96 kr',
```
sshanzel, line 126: Updated the affected test of previously rounding up.

capJavert (approved):

PR #2820, opened 2025-05-20
feat: geo resolving with maxmind

omBratteng:


.infra/index.ts
```diff
@@ -332,7 +332,30 @@ if (isAdhocEnv) {
...
+              mountOptions: 'implicit-dirs',
+            },
+          },
+        },
+      ],
+      volumeMounts: [
+        ...vols.volumeMounts,
+        {
+          name: 'geoip-data',
+          mountPath: '/usr/share/GeoIP',
```
capJavert, line 356: can we make it lowercase, just don't like uppercase in paths?

capJavert (approved):

.infra/index.ts
omBratteng: Yeah

PR #2823, opened 2025-05-20
feat: search via new mimir search

rebelchris:
Final output from endpoint might vary a bit as Viktor is still working on it.
but rest should be reviewable already.

### DIFF
--- a/src/integrations/mimir/types.ts
+++ b/src/integrations/mimir/types.ts
@@ -1,0 +1,18 @@
+// Keep the type flexible to allow for future changes
+export type MimirResponse = Partial<{
+  type: string;
+}>;
+
+export interface IMimirClient {
+  search({
+    query,
+    version,
+    offset = 0,
+    limit = 10,
+  }: {
+    query: string;
+    version: number;
+    offset: number;
+    limit: number;
+  }): Promise<MimirResponse>;
+}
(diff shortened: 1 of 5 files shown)


src/integrations/mimir/types.ts
```diff
@@ -0,0 +1,18 @@
+// Keep the type flexible to allow for future changes
+export type MimirResponse = Partial<{
+  type: string;
+}>;
+
+export interface IMimirClient {
```
capJavert, line 6: protobuf?

src/schema/search.ts
```diff
@@ -385,10 +386,19 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         limit: limit.toString(),
         offset: offset.toString(),
       });
+      let meilieSearchRes = {};
+      if (args.version >= 3) {
+        meilieSearchRes = await mimirClient.search({
```
capJavert, line 391: Oh or is the mimir client just meili search wrapper?

rebelchris: Yeah for now kinda is.
But ill check if viktor has protobuf already here

rebelchris:
@idoshamun or @omBratteng could you do a sanity check in Ante's absence

omBratteng (approved):

PR #2824, opened 2025-05-21
Revert "hotfix: adjust private and temporal memory limits"

capJavert:
This reverts commit 2747cc44189b2c1fb26b77b3c99ff6892bb8d873.

Memory is back down since we load the smaller db and only in some parts of api.

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -152,10 +152,10 @@ const bgRequests: pulumi.Input<{ cpu: string; memory: string }> = {
   memory: '256Mi',
 };
 
-const temporalLimits: pulumi.Input<{ memory: string }> = { memory: '256Mi' };
+const temporalLimits: pulumi.Input<{ memory: string }> = { memory: '320Mi' };
 const temporalRequests: pulumi.Input<{ cpu: string; memory: string }> = {
-  cpu: '50m',
-  memory: '150Mi',
+  cpu: '10m',
+  memory: '256Mi',
 };
 
 const initialDelaySeconds = 20;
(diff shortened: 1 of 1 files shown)


.infra/index.ts
```diff
@@ -427,13 +427,10 @@ if (isAdhocEnv) {
       env: [{ name: 'ENABLE_PRIVATE_ROUTES', value: 'true' }, ...commonEnv, ...jwtEnv],
       minReplicas: 2,
       maxReplicas: 2,
-      requests: {
+      limits: {
         memory: '256Mi',
         cpu: '25m',
       },
```
omBratteng, line 433: Keep these split, but we can set mem request to `192Mi` and limit to `256Mi`.
CPU can also be adjusted down to `10m`

.infra/index.ts
```diff
@@ -152,7 +152,7 @@ const bgRequests: pulumi.Input<{ cpu: string; memory: string }> = {
   memory: '256Mi',
 };
 
-const temporalLimits: pulumi.Input<{ memory: string }> = { memory: '512Mi' };
+const temporalLimits: pulumi.Input<{ memory: string }> = { memory: '256Mi' };
 const temporalRequests: pulumi.Input<{ cpu: string; memory: string }> = {
   cpu: '50m',
   memory: '150Mi',
```
omBratteng, line 158:
```suggestion
const temporalLimits: pulumi.Input<{ memory: string }> = { memory: '320Mi' };
const temporalRequests: pulumi.Input<{ cpu: string; memory: string }> = {
  cpu: '10m',
  memory: '256Mi',
```

omBratteng: Adjusted them for actual usage

omBratteng (approved):

PR #2825, opened 2025-05-21
test: error thrown and value update

sshanzel:
Fix:
- The error message thrown.
- Rounding down value.
- Removed unused tests.


__tests__/paddle.ts
```diff
@@ -935,180 +935,6 @@ describe('pricing preview by ids', () => {
   });
 });
 
-describe('core pricing preview', () => {
-  const QUERY = /* GraphQL */ `
-    query CorePricingPreview($type: PricingType, $locale: String) {
-      corePricingPreview(type: $type, locale: $locale) {
```
sshanzel, line 941: There is really no query `corePricingPreview`. We use the same query for Plus then send a different type :hide_the_pain:

Of course we have an existing test case for that.

capJavert (approved):
good vibes xD

PR #2826, opened 2025-05-21
feat: added onboarding funnel route

ilasw:
Add new funnel endpoint for onboarding page. 
Just reusing the functions from the other router with a different cookie used for saving the session.


src/routes/boot.ts
```diff
@@ -950,4 +950,14 @@ export default async function (fastify: FastifyInstance): Promise<void> {
       (req) => req.cookies[cookies.funnel.key],
     ),
   );
+
+  // Used to get the boot data for onboarding funnels
+  fastify.get(
+    '/onboarding',
+    funnelHandler(
```
capJavert, line 957: Can you explain why different endpoint for onboarding, just out of loop?

capJavert: and for example will we have more of these endpoints in the future with different cookies, maybe generic one with `:id` param would be better?

sshanzel: We are making the current /onboarding behave the same way as the Funnel does. Config based.

ilasw: Yes basically it is the same as the funnel BUT we need:
- different session id to retrieve (so we need a new cookie key)
- set a new GB feature key in order to retrieve the actual default version 

We could add a generic :id but we don't want to handle other cases then the `/onboarding` & `/funnel` for now. 
I can create an object ( similar to this `{ [endpointId]: { cookieKey, featureKey } }` ) and loop through the properties if we want to have a clearer overview of it. Wdyt?

ilasw: Updated ✔️

src/routes/boot.ts
```diff
@@ -942,12 +958,16 @@ export default async function (fastify: FastifyInstance): Promise<void> {
   });
 
   // Used to get the boot data for the web funnels
-  fastify.get(
-    '/funnel',
-    funnelHandler(
+  fastify.get('/:id', function (req, res) {
```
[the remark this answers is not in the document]
idoshamun: I agree with @capJavert `/:id` isn't a good idea.
My suggestion is to keep `/funnel` as a legacy endpoint.
Introduce two new endpoints:
* `/funnels/paid` - will have the same logic as the legacy
* `/funnels/onboarding` - the new endpoint for onboarding
Once that's done, we can point the apps to work against `/funnels/paid` and then remove the legacy one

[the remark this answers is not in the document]
capJavert: @idoshamun we can still do `/funnels/:id` with key/map that Luca mentioned above, so we don't have to create more routes each time we want new funnel, we just add to this map?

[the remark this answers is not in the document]
idoshamun: ah yes of course. i didn't to actually write these two, but more like the actual endpoints that we will support. my primary focus was that we need to keep the old one for backwards compat, and to agree on naming

src/routes/boot.ts
```diff
@@ -942,12 +958,16 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
+    const { id } = req.params as { id: string };
+    const funnel = funnelBoots[id];
+    if (!funnel) {
+      return res.status(404).send({ error: 'Funnel not found' });
```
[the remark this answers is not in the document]
capJavert: I think existing `funnel boot` tests are good example

src/routes/boot.ts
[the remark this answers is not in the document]
capJavert: Got it so:
- restore /funnel endpoint
- add /funnels/:id instead of just /:id
- and if you can add tests, I think existing `funnel boot` tests are good example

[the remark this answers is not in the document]
idoshamun: Yes!

src/cookies.ts
```diff
@@ -46,6 +46,16 @@ export const cookies: {
     },
     key: 'da4',
   },
+  onboarding: {
+    opts: {
+      maxAge: 1000 * 60 * 30,
+      httpOnly: true,
+      signed: false,
+      secure: env === 'production',
```
rebelchris, line 54: Don't we have a isProd function for this?
Otherwise the enum might be better we have somewhere

ilasw: Updated, thanks!

PR #2827, opened 2025-05-21
feat: refactor paddle webhook

capJavert:
Refactor paddle webhook so it is a bit easier to manage and add new functionality:
- main index file now only contains feature matcher and handlers, eg. for cores, default to plus
- move specific event handlers to processing files
- moved all helpers/utils to common/paddle
- moved slack helpers to common/paddle slack
- cleaned up some naming, added Paddle prefix so it is visible that type belongs to paddle

Original idea, [thread](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1747661968451329), yes its a lot of changes but mostly is if it makes the experience more readable and organized.


sshanzel (approved):
Great to see the changes.

omBratteng (commented):
Looks good.

Only remark is maybe the files in routes. It might be confusing the keep non-route files inside routes tree.
Maybe out could be

```
common/paddle/
  cores/
    cores.ts
    eventhandler.ts (index.ts)
    processing.ts
  plus/
    plus.ts?
    eventhandler.ts (index.ts)
    processing.ts
```

image

omBratteng (approved):

PR #2828, opened 2025-05-22
feat: create organization

omBratteng:
- Rename and move some enums for re-use
- Create new organization table and content preference child entity
- Create GraphQL query to list out organizations user is member of
- Update paddle webhook to handle new organization subscriptions

### DIFF
--- a/src/schema/search.ts
+++ b/src/schema/search.ts
@@ -416,6 +416,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         });
 
         const mimirSearchRes = await mimirClient.search(searchReq);
+
         const res = await meiliSearchResolver(
           source,
           {
@@ -424,8 +425,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
             pagination: {
               limit,
               offset,
-              total: mimirSearchRes.result.length,
-              current: offset + mimirSearchRes.result.length,
+              total: limit + 1,
+              current: mimirSearchRes.result.length,
             },
           },
           ctx,


__tests__/organizations.ts
```diff
@@ -0,0 +1,125 @@
...
+  );
+  client = state.client;
+});
+
+beforeEach(async () => {
+  loggedUser = null;
+  jest.clearAllMocks();
+});
+
+describe('organizations', () => {
```
sshanzel, line 37: Maybe we can let go of the `describe` that wraps everything? Just a personal preference though, non-blocking 🙏

src/common/paddle/index.ts
```diff
@@ -86,21 +84,28 @@ export const getPriceFromPaddleItem = (
...
   Plus = 'plus',
+  PlusOrganization = 'plusOrganization',
   Core = 'core',
 }
 
 export const getProductPurchaseType = ({
   id,
 }: {
-  id: string;
+  id?: string;
```
sshanzel, line 94: Is it actually possible to have id to be undefined?

sshanzel (approved):
Don't have anything blocking. Great work on setting it up 🙏 

~~One thing about the new experiment variant for the pricings, do you maybe need to update the `schema/paddle` that fetches the pricings to support this new Type?~~

Should be handled already 🚀

src/schema/organizations.ts
```diff
@@ -0,0 +1,65 @@
...
+    id: ID!
+    name: String!
+    image: String
+    seats: Int
+  }
+
+  type UserOrganization {
+    """
+    Date when the record was created
+    """
```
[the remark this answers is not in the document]
omBratteng: 🙈

__tests__/organizations.ts
omBratteng: Yeah probably

src/common/paddle/index.ts
```diff
@@ -98,9 +97,15 @@ export const getProductPurchaseType = ({
     throw new Error('Core product id is not set');
   }
 
+  if (!remoteConfig.vars.plusOrganizationProductId) {
+    throw new Error('Plus organization product id is not set');
+  }
```
rebelchris, line 102: Just need to make sure it's set when we merge else it will also break other paddle hooks?

src/common/paddle/index.ts
```diff
@@ -172,6 +177,27 @@ export const isCoreTransaction = ({
...
+    | TransactionPaymentFailedEvent
+    | SubscriptionCreatedEvent
+    | SubscriptionUpdatedEvent
+    | SubscriptionCanceledEvent;
+}): boolean => {
+  return event.data.items.some(
+    (item) =>
+      item.price?.productId &&
+      getProductPurchaseType({ id: item.price.productId }) ===
+        ProductPurchaseType.PlusOrganization,
```
rebelchris, line 197: I really wished he had a more generic approach to this 😂

src/common/paddle/organization.ts
```diff
@@ -0,0 +1,120 @@
...
+  const con = await createOrGetConnection();
+
+  const { data } = event;
+  const { user_id: userId }: PaddleCustomData = data?.customData || {};
+
+  const subscriptionType = extractSubscriptionCycle(data.items);
+  if (!subscriptionType) {
+    logger.error(
+      {
+        provider: SubscriptionProvider.Paddle,
```
rebelchris, line 58: We might have to extend these a bit to ensure we quickly know what sub provider is failing?

src/common/paddle/organization.ts
```diff
@@ -0,0 +1,120 @@
...
+    provider: SubscriptionProvider.Paddle,
+    status: SubscriptionStatus.Active,
+  };
+
+  const organization = await con.getRepository(Organization).save({
+    id: organizationId,
+    name: businessName,
+    seats: data.items[0].quantity,
+    subscriptionFlags: subscriptionFlags,
+  });
```
rebelchris, line 85: Just checking if we want to run all logic in here or rather in a background script via topic (user-organisation-created) kind of thing?

Regardless maybe wrap everything here in a transaction? (or is con a passed transaction, can't remember)

src/common/paddle/organization.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  await Promise.all([
+    // Add the user to the organization
+    con.getRepository(ContentPreferenceOrganization).save({
```
rebelchris, line 89: Missed RFC, what do we use this content pref org for?

src/common/paddle/organization.ts
```diff
@@ -0,0 +1,120 @@
...
+      },
+    }),
+    // Update the paddle subscription with the organization id
+    // This is needed to be able to update the subscription later
+    paddleInstance.subscriptions.update(data.id, {
+      customData: {
+        user_id: userId,
+        organization_id: organization.id,
+      },
+    }),
```
rebelchris, line 107: We choose not to do the frontend ID passing just so it can't fail?
But guess we can also match via user id if need be

src/common/paddle/organization.ts
```diff
@@ -0,0 +1,120 @@
...
+      con.getRepository(User).update(
+        { id: userId },
+        {
+          subscriptionFlags: updateSubscriptionFlags({
+            ...subscriptionFlags,
+            organizationId: organization.id,
+          }),
+        },
+      ),
+  ]);
```
rebelchris, line 119: This means the user creating org auto becomes a member or just gets plus for free?
eg they use up their own seat

Maybe more edge-case not sure if it would be a case where some admin person could create org, but not be a seat?
(CC: @idoshamun )

rebelchris (changes requested):
Maybe let's sync today, we use content pref to manage the active seats?

src/common/paddle/organization.ts
omBratteng: If they don't already have plus, they will get assigned a seat.
They can later (separate PR) downgrade themselves to a free user.

src/common/paddle/organization.ts
omBratteng: Yeah, just do it all on API side

src/common/paddle/organization.ts
omBratteng: To keep track of who is a member of the organization. Instead of a separate table

src/common/paddle/organization.ts
omBratteng: Probably should wrap it all in a transaction indeed. Will do that

src/common/paddle/organization.ts
omBratteng: good idea. I'll add that after #2827 is merged and I merge into this branch

src/common/paddle/index.ts
omBratteng: You have already set it I think?

src/common/paddle/index.ts
omBratteng: Yeah, I was thinking about it when I added it. Maybe I'll circle back later and clean that up

src/common/apple/subscription.ts
```diff
@@ -164,7 +165,11 @@ const renewalInfoToSubscriptionFlags = (
 
   if (isNullOrUndefined(cycle)) {
     logger.error(
-      { data, provider: SubscriptionProvider.AppleStoreKit },
+      {
+        data,
+        provider: SubscriptionProvider.AppleStoreKit,
+        processor: SubscriptionProcessor.Plus,
```
capJavert, line 171: processor means something else in UserTransaction, maybe call it `productType`, or something, just so it not overlaps

src/common/paddle/index.ts
```diff
@@ -110,9 +107,15 @@ export const getProductPurchaseType = ({
     throw new Error('Core product id is not set');
   }
 
+  if (!remoteConfig.vars.plusOrganizationProductId) {
+    throw new Error('Plus organization product id is not set');
+  }
```
capJavert, line 112: Maybe we should move these checks in specific cases, so that one does not break because other one is missing?

src/common/paddle/organization/eventHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { SubscriptionProcessor, SubscriptionProvider } from '../../plus';
+import { logger } from '../../../logger';
+import { createOrganizationSubscription } from './processing';
+
+export const processOrganizationPaddleEvent = async (event: EventEntity) => {
+  switch (event?.eventType) {
+    case EventName.SubscriptionCreated:
+      await createOrganizationSubscription({ event });
+      break;
```
capJavert, line 11: Do you want to wait for transaction completed to make sure payment went through.

src/common/plus/subscription.ts
```diff
@@ -0,0 +1,16 @@
+export enum SubscriptionProvider {
+  Paddle = 'paddle',
+  AppleStoreKit = 'storekit',
+}
+
+export enum SubscriptionProcessor {
+  Cores = 'cores',
```
capJavert, line 7: Cores is not really subscription it, is more ProductType for this enum?

capJavert:
:triggered

image

src/common/paddle/organization/processing.ts
```diff
@@ -0,0 +1,127 @@
...
+export const createOrganizationSubscription = async ({
+  event,
+}: {
+  event: SubscriptionCreatedEvent;
+}) => {
+  const organizationId = randomUUID();
+  const con = await createOrGetConnection();
+
+  const { data } = event;
+  const { user_id: userId }: PaddleCustomData = data?.customData || {};
```
capJavert, line 56: maybe add zod schema like we have for Cores, and pick up what you need from raw event, that way you don't have to do bunch of checks below if data is here, you just use zod type

for example make sure that there is one item in items, quantity exists, customData has user_id.

src/common/paddle/organization/processing.ts
```diff
@@ -0,0 +1,127 @@
...
+        },
+      }),
+      // Update the paddle subscription with the organization id
+      // This is needed to be able to update the subscription later
+      paddleInstance.subscriptions.update(data.id, {
+        customData: {
+          user_id: userId,
+          organization_id: organization.id,
+        },
+      }),
```
capJavert, line 113: maybe do this after DB stuff goes through, just to avoid issue where this goes through but then it fails on our DB side for any reason

src/common/paddle/organization/processing.ts
```diff
@@ -0,0 +1,127 @@
...
+    return defaultBusinessName;
+  }
+
+  try {
+    const business = await paddleInstance.businesses.get(
+      data.customerId,
+      data.businessId,
+    );
+    return business.name;
+  } catch (error) {
```
capJavert, line 42:
```suggestion
  } catch (
```

capJavert (commented):
Nothing blocking, left few comments to check. Overall looks straightforward 👍

src/common/paddle/organization.ts
capJavert: we could do it async as well, but transaction is a good start

src/common/apple/subscription.ts
omBratteng: Yeah, will do

src/common/plus/subscription.ts
omBratteng: You're right, will rename it

src/common/paddle/organization/processing.ts
omBratteng: I'll take a look at cores. I just based it on plus 🙈

src/common/paddle/organization/processing.ts
omBratteng: Probably a good idea yeah.

src/common/paddle/organization/eventHandler.ts
omBratteng: Will check again what we do for plus

src/common/apple/subscription.ts
omBratteng: I renamed the `SubscriptionProcessor` enum to `PurchaseType`.

src/common/plus/subscription.ts
omBratteng: I renamed the `SubscriptionProcessor` enum to `PurchaseType`.

src/common/paddle/organization/processing.ts
omBratteng: Moved to after DB

src/common/paddle/index.ts
```diff
@@ -96,28 +97,35 @@ export const getPriceFromPaddleItem = (
...
+      return PurchaseType.Cores;
+    case remoteConfig.vars.paddleProductIds?.organization:
+      return PurchaseType.Organization;
+    case remoteConfig.vars.paddleProductIds?.plus:
+      return PurchaseType.Plus;
     default:
-      return ProductPurchaseType.Plus;
+      throw new PurchaseTypeError('Product purchase type not found', id);
   }
 };
```
rebelchris, line 111: Just keep eye out when merging this that it doesn't break

rebelchris (approved):
Think it's ok for now :)

src/remoteConfig.ts
```diff
@@ -19,14 +20,14 @@ type RemoteConfigValue = {
...
   approvedStoreKitSandboxUsers: string[];
-  coreProductId: string;
   coresRoleRules: {
     regions: string[];
     role: CoresRole;
   }[];
   kvasirRequirePlus: boolean;
   paddleIps: string[];
   paddleTestDiscountIds: string[];
+  paddleProductIds: Record<PurchaseType, string>;
```
capJavert, line 30:
```suggestion
  paddleProductIds: Partial<Record<PurchaseType, string>>
```

so it denotes remote config is always optional

PR #2830, opened 2025-05-22
feat: seed system user

capJavert:
In prod it exists but I never added migration, so adding here to make sure it is created


omBratteng (commented):
maybe only insert the columns that need data? Like in the ghost user
https://github.com/dailydotdev/daily-api/blob/main/src/migration/1683721525235-DummyUser.ts

sshanzel (approved):
Don't have much to add. Looks good from my end.

rebelchris (approved):

capJavert:
Somehow clicked merge too soon, will update tests now 🙈

PR #2832, opened 2025-05-23
fix: system user test cases

capJavert:
Merged too soon and forgot tests https://github.com/dailydotdev/daily-api/pull/2830


omBratteng (approved):
I was starting to wonder 🙈

PR #2836, opened 2025-05-27
feat: option to leave a organization

rebelchris:
Option to leave a organization.

- Only members can leave, on admin/owner it will throw a error
- When not a seat user you will remain your subscription
- If seat user you'll be removed flags.


__tests__/organizations.ts
```diff
@@ -515,6 +516,176 @@ describe('mutation joinOrganization', () => {
...
+    mutation LeaveOrganization($id: ID!) {
+      leaveOrganization(id: $id) {
+        _
+      }
+    }
+  `;
+
+  beforeEach(async () => {
+    await con.getRepository(User).save({
+      id: 'org-rem-2',
```
rebelchris, line 530: To avoid issues as I use this to update user :)

PR #2837, opened 2025-05-27
feat: fetch our price metadata when getting price previews

omBratteng:
Need to be able to get the metadata we have for prices too, and can't rely on `PricingPreview` if it returns an experiment variation with a price ID that doesn't match what the organization is subscribed with

### DIFF
--- a/src/common/paddle/pricing.ts
+++ b/src/common/paddle/pricing.ts
@@ -1,8 +1,15 @@
-import { ConnectionManager, User } from '../../entity';
+import { createHmac } from 'node:crypto';
 import { EntityNotFoundError } from 'typeorm';
+import { z } from 'zod';
+import parseCurrency from 'parsecurrency';
+import {
+  ConnectionManager,
+  ExperimentVariant,
+  ExperimentVariantType,
+  User,
+} from '../../entity';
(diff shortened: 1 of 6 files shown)


src/entity/ExperimentVariant.ts
```diff
@@ -13,4 +17,7 @@ export class ExperimentVariant {
 
   @Column({ type: 'text', default: null })
   value: string;
+
+  @Column({ type: 'text' })
+  type: ExperimentVariantType;
```
omBratteng, line 22: Just added this for future proofing, if we start storing other experiment values in the table, we can store them by type.

src/common/paddle/pricing.ts
```diff
@@ -160,6 +166,34 @@ export const getPricingMetadata = async (
...
+  if (pricingIds.error) {
+    return {};
+  }
+
+  const items = await ctx.con
+    .createQueryBuilder()
+    .select('item')
+    .from(ExperimentVariant, 'ev')
+    .addFrom('jsonb_array_elements(ev.value::jsonb)', 'item')
+    .disableEscaping()
```
omBratteng, line 185: Have to disable escaping, otherwise it wraps `"` around the `jsonb_array_elements(ev.value::jsonb)` which is not valid psql :triggered:

Alternative is using a raw query

PR #2838, opened 2025-05-27
feat: add userId to subscription custom data on creation

AmarTrebinjac:
When claiming a subscription, we now add the userId to the subscriptions customdata, so that it gets properly updated on change events, i.e cancelling the sub.

Also added a log event to claiming a sub

[REF](https://dailydotdev.slack.com/archives/C02E2C3C13R/p1748340567666399?thread_ts=1748324078.181229&cid=C02E2C3C13R)

### DIFF
--- a/src/common/paddle/index.ts
+++ b/src/common/paddle/index.ts
@@ -379,14 +379,19 @@ export const logPaddleAnalyticsEvent = async (
   }
 
   const { data, occurredAt, eventId } = event;
-  const customData = data.customData as { user_id: string };
+  const customData = data.customData as {
+    user_id: string;
+    tracking_id?: string;
+  };
   const userId = await getUserId({
     userId: customData?.user_id,
     subscriptionId:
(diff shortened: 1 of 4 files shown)


sshanzel (approved):
Don't have anything to add. Although Ante is off so maybe you'll have to wait for @rebelchris to sign off the PRs before merging?

src/common/paddle/index.ts
```diff
@@ -379,14 +379,19 @@ export const logPaddleAnalyticsEvent = async (
...
+    tracking_id?: string;
+  };
   const userId = await getUserId({
     userId: customData?.user_id,
     subscriptionId:
       ('subscriptionId' in data && data.subscriptionId) || data.id,
   });
 
-  if (!userId) {
+  const analyticsId = userId || customData?.tracking_id;
```
AmarTrebinjac, line 392: we still want to send an analytics event even if the user is not registered, in which case we will use the tracking id

PR #2841, opened 2025-05-28
feat: mutation to delete organization

omBratteng:
Adds a mutation to be able to delete an organization when it passes all the checks
- Not active
- No members with assigned seats
- Owner tries to delete

### DIFF
--- a/src/schema/organizations.ts
+++ b/src/schema/organizations.ts
@@ -12,7 +12,11 @@ import {
 } from '../roles';
 import graphorm from '../graphorm';
 import {
+  notifyOrganizationUserJoined,
+  notifyOrganizationUserLeft,
+  notifyOrganizationUserRemoved,
   toGQLEnum,
+  updateFlagsStatement,
   updateSubscriptionFlags,
   uploadOrganizationImage,
 } from '../common';
@@ -45,6 +49,7 @@ import { SubscriptionStatus } from '../common/plus';
 
 export type GQLOrganizationMember = {
   role: OrganizationMemberRole;
+  seatType: ContentPreferenceOrganizationStatus;
   user: GQLUser;
 };
 export type GQLOrganization = Omit<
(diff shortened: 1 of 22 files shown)


src/schema/organizations.ts
```diff
@@ -779,5 +789,57 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+            },
+          },
+          where: { id },
+          relations: {
+            members: true,
+          },
+        });
+
+      if (
+        organization.subscriptionFlags?.status === SubscriptionStatus.Active
```
AmarTrebinjac, line 818: If the user cancels a subscription, but it still has a time period before ending, this will still prevent them from cancelling, right? 

If that's the case, we should handle that in some way to allow them.

src/schema/organizations.ts
```diff
@@ -779,5 +789,57 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        ContentPreferenceOrganization,
+        'userId' | 'status'
+      >[] = await organization.members;
+
+      if (
+        members.some(
+          (m) => m.status === ContentPreferenceOrganizationStatus.Plus,
+        )
+      ) {
+        throw new ForbiddenError(
```
AmarTrebinjac, line 835: If the organization has cancelled its plan, cant we update their plus memberships here and delete the organization afer?

src/schema/organizations.ts
rebelchris: That's a good point, maybe edge-case we eventually need to fix?

rebelchris (approved):
Nothing blocking imo

src/schema/organizations.ts
omBratteng: Yeah, can add a follow up task so fix that

PR #2842, opened 2025-05-29
feat: remove organization member

omBratteng:
- Adds new mutation for removing organization member
- Adds new mutation for updating member role
- Adds new mutation for assigning or unassigning member seat
- Updated sorting of organization members to show by role then when added to org


rebelchris (approved):
Looks good to me

PR #2843, opened 2025-05-29
feat: organization notifications

omBratteng:
- Sends notification to owner when user joins organization
- Sends notification to owner when user leaves organization
- Sends notification to user when removed from the organization

https://github.com/dailydotdev/streams/pull/48

AS-1122

### DIFF
--- a/src/schema/organizations.ts
+++ b/src/schema/organizations.ts
@@ -14,6 +14,7 @@ import graphorm from '../graphorm';
 import {
   notifyOrganizationUserJoined,
   notifyOrganizationUserLeft,
+  notifyOrganizationUserRemoved,
   toGQLEnum,
   updateFlagsStatement,
   updateSubscriptionFlags,
@@ -958,6 +959,7 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
                 }),
               },
             ),
+          notifyOrganizationUserRemoved(ctx.log, memberId, organizationId),
         ]);
       });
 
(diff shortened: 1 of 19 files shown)


src/workers/organization/organizationUserLeft.ts
```diff
@@ -0,0 +1,71 @@
...
+          { organizationId, memberId },
+          'No owner found for organization',
+        );
+        return;
+      }
+
+      const ownerUser = await owner.user;
+
+      await sendEmail({
+        send_to_unsubscribed: true,
```
rebelchris, line 53: This is right?

rebelchris (approved):
Should unsubscribed get them?

src/schema/organizations.ts
```diff
@@ -711,6 +714,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
                 },
               ),
           ]);
+
+          notifyOrganizationUserJoined(ctx.log, member.id, organization.id);
```
sshanzel, line 718: We can probably make the functions accept a single param of object. Sometimes the order gets confusing when both types are string and might get mixed up. At least with object, it is a bit harder to miss.

src/workers/index.ts
```diff
@@ -119,6 +121,8 @@ export const typedWorkers: BaseTypedWorker<any>[] = [
   postTranslated,
   transactionBalanceLogWorker,
   userBoughtCores,
+  organizationUserLeft,
+  organizationUserRemoved,
```
sshanzel, line 125: We are missing the joined one?

src/schema/organizations.ts
omBratteng: Good point, will add

PR #2845, opened 2025-05-30
feat: add last activity redis solution

rebelchris:
Added redis write on boot endpoint onResponse handler (to not block)
Then wrote a custom resolver using our dataloader system to read it back to graphQL

For now we'll just return null if redis empty or can't be found.
It seems heavily overkill to re-query in those cases.

### DIFF
--- a/src/routes/boot.ts
+++ b/src/routes/boot.ts
@@ -41,6 +41,7 @@ import {
   generateStorageKey,
   REDIS_BANNER_KEY,
   StorageTopic,
+  USER_LAST_ONLINE_KEY,
 } from '../config';
 import {
   ONE_DAY_IN_SECONDS,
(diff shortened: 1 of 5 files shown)


src/routes/boot.ts
```diff
@@ -961,6 +962,10 @@ const funnelHandler: RouteHandler = async (req, res) => {
 export default async function (fastify: FastifyInstance): Promise<void> {
   const con = await createOrGetConnection();
 
+  fastify.addHook('onResponse', async (req) => {
+    await setRedisObject(`${USER_LAST_ONLINE_KEY}${req.userId}`, Date.now());
```
omBratteng, line 966: Early return if no userid?

rebelchris: Ah yeah can do :)

omBratteng (approved):

omBratteng (approved):
LGTM. Non-blocking, but tests would be nice

omBratteng (approved):

src/schema/organizations.ts
```diff
@@ -98,6 +100,11 @@ export const typeDefs = /* GraphQL */ `
     The organization the user is a member of
     """
     organization: Organization
+
+    """
+    The date user was last active
+    """
+    lastActive: DateTime
```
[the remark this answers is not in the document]
rebelchris: It is copilot... it is...

PR #2849, opened 2025-06-02
fix: analytics events on organization paddle events

omBratteng:
AS-1132

### DIFF
--- a/src/common/paddle/slack.ts
+++ b/src/common/paddle/slack.ts
@@ -11,6 +11,7 @@ import createOrGetConnection from '../../db';
 import { User } from '../../entity/user/User';
 import { SubscriptionProvider } from '../plus';
 import { logger } from '../../logger';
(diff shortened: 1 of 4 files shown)


src/common/paddle/slack.ts
```diff
@@ -278,3 +279,137 @@ export const notifyNewPaddlePlusTransaction = async ({
 
   await webhooks.transactions.send({ blocks });
 };
+
+export const notifyNewPaddleOrganizationTransaction = async ({
```
rebelchris, line 283: Shouldn't this have a early return if it's not a org one?
I don't see it would only fire for those

rebelchris (commented):
One small remark

src/common/paddle/slack.ts
omBratteng: There is `findOneByOrFail`, so if no org is found, it will fail

PR #2853, opened 2025-06-05
feat: Squad awarding

rebelchris:
Backend for: https://github.com/dailydotdev/apps/pull/4555
Decided to go with flags.sourceId to determine if it's for a source or not.

### DIFF
--- a/src/common/njord.ts
+++ b/src/common/njord.ts
@@ -573,7 +573,7 @@ export const awardSquad = async (
 
   let firstAdminUser: SourceMember | undefined;
   for (const pm of privilegedMembers) {
-    const specialUser = { userId: pm.userId };
+    const specialUser = isSpecialUser({ userId: pm.userId });
     const canAward = checkUserCoresAccess({
       user: await pm.user,
       requiredRole: CoresRole.Creator,


src/common/njord.ts
```diff
@@ -113,6 +113,7 @@ export type TransactionProps = {
   productId: string;
   receiverId: string;
   note?: string;
+  flags?: object;
```
capJavert, line 116: let's type it as `Partial<{ sourceId: string }>`

src/common/njord.ts
```diff
@@ -531,6 +538,38 @@ const canAward = async ({
...
+    });
+    return awardUser(props, ctx);
+  } catch (error) {
+    logger.error(
+      {
+        context: 'award',
+        err: error,
+        props,
+        userId: ctx.userId,
+        transactionId: newTransactionId,
```
capJavert, line 564: you are missing transaction here, variable is not defined

capJavert: I think you can just call `awardUser` as well, you don't need to catch it since it will catch inside of itself

src/graphorm/index.ts
```diff
@@ -624,6 +624,7 @@ const obj = new GraphORM({
             totalUpvotes: value?.totalUpvotes ?? totalUpvotes,
             totalMembers: value?.totalMembers ?? totalMembers,
             featured: value?.featured ?? featured,
+            totalAwards: value?.totalAwards,
```
capJavert, line 627: add `totalAwards` to `defaultPublicSourceFlags` and fallback

src/schema/njord.ts
```diff
@@ -267,6 +268,11 @@ export const typeDefs = /* GraphQL */ `
       Note for the receiver
       """
       note: String
+
+      """
+      Extra flags for the transaction
+      """
+      flags: JSONObject
```
capJavert, line 275: let's type it as well with `{ sourceId }`

src/schema/sources.ts
```diff
@@ -1481,6 +1536,65 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         true,
       );
     },
+    sourceAwards: async (
+      _,
+      args: GQLCommentAwardArgs,
+      ctx: Context,
+      info,
+    ): Promise<Connection<any>> => {
+      await ctx.con.getRepository(Source).findOneOrFail({
```
capJavert, line 1545: Check source access, eg for private ones?

src/schema/sources.ts
```diff
@@ -1481,6 +1536,65 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        undefined,
+        true,
+      );
+    },
+    sourceAwardsTotal: async (
+      _,
+      args: GQLCommentAwardArgs,
+      ctx: Context,
+    ): Promise<{ amount: number }> => {
+      await ctx.con.getRepository(Source).findOneOrFail({
```
capJavert, line 1582: Also check source access

src/schema/sources.ts
```diff
@@ -1481,6 +1536,65 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          id: true,
+        },
+        where: { id: args.id },
+      });
+
+      const result = await ctx.con
+        .getRepository(UserTransaction)
+        .createQueryBuilder()
+        .select('COALESCE(SUM(value), 0)', 'amount')
+        .where(`flags->>'sourceId' = :sourceId`, { sourceId: args.id })
```
capJavert, line 1593: add index

src/common/njord.ts
capJavert: you can also add it to UserTransaction flags and then `Pick<UserTransactionFlags>`

capJavert (commented):
Nice, like the reuse of awardUser, some comments and tests and I think we are good 🚀

src/common/njord.ts
rebelchris: Also not to catch issues on the squad query?

src/common/njord.ts
```diff
@@ -531,6 +539,23 @@ const canAward = async ({
...
+  const sourceId = props.flags?.sourceId;
+  if (!sourceId) {
+    throw new ForbiddenError('You can not award this Squad');
+  }
+  await ctx.con.manager.getRepository(Source).findOneOrFail({
+    select: ['id'],
+    where: {
+      id: sourceId,
+    },
+  });
```
capJavert, line 555: Do we want to check if user is part of this squad? Maybe query source_member?

Also do we want to check if it is a squad, not regulard source?

src/graphorm/index.ts
```diff
@@ -1145,6 +1146,19 @@ const obj = new GraphORM({
...
+      sourceName: {
+        rawSelect: true,
+        select: (_, alias) => {
+          return `${alias}.flags->>'sourceId'`;
+        },
+        transform: async (value: string, ctx: Context) => {
+          const source = await ctx.getRepository(Source).findOne({
+            select: ['name'],
+            where: { id: value },
+          });
```
capJavert, line 1158: If no value skip query on db?

src/schema/comments.ts
```diff
@@ -101,12 +101,6 @@ interface ReportCommentArgs {
   reason: ReportReason;
 }
 
-export interface GQLUserComment {
-  vote: UserVote;
-  votedAt: Date | null;
-  awarded: boolean;
-}
```
capJavert: Where did this went?

src/schema/sources.ts
```diff
@@ -92,6 +95,12 @@ export interface GQLSourceCategory {
   updatedAt: Date;
 }
 
+export interface GQLUserReward {
```
capJavert, line 98:
```suggestion
export interface GQLUserAward {
```

src/schema/sources.ts
```diff
@@ -1481,6 +1539,67 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         true,
       );
     },
+    sourceAwards: async (
```
capJavert, line 1542: Tests for this and count query?

capJavert (changes requested):
Only blocking thing is the check if user is actually part of the squad, other then that some small comments.

src/common/njord.ts
rebelchris: Hmm, not sure it matters a lot for this notification?

src/graphorm/index.ts
rebelchris: Ah can add that yes

src/common/njord.ts
rebelchris: Yeah but it will cost you cores no?

I mean you don't need to be part of the squad to award it.
We could check that it's a squad for now?

src/common/njord.ts
```diff
@@ -531,6 +541,53 @@ const canAward = async ({
...
+    throw new ForbiddenError('You can not award this Squad');
+  }
+  await ensureSourcePermissions(ctx, sourceId);
+  // Extract the first eligble admin for this squad.
+  const privilegedMembers = await ctx.con.manager
+    .getRepository(SourceMember)
+    .find({
+      where: {
+        sourceId,
+        role: In([SourceMemberRoles.Admin, SourceMemberRoles.Moderator]),
```
capJavert, line 559: Should we just say admin?

capJavert (approved):

src/schema/sources.ts
```diff
@@ -1481,6 +1539,67 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      _,
+      args: GQLCommentAwardArgs,
+      ctx: Context,
+      info,
+    ): Promise<Connection<GQLUserAward>> => {
+      await ensureSourcePermissions(ctx, args.id);
+      await ctx.con.getRepository(Source).findOneOrFail({
```
omBratteng, line 1549: Since we already do `ensureSourcePermissions`, we don't really need the `findOneOrFail` here as long as we don't actually use it for anything. The `ensureSourcePermissions` already does a `findOneByOrFail`

src/schema/sources.ts
```diff
@@ -1481,6 +1539,67 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        true,
+      );
+    },
+    sourceAwardsTotal: async (
+      _,
+      args: GQLCommentAwardArgs,
+      ctx: Context,
+    ): Promise<{ amount: number }> => {
+      await ensureSourcePermissions(ctx, args.id);
+      await ctx.con.getRepository(Source).findOneOrFail({
```
omBratteng, line 1587: Same her as above comment, this findOneOrFail is redundant as `ensureSourcePermissions` already does the same.

src/graphorm/index.ts
```diff
@@ -624,6 +624,7 @@ const obj = new GraphORM({
             totalUpvotes: value?.totalUpvotes ?? totalUpvotes,
             totalMembers: value?.totalMembers ?? totalMembers,
             featured: value?.featured ?? featured,
+            totalAwards: value?.totalAwards ?? 0,
```
omBratteng, line 627: Minor nitpick, but group it with the others? 😅

src/common/njord.ts
```diff
@@ -531,6 +541,53 @@ const canAward = async ({
...
+
+  const firstAdminUser = privilegedMembers.find(async (pm) => {
+    const specialUser = { userId: pm.userId };
+    const canAward = await checkCoresAccess({
+      ctx,
+      userId: pm.userId,
+      requiredRole: CoresRole.Creator,
+    });
+    return !specialUser && canAward;
+  });
```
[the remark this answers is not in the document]
rebelchris: Yep! Missed it moving from map to find 🫠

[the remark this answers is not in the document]
capJavert: @rebelchris missed it as well, maybe query user with source member relation and then use checkUserCoresAccess because this will do multiple sequential queries to the db.

PR #2861, opened 2025-06-12
feat: estimated reach query and cancel post boost mutation

sshanzel:
Introduced estimated reach query and refactored the functions that are used on both ends (start post boost and this one).

Also introduced the mutation to manage cancellations. The amount to be refunded is going to applied on the worker.


rebelchris (approved):

src/schema/posts.ts
```diff
@@ -2479,6 +2506,35 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          postId,
+          userId,
+        });
+
+        await entityManager
+          .getRepository(Post)
+          .update(
+            { id: postId },
+            { flags: updateFlagsStatement<Post>({ boosted: false }) },
+          );
```
capJavert, line 2533: Better put skadi call at the end like we do with njord, it so that we first set boosted: false and then if skadi fails transaction will auto revert.

Like this skadi could go through but then DB fails, missed that in start post boost review as well.

src/schema/posts.ts
```diff
@@ -1945,6 +1987,25 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      _,
+      args: Omit<StartPostBoostArgs, 'userId'>,
+      ctx: AuthContext,
+    ): Promise<PostBoostReach> => {
+      const { postId, duration, budget } = args;
+      validatePostBoostArgs(args);
+      const post = await validatePostBoostPermissions(ctx, postId);
+      checkPostAlreadyBoosted(post);
+
+      const { estimatedReach } = await skadiBoostClient.estimatePostBoostReach({
```
capJavert, line 2000: Should we rate limit this in any way to protect skadi?

capJavert (approved):

src/schema/posts.ts
sshanzel: Good point! Let's do that.

PR #2862, opened 2025-06-12
feat: migration and bin script for syncing member count for non squad sources

AmarTrebinjac:
Currently we don't track members count for sources based on followers, but we've had some design changes where we would want this feature, for example in the new entity cards:

Screenshot 2025-06-12 at 13 57 39

### Jira ticket
MI-917


src/migration/1749656464598-SourceMemberFollowers.ts
```diff
@@ -0,0 +1,62 @@
...
+                    UPDATE source
+                    SET flags = jsonb_set(
+                                  flags,
+                                  '{totalMembers}',
+                                  to_jsonb((
+                                      SELECT COUNT(DISTINCT "userId")
+                                      FROM content_preference
+                                      WHERE "sourceId" = source_id_to_update
+                                      AND type = 'source'
+                                      AND status IN ('follow', 'subscribed')
```
omBratteng, line 34: Might need an index here

```sql
Aggregate  (cost=7426.33..7426.34 rows=1 width=8) (actual time=3257.892..3257.894 rows=1 loops=1)
  ->  Sort  (cost=7426.22..7426.28 rows=22 width=21) (actual time=3255.881..3256.482 rows=13507 loops=1)
        Sort Key: ""userId""
        Sort Method: quicksort  Memory: 385kB
        ->  Index Scan using ""IDX_content_preference_source_id"" on content_preference  (cost=0.57..7425.73 rows=22 width=21) (actual time=3.467..3208.035 rows=13507 loops=1)
              Index Cond: (""sourceId"" = 'hn'::text)
              Filter: ((status = ANY ('{follow,subscribed}'::text[])) AND (type = 'source'::text))
              Rows Removed by Filter: 668
Planning Time: 0.372 ms
Execution Time: 3257.966 ms
```

omBratteng: Something like this might work

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS "IDX_content_preference_sourceid_type_status_userid"
ON public.content_preference USING btree ("sourceId", "type", "status", "userId");
```

capJavert: Can't use `CONCURRENTLY` in migrations because it runs in transaction

capJavert: Either way this should be added manually because of table size.

bin/syncMembersForNonSquadSources.ts
```diff
@@ -0,0 +1,52 @@
+import '../src/config';
+import createOrGetConnection from '../src/db';
+
+(async (): Promise<void> => {
```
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert Yes, for squads, which looks into the `sourceMember` table instead of `contentPreferences`, I thought it appropriate to create a separate script :)

src/migration/1749656464598-SourceMemberFollowers.ts
```diff
@@ -0,0 +1,62 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class SourceMemberFollowers1749656464598 implements MigrationInterface {
+    name = 'SourceMemberFollowers1749656464598'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            CREATE OR REPLACE FUNCTION increment_source_members_count()
```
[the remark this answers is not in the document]
capJavert: One issue is that with your new trigger it would double increment because system still writes to source_member and content_preference for backward compatibility, even for squads. So we need to remove the old one or say we are gonna count followers separately. What is needed for the feature you are developing?

bin/syncMembersForNonSquadSources.ts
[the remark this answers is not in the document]
capJavert: Yes, you are right.

src/migration/1749656464598-SourceMemberFollowers.ts
```diff
@@ -0,0 +1,62 @@
...
+                END IF;
+            END;
+            $$
+        `);
+
+        await queryRunner.query(`
+            CREATE OR REPLACE TRIGGER increment_source_members_count
+            AFTER INSERT OR UPDATE OR DELETE ON "content_preference"
+            FOR EACH ROW
+            EXECUTE PROCEDURE increment_source_members_count()
```
capJavert, line 54: we should have WHEN condition here to trigger only on some rows, not all content_preference rows

src/migration/1749656464598-SourceMemberFollowers.ts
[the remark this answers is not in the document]
capJavert: I'll look tomorrow in what we look totalMembers and see if we can reuse at least the flag. Catch you tomorrow 🙌

src/migration/1749656464598-SourceMemberFollowers.ts
```diff
@@ -0,0 +1,105 @@
...
+                            )
+                WHERE id = NEW."sourceId";
+                RETURN NEW;
+            END;
+            $$
+            `);
+
+        await queryRunner.query(`
+            CREATE OR REPLACE TRIGGER blocked_source_members_count
+            AFTER UPDATE ON "content_preference"
```
capJavert, line 63: we also need after update for when row changes from blocked -> follow/subscribed, because user could first block the source and the follow it

src/migration/1749656464598-SourceMemberFollowers.ts
```diff
@@ -0,0 +1,105 @@
...
+                RETURN NEW;
+            END;
+            $$
+        `);
+
+        await queryRunner.query(`
+            CREATE OR REPLACE TRIGGER increment_source_members_count
+            AFTER INSERT ON "content_preference"
+            FOR EACH ROW
+            WHEN (NEW.type = 'source' AND NEW.status IN ('follow', 'subscribed'))
```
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert Doesn't seem to cause a double increment.

I've refactored the tests for the new changes and added an additional trigger :)

capJavert (approved):
Awesome 👍

PR #2863, opened 2025-06-12
refactor: filter url params for submitted links

omBratteng:


src/routes/private/rpc.ts
```diff
@@ -72,7 +72,7 @@ export default function (router: ConnectRouter) {
     const con = await createOrGetConnection();
 
     try {
-      req.url = standardizeURL(req.url);
+      req.url = standardizeURL(req.url).url;
```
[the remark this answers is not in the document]
omBratteng: Different uses, so OK.

src/common/links.ts
```diff
@@ -40,13 +36,70 @@ export const subscribeNotificationsLink = `${process.env.COMMENTS_PREFIX}?notify
...
+export const standardizeURL = (
+  inputUrl: string,
+): { url: string; canonicalUrl: string } => {
+  const url = new URL(inputUrl);
+
+  const isAllowedDomain = url.hostname in domainAllowedSearchParams;
+
+  const allowedSearchParams = isAllowedDomain
+    ? domainAllowedSearchParams[url.hostname as AllowedDomain]
+    : genericAllowedSearchParams;
```
capJavert, line 89: Like the concept of generic and domain filters 👌

src/entity/posts/utils.ts
```diff
@@ -298,7 +309,7 @@ export const createExternalLink = async ({
       createdAt: new Date(),
       sourceId: UNKNOWN_SOURCE,
       url,
-      canonicalUrl: url,
+      canonicalUrl: canonicalUrl || url,
```
capJavert, line 312: If I understand you correctly we are changing where we save clean and not clean URL, which is not how we did it before right?

capJavert (commented):
I Like the concept of filters you made, but seems like we are changing a lot because of one use case (medium friends link?). Would like to see tests added and carefully evaluate the changes again.

src/common/links.ts
```diff
@@ -40,13 +36,70 @@ export const subscribeNotificationsLink = `${process.env.COMMENTS_PREFIX}?notify
...
+  return filteredParams;
+};
+
+export const deduplicateURLSearchParams = (
+  params: URLSearchParams,
+): URLSearchParams => {
+  const dedupedParams = new URLSearchParams();
+  const seenValues = new Map<string, Set<string>>();
+
+  for (const [key, value] of params.entries()) {
```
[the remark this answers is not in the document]
capJavert: I understand its for specific cases, but since duplicate query params are perfectly valid not sure what the dedup is for?

src/entity/posts/utils.ts
omBratteng: That is correct. Previously we stored the URL without any params in both columns. Now we stored the filtered one in `url`.

And the one without params in canonicalUrl, which is how we did previously.

This is so that users can actually access correct and filtered URL (in cases like Medium Friend links).

src/common/links.ts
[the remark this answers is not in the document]
capJavert: ok this part I missed, you dedup only when name and value are the same?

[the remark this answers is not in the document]
omBratteng: Yeah, that is correct. Dedupe key-value, not just key

src/entity/posts/utils.ts
capJavert: I understand the reason, just think we might have some other cases where it works becaues of above, and now it will no longer work. Let's add tests and then evaluate again, probably check with Ido as well (if you did not already).

src/entity/posts/Post.ts
```diff
@@ -44,6 +44,7 @@ export type PostFlags = Partial<{
   deletedBy: string;
   vordr: boolean;
   coverVideo: string;
+  originalUrl: string;
```
idoshamun, line 47: should we add it in flags? we probably need it only for tracking purposes

src/entity/posts/utils.ts
omBratteng: Approval from Ido, so guess it's OK?

idoshamun: Could be that Ante has edge cases that I missed in mind

capJavert: I don't have any specific in mind, just the change of how we store URLs sounds it could mess something 🤷 

Asked Denis and from Yggdrasil side it should be fine since he mostly takes cleaned URL from API and only removes utm_ params. 

So let's try it and monitor 👍

omBratteng: Yeah, also checked with Denis on that regard

capJavert (approved):

PR #2867, opened 2025-06-16
feat: post boost workers

sshanzel:
Introduce two new workers that will listen on these two topics:
- api.v1.post_boost_completed (subscriber is a notification worker).
- api.v1.post_boost_canceled.

Refactored the transferCores to not inherently use the context guard to make it work with bg workers.

### Jira ticket
MI-888

### DIFF
--- a/__tests__/workers/postBoostCanceledCores.ts
+++ b/__tests__/workers/postBoostCanceledCores.ts
@@ -258,7 +258,6 @@ describe('postBoostCanceledCores worker', () => {
     const postId = 'p4';
     const refundAmountUsd = -1.5;
     const campaignId = 'campaign-negative';
-    const expectedCores = Math.floor(usdToCores(refundAmountUsd));
 
     await expectSuccessfulTypedBackground(worker, {
       userId,
@@ -271,8 +270,7 @@ describe('postBoostCanceledCores worker', () => {
       .getRepository(UserTransaction)
       .find({ where: { receiverId: userId } });
 
-    expect(transactions).toHaveLength(1);
-    const transaction = transactions[0];
-    expect(transaction.value).toBe(expectedCores);
+    // Should not create a transaction for negative refund amounts
+    expect(transactions).toHaveLength(0);
   });
 });
--- a/src/workers/postBoostCanceledCores.ts
+++ b/src/workers/postBoostCanceledCores.ts
@@ -16,6 +16,17 @@ const worker: TypedWorker<'api.v1.post-boost-canceled'> = {
     const { data } = message;
     const { userId, postId, refundAmountUsd, campaignId } = data;
 
+    if (refundAmountUsd < 0) {
+      logger.error(
+        {
+          data,
+          messageId: message.messageId,
+        },
+        'Cannot accept negative value for refund',
+      );
+      return;
+    }
+
     try {
       const { transactionId } = await con.transaction(async (entityManager) => {
         const post = await entityManager


src/common/njord.ts
```diff
@@ -192,111 +192,115 @@ const parseBalanceUpdate = ({
...
-export const transferCores = createAuthProtectedFn(
-  async ({
-    transaction,
-    entityManager,
-  }: TransferProps): Promise<TransferResult> => {
-    // TODO feat/transactions check if session is valid for real on whoami endpoint
+type TransferPropsRaw = Pick<TransferProps, 'transaction' | 'entityManager'>;
 
-    const njordClient = getNjordClient();
+export const transferCoresRaw = async ({
```
sshanzel, line 197: Nothing changed here really, just the indents and the type of the props to receive (from TransferProps to TransferPropsRaw).

capJavert: Context guard is important, you can resolve user from worker side, it only requires `userId`

src/common/njord.ts
```diff
@@ -1071,3 +1075,7 @@ export const throwUserTransactionError = async ({
 export const coresToUsd = (cores: number): number => {
   return cores / 100; // 100 Cores = 1 USD
 };
+
+export const usdToCores = (usd: number): number => {
+  return usd * 100; // 1 USD = 100 Cores
```
capJavert, line 1080: usd could be non integer so we need to make sure Cores are always integer

src/workers/notifications/postBoostCompleted.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const worker = generateTypedNotificationWorker<'api.v1.post-boost-completed'>({
+  subscription: 'api.post-boost-completed-notification',
+  handler: async ({ userId, postId, campaignId }, con) => {
+    await queryReadReplica(con, ({ queryRunner }) => {
+      return queryRunner.manager
+        .getRepository(Post)
+        .update(
+          { id: postId },
+          { flags: updateFlagsStatement<Post>({ boosted: false }) },
```
capJavert, line 16: This won't work, you can't do writes on read replica

src/workers/postBoostCanceledCores.ts
```diff
@@ -0,0 +1,84 @@
...
+  subscription: 'api.post-boost-canceled-cores',
+  handler: async (message, con, logger): Promise<void> => {
+    const { data } = message;
+    const { userId, postId, refundAmountUsd, campaignId } = data;
+
+    try {
+      const { transactionId } = await queryReadReplica(
+        con,
+        async ({ queryRunner }) => {
+          const entityManager = queryRunner.manager;
```
capJavert, line 24: Again, can't do writes on read replica

src/common/njord.ts
sshanzel: It was asking for `isTeamMember`, and I think it loses its purpose if I just override that value.

src/common/njord.ts
sshanzel: Right. Yeah, let's do the removal of decimal places here.

src/common/njord.ts
capJavert: let me take a look, probably just types

sshanzel:
@vpol Introduced two topics as below 🙏 

- api.v1.post_boost_completed (subscriber is a notification worker).
- api.v1.post_boost_canceled.

Both just needs:
- post id.
- user id.
- campaign id.

Then the cancelation topic just needs `refundAmountUsd`.

src/common/njord.ts
sshanzel: Amazing. Let me then revert this.

sshanzel: Done! 👌

src/common/njord.ts
```diff
@@ -1071,3 +1071,7 @@ export const throwUserTransactionError = async ({
 export const coresToUsd = (cores: number): number => {
   return cores / 100; // 100 Cores = 1 USD
 };
+
+export const usdToCores = (usd: number): number => {
+  return Math.floor(usd * 100); // 1 USD = 100 Cores
+};
```
capJavert, line 1077: Checked with product so they are aware we floor so users technically loos few cents?

src/workers/postBoostCanceledCores.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        if (!post) {
+          return {};
+        }
+
+        const userTransaction = await entityManager
+          .getRepository(UserTransaction)
+          .save(
+            entityManager.getRepository(UserTransaction).create({
+              id: randomUUID(),
```
capJavert, line 33: Can we check here if there is already UserTransaction for cancellation refund, technically message could fail or be executed twice and then we would return the Cores to the user twice. 

Maybe if campaignId is UUID we can just use that as transactioId which will automatically dedup because njord will not allow duplicate transaction.

__tests__/workers/postBoostCanceledCores.ts
```diff
@@ -0,0 +1,278 @@
...
+      .find({ where: { receiverId: userId } });
+
+    expect(transactions).toHaveLength(1);
+    const transaction = transactions[0];
+    expect(transaction.value).toBe(expectedCores);
+  });
+
+  it('should handle negative USD amount (edge case)', async () => {
+    const userId = '4';
+    const postId = 'p4';
```
capJavert, line 258: What do we do if negative?

src/workers/postBoostCanceledCores.ts
sshanzel: We can use the campaign id on creation, but then that would mean we won't be able to use that here during refund.

capJavert (approved):

PR #2868, opened 2025-06-17
fix: slack message anon cycle

capJavert:
For anon subs cycle can be undefined if events come in out of place, also added `anonymous` fallback for those orders to avoid undefined.

AS-1128


omBratteng (approved):

sshanzel (approved):
Thanks for the fix! 🚀

PR #2870, opened 2025-06-18
chore: bump infra deps

omBratteng:


.circleci/config.yml
```diff
@@ -1,20 +1,19 @@
...
-  gcp-gcr: circleci/gcp-gcr@0.16.9
+  gcp-gcr: circleci/gcp-gcr@0.16.10
   pulumi: pulumi/pulumi@2.1.0
-  gcp-cli: circleci/gcp-cli@3.3.0
+  gcp-cli: circleci/gcp-cli@3.3.1
 jobs:
   build:
     docker:
-      - image: cimg/node:22.13
+      - image: cimg/node:22.15
```
capJavert, line 9: Keep it the same until we update main code? I know there are not really breaking changes, just easier to update in one go with search and replace.

.infra/package.json
```diff
@@ -7,12 +7,12 @@
...
-    "@pulumi/gcp": "^8.19.1",
-    "@pulumi/kubernetes": "^4.21.1",
-    "@pulumi/pulumi": "^3.150.0"
+    "@dailydotdev/pulumi-common": "^2.10.2",
+    "@pulumi/gcp": "^8.35.0",
+    "@pulumi/kubernetes": "^4.23.0",
+    "@pulumi/pulumi": "^3.177.0"
   },
-  "packageManager": "pnpm@9.14.4+sha256.26a726b633b629a3fabda006f696ae4260954a3632c8054112d7ae89779e5f9a",
+  "packageManager": "pnpm@10.12.1",
```
capJavert, line 15: Also keep it the same?

Dockerfile
```diff
@@ -16,14 +16,14 @@ RUN apt-get update \
 RUN mkdir -p /opt/app
 WORKDIR /opt/app
 
+RUN npm install -g corepack
```
capJavert, line 19: We want a specfic version of corepack since last time it broke everything

omBratteng: Can actually replace it with `corepack enable` and just use the one from the image?

package.json
```diff
@@ -151,7 +151,7 @@
     "@types/jsonwebtoken": "^9.0.9",
     "@types/lodash": "^4.17.17",
     "@types/markdown-it": "14.1.2",
-    "@types/node": "22.15.29",
+    "@types/node": "22.15.x",
```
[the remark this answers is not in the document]
omBratteng: npm accepts `x` as a wildcard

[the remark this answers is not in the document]
omBratteng: I also prefer fixed versions, except for patch versions of @types, because they usually updates more frequent to fix minor types, and don't align with the package they're typing. Like there is not nodejs version 22.15.32

PR #2871, opened 2025-06-18
fix: types in posts test

omBratteng:
Just solves some typing errors in the posts test. Leftovers from when I tried to get Claude to fix the test but explicitly told it to ignore TS errors, and it fixed TS errors instead and ignored test errors.

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -76,6 +76,7 @@ import { rateLimiterName } from '../src/directive/rateLimit';
 import { badUsersFixture, usersFixture } from './fixture/user';
 import { PostCodeSnippet } from '../src/entity/posts/PostCodeSnippet';
 import {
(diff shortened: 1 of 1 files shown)


__tests__/posts.ts
```diff
@@ -6816,7 +6837,7 @@ describe('Source post moderation approve/reject', () => {
     await saveSquadFixtures();
     await con.getRepository(SourcePostModeration).save([
       {
-        id: pendingId,
+        id: pendingId as string,
```
capJavert, line 6840: Why are `as string` cases needed?

omBratteng: it was throwing me an error previously :triggered:, but when I removed it now, it doesn't.
Something about `string` not matching 
~~~
`${string}-${string}-${string}-${string}-${string}`
~~~

PR #2875, opened 2025-06-26
feat: move fully to mimir search

rebelchris:
Moving to mimir fully and added support for better dynamic filtering.


src/schema/search.ts
```diff
@@ -290,6 +335,111 @@ const meiliSearchResolver = feedResolver(
...
+}: {
+  contentCuration?: string[];
+  time?: SearchTime;
+}): Filter[] => {
+  const output: Filter[] = [
+    new Filter({
+      field: 'private',
+      condition: {
+        value: { value: false },
+        case: 'boolFilter' as const,
```
omBratteng, line 398: Can we put these strings in an enum or const object?

rebelchris:
bugbot run

src/integrations/retry.ts
```diff
@@ -142,8 +143,11 @@ export async function fetchParseBinary<T extends Message<T>>(
   parser: T,
 ): Promise<T> {
   const res = await fetch(url, fetchOpts);
+  if (isTest) {
+    // Jest only support mocks of JSON
+    return parser.fromJson(await res.json());
+  }
```
omBratteng, line 149: :lolsob:

omBratteng (approved):

src/integrations/retry.ts
capJavert: should be able to return binary from nock by using inverse method toBinary and json object? Or does the jest not support `res.arrayBuffer`?

PR #2878, opened 2025-06-30
refactor: skadi boost client

sshanzel:
After setting up all the integrations, it was mentioned that the URL for the Post Boost feature is going to be different from our current Skadi integration client.

The changes here basically just copy-paste the existing functions and remove the newly unused ones from `SkadiClient` to `SkadiApiClient`.

### Jira ticket
MI-888

### DIFF
--- a/.env
+++ b/.env
@@ -44,7 +44,7 @@ ANALYTICS_URL=http://localhost:5000
 NJORD_ORIGIN=http://njord-transactions-server
 FREYJA_ORIGIN=http://localhost:7800
 SKADI_ORIGIN=http://localhost:8080
-SKADI_API_ORIGIN=http://localhost:8090
+SKADI_API_ORIGIN=http://skadi-api-server.local.svc.cluster.local
 MIMIR_ORIGIN=http://localhost:7600
 
 MEILI_INDEX=dev
--- a/.env
+++ b/.env
@@ -44,6 +44,7 @@ ANALYTICS_URL=http://localhost:5000
 NJORD_ORIGIN=http://njord-transactions-server
 FREYJA_ORIGIN=http://localhost:7800
 SKADI_ORIGIN=http://localhost:8080
+SKADI_API_ORIGIN=http://localhost:8090
 MIMIR_ORIGIN=http://localhost:7600
 
 MEILI_INDEX=dev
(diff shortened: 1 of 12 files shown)


capJavert (approved):

.env
```diff
@@ -44,6 +44,7 @@ ANALYTICS_URL=http://localhost:5000
 NJORD_ORIGIN=http://njord-transactions-server
 FREYJA_ORIGIN=http://localhost:7800
 SKADI_ORIGIN=http://localhost:8080
+SKADI_API_ORIGIN=http://skadi-api-server.local.svc.cluster.local
```
rebelchris, line 47: Probably didn't mean to expose like this right?

capJavert: this is local URL, its fine, we can obfuscate the ones in prod file, but also not mandatory

.infra/Pulumi.adhoc.yaml
```diff
@@ -38,9 +38,9 @@ config:
       6S+tkxhzAgWZNOhtqvSTFxKH9c51/jTCLDrPvHNlzNoeqhvaowUmvOrRoT1TZ8hs
       7QIDAQAB
       -----END PUBLIC KEY-----
-  api:debeziumDbPass: "12345"
+  api:debeziumDbPass: '12345'
```
capJavert, line 41: let's not do this accidental changes 🙏

.env
omBratteng: It's fine, it's just the full dns name inside cluster, short hand would just be `skadi-api-server`.

src/integrations/skadi/api/clients.ts
```diff
@@ -0,0 +1,147 @@
...
+    });
+  }
+
+  getCampaigns({
+    limit,
+    offset,
+    userId,
+  }: GetCampaignsProps): Promise<PromotedPostList> {
+    return this.garmr.execute(() => {
+      return fetchParse(`${this.url}/promote/post/list`, {
```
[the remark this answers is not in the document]
sshanzel: Maybe we can do a web guild session introducing protobuf, so we can instinctively use it the next time we come across integrations.

sshanzel:
@omBratteng merging now since the concern has been addressed. In any case, this is going to be merged to an intermediate branch. Feel free to drop your thoughts if any adjustments are needed.

PR #2881, opened 2025-06-30
feat: post to user source

omBratteng:
Allow users to share links and post freeform content to their own user source. Added a compatibility check so ensure the user source is created if it does not already exist.

Also don't think we need to create it when user is created, as many users might never post to their own source, so we can limit the amount of "empty" sources.

### Jira ticket
AS-1157

### DIFF
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -12,6 +12,7 @@ import {
   PostType,
   Source,
   SourceMember,
+  SourceType,
   SquadSource,
   translateablePostFields,
   User,
@@ -241,9 +242,10 @@ export const createFreeformPost = async ({
   args,
   ctx,
 }: CreateFreeformPostArgs) => {
-  const { private: privacy } = await con
-    .getRepository(Source)
-    .findOneByOrFail({ id: args.sourceId });
+  const { private: privacy } = await con.getRepository(Source).findOneByOrFail({
+    id: args.sourceId,
+    type: In([SourceType.Squad, SourceType.User]),
+  });
 
   const createdPost = con.getRepository(FreeformPost).create({
     ...args,
--- a/src/common/post.ts
+++ b/src/common/post.ts
@@ -242,7 +242,7 @@ export const createFreeformPost = async ({
   ctx,
 }: CreateFreeformPostArgs) => {
   const { private: privacy } = await con
-    .getRepository(SquadSource)
+    .getRepository(Source)
     .findOneByOrFail({ id: args.sourceId });
 
   const createdPost = con.getRepository(FreeformPost).create({
--- a/src/cron/updateSourcePublicThreshold.ts
+++ b/src/cron/updateSourcePublicThreshold.ts
@@ -1,5 +1,9 @@
 import { Cron } from './cron';
-import { Source, SQUAD_IMAGE_PLACEHOLDER } from '../entity';
+import {
+  REPUTATION_THRESHOLD,
+  Source,
+  SQUAD_IMAGE_PLACEHOLDER,
+} from '../entity';
 import { updateFlagsStatement } from '../common';
 
 export const updateSourcePublicThreshold: Cron = {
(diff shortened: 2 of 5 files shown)


src/common/post.ts
```diff
@@ -242,7 +242,7 @@ export const createFreeformPost = async ({
   ctx,
 }: CreateFreeformPostArgs) => {
   const { private: privacy } = await con
-    .getRepository(SquadSource)
+    .getRepository(Source)
```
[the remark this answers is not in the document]
capJavert: Yeah could be good, to avoid accidentally creating those in machine sources for now.

src/cron/updateSourcePublicThreshold.ts
```diff
@@ -25,7 +29,7 @@ export const updateSourcePublicThreshold: Cron = {
             from "content_preference" cp
             join "user" u on cp."userId" = u.id
             where cp."referenceId" = "source".id and cp.type = 'source'
-                and cp.flags->>'role' = 'admin' and u.reputation >= 250
+                and cp.flags->>'role' = 'admin' and u.reputation >= ${REPUTATION_THRESHOLD}::int
```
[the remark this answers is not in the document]
omBratteng: Fixed

src/common/post.ts
```diff
@@ -241,9 +242,10 @@ export const createFreeformPost = async ({
   args,
   ctx,
 }: CreateFreeformPostArgs) => {
-  const { private: privacy } = await con
-    .getRepository(SquadSource)
-    .findOneByOrFail({ id: args.sourceId });
+  const { private: privacy } = await con.getRepository(Source).findOneByOrFail({
```
capJavert, line 245: add a test for this case?

src/schema/sources.ts
```diff
@@ -1213,6 +1213,41 @@ export const ensureSourcePermissions = async (
   throw new ForbiddenError('Access denied!');
 };
 
+export const ensureUserSourceExists = async (userId: string, con: DataSource) =>
```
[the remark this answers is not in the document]
capJavert: 👍

src/schema/sources.ts
```diff
@@ -1594,7 +1633,10 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       ctx: Context,
       info,
     ): Promise<Connection<GQLSource>> => {
-      const filter: FindOptionsWhere<Source> = { active: true };
+      const filter: FindOptionsWhere<Source> = {
+        active: true,
+        type: Not(SourceType.User),
```
capJavert, line 1638: should this be applied to stuff like trendingSource, popularSource, topVideoSource and so on?

Also mutations like leaveSource and deleteSource, not sure if they would pass and let user delete/leave?

src/schema/sources.ts
```diff
@@ -1213,6 +1213,45 @@ export const ensureSourcePermissions = async (
...
+    );
+
+    const sourceMemberExists = await entityManager
+      .getRepository(SourceMember)
+      .findOne({
+        where: { sourceId: user.id, userId: user.id },
+      });
+
+    if (!sourceMemberExists) {
+      await entityManager.getRepository(SourceMember).insert({
```
capJavert, line 1246: As https://github.com/dailydotdev/daily-api/pull/2881/files#r2175178994 this could still be race condition, better to use on conflict do nothing with query builder.

capJavert: Also, just remembered, we technically would need to add ContentPreferenceSource as well as we do for joinSource because of backward compatibility.

capJavert (commented):
Just some additional comments, sorry for raising, did not managed to get a good look at the PR yesterday.

src/schema/sources.ts
omBratteng: Wasn't supposed to commit this part 🙈
Going to do that in a separate PR

src/schema/sources.ts
omBratteng: Convert to qb with do nothing on conflict, also added ContentPreferenceSource upsert

src/schema/sources.ts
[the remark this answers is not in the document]
capJavert: just trolling at this point xD

capJavert (approved):
Just last comment on avoiding doing a transaction each time. 🚀

PR #2883, opened 2025-07-01
fix: skadi integration - data types of responses

sshanzel:
There are certain properties that needs to be formatted before returning it to the FE.

For example, Float values are sent as string instead of number, and the Date value is integer which we had to do some processing first.

After the changes, the endpoints are now working.

### Jira ticket
MI-888


.infra/Pulumi.adhoc.yaml
```diff
@@ -38,9 +38,9 @@ config:
       6S+tkxhzAgWZNOhtqvSTFxKH9c51/jTCLDrPvHNlzNoeqhvaowUmvOrRoT1TZ8hs
       7QIDAQAB
       -----END PUBLIC KEY-----
-  api:debeziumDbPass: "12345"
```
sshanzel, line 41: Accidentally committed. I'll revert quickly 🤦

src/common/utils.ts
```diff
@@ -279,3 +279,29 @@ export const getCurrencySymbol = ({
...
+
+type CamelCasedKey<S extends string> = S extends `${infer Head}_${infer Tail}`
+  ? `${Head}${Capitalize<CamelCasedKey<Tail>>}`
+  : S;
+
+export type ObjectSnakeToCamelCase<T extends object> = {
+  [K in keyof T as K extends string ? CamelCasedKey<K> : never]: T[K];
+};
+
+export function convertObjectKeysToCamelCase<T extends object>(
```
capJavert, line 297: Can we do this statically, we know which properties map to which, also this kind of snake to camel would be for free with protobuf. So basically just map it by hand.

sshanzel: Kept the type magic to make it easier for types. Moved to manual mapping now.

capJavert: would it not be more consistent to return camel cased from SkadiApiClient and then everywhere else you use camel cased? Then you have to do the mapping only once inside the client.

sshanzel: Then, if we only want the mappings from the client, should we not do the generic function since we are pretty sure that we just want to have them converted?

sshanzel: Totally agree, we should phase it out. Just to get started with production testing. After that, I'll start cleaning it up.

PR #2885, opened 2025-07-01
feat: brief stats

capJavert:


src/entity/posts/BriefPost.ts
```diff
@@ -11,4 +11,7 @@ export class BriefPost extends Post {
 
   @Column({ nullable: true })
   readTime?: number;
+
+  @Column({ type: 'text', array: true, default: [] })
+  collectionSources: string[];
```
capJavert, line 16: Reusing collections sources column because graphorm already knows how to query each source data

src/integrations/feed/types.ts
```diff
@@ -135,24 +135,3 @@ export enum BriefingType {
 export enum BriefingModel {
   Default = 'ai_briefing',
 }
-
-// TODO feat-brief replace with schema
-export type UserBriefingRequest = {
-  userId: string;
-  frequency: BriefingType;
-  modelName: string;
-};
```
capJavert, line 144: Replaced with schema

src/common/typedPubsub.ts
```diff
@@ -139,11 +138,13 @@ export type PubSubSchema = {
     organizationId: Organization['id'];
     memberId: User['id'];
   };
-  'api.v1.brief-generate': UserBriefingRequest & {
+  'api.v1.brief-generate': {
+    payload: UserBriefingRequest;
```
capJavert, line 142: changing topic signature so we can extend easier, but its fine we are still not in prod

PR #2888, opened 2025-07-02
feat: return similar shared posts

omBratteng:
Return an array of similar public posts with the shared link.

### Jira ticket
AS-1163


src/schema/posts.ts
```diff
@@ -2437,7 +2439,13 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         return fetchLinkPreview(cleanUrl);
       }
 
-      return post;
+      const similarPosts = await ctx.con.getRepository(SharePost).find({
+        where: { sharedPostId: post.id, private: false },
```
AmarTrebinjac, line 2443: "Similar" post in this case just means the same post being shared other places? Its not a "you might also like" type of similar, correct?

AmarTrebinjac: In which case I think it would be more clear if they were called "reshared", "derivative", or some other keyword that signifies its the same post, just with potentially slightly different commentary

omBratteng: Yeah, it's just public shared posts with the same url

omBratteng: Renamed to `relatedPublicPosts`

capJavert (approved):
Use graphorm, other then that good to go 👍

src/schema/posts.ts
```diff
@@ -2437,7 +2439,13 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         return fetchLinkPreview(cleanUrl);
       }
 
-      return post;
+      const similarPosts = await ctx.con.getRepository(SharePost).find({
```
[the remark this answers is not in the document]
omBratteng: Probably good, will refactor to that

AmarTrebinjac (approved):

sshanzel:
Hi @TAT2k30, thank you for your support in using the app and your interest in helping fix issues. In case you are looking for things to work on, feel free to check our main repo and filter by `help-wanted` label. The link below should point you there.

https://github.com/dailydotdev/daily/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22

We currently use [GitPod for development](https://gitpod.io/#https://github.com/dailydotdev/apps/) outside the team so you won't have to run it locally on your machine.

PR #2893, opened 2025-07-03
feat: brief crons

capJavert:
Support new digest type brief, it can be sent daily/weekly.

Brief generation running on bg workers still, we will monitor and move it to separate app if needed once number of subscriptions goes up to avoid changing infra too early.


src/workers/personalizedDigestEmail.ts
```diff
@@ -213,18 +218,34 @@ const digestTypeToFunctionMap: Record<
...
   },
 };
 
 const worker: Worker = workerToExperimentWorker({
   subscription: 'api.personalized-digest-email',
   handler: async (message, con, logger, pubsub, allocationClient) => {
-    if (process.env.NODE_ENV === 'development') {
-      return;
-    }
-
```
capJavert, line 227: Was not really needed since we have `enablePersonalizedDigest` env as well which disables the workers locally by defualt

src/common/personalizedDigest.ts
```diff
@@ -420,7 +420,7 @@ export const schedulePersonalizedDigestSubscriptions = async ({
 }: {
   queryBuilder: SelectQueryBuilder<UserPersonalizedDigest>;
   logger: FastifyBaseLogger;
-  sendType: UserPersonalizedDigestSendType;
+  sendType: UserPersonalizedDigestSendType[];
```
capJavert, line 423: just changing type so that we can support multiple sendTypes, for this function is just used for logging, so nothing breaking

src/workers/personalizedDigestEmail.ts
```diff
@@ -213,18 +218,50 @@ const digestTypeToFunctionMap: Record<
...
+
+          const post = entityManager.getRepository(BriefPost).create({
+            id: postId,
+            shortId: postId,
+            authorId: userId,
+            private: true,
+            visible: false,
+          });
+
+          await entityManager.getRepository(BriefPost).save(post);
```
omBratteng, line 240: what create and then save?

PR #2896, opened 2025-07-04
fix: allow user source type in feed

omBratteng:


omBratteng:
> Can you double check if this is the only places? Feel like we use source_types in more?

Whoops, auto-merged. I am looking to see if there's anywhere else I missed

PR #2899, opened 2025-07-04
feat: add funnelIds to remoteConfig and handle development defaults

ilasw:


src/remoteConfig.ts
```diff
@@ -28,6 +28,7 @@ type RemoteConfigValue = {
   paddleIps: string[];
   paddleTestDiscountIds: string[];
   paddleProductIds: Partial<Record<PurchaseType, string>>;
+  funnelIds: Record<string, string>;
```
capJavert, line 31: Can we make this typed `web_funnel_id`, `onboarding_funnel_id`, I don't think we will introduce many different funnel types as much + also they need other changes so better to have it typed

capJavert (approved):
just the type change

rebelchris (approved):
Nothing besides Ante's feedback

PR #2900, opened 2025-07-07
fix: update `updateSourcePublicThreshold` cron to set public threshold on user source

omBratteng:
Re-used the existing cron job, just adjusted the SQL so that it accepts both squad and user source, then wrapped some squad specific requirements

### Jira ticket
AS-1165

### DIFF
--- a/src/cron/updateSourcePublicThreshold.ts
+++ b/src/cron/updateSourcePublicThreshold.ts
@@ -18,29 +18,33 @@ export const updateSourcePublicThreshold: Cron = {
       .set({ flags: updateFlagsStatement({ publicThreshold: true }) })
       .where(
         /* sql */ `
-          "type" IN (:...sourceTypes) AND
-          (
+          "type" IN (:...sourceTypes)
+          AND (
             "type" != :squadSourceType
             OR (
-              "type" = :squadSourceType AND
-              "image" IS NOT NULL AND
-              "image" != :imagePlaceholder AND
-              "description" IS NOT NULL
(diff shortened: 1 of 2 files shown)


src/cron/updateSourcePublicThreshold.ts
```diff
@@ -16,24 +17,43 @@ export const updateSourcePublicThreshold: Cron = {
...
+            "type" != :squadSourceType
+            OR (
+              "type" = :squadSourceType
+              AND "image" IS NOT NULL
+              AND "image" != :imagePlaceholder
+              AND "description" IS NOT NULL
+            )
+          )
+          AND (flags->>'publicThreshold')::boolean IS NOT TRUE
+          AND (flags->>'vordr')::boolean IS NOT TRUE
```
capJavert, line 32: Can you explain what changes you did for reference?

omBratteng: Ah, yeah, all lines changed since I made it more readable 🙈 

Changed the first check for type to be `IN` instead of `=` so that it checks for squad and user source.

I wrapped the `image` and `description` to only be checked when source type is `squad` (or `:squadSourceType`).

Then I also wrapped the `flags.totalMembers` and `flags.totalPosts` to only be checked when source type is `squad`

omBratteng: Can check the commit 9f1aba35e4867c059d9419a47be0a5ec6f8ce5ed, which is where I did the change for checks.

PR #2905, opened 2025-07-08
feat: send brief to slack

capJavert:
- added separate subscription to not mix up logic for other slack posts
  - brief worker sends only the posts that user is author of, eg. their own briefs
- added exclusion for brief on other worker
- updated tests to reflect new guards

### DIFF
--- a/src/workers/postAddedSlackChannelSendBrief.ts
+++ b/src/workers/postAddedSlackChannelSendBrief.ts
@@ -3,42 +3,42 @@ import { TypedWorker } from './worker';
 import fastq from 'fastq';
 import {
   BRIEFING_SOURCE,
-  Post,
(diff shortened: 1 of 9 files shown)


src/workers/postAddedSlackChannelSendBrief.ts
```diff
@@ -0,0 +1,183 @@
...
+              source: true,
+              author: true,
+            },
+          }),
+          con.getRepository(UserSourceIntegrationSlack).find({
+            where: {
+              sourceId: data.post.sourceId,
+              userIntegration: {
+                userId: data.post.authorId,
+              },
```
capJavert, line 55: Only send brief post to integrations created by author

PR #2913, opened 2025-07-10
refactor: opt-in for squad post added notifications

omBratteng:
We move squad new post added notifications to opt-in instead of opt-out


src/workers/notifications/postAdded.ts
```diff
@@ -79,11 +79,13 @@ const worker: NotificationWorker = {
...
         });
-        const members = await getSubscribedMembers(
+
+        const members = await getSubscribedMembers({
           con,
-          NotificationType.SquadPostAdded,
-          source.id,
-          {
+          type: NotificationType.SquadPostAdded,
+          byStatus: NotificationPreferenceStatus.Subscribed,
```
rebelchris, line 86: Shouldn't this also be byNotMuted?

rebelchris (commented):
Not sure why the one is specific?

src/workers/notifications/postAdded.ts
omBratteng: No, here we want to select all members with status subscribed, i.e. opt-in.

`byNotStatus` is selecting all users who _do not_ have the status

rebelchris: Ah ok, but don't you want subscribed OR notify status?

omBratteng: No, just need notification preference status. It selects all members from source_member, then filters out where preference status is subscribed.

Ended up not using content_preference table, since we'd need to switch everything over so we don't accidentally miss somewhere 🙈

omBratteng: Also uses "same" logic as notifications from machine sources

rebelchris: Ah ok, I thought we had 2 status, subscribed or notified?

omBratteng: No, `NotificationPreferenceStatus` is just `muted` or `subscribed`

```ts
export enum NotificationPreferenceStatus {
  Muted = 'muted',
  Subscribed = 'subscribed',
}
```

src/workers/notifications/utils.ts
```diff
@@ -54,11 +75,13 @@ export const getSubscribedMembers = (
       notificationType: type,
       referenceId,
       type: notificationPreferenceMap[type],
-      status: NotificationPreferenceStatus.Muted,
+      status: byStatus || byNotStatus,
```
capJavert, line 78: How are both under same where?

omBratteng: I can split it into two functions instead

capJavert: let's do that please, I think it will be easier

omBratteng:
bugbot run

src/workers/notifications/utils.ts
```diff
@@ -151,10 +201,16 @@ export async function articleNewCommentHandler(
       : NotificationType.ArticleNewComment;
 
   if (source.type === SourceType.Squad) {
-    const members = await getSubscribedMembers(con, type, post.id, {
-      userId: In(users),
-      sourceId: source.id,
-      role: Not(SourceMemberRoles.Blocked),
+    const members = await getOptOutMembers({
```
rebelchris, line 204: I'm a bit confused by the naming or the function maybe?
you get opted out members, but those get the email?

rebelchris (commented):
think i'm confused with naming now.

src/workers/notifications/utils.ts
```diff
@@ -37,6 +37,38 @@ export const uniquePostOwners = (
...
+  const subscribedquery = builder
+    .subQuery()
+    .select('np."userId"')
+    .from(NotificationPreference, 'np')
+    .where(`"np"."userId" = "${memberQuery.alias}"."userId"`)
+    .andWhere({
+      notificationType: type,
+      referenceId,
+      type: notificationPreferenceMap[type],
+      status: NotificationPreferenceStatus.Subscribed,
```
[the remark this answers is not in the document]
omBratteng: Yeah, based on the AND WHERE EXISTS query just a few lines further down.

rebelchris (approved):
This way looks good to me

PR #2915, opened 2025-07-10
fix: fallback values just in case any is null

sshanzel:
Just to be sure we won't get any runtime error when using the values to calculate the estimated reach after the client fetches.

### Jira ticket
MI-888

### DIFF
--- a/src/integrations/skadi/api/clients.ts
+++ b/src/integrations/skadi/api/clients.ts
@@ -139,9 +139,9 @@ export class SkadiApiClient implements ISkadiApiClient {
       );
 
       return {
-        impressions: response.impressions,
-        clicks: response.clicks,
-        users: response.users,
+        impressions: response.impressions ?? 0,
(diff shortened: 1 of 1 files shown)


src/integrations/skadi/api/clients.ts
```diff
@@ -139,9 +139,9 @@ export class SkadiApiClient implements ISkadiApiClient {
       );
 
       return {
-        impressions: response.impressions,
-        clicks: response.clicks,
-        users: response.users,
+        impressions: response.impressions ?? 0,
+        clicks: response.clicks ?? 0,
+        users: response.users ?? 0,
```
capJavert, line 144: again, ideally tests should test this

capJavert (approved):

src/integrations/skadi/api/clients.ts
sshanzel: Will make tests to mock requests in the future 🙏

PR #2917, opened 2025-07-11
refactor: canceling a boost should refund immediately

sshanzel:
Previously, it was not possible to return the amount to be refunded directly after sending the cancellation request.

@vpol and I have discussed and realized that it is now possible, so instead of having a sole worker to refund the remaining amount, we will do it synchronously instead.

Thankfully nothing in production is using this yet.

### Jira ticket
MI-888


__tests__/schema/posts/boost.ts
```diff
@@ -105,6 +106,12 @@ beforeEach(async () => {
   jest.clearAllMocks();
   await ioRedisPool.execute((client) => client.flushall());
 
+  // Mock transferCores to succeed by default - will be overridden in specific tests
+  (transferCores as jest.Mock).mockResolvedValue({
+    senderBalance: { newBalance: '10000' },
+    receiverBalance: { newBalance: '5000' },
+  });
```
capJavert, line 113: There is a `createMockNjordTransport` inside which you can setup accounts balance through njord calls and the e2e test it without faking anything

src/schema/posts.ts
```diff
@@ -2768,29 +2768,67 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          entityManager,
+        });
+
+        return {
+          transfer,
+          transaction: {
+            referenceId: campaignId,
+            transactionId: userTransaction.id,
+            balance: {
+              amount: parseBigInt(transfer.receiverBalance!.newBalance),
```
capJavert, line 2825:
```suggestion
              amount: parseBigInt(transfer.receiverBalance?.newBalance),
```

src/schema/posts.ts
```diff
@@ -2768,29 +2768,67 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const result = await ctx.con.transaction(async (entityManager) => {
         await entityManager
           .getRepository(Post)
           .update(
             { id: postId },
             { flags: updateFlagsStatement<Post>({ campaignId: null }) },
           );
 
-        await skadiApiClient.cancelPostCampaign({
+        const { currentBudget } = await skadiApiClient.cancelPostCampaign({
```
capJavert, line 2788: we should do db update only once skadi call goes through

src/schema/posts.ts
```diff
@@ -2768,29 +2768,67 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+              value: usdToCores(toRefund),
+              valueIncFees: 0,
+              fee: 0,
+              flags: { note: 'Post boost canceled' },
+              referenceId: campaignId,
+              referenceType: UserTransactionType.PostBoost,
+            }),
+          );
+
+        const transfer = await transferCores({
```
capJavert, line 2813: try catch missing here with `TransferError` so that its handled correctly.

capJavert: what happens if cores refund fails on njord for any reason, should skadi campaign stay cancelled?

sshanzel: Yes, it should stay cancelled. We'll have to do some intervention if that happens.

capJavert: do you want to log something specific and setup bew critical alert over those logs?

src/schema/posts.ts
```diff
@@ -2768,29 +2768,81 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+              referenceId: campaignId,
+              transactionId: userTransaction.id,
+              balance: {
+                amount: parseBigInt(transfer.receiverBalance?.newBalance),
+              },
+            },
+          };
+        } catch (error) {
+          if (error instanceof TransferError) {
+            await throwUserTransactionError({
```
capJavert, line 2833: Just to check this will also rollback `campaignId` on Post if transaction fails, you ok with that?

capJavert: then you need to handle that manually in catch block

sshanzel: I mean, we should not rollback the campaign id, and keep it as it is. So that when the user tries to send another request, the result would be as intended. Where the campaign is already not boosted, which is true.

capJavert: ok 👍

PR #2919, opened 2025-07-11
fix: estimated boost reach - reduced params and returning pure int

sshanzel:
This was the source of the issue. However, we want to know more about it by wrapping it in a try-catch instead.

Only thing I can think of where this can break was if the `impressions` is actually a string. We'll find out.

### DIFF
--- a/__tests__/schema/posts/boost.ts
+++ b/__tests__/schema/posts/boost.ts
@@ -2278,15 +2278,15 @@ describe('mutation cancelPostBoost', () => {
 
 describe('query boostEstimatedReach', () => {
   const QUERY = `
-    query BoostEstimatedReach($postId: ID!, $duration: Int!, $budget: Int!) {
-      boostEstimatedReach(postId: $postId, duration: $duration, budget: $budget) {
+    query BoostEstimatedReach($postId: ID!) {
+      boostEstimatedReach(postId: $postId) {
         min
         max
       }
     }
   `;
 
-  const params = { postId: 'p1', duration: 7, budget: 5000 };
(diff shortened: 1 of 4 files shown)


src/schema/posts.ts
```diff
@@ -2153,14 +2153,29 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-      // const difference = impressions * 0.08;
-      // const estimatedReach = {
-      //   min: Math.max(impressions - difference, 0),
-      //   max: impressions + difference,
-      // };
-
-      return { min: impressions, max: impressions };
+      // Parse impressions if it's a string, or use as-is if it's already a number
+      const parsedImpressions =
+        typeof impressions === 'string' ? parseFloat(impressions) : impressions;
```
capJavert, line 2158: Who controls the response, Viktor?

capJavert: It is due to new logic returning float and in gql its defined as int

__tests__/schema/posts/boost.ts
```diff
@@ -2300,56 +2300,6 @@ describe('query boostEstimatedReach', () => {
...
-    loggedUser = '1';
-
-    return testQueryErrorCode(
-      client,
-      { query: QUERY, variables: { ...params, budget: 100001 } },
-      'GRAPHQL_VALIDATION_FAILED',
-    );
-  });
-
-  it('should return an error if budget is not divisible by 1000', async () => {
```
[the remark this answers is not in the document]
sshanzel: These tests are based on the params we just removed.

PR #2925, opened 2025-07-14
feat: check brief updated time before generation

capJavert:
- fetch brief generation time before scheduling briefs
- check with each brief (digest) `lastUpdatedAt`
- log error which we can use for alerting

### DIFF
--- a/src/cron/dailyDigest.ts
+++ b/src/cron/dailyDigest.ts
@@ -16,6 +16,7 @@ import {
 } from '../entity';
 import { Cron } from './cron';
 import { addHours, startOfHour, subDays } from 'date-fns';
+import { briefFeedClient } from '../common/brief';
 
 const sendTypes = [
   UserPersonalizedDigestSendType.workdays,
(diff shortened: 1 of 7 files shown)


src/cron/personalizedDigest.ts
```diff
@@ -81,6 +84,24 @@ const cron: Cron = {
...
+          logger.error(
+            {
+              briefingUptime,
+              personalizedDigest,
+              emailSendTimestamp,
+              previousSendTimestamp,
+              emailBatchId,
+            },
+            'Brief generation skipped, outdated',
+          );
```
capJavert, line 100: Will add critical alert on this so we can handle per case at the begining

src/cron/personalizedDigest.ts
```diff
@@ -61,6 +62,8 @@ const cron: Cron = {
     // Make sure digest is sent at the beginning of the hour
     const timestamp = startOfHour(new Date());
 
+    const briefingUptime = await briefFeedClient.getBriefLastUpdate();
```
capJavert, line 65: at the end managed to fetch once per cron execution, and then check below for each row

src/cron/dailyDigest.ts
```diff
@@ -53,6 +54,14 @@ const cron: Cron = {
     // Make sure digest is sent at the beginning of the hour
     const timestamp = startOfHour(new Date());
 
+    const briefingUptime = await briefFeedClient
+      .getBriefLastUpdate()
+      .catch(() => {
+        return {
+          updatedAt: new Date(0),
+        };
+      });
```
capJavert, line 63: if not available set to 1-1-1970 and skip

PR #2929, opened 2025-07-16
fix: exclude show on feed from new post worker

rebelchris:
We should exclude posts with showOnFeed: false, from triggering the new post worker.

I've only applied it for machine sources, my thinking is that private squads could have this flag, but that should not stop the notifications.

### DIFF
--- a/src/workers/notifications/postAdded.ts
+++ b/src/workers/notifications/postAdded.ts
@@ -51,7 +51,7 @@ const worker: NotificationWorker = {
 
     if (
       source?.type === SourceType.Machine &&
-      (post.flags as PostFlags)?.showOnFeed?.toString() === 'false'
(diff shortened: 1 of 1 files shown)


src/workers/notifications/postAdded.ts
```diff
@@ -49,6 +49,13 @@ const worker: NotificationWorker = {
       return;
     }
 
+    if (
+      source.type === SourceType.Machine &&
+      (post.flags as PostFlags)?.showOnFeed === false
```
capJavert, line 54: do check if they are `string`, because from cdc they can be as type suggests

rebelchris: Parsing as string now.

capJavert (approved):

PR #2930, opened 2025-07-16
fix: allow connecting multiple briefing sources on same org

capJavert:
If users from same organization try to connect briefing it will fail due to existing restriction of one integration per source (that we apply for squads). For now I just ignore this for briefing source, if we add more sources for slack we can move this to more reusable logic.


capJavert:
bugbot run

PR #2932, opened 2025-07-17
feat: add resume upload for users

ilasw:
- Adding new `uploadResume` mutation
- add Google Cloud Storage functions for upload and deletion
- Integrates resume cleanup into user deletion workflow

### Jira ticket
MI-944

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -134,7 +134,6 @@ import { SubscriptionProvider, SubscriptionStatus } from '../src/common/plus';
 import * as njordCommon from '../src/common/njord';
 import { createClient } from '@connectrpc/connect';
 import { Credits, EntityType } from '@dailydotdev/schema';
-import * as googleCloud from '../src/common/googleCloud';
 import { RESUMES_BUCKET_NAME } from '../src/common/googleCloud';
 import { fileTypeFromBuffer } from './setup';
 
@@ -145,6 +144,7 @@ jest.mock('../src/common/geo', () => ({
 
 const uploadFileFromBuffer = jest.fn();
 const uploadResumeFromBuffer = jest.fn();
+const deleteResumeByUserId = jest.fn();
 jest.mock('../src/common/googleCloud', () => ({
   ...(jest.requireActual('../src/common/googleCloud') as Record<
     string,
@@ -153,6 +153,7 @@ jest.mock('../src/common/googleCloud', () => ({
   uploadFileFromBuffer: (...args: unknown[]) => uploadFileFromBuffer(...args),
   uploadResumeFromBuffer: (...args: unknown[]) =>
     uploadResumeFromBuffer(...args),
+  deleteResumeByUserId: (...args: unknown[]) => deleteResumeByUserId(...args),
 }));
 
 let con: DataSource;
@@ -4385,24 +4386,24 @@ describe('mutation deleteUser', () => {
     loggedUser = '1';
 
     // Mock that the resume file exists
-    jest.spyOn(googleCloud, 'deleteResumeByUserId').mockResolvedValue(true);
+    deleteResumeByUserId.mockResolvedValue(true);
 
     await client.mutate(MUTATION);
 
     // Verify the resume was deleted
-    expect(googleCloud.deleteResumeByUserId).toHaveBeenCalledWith('1');
+    expect(deleteResumeByUserId).toHaveBeenCalledWith('1');
   });
 
   it('should handle case when user has no resume', async () => {
     loggedUser = '1';
 
     // Mock that the resume file doesn't exist
-    jest.spyOn(googleCloud, 'deleteResumeByUserId').mockResolvedValue(false);
+    deleteResumeByUserId.mockResolvedValue(false);
 
     await client.mutate(MUTATION);
 
     // Verify the function was called but no error was thrown
-    expect(googleCloud.deleteResumeByUserId).toHaveBeenCalledWith('1');
+    expect(deleteResumeByUserId).toHaveBeenCalledWith('1');
 
     // User should still be deleted
     const userOne = await con.getRepository(User).findOneBy({ id: '1' });


ilasw:
bugbot run

ilasw:
bugbot run

src/common/googleCloud.ts
```diff
@@ -1,9 +1,14 @@
...
 import { PropsParameters } from '../types';
 import path from 'path';
 import { BigQuery } from '@google-cloud/bigquery';
 import { Query } from '@google-cloud/bigquery/build/src/bigquery';
 import { subDays } from 'date-fns';
+import { Readable } from 'stream';
+import { logger } from '../logger';
+
+export const RESUMES_BUCKET_NAME =
+  process.env.GCS_PDF_BUCKET || 'daily-dev-resumes';
```
capJavert, line 11: why do you need the `||` why not just use env?

src/schema/users.ts
```diff
@@ -2364,6 +2377,43 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        throw new ValidationError('File is missing!');
+      }
+
+      const upload = await resume;
+
+      // Validate file extension
+      const extension = upload.filename?.split('.')?.pop()?.toLowerCase();
+      if (extension !== 'pdf') {
+        throw new ValidationError('Extension must be .pdf');
+      }
```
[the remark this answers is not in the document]
sshanzel: I agree with this. Let's check the type rather than the extension.

src/common/googleCloud.ts
```diff
@@ -33,6 +38,110 @@ export const downloadJsonFile = async <T>({
...
+    contentType,
+    resumable: false,
+    ...options,
+  });
+
+  // Return a promise that resolves when the upload is complete
+  return new Promise((resolve, reject) => {
+    fileStream
+      .pipe(writeStream)
+      .on('error', (error) => reject(error))
```
sshanzel, line 73:
```suggestion
      .on('error', reject)
```

src/common/googleCloud.ts
```diff
@@ -33,6 +38,110 @@ export const downloadJsonFile = async <T>({
...
+  try {
+    const storage = new Storage();
+    const bucket = storage.bucket(bucketName);
+    const file = bucket.file(fileName);
+
+    // Check if the file exists before deleting
+    const [exists] = await file.exists();
+    if (!exists) {
+      return false;
+    }
```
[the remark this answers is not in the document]
sshanzel: Makes sense. Let's do this recommendation.

sshanzel (approved):
Just the comment, but looks good to me. Also, wait for Ante's final approval.

src/schema/users.ts
```diff
@@ -2364,6 +2377,43 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      // Buffer the stream
+      const chunks: Buffer[] = [];
+      const stream = upload.createReadStream();
```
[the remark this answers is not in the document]
capJavert: +1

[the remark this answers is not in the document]
capJavert: look at `uploadFile` for references

src/common/utils.ts
```diff
@@ -287,3 +287,32 @@ type CamelCasedKey<S extends string> = S extends `${infer Head}_${infer Tail}`
...
+      case '25504446':
+        return 'application/pdf';
+      case 'FFD8FFDB':
+      case 'FFD8FFE0':
+        return 'image/jpeg';
+      case '504B0304':
+        return 'application/zip';
+      default:
+        return null;
+    }
```
capJavert, line 315: Any reference for this, was expecting we are gonna use a package?

src/schema/users.ts
```diff
@@ -2364,6 +2377,43 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      // Validate MIME type using buffer
+      const detectedFileType = getMimeTypeFromArrayBuffer(buffer);
+      if (detectedFileType !== 'application/pdf') {
+        throw new ValidationError('File is not a PDF');
+      }
+
+      // Actual upload using buffer as a stream
+      const filename = `${ctx.userId}.pdf`;
+      await uploadResumeFromStream(filename, Readable.from(buffer));
```
capJavert, line 2413: is it a little too much of juggling between different stream/buffer/readable? Seems like we should not need to juggle around types that much.

src/schema/users.ts
```diff
@@ -2364,6 +2377,43 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         .update({ id: ctx.userId }, { cover: imageUrl });
       return getCurrentUser(ctx, info);
     },
+    uploadResume: async (
```
capJavert, line 2380: Where do we check user already uploaded a file, is he supposed to upload multiple times?

src/schema/users.ts
```diff
@@ -1111,6 +1114,16 @@ export const typeDefs = /* GraphQL */ `
...
 
+    """
+    Upload user resume
+    """
+    uploadResume(
+      """
+      Asset to upload
+      """
+      resume: Upload!
+    ): EmptyResponse @auth @rateLimit(limit: 5, duration: 60)
```
capJavert, line 1125: 5 in a minute sounds a bit much?

capJavert (commented):
Agree with other comments, left some below as well. Let's fix up tests and iron it out and look again after 👌

src/common/googleCloud.ts
capJavert: If we don't use `GCS_PDF_BUCKET` env lets remove it

src/common/utils.ts
capJavert: tried this one? https://www.npmjs.com/package/file-type

ilasw: Yes, it was the last one that made the test failing

capJavert: Can you say what happened exactly, I can help with fixing, I have no idea what the above switch case does and would rather use this package. If tests fail we can always mock the whole package as we did with others.

src/schema/users.ts
ilasw: We set file upload size limit in `/src/index.ts`

```
app.register(MercuriusGQLUpload, {
    maxFileSize: 1024 * 1024 * 20,
    maxFiles: 1,
  });
```

src/common/utils.ts
capJavert: our jest does not support ES modules, since file-type is ESM only

Can you confirm in the app it works for this feature and only tests break, we can then fix up tests by mocking it.

src/schema/users.ts
capJavert: Awesome 🚀

PR #2933, opened 2025-07-18
feat: user follow notification

capJavert:
Follow notification for user. Ignore notification if user is vordr or blocked by followed user.

### Jira ticket
AS-1170


src/workers/cdc/primary.ts
```diff
@@ -1249,6 +1284,9 @@ const worker: Worker = {
         case getTableName(con, UserTransaction):
           await onUserTransactionChange(con, logger, data);
           break;
+        case getTableName(con, ContentPreference):
```
rebelchris, line 1287: Maybe just keep eye on this cdc load

rebelchris (approved):
Looks good

src/workers/cdc/primary.ts
capJavert: yes will do

PR #2936, opened 2025-07-21
feat: scheduling notifications

capJavert:
Ability to schedule email and slack notifications sends as we do with digest. In app would still be published instant since we don't have a concept of scheduling those. We could introduce `createdAt` in the future for UserNotification but not sure its needed for now?


rebelchris (approved):
Looks good to me

PR #2940, opened 2025-07-23
fix: missing worker

sshanzel:


capJavert (commented):
was "should be registeres" test missing?

sshanzel:
> was "should be registered" test missing?

It is the other way around. When a worker is listed but not found. I am writing another test to catch this now.

PR #2942, opened 2025-07-23
fix: slack message structure to two columns and include budget property

sshanzel:
- Made the Slack message structure to be two-columned.
- Added budget property on the query.
- Ensures spend and budget to be integer.

### DIFF
--- a/src/common/post/boost.ts
+++ b/src/common/post/boost.ts
@@ -185,8 +185,8 @@ export const getFormattedCampaign = ({
   ...campaign
 }: GetCampaignResponse): GQLPromotedPost => ({
   ...campaign,
-  spend: Math.floor(usdToCores(parseFloat(spend))),
-  budget: Math.floor(usdToCores(parseFloat(budget))),
+  spend: usdToCores(parseFloat(spend)),
+  budget: usdToCores(parseFloat(budget)),
   startedAt: debeziumTimeToDate(startedAt),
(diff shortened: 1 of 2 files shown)


src/common/post/boost.ts
```diff
@@ -185,8 +185,8 @@ export const getFormattedCampaign = ({
   ...campaign
 }: GetCampaignResponse): GQLPromotedPost => ({
   ...campaign,
-  spend: usdToCores(parseFloat(spend)),
-  budget: usdToCores(parseFloat(budget)),
+  spend: Math.floor(usdToCores(parseFloat(spend))),
+  budget: Math.floor(usdToCores(parseFloat(budget))),
```
capJavert, line 189: Why do you need `Math.floor` when Cores is already integer?

src/common/slack.ts
```diff
@@ -54,46 +59,41 @@ export const notifyNewPostBoostedSlack = async ({
...
-              debeziumTimeToDate(campaign.startedAt),
-            ).toString(),
-          ),
-        },
+        fields: [
+          {
+            type: 'mrkdwn',
+            text: concatTextToNewline(
+              '*Budget:*',
+              `${Math.floor(usdToCores(parseFloat(campaign.budget)))} :cores:`,
```
capJavert, line 86: also here

capJavert (approved):

PR #2944, opened 2025-07-23
feat: add user flags, profile experiences and autocomplete for profile interim

ilasw:
Added user profile user experiences, skills, flags, job preferences, and a new autocomplete endopoint.

### Jira ticket
MI-953


src/common/utils.ts
```diff
@@ -298,3 +298,11 @@ export const getBufferFromStream = async (
     stream.on('error', (err) => reject(err));
   });
 };
+
+export const slugify = (text: string): string => {
+  return text
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, '-') // Replace non-alphanumeric characters with hyphens
+    .replace(/^-|-$/g, '') // Remove leading and trailing hyphens
+    .substring(0, 100); // Limit to 100 characters
```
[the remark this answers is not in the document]
capJavert: we have a slug logic we can use through generated columns in psql, see `post.slug` column

src/entity/Company.ts
```diff
@@ -25,4 +30,11 @@ export class Company {
 
   @Column({ type: 'text', array: true, default: [] })
   domains: string[];
+
+  @Column({
+    type: 'enum',
+    enum: CompanyType,
+    default: CompanyType.Business,
+  })
```
capJavert, line 38: Type should be string on psql side, we only use enum to validate in code and gql. This allows us to adjust the type more freely without changing column type.

capJavert: See other examples eg. `user_transacton.status`

src/entity/user/User.ts
```diff
@@ -281,4 +285,12 @@ export class User {
...
   @Column({ type: 'boolean', default: true })
   awardNotifications: boolean;
+
+  @OneToOne(() => UserJobPreferences, (preferences) => preferences.user, {
+    onDelete: 'CASCADE',
+  })
+  jobPreferences: Promise<UserJobPreferences>;
+
+  @OneToMany(() => UserExperience, (experience) => experience.user)
+  experiences: Promise<UserExperience[]>;
```
capJavert, line 295: `onDelete` cascade?

src/entity/user/UserJobPreferences.ts
```diff
@@ -0,0 +1,43 @@
...
+  userId: string;
+
+  @OneToOne(() => User, (user) => user.jobPreferences)
+  @JoinColumn({ name: 'userId' })
+  user: Promise<User>;
+
+  @Column({ default: false })
+  openToOpportunities: boolean;
+
+  @Column({ type: 'jsonb', default: () => [] })
```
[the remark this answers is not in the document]
capJavert: if you want array I think this:
```suggestion
  @Column({ type: 'text', array: true, default: [] })
```

src/entity/user/UserJobPreferences.ts
```diff
@@ -0,0 +1,43 @@
...
+  })
+  preferredLocationType: WorkLocationType;
+
+  @Column({ default: false })
+  openToRelocation: boolean;
+
+  // !! Never send this field to FE, is only stored for better recommendation !!
+  // Currency must be "ISO-4217" compliant
+  // Amount is yearly based
+  @Column({ type: 'jsonb', nullable: true })
```
capJavert, line 41: we can make it `{}` as other jsonb fields I think

src/entity/user/UserJobPreferences.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  @Column({
+    nullable: true,
+  })
+  preferredLocationType: WorkLocationType;
+
+  @Column({ default: false })
+  openToRelocation: boolean;
+
+  // !! Never send this field to FE, is only stored for better recommendation !!
```
capJavert, line 38: this is controlled on gql type, so make sure its not included there so clients can't request it

src/entity/user/UserSkill.ts
```diff
@@ -0,0 +1,31 @@
...
+  @Column({ type: 'text', nullable: true })
+  description: string;
+
+  @ManyToMany(() => UserExperience, (experience) => experience.skills)
+  experiences: Promise<UserExperience[]>;
+
+  @BeforeInsert()
+  @BeforeUpdate()
+  generateSlug() {
+    this.slug = slugify(this.name);
```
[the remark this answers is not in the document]
capJavert: see my comment about generated column, much better then before hooks of typeorm with only work per app instance

src/entity/user/experiences/UserExperience.ts
```diff
@@ -0,0 +1,64 @@
...
+    nullable: false,
+  })
+  type: UserExperienceType;
+
+  @Column({
+    type: 'enum',
+    enum: ExperienceStatus,
+    default: ExperienceStatus.Draft,
+  })
+  status: ExperienceStatus;
```
capJavert, line 52: Same comment as above, type should be string

src/entity/user/experiences/UserExperience.ts
```diff
@@ -0,0 +1,64 @@
...
+@Entity()
+@TableInheritance({ column: { type: 'text', name: 'type' } })
+export class UserExperience {
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @Column()
+  userId: string;
+
+  @ManyToOne(() => User, (user) => user.experiences)
```
capJavert, line 24: for relentionships use `import type { User }` and string for entity:
```suggestion
  @ManyToOne('User', (user) => user.experiences)
```

it helps with dependency cycles, please update other places for new entities

src/entity/user/experiences/UserProjectExperience.ts
```diff
@@ -0,0 +1,24 @@
...
+  // user ids
+  @Column({ type: 'jsonb', default: () => [] })
+  contributors: string[];
+
+  // not adding a relationship, maybe is not important for now
+  @Column({ type: 'jsonb', nullable: true })
+  associatedWith: {
+    type: ExperienceAssociationType;
+    id: string;
+  };
```
capJavert, line 23: FK can't be added on jsonb columns I think. Ideally we implement it through referenceId and specific columns workId, and educationId. Is it used anywhere in code, maybe we don't need to define it at all and see later?

src/schema/profileAutocomplete.ts
```diff
@@ -0,0 +1,228 @@
...
+        query,
+        limit = DEFAULT_LIMIT,
+      }: { type: AutocompleteType; query: string; limit?: number },
+      ctx: Context,
+    ) => {
+      if (!type || !(type in autocompleteQueryMap)) {
+        throw new ValidationError(`Invalid autocomplete type: ${type}`);
+      }
+
+      if (!query || query.length < 2 || query.length > 100) {
```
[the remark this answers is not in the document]
capJavert: can you do validations over zod?

src/entity/user/experiences/UserProjectExperience.ts
ilasw: This is a recurring association field for projects/publications/[...]. 
IMO we should define some kind of relationship, now or during [MI-958](https://dailydotdev.atlassian.net/jira/software/projects/MI/boards/31?selectedIssue=MI-958), because that task will require it.

Maybe we can leave as is and check for it during deletion of experiences using CDC, wdyt?

src/entity/user/experiences/UserAwardExperience.ts
```diff
@@ -0,0 +1,30 @@
+import { ChildEntity, Column, JoinColumn, OneToOne } from 'typeorm';
+import { UserExperienceType } from './types';
+import { UserExperience } from './UserExperience';
```
capJavert, line 3: import `type`

src/entity/user/UserSkill.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  @Column({ type: 'text', nullable: true })
+  description: string;
+
+  @ManyToMany(
+    'UserExperience',
+    (experience: UserExperience) => experience.skills,
+  )
+  experiences: Promise<UserExperience[]>;
+}
```
capJavert, line 28: Entities look good, please just verify all have cascades so that everything is cleaned up when user, skill and so on is deleted

src/graphql.ts
```diff
@@ -74,6 +75,7 @@ export const schema = urlDirective.transformer(
               paddle.typeDefs,
               njord.typeDefs,
               organizations.typeDefs,
+              profileAutocomplete.typeDefs,
```
capJavert, line 78: What queries you image having in this new schema, kinda feels to specific to be generic, maybe call it userExperience schema so it can have other stuff related to user experience profile and so on?

capJavert: No strong preference just observation.

src/schema/profileAutocomplete.ts
```diff
@@ -0,0 +1,248 @@
...
+      const typeNameByEntity = {
+        [UserWorkExperience.name]: 'ExperienceHit',
+        [Company.name]: 'CompanyHit',
+        [UserSkill.name]: 'SkillHit',
+        [UserCertificationExperience.name]: 'ExperienceHit',
+        [UserAwardExperience.name]: 'ExperienceHit',
+        [UserPublicationExperience.name]: 'ExperienceHit',
+        [UserCourseExperience.name]: 'ExperienceHit',
+        [UserEducationExperience.name]: 'ExperienceHit',
+      } as const;
```
capJavert, line 217: Why exactly do you need this remapping?

src/schema/profileAutocomplete.ts
```diff
@@ -0,0 +1,248 @@
...
+
+      const hits = await queryReadReplica(ctx.con, ({ queryRunner }) =>
+        queryRunner.manager
+          .getRepository(entity)
+          .createQueryBuilder('entity')
+          .select(select ?? ['id', propertyName])
```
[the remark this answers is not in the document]
capJavert: +1

src/schema/profileAutocomplete.ts
```diff
@@ -0,0 +1,248 @@
...
+          .where({
+            [propertyName]: Raw(
+              (alias) => `LOWER(${alias}) LIKE LOWER(:query)`,
+              { query: `%${query}%` },
+            ),
+            // extending where condition using the autocompleteQueryMap
+            ...where,
+          })
+          // sort by selected property name
+          .orderBy(`entity.${propertyName}`, 'ASC')
```
capJavert, line 233: do we have order by index?

src/schema/profileAutocomplete.ts
```diff
@@ -0,0 +1,248 @@
...
+      } as const;
+
+      const hits = await queryReadReplica(ctx.con, ({ queryRunner }) =>
+        queryRunner.manager
+          .getRepository(entity)
+          .createQueryBuilder('entity')
+          .select(select ?? ['id', propertyName])
+          .where({
+            [propertyName]: Raw(
+              (alias) => `LOWER(${alias}) LIKE LOWER(:query)`,
```
capJavert, line 226: `ILIKE` so you don't have to use LOWER

src/schema/profileAutocomplete.ts
```diff
@@ -0,0 +1,248 @@
...
+          .orderBy(`entity.${propertyName}`, 'ASC')
+          .limit(limit)
+          .getRawMany(),
+      );
+
+      return {
+        query,
+        limit,
+        hits: hits.map((hit) => ({
+          __typename: typeNameByEntity[entity.name],
```
capJavert, line 242: why do we need `__typename`?

src/schema/profileAutocomplete.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  extend type Query {
+    """
+    Get autocomplete suggestions for various fields like job titles, companies, skills, etc.
+    """
+    profileAutocomplete(
+      type: String!
+      query: String!
+      limit: Int
+    ): AutocompleteResult!
```
capJavert, line 151: does this need to be accessible for anon users, we can add `@auth` if not

src/entity/Company.ts
```diff
@@ -25,4 +30,10 @@ export class Company {
 
   @Column({ type: 'text', array: true, default: [] })
   domains: string[];
+
+  @Column({
```
capJavert, line 34: Overall note you are missing migration file for psql changes

capJavert (commented):
Should be good to go once comments are addressed 👌

src/entity/user/experiences/UserAwardExperience.ts
ilasw: I'm actually extending this `UserExperience`, I cannot import as type

src/graphql.ts
ilasw: Fine to me ✔️

src/entity/Company.ts
ilasw: I'm missing something from [this](https://github.com/dailydotdev/daily-api/pull/2944/files#diff-e06fe0d4808cc58d47217607e972ee2d664d161fb0605b65fc420ed45d60496c)?

src/schema/profileAutocomplete.ts
capJavert: Ok, with this kind of dynamic column building its harder to see what columns are used from code, so that is why I asked. For example you change the mappings above but forget to add index. 

If it was just `orderBy('createdAt', 'ASC')` (just an example) you immediately know what is used.

PR #2953, opened 2025-07-31
feat: add user job preferences support

ilasw:
Added user job preferences support.

- Added GraphQL schema and resolvers for user job preferences
- Implemented profile completeness validation for job opportunities
- Added test coverage for the new functionality

### Jira ticket
MI-953


src/schema/users.ts
```diff
@@ -2088,6 +2195,43 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        // if no job preferences are set, return default values
+        return {
+          userId: ctx.userId,
+          openToOpportunities: false,
+          currentTotalComp: {},
+          preferredRoles: [],
+          preferredLocationType: null,
+          openToRelocation: false,
+        };
+      }
```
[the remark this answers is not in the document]
ilasw: I was foreseeing a case where the user have no preferences and we fill the frontend form with default ones, adding them to API we can have the same logic for filled/default values.
 
If the default one changes overtime we can update only in one place (API). WDYT?

src/schema/users.ts
```diff
@@ -286,6 +293,17 @@ export interface SendReportArgs {
   tags?: string[];
 }
 
+export interface GQLUserJobPreferences {
+  userId: User['id'];
```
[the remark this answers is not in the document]
ilasw: Removed ✔️

src/common/users.ts
```diff
@@ -636,3 +639,65 @@ export const bskySocialUrlMatch =
...
+  openToRelocation: z.boolean().optional().default(false),
+  currentTotalComp: z
+    .object({
+      currency: z.string().length(3),
+      amount: z.number().int().positive(),
+    })
+    .partial()
+    .default({}),
+});
+export const checkJobPreferenceParamsValidity = (params: unknown) => {
```
sshanzel, line 694: Why is the params here `unknown`?

src/schema/users.ts
```diff
@@ -2088,6 +2190,42 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         coresRole: userCoresRole,
       };
     },
+    userJobPreferences: async (
+      _,
+      __,
+      ctx: AuthContext,
+    ): Promise<GQLUserJobPreferences> => {
+      // important: if you need to implement public user job preferences queries, we need to remove the "currentTotalComp" for other users
```
capJavert, line 2198: Usually we would do this in graphorm, with computed field where we check if user requesting is the same as user owning job.

src/schema/users.ts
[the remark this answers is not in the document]
capJavert: I am just worried this returning accidentally without us realizing

src/common/users.ts
```diff
@@ -636,3 +639,65 @@ export const bskySocialUrlMatch =
...
 
 export const portfolioLimit = 500;
+
+const MIN_WORK_EXPERIENCE = 1; // Minimum number of work experiences required for profile completion
+export const isProfileCompleteById = async (
+  con: DataSource,
+  userId: User['id'],
+) => {
+  try {
+    const [user, experiencesCount] = await queryReadReplica(
```
rebelchris, line 649: From when does the FE call this, because it's RR it might not be instant available.

src/common/users.ts
```diff
@@ -636,3 +639,65 @@ export const bskySocialUrlMatch =
...
+            .getRepository('UserExperience')
+            .count({ where: { userId, type: UserExperienceType.Work } }),
+        ]),
+    );
+
+    const hasCountry =
+      typeof user.flags.country === 'string' && user.flags.country.length >= 2;
+    const hasEnoughExperiences = experiencesCount >= MIN_WORK_EXPERIENCE;
+
+    return hasCountry && hasEnoughExperiences;
```
rebelchris, line 666: We don't care here about which one is not completed?

src/schema/users.ts
```diff
@@ -2088,6 +2195,43 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const userJobPreferences = await queryReadReplica(
+        ctx.con,
+        ({ queryRunner }) =>
+          queryRunner.manager.getRepository(UserJobPreferences).findOne({
```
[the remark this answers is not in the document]
rebelchris: Also the default below can be done with graphORM actually.

rebelchris (approved):
No blockers, but good feedback from Ante to consider.

src/common/users.ts
ilasw: Not really, for now both are required so we do't need to go further on checks

src/common/users.ts
ilasw: Because this will come from the mutation, I can better type it but we never use any field before the validation (after that all the right types are infered)

PR #2954, opened 2025-07-31
feat: notificationSettings

AmarTrebinjac:
Implements the new notification flags system detailed in this [RFC](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1693417555/RFC-128+-+Notification+updates+V2) 

~~Still need to create a script to migrate the users current preferences to this new column. Wondering if I should maybe split off the migration file adding the column + the script into its own PR and then merge that first. Would love to hear your input on that @capJavert~~ (completed)

### Jira ticket
MI-949

### DIFF
--- a/bin/migrateNotificationFlags.ts
+++ b/bin/migrateNotificationFlags.ts
@@ -1,3 +1,4 @@
+import fastq from 'fastq';
 import '../src/config';
 import createOrGetConnection from '../src/db';
 import {
@@ -7,6 +8,8 @@ import {
 } from '../src/notifications/common';
 import { User } from '../src/entity/user/User';
 import type { UserNotificationFlags } from '../src/entity/user/User';
+
+const QUEUE_CONCURRENCY = 1;
 
 interface UserData {
   id: string;
@@ -73,50 +76,53 @@ function buildNotificationFlags(user: UserData): UserNotificationFlags {
   const con = await createOrGetConnection();
 
   try {
-    const userRepo = con.getRepository(User);
-
-    const users = await userRepo
-      .createQueryBuilder('user')
-      .select([
-        'user.id',
-        'user.notificationEmail',
-        'user.followingEmail',
-        'user.followNotifications',
-        'user.awardEmail',
-        'user.awardNotifications',
-        'user.notificationFlags',
-      ])
-      .orderBy('user.id')
-      .limit(limit)
-      .offset(offset)
-      .getMany();
-
     console.log(
-      `Processing ${users.length} users (offset ${offset} to ${offset + users.length - 1})...`,
+      `Processing users starting from offset ${offset} (limit ${limit})...`,
     );
 
-    await con.transaction(async (manager) => {
-      for (const user of users) {
-        const userData: UserData = {
-          id: user.id,
-          notificationEmail: user.notificationEmail,
-          followingEmail: user.followingEmail,
-          followNotifications: user.followNotifications,
-          awardEmail: user.awardEmail,
-          awardNotifications: user.awardNotifications,
-          notificationFlags: user.notificationFlags,
-        };
+    let processedCount = 0;
 
-        const newFlags = buildNotificationFlags(userData);
+    await con.transaction(async (manager) => {
+      const userRepo = manager.getRepository(User);
+
+      const builder = userRepo
+        .createQueryBuilder('user')
+        .select('user.id', 'id')
+        .addSelect('user.notificationEmail', 'notificationEmail')
+        .addSelect('user.followingEmail', 'followingEmail')
+        .addSelect('user.followNotifications', 'followNotifications')
+        .addSelect('user.awardEmail', 'awardEmail')
+        .addSelect('user.awardNotifications', 'awardNotifications')
+        .addSelect('user.notificationFlags', 'notificationFlags')
+        .orderBy('user.id')
+        .limit(limit)
+        .offset(offset);
+
+      const stream = await builder.stream();
+
+      const insertQueue = fastq.promise(async (user: UserData) => {
+        const newFlags = buildNotificationFlags(user);
 
         await manager
           .getRepository(User)
           .update({ id: user.id }, { notificationFlags: newFlags });
-      }
+
+        processedCount++;
+      }, QUEUE_CONCURRENCY);
+
+      stream.on('data', (user: UserData) => {
+        insertQueue.push(user);
+      });
+
+      await new Promise((resolve, reject) => {
+        stream.on('error', reject);
+        stream.on('end', () => resolve(true));
+      });
+      await insertQueue.drained();
     });
 
     console.log(
-      `Migration completed successfully. Updated ${users.length} users (offset ${offset} to ${offset + users.length - 1}).`,
+      `Migration completed successfully. Updated ${processedCount} users (offset ${offset} to ${offset + processedCount - 1}).`,
     );
   } catch (error) {
     console.error('Migration failed:', error);


.infra/application.properties
```diff
@@ -9,7 +9,7 @@ debezium.source.database.password=%database_pass%
 debezium.source.database.dbname=%database_dbname%
 debezium.source.database.server.name=api
 debezium.source.table.include.list=public.comment,public.user_comment,public.comment_mention,public.source_request,public.post,public.user,public.post_report,public.source_feed,public.settings,public.reputation_event,public.submission,public.user_state,public.notification_v2,public.source_member,public.feature,public.source,public.post_mention,public.content_image,public.comment_report,public.user_post,public.banner,public.post_relation,public.marketing_cta,public.squad_public_request,public.user_streak,public.bookmark,public.user_company,public.source_report,public.user_top_reader,public.source_post_moderation,public.user_report,public.user_transaction,public.content_preference
-debezium.source.column.exclude.list=public.post.tsv,public.post.placeholder,public.source.flags,public.user_top_reader.image
+debezium.source.column.exclude.list=public.post.tsv,public.post.placeholder,public.source.flags,public.user_top_reader.image,public.user.notification_flags
```
sshanzel, line 12: iirc this does not actually work.

src/entity/user/User.ts
```diff
@@ -55,6 +56,16 @@ export type UserSubscriptionFlags = Partial<{
   PaddleUserSubscriptionFlags &
   StoreKitUserSubscriptionFlags;
 
+export type UserNotificationFlags = Partial<
+  Record<
+    string,
```
sshanzel, line 61: Should we not use the `NotificationType` here?

src/schema/users.ts
```diff
@@ -2269,14 +2336,19 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-        resubscribeUser(cio, ctx.userId),
-        identifyUserPersonalizedDigest({
-          userId: ctx.userId,
-          cio,
-          subscribed: true,
-        }),
-      ]);
+      const promises = [resubscribeUser(cio, ctx.userId)];
+
+      if (process.env.NODE_ENV !== 'development') {
```
sshanzel, line 2341: I think these checks should be handled within the function itself, not on every instance where we call it.

src/schema/users.ts
```diff
@@ -2815,6 +2888,32 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+    updateNotificationSettings: async (
+      _,
+      { notificationFlags }: { notificationFlags: UserNotificationFlags },
+      ctx: AuthContext,
+    ): Promise<GQLEmptyResponse> => {
+      try {
+        notificationFlagsSchema.parse(notificationFlags);
+      } catch (error) {
+        if (error instanceof z.ZodError) {
+          console.log('**** zod errors', error.errors);
```
sshanzel, line 2900: I think we can now remove the console log.

sshanzel (commented):
Looking great so far. Just some comments, nothing major.

src/schema/users.ts
AmarTrebinjac: I thought I already did 🙈

src/schema/users.ts
```diff
@@ -1017,6 +1063,11 @@ export const typeDefs = /* GraphQL */ `
     Check and apply Cores role
     """
     checkCoresRole: CheckCoresRole! @auth
+
+    """
+    Get current user's notification preferences
+    """
+    notificationSettings: JSON @auth
```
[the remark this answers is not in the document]
AmarTrebinjac: Maybe we can infer the type from the zod schema. However, the zod schema is strict, so if there's an invalid property in it, it will throw an error anyway

.infra/application.properties
AmarTrebinjac: I think this is what @capJavert suggested we could do, but I might have misunderstood him. We'll see what he says 😄

src/schema/users.ts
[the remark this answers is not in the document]
capJavert: I don't know of a way to infer it since type is not something you can use as value, we will have to define in in gql, its important for validation. There is a zod plugin for gql but did not have time to look into it.

src/notifications/index.ts
```diff
@@ -145,11 +146,29 @@ export async function storeNotificationBundleV2(
...
+  for (const chunk of userChunks) {
+    const users = await entityManager.getRepository(User).find({
+      select: ['id', 'notificationFlags'],
+      where: { id: In(chunk) },
+    });
+
+    users.forEach((user) => {
+      userPreferences.set(user.id, user);
+    });
+  }
```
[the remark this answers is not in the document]
AmarTrebinjac: @sshanzel I had updated this, but forgot to push (that's also why the console log was still there 😂)

src/schema/common.ts
```diff
@@ -37,6 +41,8 @@ export const typeDefs = /* GraphQL */ `
 
   scalar JSONObject
 
+  scalar JSON
```
capJavert, line 44: I don't think we will need this once type is set

__tests__/users.ts
```diff
@@ -6991,3 +7123,53 @@ describe('mutation uploadResume', () => {
...
   });
 });
+describe('mutation updateNotificationSettings', () => {
+  const MUTATION = `mutation UpdateNotificationSettings($notificationFlags: JSON!) {
+    updateNotificationSettings(notificationFlags: $notificationFlags) {
+      _
+    }
+  }`;
+
+  it('should update notification settings', async () => {
```
capJavert, line 7133:
```suggestion
  it('should overwrite notification settings during update', async () => {
```

just so its clear it is supposed to happen as you defined in RFC

src/workers/newNotificationV2Mail.ts
```diff
@@ -1054,46 +1052,16 @@ const worker: Worker = {
...
-                notification.public
-                  ? true
-                  : undefined,
-              awardEmail:
-                !isFollowNotification &&
-                isAwardNotification &&
-                !digestNotification &&
-                notification.public
-                  ? true
-                  : undefined,
```
capJavert, line 1096: Are this settings replaced everywhere else in the code, for example I don't see adjustments to cio or where those columns are set on old user columns.

src/schema/users.ts
```diff
@@ -147,6 +150,48 @@ import {
...
 import { fileTypeFromBuffer } from 'file-type';
 
+const notificationPreferenceSchema = z.object({
+  email: z.enum(['muted', 'subscribed']),
+  inApp: z.enum(['muted', 'subscribed']),
+});
+
+const notificationFlagsSchema = z
+  .object({
+    [NotificationType.ArticleNewComment]: notificationPreferenceSchema,
```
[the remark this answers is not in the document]
capJavert: To avoid repetitive code

src/schema/users.ts
```diff
@@ -147,6 +150,48 @@ import {
...
+    [NotificationType.ArticlePicked]: notificationPreferenceSchema,
+    [NotificationType.UserReceivedAward]: notificationPreferenceSchema,
+    [NotificationType.BriefingReady]: notificationPreferenceSchema,
+    [NotificationType.SquadNewComment]: notificationPreferenceSchema,
+    [NotificationType.ArticleAnalytics]: notificationPreferenceSchema,
+    [NotificationType.SquadMemberJoined]: notificationPreferenceSchema,
+    [NotificationType.SquadReply]: notificationPreferenceSchema,
+    [NotificationType.SquadBlocked]: notificationPreferenceSchema,
+  })
+  .strict();
```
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert Are you sure? Its not giving me a deprecation warning, and there's no strictObject prop as far as I can see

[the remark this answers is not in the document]
capJavert: yes I am, but we are on lower version of zod, I'll update us separate

src/migration/1753092411000-UserNotificationFlags.ts
```diff
@@ -0,0 +1,20 @@
...
+  name = 'UserNotificationFlags1753092411000';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "user" ADD "notificationFlags" jsonb NOT NULL DEFAULT '{}'`,
+    );
+    
+    await queryRunner.query(
+      `CREATE INDEX "IDX_user_notificationFlags_path_ops" ON "user" USING GIN ("notificationFlags" jsonb_path_ops)`,
+    );
```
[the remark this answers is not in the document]
capJavert: Because of `jsonb_path_ops`? It does not work that way.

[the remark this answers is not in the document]
capJavert: Never used jsonb_path_ops specifically, but looking at the docs it only works on part of the operators and its still on the whole value of your column, not nested ones. It would work if you use `@>` but I don't see you using it.

src/graphorm/index.ts
```diff
@@ -190,15 +190,9 @@ const obj = new GraphORM({
       followingEmail: {
         transform: nullIfNotSameUser,
       },
-      followNotifications: {
-        transform: nullIfNotSameUser,
-      },
       awardEmail: {
         transform: nullIfNotSameUser,
       },
-      awardNotifications: {
```
capJavert, line 199: Check please if something in extension uses this fields, we can't break old queries.

src/schema/users.ts
```diff
@@ -1234,6 +1285,11 @@ export const typeDefs = /* GraphQL */ `
     Claim unclaimed user ClaimableItem
     """
     claimUnclaimedItem: UserClaim @auth
+
+    """
+    Update user's notification preferences
+    """
+    updateNotificationSettings(notificationFlags: JSON!): EmptyResponse @auth
```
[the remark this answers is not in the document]
capJavert: fair, let's just address my comments below on zod handling and see if you need JSON because we already have JSONObject in gql.

src/schema/users.ts
```diff
@@ -2269,14 +2267,17 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      promises.push(
         identifyUserPersonalizedDigest({
           userId: ctx.userId,
           cio,
           subscribed: true,
         }),
-      ]);
+      );
+
+      await Promise.all(promises);
```
capJavert, line 2280: This change you don't actually need, can we revert how it was?

src/workers/notifications/utils.ts
```diff
@@ -64,12 +87,21 @@ export const getOptInSubscribedMembers = async ({
...
+  const members = await memberQuery
     .andWhere(`EXISTS(${subscribedquery.getQuery()}) IS TRUE`)
     .getRawMany<SourceMember>();
+
+  const allowedUserIds = await filterUsersWithGlobalPreferences(
+    con,
+    members.map((m) => m.userId),
+    type,
+    'inApp',
+  );
```
capJavert, line 99: Here we introduce another query on user table, I think it should be possible to fetch notificationFlags from sourceMember by joining with user table there.

src/notifications/index.ts
```diff
@@ -145,38 +145,47 @@ export async function storeNotificationBundleV2(
...
+      });
+
+    const [query, params] = selectQuery.getQueryAndParameters();
+
+    await entityManager.query(
+      `INSERT INTO "user_notification" ("userId", "notificationId", "createdAt", "public", "uniqueKey")
+       ${query}
+       ON CONFLICT ("userId", "uniqueKey") WHERE "uniqueKey" IS NOT NULL DO NOTHING`,
+      params,
+    );
```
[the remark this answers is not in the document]
capJavert: ok, let's just make sure there are tests for it

src/workers/newNotificationV2Mail.ts
AmarTrebinjac: @capJavert I've made updates to CIO now. The marketing aspect is unhandled for now and will be added right after this part because it will require changing the frontend in unexpected places like onboarding.

[REF](https://dailydotdev.slack.com/archives/C03RA0WEGEQ/p1754554443064409)

src/schema/users.ts
AmarTrebinjac: Yeah my bad lol, I initially wrapped it in a env check and moved it inside the function, but forgot to und othe promises lol 🙈

src/graphorm/index.ts
capJavert: can you send me the code change from the frontend? 🙏

src/schema/users.ts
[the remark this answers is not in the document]
AmarTrebinjac: @rebelchris Its a record of channel -> subscribed / muted

```
{ 
	"article_upvote_milestone": { 
		"email": "muted",
		"inApp": "subscribed",
		 },
}
```

From what I understand you cannot dynamically type these property names. I think it would work if we typed out the entire notificationFlags object, but that would kind of defeat the purpose of having zod validation I think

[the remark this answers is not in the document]
rebelchris: @claude can you suggest how this could be stronger typed based on the dynamic need?

src/cio.ts
```diff
@@ -180,6 +277,33 @@ export async function identifyUser({
...
+    return;
+  }
+
+  const cioTopics = getNotificationFlagsCioTopics(notificationFlags);
+
+  try {
+    await cio.identify(userId, cioTopics);
+  } catch (err) {
+    if (err instanceof CustomerIORequestError && err.statusCode === 400) {
+      logger.warn(
```
capJavert, line 297:
```suggestion
      logger.error(
```

and let's monitor for this please

src/common/mailing.ts
```diff
@@ -16,20 +16,18 @@ import {
...
+  generateIdentifyObject,
+  getCioTopicsToNotificationFlags,
+  CioUnsubscribeTopic,
+} from '../cio';
 import { setTimeout } from 'node:timers/promises';
 import { toChangeObject, updateFlagsStatement } from './utils';
 import { GetUsersActiveState } from './googleCloud';
 import { logger } from '../logger';
 
-export enum CioUnsubscribeTopic {
```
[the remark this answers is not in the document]
capJavert: I think moved to cio filed

src/schema/users.ts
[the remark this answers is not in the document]
capJavert: For me its fine for now, since we validate with zod, but lets move to JSONObject if we can

src/schema/users.ts
```diff
@@ -2190,6 +2175,20 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      ctx: AuthContext,
+    ): Promise<UserNotificationFlags> => {
+      const user = await ctx.con.getRepository(User).findOne({
+        where: { id: ctx.userId },
+        select: ['notificationFlags'],
+      });
+      return {
+        ...DEFAULT_NOTIFICATION_SETTINGS,
+        ...user?.notificationFlags,
+      };
```
capJavert, line 2190: since notification flags are nested, it would not be deep merged so I don't see any point in having default settings, but non blocking, something to maybe adjust down the line

src/workers/notifications/utils.ts
```diff
@@ -44,54 +44,94 @@ type GetMembersParams = {
...
+  channel: 'email' | 'inApp',
+): Promise<string[]> => {
+  if (!userIds.length) {
+    return [];
+  }
+
+  const users = await con.getRepository(User).find({
+    select: ['id', 'notificationFlags'],
+    where: { id: In(userIds) },
+  });
```
capJavert, line 60: I think I commented before that this we could get my join over source member, no need for another query on user table?

capJavert (approved):
image

PR #2956, opened 2025-08-04
feat: add hasUnreadPosts flag and clearUnreadPosts mutation

omBratteng:
- Adds hasUnreadPosts to SourceMember flags and expose it in boot
- Create new mutation clearUnreadPosts that sets new flag to false

TODO:
- [x] Add worker to set flag to true on new posts
- [x] Add tests for worker

### Jira ticket
AS-1172


src/entity/SourceMember.ts
```diff
@@ -6,6 +6,7 @@ import { SourceMemberRoles } from '../roles';
 export type SourceMemberFlags = Partial<{
   hideFeedPosts: boolean;
   collapsePinnedPosts: boolean;
+  hasUnreadPosts: boolean;
```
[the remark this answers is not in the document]
capJavert: I mean ideally we should store it in ContentPreferenceSource 🙈 since SourceMember is partially deprecated on other places.

As for flags, if we don't use them/listen for them in CDC sounds good to exclude.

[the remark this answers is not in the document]
omBratteng: Ideally, yeah 😬 But boot still queries from SourceMember, and don't really want to add extra read/write to ContentPreferenceSource just to keep in sync.

Currently we don't listen to flags from SourceMember, so I'll add exclude them from CDC.

src/workers/postAddedSquadUnreadPosts.ts
```diff
@@ -0,0 +1,69 @@
...
+              {
+                flags: updateFlagsStatement({
+                  hasUnreadPosts: true,
+                }),
+              },
+            );
+          },
+          CONCURRENCY,
+          BATCH_SIZE,
+        );
```
omBratteng, line 60: Went with batching for when it's one of those really big squads with several tens of thousands of members, so to not do a giant update

src/workers/postAddedSquadUnreadPosts.ts
```diff
@@ -0,0 +1,69 @@
...
+        const squadMembersStream = await con
+          .createQueryBuilder()
+          .select('sm."userId"')
+          .from(SourceMember, 'sm')
+          .where('sm.sourceId = :sourceId', { sourceId: source.id })
+          .andWhere('sm.flags @> :flags', {
+            flags: {
+              hasUnreadPosts: false,
+            },
+          })
```
omBratteng, line 40: Only select the source members that doesn't have unread posts

src/workers/postAddedSquadUnreadPosts.ts
```diff
@@ -0,0 +1,69 @@
...
+            );
+          },
+          CONCURRENCY,
+          BATCH_SIZE,
+        );
+      } catch (_err) {
+        const err = _err as Error;
+        logger.error(
+          { err, sourceId: data.post.sourceId, message },
+          'Error updating flags',
```
capJavert, line 65:
```suggestion
          'Error updating  source hasUnreadPosts flag',
```

just so its easier to find what broke

capJavert (commented):
Looks good, just few comments to check after you add tests 👍

src/workers/postAddedSquadUnreadPosts.ts
omBratteng: Yeah, can do it like that too

omBratteng: Refactored to use that, and added index

rebelchris (approved):
Would prefer optimized mutation, but it's not a blocker.

src/schema/sources.ts
```diff
@@ -2623,6 +2637,25 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      await ctx.con.getRepository(SourceMember).update(
+        { sourceId: source.id, userId: ctx.userId },
+        {
+          flags: updateFlagsStatement<SourceMember>({
+            hasUnreadPosts: false,
+          }),
+        },
+      );
+
+      return getSourceById(ctx, info, sourceId);
```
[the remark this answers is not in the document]
omBratteng: I've refactorer it now to be a smaller mutation, it just returns true or false depending on affected rows in the update.
Then on front-end, if response is true, I update object

PR #2958, opened 2025-08-05
feat: notificationFlags migration

AmarTrebinjac:
### Jira ticket

Migration file and script necessary to migrate notification columns to the new notificationFlags

MI-964


src/notifications/common.ts
```diff
@@ -96,6 +97,137 @@ export const notificationPreferenceMap: Partial<
...
+  },
+  [NotificationType.ArticleReportApproved]: {
+    email: NotificationPreferenceStatus.Subscribed,
+    inApp: NotificationPreferenceStatus.Subscribed,
+  },
+  [NotificationType.StreakResetRestore]: {
+    email: NotificationPreferenceStatus.Subscribed,
+    inApp: NotificationPreferenceStatus.Subscribed,
+  },
+  ['streak_reminder']: {
```
omBratteng, line 141: No enum for this?

AmarTrebinjac: @omBratteng There is in `UserPersonalizedDigestType`, but causes a circular dependency error. Can move that enum out to the common file as well if that's preferable

src/migration/1754302649060-UserNotificationFlags.ts
```diff
@@ -8,13 +8,13 @@ export class UserNotificationFlags1754302649060 implements MigrationInterface {
       `ALTER TABLE "user" ADD "notificationFlags" jsonb NOT NULL DEFAULT '{}'`,
     );
     
-    await queryRunner.query(
-      `CREATE INDEX "IDX_user_notificationFlags_ops" ON "user" USING GIN ("notificationFlags" jsonb_ops)`,
+   await queryRunner.query(
+      `CREATE INDEX "IDX_user_notificationFlags_path_ops" ON "user" USING GIN ("notificationFlags" jsonb_path_ops)`,
```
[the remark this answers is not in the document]
capJavert: I don't think he knows what he is talking about :) 

Skip the index for now, once you deploy this and do database migration you will be able to test the queries on production and check if index is used in query plan. If yes then it is ok solution, if not you will need to find one that works.

bin/migrateNotificationFlags.ts
```diff
@@ -0,0 +1,131 @@
...
+        };
+
+        const newFlags = buildNotificationFlags(userData);
+
+        await manager
+          .getRepository(User)
+          .update({ id: user.id }, { notificationFlags: newFlags });
+      }
+
+      await manager.query(`SET session_replication_role = DEFAULT;`);
```
capJavert, line 119: For what do you use this?

bin/migrateNotificationFlags.ts
```diff
@@ -0,0 +1,131 @@
...
+          awardEmail: user.awardEmail,
+          awardNotifications: user.awardNotifications,
+          notificationFlags: user.notificationFlags,
+        };
+
+        const newFlags = buildNotificationFlags(userData);
+
+        await manager
+          .getRepository(User)
+          .update({ id: user.id }, { notificationFlags: newFlags });
```
capJavert, line 116: Even though its in transaction better to batch it and do less update queries overall. It will be faster, for example take 1000 users, do update query, loop.

Still in single transaction.

capJavert (approved):

rebelchris (approved):
Ideally we fix the circular deps on type import, but non blocking.

PR #2960, opened 2025-08-05
fix: re-create index so that it actually works

omBratteng:
Previous index did not actually do anything, so new migration to make it actually work.

### Jira ticket
AS-1172


omBratteng:
@claude compare the new index I created, to the query I intend to run. "sourceId" is dynamic based on the query.

```sql
SELECT
	*
FROM
	source_member
WHERE
	"sourceId" = '4c4a674b-c7d4-45ab-9333-1c79ef6b1d99'
	AND COALESCE((flags ->> 'hasUnreadPosts')::boolean, FALSE) != TRUE;
```

PR #2961, opened 2025-08-05
feat: claude github workflow

omBratteng:


omBratteng:
> Looks ok to me, but not sure we even want it on all PR's for cost side of thingS?
> CC: @idoshamun 

You need to invoke it manually for it to trigger

PR #2975, opened 2025-08-11
fix: syncSubscriptions should no longer malform notificationFlags on user

AmarTrebinjac:
Contains the reverted changes + fixes for the cause of the revert.

[fix commit](https://github.com/dailydotdev/daily-api/commit/caa1c59cd927207b94aefe5ebf4dd7aa3ca14cef)

* updated syncSubscriptions to preserve existing flags
* If we add new flags, syncSubscription should still pass because we spread them
* added logging in case updating a users flags fails
* while not strictly necessary, added the default flags on user creation
* added some tests to ensure desired behavior when syncing


rebelchris (approved):
I don't see any issues :)

src/common/mailing.ts
```diff
@@ -138,15 +138,49 @@ export const syncSubscription = async function (
...
+        const mergedNotificationFlags = getCioTopicsToNotificationFlags(
+          subs,
+          existingFlags,
+        );
+
+        const validation = notificationFlagsSchema.safeParse(
+          mergedNotificationFlags,
         );
+        if (!validation.success) {
+          logger.warn(
```
capJavert, line 160: this should be logger.error no, and why don't we throw, since if schema is not valid we don't want to save to db?

capJavert (changes requested):
Only comment is if we should throw when schema is not ok.

src/common/mailing.ts
AmarTrebinjac: @capJavert The only other relevant field here that is not gone (not sure why I didn't clean it up, slipped my mind), is acceptedMarketing, but that will also disappear after the p2 updates. Thought we could let that still be updated if it fails, but I can make the change so that it just throws, since that will be the defacto soon either way.

capJavert (approved):

PR #2976, opened 2025-08-12
refactor: ignore rate limit for plus users and optimize querying

idoshamun:
Follow up on this:  https://github.com/dailydotdev/daily/issues/1236, I decided to ignore rate limit for Plus users. I feel confident that they're not here for spam

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -115,11 +115,7 @@ import { identifyUserPersonalizedDigest } from '../src/cio';
 import type { GQLUser } from '../src/schema/users';
 import { cancelSubscription } from '../src/common/paddle';
 import { isPlusMember, SubscriptionCycles } from '../src/paddle';
-import {
-  acceptedResumeExtensions,
-  CoresRole,
-  StreakRestoreCoresPrice,
-} from '../src/types';
+import { CoresRole, StreakRestoreCoresPrice } from '../src/types';
 import {
   UserTransaction,
   UserTransactionProcessor,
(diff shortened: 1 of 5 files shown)


src/entity/posts/utils.ts
```diff
@@ -414,14 +414,16 @@ export const createSharePost = async ({
...
-      { con, userId, req: ctx?.req },
-    );
+    const vordrStatus = ctx
+      ? await checkWithVordr(
+          {
+            id: createdPost.id,
+            content: createdPost.title || undefined,
+            type: VordrFilterType.Post,
+          },
+          ctx,
```
capJavert, line 424: I don't think this is right, `userId` before did not come from `ctx` but from `authorId` param, not sure if it can be different user but technically it could

src/common/vordr.ts
```diff
@@ -34,19 +34,20 @@ type CheckWithVordrInput = {
...
-  req?: Pick<FastifyRequest, 'ip'>;
-};
-
 export const checkWithVordr = async (
   { id, type, content }: CheckWithVordrInput,
-  { userId, con, req }: CheckWithVordrContext,
+  ctx: Context,
 ): Promise<boolean> => {
-  if (req && validateVordrIPs(req.ip)) {
+  const { userId } = ctx;
```
capJavert, line 41: also I don't think it always comes from ctx, sometimes its used in different scope

src/entity/posts/utils.ts
idoshamun: It can't be. I'll check if it's easy to fix. No one can write a post on behalf of someone else, and also if they did, you still want to rate limit the logged in user

PR #2977, opened 2025-08-12
fix: remove file suffix from user CV blob

omBratteng:
To make it easier to know where to read the blob from, we remove the suffix from the uplaoded blob.

Will do a migration on the GCS bucket to update the existing blobs after PR is merged.

### Jira ticket
AS-1177


ilasw (approved):
LGTM ✅

PR #2982, opened 2025-08-13
feat: post analytics history

capJavert:


src/entity/posts/PostAnalyticsHistory.ts
```diff
@@ -0,0 +1,33 @@
...
+  OneToOne,
+  PrimaryColumn,
+  UpdateDateColumn,
+} from 'typeorm';
+import type { Post } from './Post';
+
+@Entity()
+export class PostAnalyticsHistory {
+  @PrimaryColumn({ type: 'text' })
+  id: string;
```
capJavert, line 14: opted to do another table, to conserve space and also not create dependency between between two tables if we need to back import data

src/entity/posts/PostAnalyticsHistory.ts
```diff
@@ -0,0 +1,33 @@
...
+  createdAt: Date;
+
+  @UpdateDateColumn()
+  updatedAt: Date;
+
+  @Column()
+  date: string;
+
+  @Column({ default: 0 })
+  impressions: number;
```
capJavert, line 26: Also storing only impressions for now, adding other columns will be possible since we store raw events and can always compute historical data if we want to

rebelchris (approved):

src/cron/postAnalyticsHistoryDayClickhouse.ts
```diff
@@ -0,0 +1,112 @@
...
+      postAnalyticsHistoryDayClickhouseCron.name,
+      'config',
+    );
+
+    const cronConfig: Partial<PostAnalyticsHistoryClickhouseCronConfig> =
+      await getRedisHash(redisStorageKey);
+
+    const lastRunAt = cronConfig.lastRunAt
+      ? new Date(cronConfig.lastRunAt)
+      : startOfToday(); // for now use start of today if no last run time is set
```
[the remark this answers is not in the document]
capJavert: yes

PR #2987, opened 2025-08-17
feat: add marketing and billing related notification flags

AmarTrebinjac:
* Adds the remaining `billing` and `daily.dev` sections to the email tab in marketing.
* Now set "Muted" on `NotificationType.Marketing` flag if user does not accept marketing on user creation
* Handle not sending email to user in `userboughtCores` if they have toggled off `NotificationType.InAppPurchases` email prop.

### Jira ticket
MI-971


src/notifications/generate.ts
```diff
@@ -535,4 +539,8 @@ export const generateNotificationMap: Record<
       .avatarUser(ctx.user)
       .targetUser(ctx.user);
   },
+  marketing: (builder) => builder.systemNotification(),
```
AmarTrebinjac, line 542: Not sure if this is the right way to basically return notifications that we aren't going to be sending?

src/workers/transactions/userBoughtCores.ts
```diff
@@ -54,6 +59,16 @@ export const userBoughtCores: TypedWorker<'api.v1.user-transaction'> = {
...
 
+    const shouldReceiveEmail = isSubscribedToNotificationType(
+      user.notificationFlags,
+      NotificationType.InAppPurchases,
+      NotificationChannel.Email,
+    );
+
+    if (!shouldReceiveEmail) {
+      return;
+    }
```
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert 
Yes, its tied to the "in app purchases" toggle 
Screenshot 2025-08-19 at 09 18 42

src/workers/userUpdatedCio.ts
```diff
@@ -22,10 +23,14 @@ const worker: TypedWorker<'user-updated'> = {
...
-      if (
-        (user.notificationEmail && !oldUser.notificationEmail) ||
-        (user.acceptedMarketing && !oldUser.acceptedMarketing)
-      ) {
+      const oldUserReceivedEmail = isSubscribedToEmails(
+        JSON.parse(oldUser.notificationFlags),
+      );
+      const newUserReceivesEmail = isSubscribedToEmails(
+        JSON.parse(user.notificationFlags),
+      );
```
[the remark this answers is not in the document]
AmarTrebinjac: @AmarTrebinjac The old check was notificationEmail || acceptedMarketing, since marketing is also a flag now, doesn't it just make sense to just check any email?

src/workers/transactions/userBoughtCores.ts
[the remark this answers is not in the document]
capJavert: Yeah but Cores is not in app purchases I would say, its buying the currency. But ok if you got a green light from product 👍

src/workers/userUpdatedCio.ts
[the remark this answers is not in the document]
AmarTrebinjac: I think it would be the same result, but since you are unsure, I just reverted it back to the old type of check 😄

src/notifications/generate.ts
AmarTrebinjac: I checked, but I think you might be referring to `notificationTitleMap` which can return string or undefined. `generateNotificationMap` specifically returns `NotificationBuilder`

rebelchris: Ah ok, but these don't persist to storage then?
Or are they supposed to?

I'm just not sure on what the user will see here. (if anything)

rebelchris (approved):
More validation question on what the user will see?

src/notifications/generate.ts
AmarTrebinjac: The builder.systemNotification() sets the "public" property to "false", so the user will not see these 😄

rebelchris: Oh good! then we should be all good on these

src/workers/userUpdatedCio.ts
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert I asked Nimrod, who sent me to Ido, who [clarified the issue](https://dailydotdev.slack.com/archives/C03RA0WEGEQ/p1755703679412319). 😄 

Reverted back to simple check

src/schema/users.ts
```diff
@@ -2242,6 +2246,24 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          };
+        } else {
+          updatedUser.notificationFlags = {
+            ...user.notificationFlags,
+            marketing: {
+              email: NotificationPreferenceStatus.Subscribed,
+              inApp: NotificationPreferenceStatus.Subscribed,
+            },
+          };
+        }
```
capJavert, line 2265: can you use method similar to updateFlagsStatement to avoid overwriting the whole object?

src/schema/users.ts
```diff
@@ -446,10 +449,6 @@ export const typeDefs = /* GraphQL */ `
     """
     reputation: Int
     """
-    If the user has accepted marketing
-    """
-    acceptedMarketing: Boolean
-    """
```
capJavert, line 452: need to restore this?

src/schema/users.ts
capJavert: also can we add a test for this?

capJavert (commented):
Almost there 👍

src/schema/users.ts
```diff
@@ -498,6 +497,11 @@ export const typeDefs = /* GraphQL */ `
     Role for Cores access
     """
     coresRole: Int
+
+    """
+    User's notification preferences
+    """
+    notificationFlags: JSON
```
capJavert, line 504: How do we get notificationFlags currently on the frontend?

capJavert: If we need to expose this on User object we need to apply `nullIfNotSameUser` as we have for `acceptedMarketing` in graphorm

AmarTrebinjac: Purely for the whoami data. Its not used on the user object anywhere else.

capJavert: ah, whoami is not really used for anything, so we can remove it

capJavert (approved):

PR #2992, opened 2025-08-19
feat: brief notification settings

capJavert:
- remove flags.email and flags.slack from user_personalized_digest
- email setting is now managed through `notificationFlags->briefing_ready->email` and https://github.com/dailydotdev/daily-api/pull/2990 will sync it for existing users
- slack setting is now exclusively managed by user_integration, and will send if integration exists for brief source (no more checkbox on frontend)
- remove flags from subscribePersonalizedDigest mutation 
- frontend will handle changes to manage digest settings through notificationFlags mutation instead of subscribePersonalizedDigest mutation

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -4653,9 +4653,7 @@ describe('query personalizedDigest', () => {
       personalizedDigest {
         type
         flags {
-          email
           sendType
-          slack
         }
       }
   }`;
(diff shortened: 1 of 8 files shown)


rebelchris (approved):
Looks good, trust you guys to have checked the edge-cases.
But seemed this would only make things easier.

PR #2994, opened 2025-08-20
feat: post metrics

capJavert:
- save post metrics to analytics table


src/workers/postAnalytics/postAnalyticsUpdate.ts
```diff
@@ -0,0 +1,27 @@
+import { PostAnalytics } from '../../entity/posts/PostAnalytics';
+import type { TypedWorker } from '../worker';
+
+export const postAnalyticsUpdate: TypedWorker<'api.v1.post-metrics-updated'> = {
+  subscription: 'api.post-analytics-update',
+  handler: async ({ data }, con): Promise<void> => {
+    const { postId, payload } = data;
```
omBratteng, line 7: should we use zod to validate the payload?

src/workers/cdc/primary.ts
```diff
@@ -622,6 +622,29 @@ const onPostChange = async (
...
+    const postMetrics: (keyof ChangeObject<Post>)[] = [
+      'upvotes',
+      'downvotes',
+      'comments',
+      'awards',
+    ];
+
+    if (
+      postMetrics.some((field) =>
+        isChanged(data.payload.before!, data.payload.after!, field),
```
omBratteng, line 635: Should we make `isChanged` be able to accept field as array as well, and embed `some` in `isChanged` when field is an array?

omBratteng (approved):
LGTM, two non-blocking comments

src/workers/postAnalytics/postAnalyticsUpdate.ts
capJavert: yeah we can, actually `parseMessage` argument on worker object makes it easy 👌

src/workers/cdc/primary.ts
capJavert: good idea

src/workers/postAnalytics/postAnalyticsUpdate.ts
```diff
@@ -0,0 +1,31 @@
...
+        ...payload,
+      },
+      {
+        conflictPaths: ['id'],
+      },
+    );
+  },
+  parseMessage: (message) => {
+    return postMetricsUpdatedTopic.parse(message.data);
+  },
```
omBratteng, line 30: Nice, forgot we had this

PR #3015, opened 2025-08-26
feat: handle redirector and post page links on share post

idoshamun:
Special care of our own links to make sure we handle them correctly. When user drops a redirector link, we automatically fetch the post itself and link them. Same goes for the post page. Many times, I've seen people actually sharing post pages instead of the post itself, this now fixes it making it bullet proof.

It also looks twice for existing posts, one with original link and another after following the redirects. So if user shares link to dly.to or youtu.be it should handle it and find existing posts.

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -52,7 +52,6 @@ import {
   systemUser,
   parseBigInt,
   triggerTypedEvent,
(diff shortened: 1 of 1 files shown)


rebelchris (approved):
Finally!

PR #3018, opened 2025-08-27
fix: we should filter exact matches, not partial ones.

rebelchris:
We filtered out something like `sendlume.com` because we had `me.com`

Since we split on @ I think explicit match is better


capJavert (approved):
Works for me, if you can add a dedicated test that you specified in description that would be 👌

PR #3019, opened 2025-08-27
fix: only decrease on banned

rebelchris:
We should only decrease reputation on deletion of banned posts.
(We hardly hard-delete,. only on squad deletion)

The only edge-case is welcome posts as they are always banned, but since it's 99% the squad admin should be acceptable for now.

Solves:
https://github.com/dailydotdev/daily/issues/1938#issuecomment-3228353239

### DIFF
--- a/__tests__/workers/postBannedRep.ts
+++ b/__tests__/workers/postBannedRep.ts
@@ -55,7 +55,10 @@ beforeEach(async () => {
 it('should create a reputation event that increases reputation', async () => {
   const post = await con.getRepository(Post).findOneBy({ id: 'p1' });
   await expectSuccessfulBackground(worker, {
-    post,
+    post: {
+      ...post,
+      banned: true,
+    },
   });
   const events = await con
     .getRepository(ReputationEvent)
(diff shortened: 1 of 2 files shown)


src/workers/postBannedRep.ts
```diff
@@ -17,11 +17,17 @@ const worker: Worker = {
   subscription: 'post-banned-rep',
   handler: async (message, con, logger): Promise<void> => {
     const data: Data = messageToJson(message);
-    const { id, authorId, scoutId, flags } = data.post;
+    const { id, authorId, scoutId, flags, banned } = data.post;
```
rebelchris, line 20: We could also just ignore welcome posts here, not sure why we penalise those 😅

__tests__/workers/postBannedRep.ts
```diff
@@ -149,7 +152,7 @@ it('should create a reputation decrease event for the author of the freeform pos
 it('should create a reputation event that decreases reputation', async () => {
```
omBratteng, line 152: add a test to make sure it doesn't decrease on non-banned

src/workers/postBannedRep.ts
rebelchris: Solved :)

idoshamun:
> We should only decrease reputation on deletion of banned posts.
I think you got my message wrong. We decrease reputation when banning or when soft deleting (`deleted` flag).
When hard deleting (e.g the row gets removed completely from db), we should not penalize

idoshamun:
> Did end up excluding welcome post as well, not sure why we where decreasing reputation for those.

Not really, since those are created banned they never trigger this topic

capJavert (approved):

PR #3021, opened 2025-08-27
feat: clickhouse migrations support

capJavert:
Added a small script to run clickhouse migrations. Based it on lib I found, but opted to just copy small portion of functionality vs using the whole lib. Copied naming/structure of migrations from go projects.

Logic is:
- Discovers migrations named: {id}_{snake_case_name}.{up|down}.sql
- Records migration as dirty in the migrations_ch table on API
- Executes pending *.up.sql in ascending id order
- If any up fails blocks further execution
- Records each applied up into migrations_ch table (id, name, timestamp, dirty = true)

### DIFF
--- a/bin/createClickhouseMigration.ts
+++ b/bin/createClickhouseMigration.ts
@@ -3,6 +3,7 @@ import z from 'zod';
 import fs from 'node:fs/promises';
 import path from 'node:path';
 import { clickhouseMigrationsDir } from '../src/types';
+import { logger } from '../src/logger';
 
 const main = async () => {
   try {
@@ -32,7 +33,7 @@ const main = async () => {
 
     const migrationName = `${Date.now()}_${dataResult.data.name}`;
 
-    console.log(`Created migration ${migrationName} files 🎉`);
+    logger.info(`Created migration ${migrationName} files 🎉`);
 
     await Promise.all([
       fs.writeFile(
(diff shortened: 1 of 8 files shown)


bin/createClickhouseMigration.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    console.log(`Created migration ${migrationName} files 🎉`);
+
+    await Promise.all([
+      fs.writeFile(
+        path.join(clickhouseMigrationsDir, `${migrationName}.up.sql`),
+        '-- up\n\n',
+        'utf-8',
+      ),
+      fs.writeFile(
```
idoshamun, line 43: We never ever use the down script. So up to you, but feels unnecessary

idoshamun (commented):
I don't like the idea of a custom migration runner. I know that whatever we use on Go, flags the db as dirty for example so in case it fails midway, we need to manually fix it, etc. Is there no generic migration package we can use? I'm afraid of all the edge cases, etc

clickhouse/migrations/1756316678212_post_analytics.up.sql
```diff
@@ -0,0 +1,81 @@
+-- up
```
idoshamun, line 1: This is already battle tested right? It exists but we just add a proper migration now?

vpol:
We use this: https://github.com/golang-migrate/migrate

it's actually a standalone tool, I just wrote a small wrapper to allow to use env convention.

capJavert:
Yes found it with Ido, I coped pieces from it into this, mostly dirty behaviour 👍

PR #3024, opened 2025-08-28
feat: add opportunity tables

omBratteng:
Before merging, I will merge the separate migration files, but kept them separate for now to make it easier to review


src/entity/Organization.ts
```diff
@@ -61,6 +122,33 @@ export class Organization {
...
+  founded: number;
+
+  @Column({ type: 'text', default: null })
+  location: string;
+
+  @Column({ type: 'text', default: null })
+  size: CompanySize;
+
+  @Column({ type: 'text', default: null })
+  category: CompanyCategory;
```
rebelchris, line 147: I think string will work better here, else we probably end up runing tons of changes over and over.

rebelchris (commented):
Don't see anything blocking for now.

rebelchris (approved):
Looks good to me :)

PR #3025, opened 2025-08-28
feat: skadi v2 workers

sshanzel:
Workers for when the campaign is:
- Completed.
- Started.
- Updated stats.
- Used budget updated.

### Jira ticket
MI-1006


src/common/campaign/common.ts
```diff
@@ -131,3 +135,63 @@ export const getReferenceTags = (
       throw new ValidationError('Unknown campaign type to estimate reach');
   }
 };
+
+export enum CampaignUpdateEvent {
```
rebelchris, line 139: Does this come from Skadi or is it API only?
Asking because ideally we should leverage proto then

src/workers/notifications/campaignUpdatedAction.ts
```diff
@@ -0,0 +1,66 @@
...
+        .findOneByOrFail({ id: campaignId }),
+      queryRunner.manager
+        .getRepository(User)
+        .findOneByOrFail({ id: campaign.userId }),
+    ]);
+
+    return promises;
+  });
+
+  let source = undefined;
```
rebelchris, line 46: Can't we just omit the let and do `const source` below and something like `source || undefined` in the return?

rebelchris (commented):
I see very few validations, don't you expect required fields etc?

src/workers/notifications/campaignUpdatedAction.ts
sshanzel: The source is not always necessary to be present. Are you asking it to be inline like `source: campaign.type === CampaignType.Squad ? query || undefined`?

src/common/campaign/common.ts
sshanzel: This comes from Skadi.

src/workers/notifications/campaignUpdatedAction.ts
rebelchris: Uhh in the return below yeah, never big fan of lets :P 
But it's not a big thing

sshanzel: Removed the let now but moving the object above and setting the property only when necessary.

src/workers/campaignUpdatedAction.ts
```diff
@@ -0,0 +1,105 @@
...
+  );
+};
+
+const handleCampaignStatsUpdate = async (
+  con: DataSource,
+  { data, campaignId, d_update }: CampaignUpdateEventArgs,
+) => {
+  const { impressions, clicks, unique_users } = data as CampaignStatsUpdate;
+
+  await con.getRepository(Campaign).update(
```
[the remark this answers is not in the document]
sshanzel: Updated and added the tests now.

src/entity/campaign/Campaign.ts
```diff
@@ -29,6 +29,7 @@ export interface CampaignFlags {
   impressions: number;
   clicks: number;
   users: number;
+  lastUpdatedAt?: Date;
```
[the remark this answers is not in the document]
capJavert: Feels redundant yes

capJavert (approved):
Feels like we don't need updatedAt in jsonb as well, so see if we can remove it.

PR #3027, opened 2025-08-28
feat: close clickhouse connection on teardown

capJavert:
- it was running only in crons so its fine, but it does not hurt to add this to have graceful shutdown

### DIFF
--- a/__tests__/teardown.ts
+++ b/__tests__/teardown.ts
@@ -1,3 +1,4 @@
+import { closeClickHouseClient } from '../src/common/clickhouse';
 import '../src/config';
 import createOrGetConnection from '../src/db';
 import { ioRedisPool, redisPubSub, singleRedisClient } from '../src/redis';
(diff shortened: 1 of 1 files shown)


src/common/clickhouse.ts
```diff
@@ -5,11 +5,19 @@ let client: ReturnType<typeof createClient> | null = null;
 export const getClickHouseClient = () => {
   if (!client) {
     client = createClient({
-      url: process.env.CLICKHOUSE_URL,
+      url: 'http://localhost:18123',
```
rebelchris, line 8: Don't we want this to be env based still?
Probably you added it for testing locally

rebelchris (changes requested):
Minor change needed

src/common/clickhouse.ts
capJavert: ah yeah, slipped through

PR #3028, opened 2025-08-29
refactor: campaign by id fetched by regular users

sshanzel:
We use the `campaignById` query to fetch the information about who boosted the Squad and more details, such as when it is going to end. Also removed the `@auth` directive as anonymous users are able to view Squads directory that has this feature.

For other Squad mods/admins:

Screenshot 2025-08-29 at 5 12 20 PM

For anonymous user:

Screenshot 2025-08-29 at 5 10 54 PM

### DIFF
--- a/__tests__/campaigns.ts
+++ b/__tests__/campaigns.ts
@@ -344,21 +344,8 @@ describe('query campaignById', () => {
     });
   });
 
-  it('should throw error when campaign not found', async () => {
-    loggedUser = '1';
-
-    return testQueryErrorCode(
-      client,
-      {
-        query: CAMPAIGN_BY_ID_QUERY,
-        variables: { id: '550e8400-e29b-41d4-a716-446655440011' },
-      },
-      'NOT_FOUND',
-    );
-  });
-
-  it('should allow any user to access any campaign', async () => {
-    loggedUser = '2'; // User 2 accessing User 1's campaign
+  it('should return full flags data for campaign owners', async () => {
+    loggedUser = '1'; // User 1 accessing their own campaign
 
     const res = await client.query(CAMPAIGN_BY_ID_QUERY, {
       variables: { id: CAMPAIGN_UUID_1 }, // This campaign belongs to user '1'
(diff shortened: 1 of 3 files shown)


rebelchris (approved):

AmarTrebinjac (approved):

__tests__/campaigns.ts
```diff
@@ -357,28 +357,94 @@ describe('query campaignById', () => {
     );
   });
 
-  it("should throw error when user tries to access another user's campaign", async () => {
-    loggedUser = '2'; // User 2 trying to access User 1's campaign
+  it('should allow any user to access any campaign', async () => {
```
capJavert, line 360: Should this really be the case, I don't think anonymous users would really need to know what is boosted, if they must know maybe we can give them access on just the part of the fields, eg. name of the admin that boosted, other data they don't need

capJavert: You can use graphorm to limit which properties are available to who, we have helpers to check current user eg. if you want to check if user is owner and so on.

capJavert: @sshanzel where do you check if user owns the campaign, I don't see it in code? Let's say for `campaignById`

capJavert: Gotcha, let's allow only `id` and `user` fields for non owners. That would be enough to show data you need?

sshanzel: I'll add type too as it is pretty important in many cases.

sshanzel: Basically we just removed `createdAt` and `state` which I think doesn't sound too bad. But if you really think we should hide them, then sure I'll remove them.

sshanzel: @capJavert should be updated by now.

src/graphorm/index.ts
```diff
@@ -1354,10 +1355,30 @@ const obj = new GraphORM({
...
+        transform: (flag, ctx, parent) => {
+          if (ctx.userId === (parent as Campaign).userId) {
+            return flag;
+          }
+
+          return null;
+        },
+      },
+      createdAt: {
+        rawSelect: true,
```
capJavert, line 1371: I think you can remove this

capJavert (approved):

__tests__/campaigns.ts
capJavert: Perfect 👌 thank you!

PR #3030, opened 2025-08-29
feat: add generation id for local env

capJavert:
To be able to test locally if no skadi instance.


src/routes/localAds.ts
```diff
@@ -61,12 +62,25 @@ const ads = [
   },
 ];
 
+const skadiGenerationIdHeader = 'x-generation-id';
+
+const addSkadiGenerationHeaders = (reply: FastifyReply) => {
+  return reply.headers({
+    [skadiGenerationIdHeader]: randomUUID(),
```
capJavert, line 69: just randomUUID we don't really care about contents locally

PR #3035, opened 2025-08-30
refactor: post boost legacy endpoints to utilize v2 flow

sshanzel:
**Note:** This must not be merged only when we shipped Frontend. Then we should import data from Clickhouse.

One of the strongest indicators that we were able to successfully port the legacy endpoints to use the v2 client is seeing the tests turn green for the `boost.ts` file.

It is important to look at the shape of the objects that are returned by the queries, because that would tell us that old extension users are going to be served correctly.

Notable changes:
- startPostBoost to use v2 client.
- stopPostBoost to use v2 client.
- postCampaigns to use v2 client.
- postCampaignById to use v2 client.
- Removed a lot of functions that are not gonna be used anymore (soon the whole v1 client).
- The `engagements` property is not being used for quite a while now - we are returning a value of 0 to not break any existing queries.

### Jira ticket
MI-983

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -1357,7 +1357,7 @@ export const typeDefs = /* GraphQL */ `
       ID of the exisiting post
       """
       postId: ID
-    ): SourcePostModeration! @auth
+    ): SourcePostModeration! @auth @rateLimit(limit: 1, duration: 30)
 
     """
     Hide a post from all the user feeds
@@ -2473,11 +2473,14 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       ctx: AuthContext,
       info,
     ): Promise<GQLSourcePostModeration> => {
-      await ensureSourcePermissions(
-        ctx,
-        props.sourceId,
-        SourcePermissions.PostRequest,
-      );
+      await Promise.all([
+        ensureSourcePermissions(
+          ctx,
+          props.sourceId,
+          SourcePermissions.PostRequest,
+        ),
+        ensurePostRateLimit(ctx.con, ctx.userId),
+      ]);
 
       const pendingPost = await validateSourcePostModeration(ctx, props);
       const moderatedPost = await createSourcePostModeration({
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -10,6 +10,10 @@ const REMOVE_SPECIAL_CHARACTERS_REGEX = /[^a-zA-Z0-9-_#.]/g;
 
 export const DELETED_BY_WORKER = 'worker';
 
+export const playwrightUser = {
+  id: '8bf2UpTsHFnczzOlk0mtg',
+};
+
 export const ghostUser = {
   id: '404',
   username: 'ghost',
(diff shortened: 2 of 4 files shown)


src/integrations/skadi/api/v2/clients.ts
```diff
@@ -131,20 +131,25 @@ export class SkadiApiClientV2 implements ISkadiApiClientV2 {
     userId,
   }: CancelCampaignArgs): Promise<{ budget: string }> {
     return this.garmr.execute(async () => {
-      const { budget, error } = await fetchParse<CancelPostCampaignResponse>(
```
sshanzel, line 134: The `fetchParse` function does not really check `response.ok`, so error 400 and such are not handled properly.

src/schema/posts.ts
```diff
@@ -296,6 +299,49 @@ export type GQLPostAnalyticsHistory = Omit<
...
+        SELECT json_build_object(
+          'id', p1.id,
+          'shortId', p1."shortId",
+          'slug', p1.slug,
+          'image', p1.image,
+          'title', p1.title,
+          'type', p1.type,
+          'sharedTitle', p2.title,
+          'sharedImage', p2.image
+        )
```
sshanzel, line 337: To avoid creating more types/interfaces since all these are going to be deleted anyway once we remove support for v1 after getting the right amount of adoption.

src/common/campaign/common.ts
```diff
@@ -222,3 +232,27 @@ export const generateCampaignCompletedEmail: TemplateDataFunc = async (
       return null;
   }
 };
+
+export const getUserCampaignStats = async (
```
capJavert, line 236: is this duplicate now with other PR?

src/schema/posts.ts
capJavert: why do we even do it, why are we not just keeping old queries and once v2 adoption goes through we just kill the old endpoints? I am a bit confused with all the back and forth about v1 and v2

capJavert: Also I am pretty sure you can get this query with repo and query builder, don't see why you need this.

capJavert (commented):
Not sure if we need this, we went from v1 to v2 to v1 multiple times feels like 😆 Feel free to merge if you think its needed.

sshanzel:
> I am not sure of all the logic and fallback, trust you on this one.
> 
> https://media4.giphy.com/media/v1.Y2lkPTc5MGI3NjExNDRvbWYweHI3OTkxZDdvMTducGFqZWM1dm5mNmFiMTBtaDhvY285cyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/TrTHcHiO1jUpZVtOzl/giphy.gif

Yes, I shall do local testings too for sure before I merge 🙏 

The tests turning green is a good indicator that we were able to port it successfully. Hopefully that's going to be the case in prod 🤣

sshanzel:
Local dev extension build with this PR as the API. The items listed below were created using the v2 flow but can be listed as v1.

Screenshot 2025-09-03 at 10 01 11 PM

PR #3036, opened 2025-08-30
feat: query campaign stats as a separate query

sshanzel:
We've separated the stats call from the list to avoid coupling things together.

### DIFF
--- a/src/common/campaign/common.ts
+++ b/src/common/campaign/common.ts
@@ -138,16 +138,16 @@ export const getReferenceTags = (
   }
 };
 
-export interface UserCampaignData {
+export interface UserCampaignStats {
   impressions: number;
   clicks: number;
   spend: number;
   users: number;
 }
 
-export const getUserCampaignData = async (
+export const getUserCampaignStats = async (
   ctx: AuthContext,
-): Promise<UserCampaignData> => {
+): Promise<UserCampaignStats> => {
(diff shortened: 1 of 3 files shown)


src/common/campaign/common.ts
```diff
@@ -131,3 +137,33 @@ export const getReferenceTags = (
...
+  impressions: number;
+  clicks: number;
+  spend: number;
+  users: number;
+}
+
+export const getUserCampaignData = async (
+  ctx: AuthContext,
+): Promise<UserCampaignData> => {
+  const result = await queryReadReplica(ctx.con, ({ queryRunner }) =>
```
rebelchris, line 151: You ok with this coming from read replica? I'm not sure how often it processes this/how realtime it needs to be?

rebelchris (approved):
No blocker, just checking if the RR is ok for this query.

src/schema/campaigns.ts
```diff
@@ -85,6 +87,13 @@ export const typeDefs = /* GraphQL */ `
     max: Int!
   }
 
+  type UserCampaignData {
```
capJavert, line 90:
```suggestion
  type UserCampaignStats {
```

Data is generic terms, maybe stats or analytics is better?

sshanzel: Yeah, my mind was battling which one to use 😅

PR #3037, opened 2025-09-01
fix: missing data from workers

sshanzel:
Campaign v2 workers: if campaign does not exist, ignore it as [Viktor have said](https://dailydotdev.slack.com/archives/C07LHS7RXA4/p1756719562719959?thread_ts=1756718292.452039&cid=C07LHS7RXA4).

Campaign v1 workers: when a user is not found, log an error instead of throwing it.

### DIFF
--- a/__tests__/workers/campaignUpdatedAction.ts
+++ b/__tests__/workers/campaignUpdatedAction.ts
@@ -9,7 +9,7 @@ import { DataSource } from 'typeorm';
 import createOrGetConnection from '../../src/db';
 import {
   CampaignUpdateEvent,
-  type CampaignStatsUpdateEvent,
+  type CampaignUpdateEventArgs,
 } from '../../src/common/campaign/common';
 import { CampaignState } from '../../src/entity/campaign/Campaign';
 import { randomUUID } from 'crypto';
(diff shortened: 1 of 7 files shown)


src/workers/campaignUpdatedAction.ts
```diff
@@ -24,7 +24,7 @@ const worker: TypedWorker<'skadi.v2.campaign-updated'> = {
       .findOneBy({ id: message.data.campaignId });
 
     if (!campaign) {
-      throw new Error(`Campaign not found! ${message.data.campaignId}`);
+      return;
```
capJavert, line 27: but should we not save it in case we don't have it already?

src/workers/notifications/postBoostAction.ts
```diff
@@ -4,12 +4,18 @@ import { generateTypedNotificationWorker } from './worker';
...
+import { logger } from '../../logger';
 
 const worker = generateTypedNotificationWorker<'skadi.v1.campaign-updated'>({
   subscription: 'api.campaign-updated-notification',
   handler: async (params, con) => {
     const { userId, postId, campaignId, action } = params;
 
+    if (!userId) {
+      logger.error({ data: params }, `skadi v1 worker: user id is empty!`);
+      return;
```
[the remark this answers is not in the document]
sshanzel: I was also under the impression that we should only receive squad/post boost campaigns, hence this should be looked at in case the user id is missing. But after the discussion in Slack, we should also ignore. Also not a fan of this whole ignore thing 😅

__tests__/workers/campaignUpdatedAction.ts
```diff
@@ -9,7 +9,7 @@ import { DataSource } from 'typeorm';
 import createOrGetConnection from '../../src/db';
 import {
   CampaignUpdateEvent,
-  type CampaignStatsUpdateEvent,
+  type CampaignUpdateEventArgs,
```
sshanzel, line 12: Some merge conflict type error. Shouldn't affect anything, just fixes TS lints.

PR #3040, opened 2025-09-01
feat: import skadi campaigns

capJavert:
- already run initial import in prod db, committing to keep the script for future use if needed

### DIFF
--- a/bin/importCampaignsFromSkadi.ts
+++ b/bin/importCampaignsFromSkadi.ts
@@ -1,0 +1,192 @@
+import '../src/config';
+import { z } from 'zod';
+import createOrGetConnection from '../src/db';
+import { logger } from '../src/logger';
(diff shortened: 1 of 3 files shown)


bin/importCampaignsFromSkadi.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    chunks[chunks.length - 1].push({
+      id: item.campaign_id,
+      referenceId: item.post_id,
+      userId: item.user_id,
+      type: CampaignType.Post,
+      endedAt: item.end_date,
+      state: campaignStateFromSkadi({ item }),
+      flags: {
+        budget: usdToCores(item.budget),
```
sshanzel, line 161: On our side, we store the `budget` as the full budget. We can maybe multiply this with the result of `getAbsoluteDifferenceInDays` for the start date and end date.

PR #3044, opened 2025-09-02
refactor: exclude system mods and some extra from leaderboard

idoshamun:
Following our hug a user post, a user asked to remove ourselves from the leaderboard, and it makes sense. So here we are 😬

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -60,3 +60,10 @@ export const FEED_SURVEY_INTERVAL = 30;
 export const WEBAPP_MAGIC_IMAGE_PREFIX = `/image-generator`;
 
 export const SQL_QUERIES_PATH = '/opt/app/queries';
+
+export const MODERATORS = [
+  '1d339aa5b85c4e0ba85fdedb523c48d4',
+  '28849d86070e4c099c877ab6837c61f0',
+  '5e0af68445e04c02b0656c3530664aff',
+  'a491ef61599a4b3e84b6dfa602e6bdfe',
+];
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -60,3 +60,10 @@ export const FEED_SURVEY_INTERVAL = 30;
 export const WEBAPP_MAGIC_IMAGE_PREFIX = `/image-generator`;
 
 export const SQL_QUERIES_PATH = '/opt/app/queries';
+
+export const MODERATORS = [
+  '1d339aa5b85c4e0ba85fdedb523c48d4',
+  '28849d86070e4c099c877ab6837c61f0',
+  '5e0af68445e04c02b0656c3530664aff',
+  'a491ef61599a4b3e84b6dfa602e6bdfe',
+];
```
omBratteng, line 69: Could this be stored in remote config?

src/schema/leaderboard.ts
```diff
@@ -100,6 +101,15 @@ export const typeDefs = /* GraphQL */ `
   }
 `;
 
+const excludedUsers = [
+  ghostUser.id,
```
omBratteng, line 105: iirc there's also `system` user

capJavert: we jave `isSpecialUser` helper for ghost, system and so on

src/config.ts
idoshamun: i was afraid of this comment 🙈 
it's not new, i just moved it to a new place 😬

omBratteng: I know, just always thought it'd be useful to keep in remote config if you wanted to add extra users quickly.

PR #3045, opened 2025-09-02
chore: cleanup show_squad_unread_posts

nensidosari:
Cleanup experiment https://dailydotdev.slack.com/archives/C02JAUF8HJL/p1755767049664219?thread_ts=1755766562.964589&cid=C02JAUF8HJL


src/migration/1754321021916-RollbackSourceMemberHasUnreadPostsIndexFix.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class RollbackSourceMemberHasUnreadPostsIndexFix1754321021916 implements MigrationInterface {
```
capJavert, line 3: you can rollback only current state, no need to do multiple rollbacks of all migrations

capJavert: so basically just drop the index in its latest state

capJavert:
@sshanzel streams topic is `api.v1.post-visible`, we should not remove it since its used for other subscriptions.

src/migration/1754321021916-RollbackSourceMemberHasUnreadPostsIndexFix.ts
nensidosari: updated

capJavert (approved):

nensidosari:
Will wait for extension release and adoption to merge this just to rule out any possible issues.

rebelchris (approved):

nensidosari:
We can merge this now - extension adoption is more then 95%

PR #3046, opened 2025-09-02
feat: new event for members count

sshanzel:
Reference: https://github.com/dailydotdev/skadi/blob/main/pkg/update_publisher/types.go#L36

Sample event shape:

```
{
    "campaignId": "some_campaign_id",
	"event": "EXTRA_STATS_UPDATED",
	"data": {
        "complete joining squad": {
            "events_count": 12345 <- you ignore this
            "unique_events_count": 12345 <- this is what you need
        }
    },
	"d_update": <usual_value_here>
}
```

### DIFF
--- a/src/workers/campaignUpdatedAction.ts
+++ b/src/workers/campaignUpdatedAction.ts
@@ -118,7 +118,7 @@ const handleExtraCampaignStatsUpdate = async ({
   params: { data, campaignId },
 }: HandlerEventArgs) => {
   const update = data as CampaignExtraStatsUpdate;
-  const newMembers = update?.['complete joining squad']?.unique_events_count;
+  const newMembers = update['complete joining squad']?.unique_events_count;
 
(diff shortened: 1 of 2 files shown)


src/workers/campaignUpdatedAction.ts
```diff
@@ -117,15 +117,15 @@ const handleExtraCampaignStatsUpdate = async ({
...
-    { id: campaignId },
-    {
-      flags: updateFlagsStatement<Campaign>({
-        newMembers: (data as CampaignExtraStatsUpdate)['complete joining squad']
-          .unique_events_count,
-      }),
-    },
-  );
+  const update = data as CampaignExtraStatsUpdate;
+  const newMembers = update?.['complete joining squad']?.unique_events_count;
```
capJavert, line 121: make it optional please on the type as well then

PR #3055, opened 2025-09-03
fix: update the column types for Opportunity and UserCandidatePreference to be int

omBratteng:
Because we use the enum values from protobuf schema, the value is now a int and not text
And replace Set with Array because it causes more issue than it solved


omBratteng:
> Nice idea with comments

Yeah, ideally I'd dump the protobuf schema enum directly into the comment, so you can see the enum values

capJavert:
@omBratteng yeah Object.entries then each time its updated you have to do migration to update comment :hidethepain:

PR #3056, opened 2025-09-03
feat: get campaigns per entity

capJavert:
I need to query campaigns for post Id to show them on analytics graph, so I exposed `entityId` so we can fetch it in the future for squads and so on.

### DIFF
--- a/src/schema/campaigns.ts
+++ b/src/schema/campaigns.ts
@@ -113,6 +113,11 @@ export const typeDefs = /* GraphQL */ `
       Paginate first
       """
       first: Int
+
+      """
+      ID of the entity to fetch campaigns for
+      """
+      entityId: ID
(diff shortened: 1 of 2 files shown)


src/schema/campaigns.ts
```diff
@@ -220,6 +227,15 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
             userId,
           });
 
+          if (args.entityId) {
+            builder.queryBuilder.andWhere(
+              `"${alias}"."referenceId" = :entityId`,
+              {
+                entityId: args.entityId,
```
capJavert, line 234: Not checking entity access since its implied on campaign level

omBratteng (approved):

sshanzel (approved):
🚀

PR #3061, opened 2025-09-04
intermediate: recruiter

rebelchris:
Intermediate branch


omBratteng:
1. no data, no loss
2. default value comes from db
3. globally handled from cdc worker
4. this will be dropped in next PR
5. no, that's the intent
6. same as 3.
7. it's fine, we don't want to update comments each time schema is updated
8. stop hallucinating

PR #3062, opened 2025-09-04
fix: playwright user should be excluded from marketing CTA's

rebelchris:
We should avoid having the playwright user receiving marketing CTA's to avoid heart attacks on devs :)

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -10,6 +10,10 @@ const REMOVE_SPECIAL_CHARACTERS_REGEX = /[^a-zA-Z0-9-_#.]/g;
 
 export const DELETED_BY_WORKER = 'worker';
 
+export const playwrightUser = {
+  id: '8bf2UpTsHFnczzOlk0mtg',
+};
+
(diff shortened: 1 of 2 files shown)


omBratteng (approved):
testing? 🙈  (if there already exist any)

src/common/utils.ts
```diff
@@ -10,6 +10,10 @@ const REMOVE_SPECIAL_CHARACTERS_REGEX = /[^a-zA-Z0-9-_#.]/g;
 
 export const DELETED_BY_WORKER = 'worker';
 
+export const playwrightUser = {
+  id: '8bf2UpTsHFnczzOlk0mtg',
+};
```
capJavert, line 15: you can add him to `isSpecialUser` as well

rebelchris: Mostly used for njord now right?
I think the only issue for now was the random popups it would get.

capJavert: we use it on few places, to ignore those kind of special users, but yeah no blocker

PR #3065, opened 2025-09-04
refactor: boost notification

sshanzel:
Notifications are now generated through the new type.

Props have changed, but they remain pretty similar (removed stats from email).


src/workers/newNotificationV2Mail.ts
```diff
@@ -109,15 +117,26 @@ export const notificationToTemplateId: Record<NotificationType, string> = {
...
   user_follow: '',
   marketing: '',
   new_user_welcome: '',
   announcements: '',
   in_app_purchases: '',
-  campaign_completed: '', // TODO: MI-1007 - wait for design's template id
+  campaign_completed: async (con, notification) => {
+    const campaign = await con
+      .getRepository(Campaign)
+      .findOneBy({ id: notification.referenceId });
```
capJavert, line 129: I don't like this, we add additional query which will fire when resolving, can we keep this map as is, you can split the campaigns in different notification types if they require different templates

capJavert (changes requested):
Some comments

src/workers/newNotificationV2Mail.ts
sshanzel: It is the worker. I don't think an additional query should hurt when fetching the template ID since we only fetch it once, regardless of the number of users to send to.

capJavert: Either way I would prefer not to add async complexity here, we have different type of notifications for this reason no?

capJavert: Is there is some other issue to having this not be same notification type that I am not aware of?

src/workers/newNotificationV2Mail.ts
```diff
@@ -1106,7 +1164,7 @@ const worker: Worker = {
                 return;
               }
               const formattedData = formatTemplateDate(templateData);
-              const templateId = notificationToTemplateId[notification.type];
```
capJavert, line 1109: just return this to the original lines to minimize unrelated changes 🙏

src/workers/newNotificationV2Mail.ts
```diff
@@ -132,9 +132,65 @@ export type TemplateDataFunc = (
...
+      .getRepository(CampaignPost)
+      .findOneBy({ id: notification.referenceId });
+
+    if (!campaign) {
+      return null;
+    }
+
+    const post = await con.getRepository(Post).findOneOrFail({
+      where: { id: campaign.postId },
+    });
```
capJavert, line 146: potentially preload post on campaign query above with relation

capJavert (approved):
Tnx for adjusting. Just minor comments, and see tests fail, but approving.

src/workers/newNotificationV2Mail.ts
sshanzel: Right. Let me see if we can also preload the sharedPost 🤯

PR #3070, opened 2025-09-07
feat: create poll

AmarTrebinjac:
### Jira ticket

Poll creation PR. Handles entity updates and creation of polls. Fetching, voting and moderation will be separate tasks.

Not sure if it would be preferable to merge all to a intermediate branch before merging to main. Your call @capJavert 

MI-1009

### DIFF
--- a/__tests__/posts.ts
+++ b/__tests__/posts.ts
@@ -9081,6 +9081,10 @@ describe('mutate polls', () => {
     });
   });
 
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
   const MUTATION = `
     mutation CreatePollPost($sourceId: ID!, $title: String!, $options: [PollOptionInput!]!, $duration: Int) {
       createPollPost(sourceId: $sourceId, title: $title, options: $options, duration: $duration) {
@@ -9184,6 +9188,20 @@ describe('mutate polls', () => {
 
   it('should create a poll that ends in 7 days', async () => {
     loggedUser = '1';
+    const fakeNow = new Date(2025, 5, 5);
+    const futureDate = addDays(fakeNow, 7);
+
+    // Test fails due to timeout if we do not exclude the APIs we don't need.
+    const doNotFake: FakeableAPI[] = [
+      'nextTick',
+      'setImmediate',
+      'clearImmediate',
+      'setInterval',
+      'clearInterval',
+      'setTimeout',
+      'clearTimeout',
+    ];
+    jest.useFakeTimers({ doNotFake }).setSystemTime(fakeNow);
 
     const poll = {
       ...defaultPoll,
@@ -9195,13 +9213,14 @@ describe('mutate polls', () => {
       variables: poll,
     });
 
+    jest
+      .useFakeTimers({ doNotFake, advanceTimers: true })
+      .setSystemTime(futureDate);
+
     expect(res.errors).toBeFalsy();
     expect(res.data.createPollPost.endsAt).toBeTruthy();
     expect(
-      isSameDay(
-        new Date(res.data.createPollPost.endsAt),
-        addDays(new Date(), 7),
-      ),
+      isSameDay(new Date(res.data.createPollPost.endsAt), futureDate),
     ).toBeTruthy();
   });
 
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -232,7 +232,7 @@ export interface GQLPost {
   permalink?: string;
   endsAt?: Date;
   pollOptions?: GQLPollOption[];
-  totalPollVotes?: number;
+  numPollVotes?: number;
 }
 
 interface PinPostArgs {


src/entity/polls/PollOption.ts
```diff
@@ -0,0 +1,40 @@
...
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @Column({ type: 'text' })
+  text: string;
+
+  @Column({ type: 'integer' })
+  order: number;
+
+  @Index('poll_option_post_id_index')
```
omBratteng, line 23: Prefix IDX_

src/entity/polls/PollOption.ts
```diff
@@ -0,0 +1,40 @@
...
+  ManyToOne,
+  OneToMany,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import type { Post } from '../posts';
+import type { UserPost } from '../user';
+
+@Entity()
+export class PollOption {
+  @PrimaryGeneratedColumn('uuid')
```
omBratteng, line 14: See my comment on slack regarding manfully specifying primary keys.

https://dailydotdev.slack.com/archives/C02E2C3C13R/p1756725897378429

omBratteng (changes requested):

src/migration/1757069298710-PollPost.ts
```diff
@@ -5,10 +5,10 @@ export class PollPost1757069298710 implements MigrationInterface {
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
-      `CREATE TABLE "poll_option" ("id" uuid NOT NULL DEFAULT uuid_generate_v4(), "text" text NOT NULL, "order" integer NOT NULL, "postId" text NOT NULL, "numVotes" integer NOT NULL DEFAULT '0', CONSTRAINT "PK_5fdd46d449ddcc8201aed9b5a1b" PRIMARY KEY ("id"))`,
+      `CREATE TABLE "poll_option" ("id" uuid NOT NULL DEFAULT uuid_generate_v4(), "text" text NOT NULL, "order" integer NOT NULL, "postId" text NOT NULL, "numVotes" integer NOT NULL DEFAULT '0', CONSTRAINT "PK_Poll_Option_Id" PRIMARY KEY ("id"))`,
     );
     await queryRunner.query(
-      `CREATE INDEX "poll_option_post_id_index" ON "poll_option" ("postId") `,
+      `CREATE INDEX "IDX_poll_option_post_id_index" ON "poll_option" ("postId") `,
```
omBratteng, line 11: CREATE INDEX IF NOT EXITS

src/migration/1757069298710-PollPost.ts
```diff
@@ -0,0 +1,51 @@
...
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "user_post" DROP CONSTRAINT "FK_User_Post_Poll_Vote_Option_Id"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "poll_option" DROP CONSTRAINT "FK_Poll_Option_Post_Id"`,
+    );
```
omBratteng, line 39: can be probably removed, since it should be implicitly dropped via drop table

src/migration/1757069298710-PollPost.ts
AmarTrebinjac: @omBratteng You should make a PR on the TYPEORM repo :lolsob:

src/migration/1757069298710-PollPost.ts
```diff
@@ -0,0 +1,51 @@
...
+      `ALTER TABLE "post" ADD "numPollVotes" integer DEFAULT '0'`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "user_post" ADD "pollVoteOptionId" uuid`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "poll_option" ADD CONSTRAINT "FK_Poll_Option_Post_Id" FOREIGN KEY ("postId") REFERENCES "post"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
+    );
```
omBratteng, line 27: This can also be added in the create table, see for reference
https://github.com/dailydotdev/daily-api/blob/a1982fa538213511d1ae695a54f7b2242e8cb8e8/src/migration/1756461014954-Opportunity.ts#L21-L25

src/migration/1757069298710-PollPost.ts
omBratteng: 😂

src/common/post.ts
```diff
@@ -239,6 +242,76 @@ interface CreateFreeformPostArgs {
...
+  const { private: privacy } = await con.getRepository(Source).findOneByOrFail({
+    id: restArgs.sourceId,
+    type: In([SourceType.Squad, SourceType.User]),
+  });
+
+  const createdPost = con.getRepository(PollPost).create({
+    ...restArgs,
+    shortId: restArgs.id,
+    endsAt: restArgs?.duration && addDays(new Date(), restArgs.duration),
+    visible: true,
```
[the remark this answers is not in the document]
AmarTrebinjac: I used `createFreeformPost` as reference, and we do the same there:
Screenshot 2025-09-08 at 14 56 20

src/migration/1757069298710-PollPost.ts
AmarTrebinjac: You are making me sweat writing all this string templated SQL 😭

capJavert:
@AmarTrebinjac also please check if you need to add this post type to `nullableImageType` and `editablePostTypes`

omBratteng (approved):
migration looks good now

__tests__/posts.ts
```diff
@@ -9067,3 +9068,183 @@ describe('query history for post analytics', () => {
...
+      variables: poll,
+    });
+
+    expect(res.errors).toBeFalsy();
+    expect(res.data.createPollPost.endsAt).toBeTruthy();
+    expect(
+      isSameDay(
+        new Date(res.data.createPollPost.endsAt),
+        addDays(new Date(), 7),
+      ),
```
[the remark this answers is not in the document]
AmarTrebinjac: But this is testing that the duration is being converted and set properly 😅 

Might as well just check that the prop is being set instead then ?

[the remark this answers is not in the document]
capJavert: your testing is fine, what I mean that tests involving current time can be flaky, let's say your test runs close to midnight, isSameDay could technically fail because of time difference between code writing to db and then calling `new Date()`, so its better to set fixed `endsAt` and then also expecting fixed date, not just any day + 7.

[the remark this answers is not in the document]
AmarTrebinjac: I understand the potential issue, but not the solution. I might be missing something obvious 😅 

Since we are passing a duration number (1,3,7,14,30) to the mutation, which then resolves it into a date - how can we then pass a static endsAt date?

Maybe we could use the `isWithinInterval` function from date-fns in the test instead?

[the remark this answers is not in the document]
capJavert: its not about mutation and real code, its about the test setup, you want to expect static date values and not dynamic like current date

nensidosari (approved):

PR #3071, opened 2025-09-08
feat: introduce pre-insert and pre-update hooks to posts

idoshamun:
Initially I started introducing a block by words in title feature, and it turned out to be a mess. So I decided to introduce hooks that will allow us to easily do that and also expand in the future. As part of the initial setup, I also do the vordr check there as well as added title check.

I also removed some old stuff of community picks when I encountered it

### DIFF
--- a/src/common/vordr.ts
+++ b/src/common/vordr.ts
@@ -1,4 +1,5 @@
 import { isInSubnet, isIP } from 'is-in-subnet';
+import emojiRegex from 'emoji-regex';
 import { User } from '../entity';
 import { logger } from '../logger';
 import { counters } from '../telemetry';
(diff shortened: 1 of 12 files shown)


src/common/vordr.ts
```diff
@@ -62,6 +83,20 @@ export const checkWithVordr = async (
     return true;
   }
 
+  // Check post title for vordr conditions
+  if (title && validatePostTitle(title)) {
+    logger.info(
+      { id, type, userId },
+      `Prevented ${type} because title contains banned content`,
+    );
```
capJavert, line 91: if we are logging, I think it would be useful to say what the title was and what it was set to, not much use of logs without that info.

src/entity/posts/utils.ts
```diff
@@ -285,6 +288,96 @@ interface CreateExternalLinkArgs {
...
+      id: post.id || 'new-post',
+      type: VordrFilterType.Post,
+      title: post.title || undefined,
+      content: postContent,
+    },
+    context,
+  );
+
+  if (shouldVordr) {
+    logger.info({ postId: post.id }, 'Post will be vordr on insert');
```
capJavert, line 318: again not sure its useful to have logs, you can always see in the DB if post is vordr?

PR #3072, opened 2025-09-08
refactor: remove gql enums of protobf enums

omBratteng:
Remove GQL enums from Protobuf enums, use Int instead via custom scalar ProtoEnumValue, that way we can differentiate between actual Int and when we intend for proto enums.

The scalar ProtoEnumValue just accepts whole ints above and 0

### DIFF
--- a/src/schema/common.ts
+++ b/src/schema/common.ts
@@ -6,7 +6,12 @@ import {
   getOffsetWithDefault,
   offsetToCursor,
 } from 'graphql-relay';
-import { GraphQLResolveInfo } from 'graphql';
+import {
+  GraphQLResolveInfo,
+  GraphQLScalarType,
+  Kind,
+  ValueNode,
+} from 'graphql';
 // @ts-expect-error - no types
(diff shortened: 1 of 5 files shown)


src/schema/common.ts
```diff
@@ -96,11 +103,38 @@ export const typeDefs = /* GraphQL */ `
...
 
+function toPositiveInt(value: unknown): number {
+  const n = Number(value);
+  if (!Number.isFinite(n) || !Number.isInteger(n) || n < 0) {
+    throw new TypeError('ProtoEnumValue must be a positive integer (>= 0)');
+  }
+
+  if (!Number.isSafeInteger(n)) {
+    throw new TypeError('ProtoEnumValue must be a safe integer');
+  }
```
capJavert, line 114: Thinking would `z.coerce.int().positive()` just work?

capJavert: I mean replace all your custom checks with zod?

capJavert: basically `toPositiveInt` helper

omBratteng: done

PR #3074, opened 2025-09-08
fix: replication identity of organization table

omBratteng:


src/workers/cdc/primary.ts
```diff
@@ -1283,6 +1283,10 @@ const onOrganizationChange = async (
     return;
   }
 
+  if (!data.payload.before) {
+    return;
+  }
```
omBratteng, line 1288: This is just to handle existing messages that don't have before payload

rebelchris (approved):

PR #3076, opened 2025-09-08
feat: post analytics ads metrics

capJavert:
- refactor boostImpressions and boostReach to be more generic impressionsAds and impressionsReach
- add reachAll column
- add impressionsAds for history
- update MVs to aggregate reach_all on clickhouse side
- add new MVs for skadi impressions to count them as well
- removed separate boost cron since now post_analytics table has the data, so existing crons can fetch it

### DIFF
--- a/clickhouse/migrations/1757338344011_post_analytics_events_ads_mv.up.sql
+++ b/clickhouse/migrations/1757338344011_post_analytics_events_ads_mv.up.sql
@@ -6,7 +6,6 @@ AS
 SELECT
 	c."postId" AS post_id,
 	sum(toUInt64(1)) AS impressions_ads,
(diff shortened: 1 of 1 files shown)


src/graphorm/index.ts
```diff
@@ -1442,6 +1442,22 @@ const obj = new GraphORM({
...
+          return `
+            GREATEST(${alias}.impressions + ${alias}."impressionsAds", 0)
+          `;
+        },
+      },
+      reach: {
+        rawSelect: true,
+        select: (_, alias) => {
+          return `
+            GREATEST(${alias}."reachAll", ${alias}.reach, 0)
```
capJavert, line 1457: fallback in case of being empty

PR #3077, opened 2025-09-09
feat: expose organization links

omBratteng:
- Fixes the opportunity match description key
- Exposes custom, social and press links from organization


src/graphorm/index.ts
```diff
@@ -177,6 +179,15 @@ const createSmartTitleField = ({ field }: { field: string }): GraphORMField => {
...
 };
 
+const organizationLink = (type: OrganizationLinkType) => ({
+  jsonType: true,
+  select: 'links',
+  transform: (
+    links: OrganizationLink[] | null,
+  ): OrganizationLink[] | undefined =>
+    links?.filter((link) => link.type === type),
+});
```
omBratteng, line 189: Alternative is a more complex select that only selects the jsonb objects with the specifc type. But since it's maybe a couple few objects, I think a quick transform with filter is fine.

rebelchris (approved):

PR #3080, opened 2025-09-09
feat: allow estimating reach for boosted posts

capJavert:
- we want to show estimate reach on analytics page while boost is active


capJavert:
cc: @sshanzel

PR #3090, opened 2025-09-10
feat: dedup key for posts

idoshamun:
The new deduplication key helps the feed to well, dedup posts. We already have a solution place for shared post on feed side, but this is a more generic solution that supports also freeform.

For freeform we generate a hash so we can get rid of all spams

BONUS: CLAUDE.md file 😬

### DIFF
--- a/src/workers/postUpdated.ts
+++ b/src/workers/postUpdated.ts
@@ -363,11 +363,14 @@ const updatePost = async ({
   const title = data?.title || databasePost.title;
   data.title = title;
 
+  const metadataChangedAt = data?.metadataChangedAt;
+
   // Apply vordr checks before updating the post
   data = await preparePostForUpdate(data, databasePost, {
     con: entityManager,
     userId: data.authorId || undefined,
   });
+  data.metadataChangedAt = metadataChangedAt;
 
   data.id = databasePost.id;
   data.sourceId = data.sourceId || databasePost.sourceId;


CLAUDE.md
```diff
@@ -0,0 +1,85 @@
+# CLAUDE.md
```
capJavert, line 1: thank you for this, the task to add this is on me but got wrapped up in other s*** 🍻

src/entity/posts/hooks.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (post.title) {
+      const normalizedTitle = normalizeContentForDeduplication(post.title);
+      if (normalizedTitle) {
+        return generateContentHash(normalizedTitle);
+      }
+    }
+  }
+
+  // For other post types, no deduplication key
+  return undefined;
```
rebelchris, line 161: Probably better to early return if not x post types?

src/entity/posts/hooks.ts
```diff
@@ -0,0 +1,183 @@
...
+  post: T,
+): Promise<T> => {
+  const dedupKey = generateDeduplicationKey(post);
+
+  if (dedupKey) {
+    return {
+      ...post,
+      flags: {
+        ...post.flags,
+        dedupKey,
```
rebelchris, line 177: Just adding so far?
They won't get vordr I assume?

src/entity/posts/hooks.ts
```diff
@@ -0,0 +1,183 @@
+import { DataSource, DeepPartial, EntityManager } from 'typeorm';
```
rebelchris, line 1: Generic:
Time for full blown vordr service when...

rebelchris (approved):
No blocks on my end.

src/entity/posts/hooks.ts
```diff
@@ -0,0 +1,183 @@
...
+ */
+export const normalizeContentForDeduplication = (content: string): string => {
+  return removeAllSpecialCharactersForDedup(content.toLowerCase().trim());
+};
+
+/**
+ * Generate SHA-256 hash from normalized content
+ */
+export const generateContentHash = (content: string): string => {
+  const hash = createHash('sha256');
```
capJavert, line 118: seen Viktor recommended sha1, probably is good as well, we used it for hashing code snippets

capJavert (approved):

src/entity/posts/hooks.ts
omBratteng: Think sha256 or newer is better. Usually faster and less chance for collision

PR #3092, opened 2025-09-10
feat: campaign post analytics notification

capJavert:
- [x] add generic temporal workflow to schedule delayed notifications for different entities
- [x] implement notification to send impressions 24 hours after campaign creation
- [x] add new notification type and icons
- [x] add notification settings mapping for new type
- [x] tests

### DIFF
--- a/__tests__/notifications/index.ts
+++ b/__tests__/notifications/index.ts
@@ -372,6 +372,83 @@ describe('generateNotification', () => {
     expect(actual.attachments).toEqual([]);
   });
 
+  it('should generate campaign_post_first_milestone notification', () => {
+    const type = NotificationType.CampaignPostFirstMilestone;
+    const ctx: NotificationCampaignContext = {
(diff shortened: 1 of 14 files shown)


src/workers/notifications/campaignPostAnalyticsNotification.ts
```diff
@@ -0,0 +1,76 @@
+import { CampaignPost } from '../../entity/campaign/CampaignPost';
+import { PostAnalytics } from '../../entity/posts/PostAnalytics';
+import { User } from '../../entity/user/User';
+import type { NotificationPostAnalyticsContext } from '../../notifications';
+import { NotificationType } from '../../notifications/common';
+import { buildPostContext } from './utils';
+import { generateTypedNotificationWorker } from './worker';
+
+const impressionsThreshold = 10_000;
```
capJavert, line 9: we decided we won't send notification if post got less then 10k impressions

PR #3093, opened 2025-09-10
fix: reasoning in opportunity match notification

omBratteng:
- Expand reasoning on message
- Make it work in local environment
- Unique notification per user recipient

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -50,3 +50,33 @@ cmd_button(
     '--skip-preview',
   ],
 )
+
+# Add a button to API to seed the database
+cmd_button(
+  name="db_seed",
+  resource="api",
+  text="Seed Database",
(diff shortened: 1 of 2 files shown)


src/workers/notifications/candidateOpportunityMatchNotification.ts
```diff
@@ -7,10 +7,6 @@ export const candidateOpportunityMatchNotification =
   generateTypedNotificationWorker<'gondul.v1.candidate-opportunity-match'>({
     subscription: 'api.candidate-opportunity-match-notification',
     handler: async (data, _, logger) => {
-      if (process.env.NODE_ENV === 'development') {
-        return;
-      }
-
```
omBratteng, line 13: @rebelchris not sure why this was in the worker?

PR #3094, opened 2025-09-10
feat: show new opportunity

omBratteng:
Store the ID of the last opportunity match in alerts, so we can show the "new opportunity" button in front-end


omBratteng:
@claude review this pl0x as a 1337 h3x0r

PR #3107, opened 2025-09-14
feat: vote on polls

AmarTrebinjac:


src/schema/posts.ts
```diff
@@ -3493,6 +3520,50 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        .createQueryBuilder(PollPost, 'post')
+        .leftJoin(
+          UserPost,
+          'userPost',
+          'userPost.postId = post.id AND userPost.userId = :userId',
+        )
+        .select(['post.id', 'post.endsAt', 'userPost.pollVoteOptionId'])
+        .where('post.id = :postId')
+        .setParameters({ postId: args.postId, userId: ctx.userId })
+        .getRawOne();
```
AmarTrebinjac, line 3543: I am not entirely sure if this is a great way to do this.

I initially tried to fetch the `userPost` with a relation `post`, but seemingly you still have to await it, even though the relation property explicitly says it is used to decide what to load?

Also tried creating FK constraints and unique indexes, but at least the ways I tried it, it always messed with the fact if the `userPost` already existed when you tried to insert the `pollVoteOptionId`, it would error.

Open to better suggestions on this if anyone has any.

src/migration/1757630994202-PollVotes.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            CREATE OR REPLACE FUNCTION user_post_poll_vote_insert_trigger_function()
+            RETURNS TRIGGER AS $$
+            BEGIN
+                IF NEW."pollVoteOptionId" IS NOT NULL THEN
+                    UPDATE poll_option SET "numVotes" = "numVotes" + 1 WHERE id = NEW."pollVoteOptionId";
+                    
+                    UPDATE post SET "numPollVotes" = "numPollVotes" + 1
```
rebelchris, line 14: This is just the generic total then right? so total amount of votes on a post.

src/schema/posts.ts
rebelchris: Can you describe maybe what we try to achieve with these?
Get post + user_state of that post correct?

Maybe we can extract them and do ensourceSourcePermissions and ensurePostVotePermissions?

Then that could yield back the post?

src/migration/1757630994202-PollVotes.ts
rebelchris: Ah no wait this is the individual option count?
Why do you need to weird select, isn't it just pollVoteOptionId?

src/schema/posts.ts
```diff
@@ -1350,6 +1360,23 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      postId: ID!
+      """
+      ID of the option to vote for
+      """
+      optionId: ID!
+      """
+      ID of the source the post belongs to
+      """
+      sourceId: ID
```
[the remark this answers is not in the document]
AmarTrebinjac: @capJavert I assumed that if you cant post, you cant vote either, but I clarified with product, and you should be able to vote on any poll irrespective of that. Will change.

src/schema/posts.ts
AmarTrebinjac: @rebelchris We need the userPost to see if the user has already voted, so would be ideal to fetch the post at the same time, which we need to check if the poll has ended. But if userPost doesnt exist, we need to create it and fetch post separately.

But with what @capJavert said I changed it up to be a little simpler now.

src/schema/posts.ts
```diff
@@ -553,6 +555,10 @@ export const typeDefs = /* GraphQL */ `
     The transaction that was created for the award
     """
     awardTransaction: UserTransactionPublic
+    """
+    If the user voted on the poll, this is the option they voted for
+    """
+    pollVoteOptionId: String
```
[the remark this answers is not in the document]
capJavert: thats how we usually do it in graphql, because if for example you will need title you can then get from frontend easily

src/schema/posts.ts
```diff
@@ -3493,6 +3516,47 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      });
+
+      let post: PollPost | undefined;
+      if (userPost) {
+        post = (await userPost.post) as PollPost;
+      } else {
+        post = await ctx.con.getRepository(PollPost).findOneByOrFail({
+          id: args.postId,
+        });
+      }
```
capJavert, line 3540: why do you need double logic for this, would below work?

```suggestion
      const post: PollPost | null = await userPost.post;
      
      if (!post) {
          throw new NotFoundError()
      }
```

src/schema/posts.ts
```diff
@@ -3493,6 +3516,47 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      if (post.endsAt && isBefore(post.endsAt, new Date())) {
+        throw new ValidationError('Poll has ended');
```
capJavert, line 3543:
```suggestion
        throw new ConflictError('Poll has ended');
```

src/schema/posts.ts
```diff
@@ -3493,6 +3516,47 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      }
+
+      const hasAlreadyVoted = !!userPost?.pollVoteOptionId;
+
+      if (hasAlreadyVoted) {
+        throw new ValidationError('User has already voted on this poll');
```
capJavert, line 3549:
```suggestion
        throw new ConflictError('User has already voted on this poll');
```

src/schema/posts.ts
AmarTrebinjac: @capJavert If the userPost does not exist, we need to still fetch the post

__tests__/posts.ts
```diff
@@ -9271,3 +9273,165 @@ describe('mutate polls', () => {
...
+      client,
+      {
+        mutation: MUTATION,
+        variables: vote,
+      },
+      'GRAPHQL_VALIDATION_FAILED',
+    );
+  });
+
+  it('should remove vote from count if user is deleted', async () => {
```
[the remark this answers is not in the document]
AmarTrebinjac: Keep in mind this is for users being deleted. I don't think anyone's going to be keeping a close enough eye to notice a vote or two drop off over the course of time.

Besides, we do the same with upvotes and other post interactions, so this is in line with how we handle everything else

src/schema/posts.ts
capJavert: why not make it other way around, fetch the post and userPost as relation, then its a single line?

capJavert: ignore me, ok 👍

src/migration/1757630994202-PollVotes.ts
```diff
@@ -0,0 +1,96 @@
...
+            AFTER DELETE ON user_post
+            FOR EACH ROW
+            EXECUTE FUNCTION user_post_poll_vote_delete_trigger_function();
+        `);
+
+    await queryRunner.query(`
+            CREATE OR REPLACE FUNCTION user_post_poll_vote_update_trigger_function()
+            RETURNS TRIGGER AS $$
+            BEGIN
+                IF OLD."pollVoteOptionId" IS NULL AND NEW."pollVoteOptionId" IS NOT NULL THEN
```
capJavert, line 57: should handle the decrement case as well, if `pollVoteOptionId` becomes null?

src/schema/posts.ts
```diff
@@ -3493,6 +3517,47 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+          userId: ctx.userId,
+        },
+        relations: ['post'],
+      });
+
+      let post: PollPost | undefined;
+      if (userPost) {
+        post = (await userPost.post) as PollPost;
+      } else {
+        post = await ctx.con.getRepository(PollPost).findOneByOrFail({
```
capJavert, line 3538: I know that you removed sourceId from params, but still now that we have post should we not do `ensureSourcePermission` to avoid user answering polls they don't have access to?

src/migration/1757630994202-PollVotes.ts
capJavert: yeah but good to have fallbacks, we will easily forget trigger needs updating so ideally would like to add it now?

src/schema/posts.ts
AmarTrebinjac: We can add the check back if you feel necessary. However, I asked about "permissions" in this [thread](https://dailydotdev.slack.com/archives/C07GYGWLLKS/p1757936955009659) and it doesnt seem like we care if they can post or not in the squad.

I can add it back for access control if you want, but realistically they should never see the poll in the first place

PR #3118, opened 2025-09-16
refactor: rename `type` to `opportunityType` on `Opportunity`

omBratteng:
Encountered some seed issues when column was named `type` :lolsob:, so renaming type to be more specific on `Opportunity` and `OpportunityUser` table


omBratteng:
Might actually not need it, I solved the seed by using the proper entity :fml:
770208e9e79a19feba3a6b7495d5cb5f55eef19b

PR #3123, opened 2025-09-17
feat: return candidate keywords

omBratteng:
- Return candidate keywords in `getCandidatePreferences` query
- Add mutation to add keyword to candidate
- Add mutation to remove keyword from candidate

### Jira ticket
AS-1210


omBratteng:
@claude re-review, added keywords to getCandidatePreferences test

PR #3124, opened 2025-09-17
feat: poll worker

AmarTrebinjac:
Workers for the poll result notifications. It doesn't seem like we can effectively change the "title" in a notification based on data with the way we generate it now. Since the author and voters receive different notification titles, it seems like we have to make two, unless I've missed something.

There's also supposed to be email templates, but I haven't received those yet, so I'll add them in a separate task.

### Jira ticket
MI-1012

### DIFF
--- a/__tests__/notifications/index.ts
+++ b/__tests__/notifications/index.ts
@@ -2065,31 +2065,4 @@ describe('poll result notifications', () => {
       '<b>Your poll has ended!</b> Check the results for: <b>What is your favorite programming language?</b>',
     );
   });
-
-  it('should handle poll with null title', async () => {
-    const pollPost: ChangeObject<PollPost> = {
-      id: 'poll2',
-      shortId: 'sp2',
-      title: null,
-      type: PostType.Poll,
-      sourceId: 'a',
-      createdAt: 0,
-      endsAt: null,
-    };
-
-    const type = NotificationType.PollResult;
-    const ctx: NotificationPostContext = {
-      userIds: ['1'],
-      source: sourcesFixture.find(
-        (item) => item.id === 'a',
-      ) as Reference<Source>,
-      post: pollPost,
-    };
-
-    const actual = generateNotificationV2(type, ctx);
-    expect(actual.notification.title).toEqual(
-      '<b>Poll you voted on has ended!</b> See the results for: <b>null</b>',
-    );
-    expect(actual.attachments).toEqual([]);
-  });
 });


src/notifications/generate.ts
```diff
@@ -552,4 +556,21 @@ export const generateNotificationMap: Record<
...
+  poll_result_author: (
+    builder: NotificationBuilder,
+    ctx: NotificationPostContext,
+  ) =>
+    builder
+      .icon(NotificationIcon.Bell)
+      .objectPost(ctx.post, ctx.source, ctx.sharedPost!)
+      .avatarSource(ctx.source)
+      .targetUrl(`${process.env.COMMENTS_PREFIX}/posts/${ctx.post.id}`)
+      .uniqueKey(`${ctx.post.id}-poll-result`),
```
rebelchris, line 575: Same here.

src/common/typedPubsub.ts
```diff
@@ -177,6 +177,7 @@ export type PubSubSchema = {
   'gondul.v1.candidate-opportunity-match': MatchedCandidate;
   'api.v1.candidate-preference-updated': CandidatePreferenceUpdated;
   'api.v1.delayed-notification-reminder': z.infer<typeof entityReminderSchema>;
+  'api.v1.delayed-poll-result': z.infer<typeof entityReminderSchema>;
```
capJavert, line 180: think you don't need this

src/notifications/generate.ts
```diff
@@ -552,4 +556,21 @@ export const generateNotificationMap: Record<
       )
       .uniqueKey(`${ctx.post.id}-metrics-${new Date().toISOString()}`);
   },
+  poll_result: (builder: NotificationBuilder, ctx: NotificationPostContext) =>
+    builder
+      .icon(NotificationIcon.Bell)
+      .objectPost(ctx.post, ctx.source, ctx.sharedPost!)
+      .avatarSource(ctx.source)
+      .targetUrl(`${process.env.COMMENTS_PREFIX}/posts/${ctx.post.id}`)
+      .uniqueKey(`${ctx.post.id}-poll-result`),
```
capJavert, line 565: you can use `referencePost` which will put `referenceId` to postId and then you don't need custom uniqueKey per this notification type

src/notifications/generate.ts
capJavert, line 575: same as above

src/workers/cdc/primary.ts
```diff
@@ -574,6 +590,18 @@ const onPostChange = async (
...
 
+    if (
+      data.payload.after!.type === PostType.Poll &&
+      data.payload.after!.deleted
+    ) {
+      cancelEntityReminderWorkflow({
+        entityId: data.payload.after!.id,
+        entityTableName: getTableName(con, PollPost),
+        scheduledAtMs: 0,
+        delayMs: 0,
```
capJavert, line 601: you will have to calculate same `delayMs` as for `runEntityReminderWorkflow` since `delayMs` is part of the job id so only then it will invalidate correct job

src/workers/notifications/pollResult.ts
```diff
@@ -0,0 +1,92 @@
...
+      return [
+        {
+          type: NotificationType.PollResult,
+          ctx: notificationCtx,
+        },
+      ];
+    },
+  };
+
+export const pollResultAuthorNotification: TypedNotificationWorker<'api.v1.delayed-notification-reminder'> =
```
capJavert, line 56: let's keep it one worker per file as we do usually 🙏

__tests__/notifications.ts
```diff
@@ -1465,3 +1483,173 @@ describe('streamNotificationUsers', () => {
...
+          entityTableName: getTableName(con, PollPost),
+          scheduledAtMs: Date.now(),
+          delayMs: 1000,
+        },
+      );
+
+    expect(result).toBeUndefined();
+  });
+
+  it('should cancel entity reminder workflow when poll is deleted', async () => {
```
capJavert, line 1633: cdc tests got to `__tests__/workers/cdc/primary`, and also if you can add a test for `runEntityReminderWorkflow`, you can check how I did it for campaign

src/workers/notifications/pollResult.ts
```diff
@@ -0,0 +1,92 @@
...
+        where: {
+          postId: poll.id,
+          pollVoteOptionId: Not(IsNull()),
+          userId: Not(poll.authorId!),
+        },
+        select: ['userId'],
+      });
+
+      const postCtx = await buildPostContext(con, poll.id);
+      if (!postCtx || !userPosts.length) {
```
capJavert, line 38: you can even early return above immediately if userPosts length is 0, to save buildPostContext call

src/workers/notifications/pollResult.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      if (!poll) {
+        return;
+      }
+
+      const userPosts = await con.getRepository(UserPost).find({
+        where: {
+          postId: poll.id,
+          pollVoteOptionId: Not(IsNull()),
+          userId: Not(poll.authorId!),
```
capJavert, line 32: instead of `poll.authorId` you should check above `if (!poll.authorId)` and return early if no author, want to avoid `!` since they break type safety if not really needed

src/notifications/generate.ts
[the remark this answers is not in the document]
capJavert: yeah just commented in my feedback

src/common/typedPubsub.ts
AmarTrebinjac: Youre right! Added this and forgot about it 😅

src/workers/notifications/pollResult.ts
AmarTrebinjac: Right, I guess poll author is technically not optional anymore?

capJavert: it is optional, that is why I recommend adding if check.

Because like this, if `poll.authorId` was null, query would be `WHERE userId IS NOT NULL`, which would then return random row 😆

src/workers/cdc/primary.ts
```diff
@@ -574,6 +591,26 @@ const onPostChange = async (
...
+    ) {
+      const poll = data as ChangeMessage<PollPost>;
+      const after = poll.payload.after!;
+
+      const createdDate = debeziumTimeToDate(after.createdAt).getTime();
+      const notificationTime =
+        after!.endsAt?.getTime() ||
+        new Date(createdDate + 14 * 24 * 60 * 60 * 1000).getTime();
+
+      cancelEntityReminderWorkflow({
```
capJavert, line 606: missed it in review but `await` is missing here

src/workers/notifications/pollResultNotification.ts
```diff
@@ -0,0 +1,58 @@
...
+        userIds: userPosts.map((up) => up.userId),
+      };
+
+      return [
+        {
+          type: NotificationType.PollResult,
+          ctx: notificationCtx,
+        },
+      ];
+    },
```
capJavert, line 57: ahh also @rebelchris typenotification workers need parse message no due to temp fix?
```

parseMessage(message) {
  return messageToJson(message);
},
```

PR #3131, opened 2025-09-18
chore: remove popular feed client

capJavert:
Unused feed client, its all config based now, also removed from feed side in https://github.com/dailydotdev/daily-feed/pull/696

### DIFF
--- a/.infra/Pulumi.prod.yaml
+++ b/.infra/Pulumi.prod.yaml
@@ -95,7 +95,6 @@ config:
     personalizedDigestFeed: http://feed-digest-server.feed/api/personalised
     personalizedDigestSecret:
       secure: AAABAN1TYaqznVMVbKDZtYEnLGB0VIzXDrpcmB6rBsImneobrDOS12T4XwbVvnexdZt1NBCnZRBwH4AL7OgpZkj/2UgT8Lzj0o7TdVSBeIeVDzKa/V6jlvJ4AskEC9rg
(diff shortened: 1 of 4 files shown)


rebelchris (approved):
:feels-good:

.infra/Pulumi.prod.yaml
```diff
@@ -95,7 +95,6 @@ config:
     personalizedDigestFeed: http://feed-digest-server.feed/api/personalised
     personalizedDigestSecret:
       secure: AAABAN1TYaqznVMVbKDZtYEnLGB0VIzXDrpcmB6rBsImneobrDOS12T4XwbVvnexdZt1NBCnZRBwH4AL7OgpZkj/2UgT8Lzj0o7TdVSBeIeVDzKa/V6jlvJ4AskEC9rg
-    popularFeed: http://feed.feed.svc.cluster.local/api/popular
```
idoshamun, line 98: We're still using this endpoint though. Check the anonymous feed query, it uses the `feedGenerators['popular']` which refer to this endpoint

idoshamun (approved):
Are you sure we don't use it?

.infra/Pulumi.prod.yaml
capJavert: I search and I don't see any reference to `env.POPULAR_FEED` in codebase aside from what I removed

PR #3135, opened 2025-09-19
fix: poll worker

AmarTrebinjac:
I think the workers were causing issues because they were missing the `parsedMessage` prop. I thought this was not needed since it was a typed worker, but I think I misunderstood the TODO comment 😅

### DIFF
--- a/src/workers/cdc/primary.ts
+++ b/src/workers/cdc/primary.ts
@@ -154,6 +154,7 @@ import {
 import { Opportunity } from '../../entity/opportunities/Opportunity';
 import { notifyJobOpportunity } from '../../common/opportunity/pubsub';
 import { UserCandidatePreference } from '../../entity/user/UserCandidatePreference';
+import { PollPost } from '../../entity/posts/PollPost';
 
 const isFreeformPostLongEnough = (
   freeform: ChangeMessage<FreeformPost>,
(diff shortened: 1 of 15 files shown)


src/workers/notifications/pollResultNotification.ts
```diff
@@ -0,0 +1,62 @@
...
+      };
+
+      return [
+        {
+          type: NotificationType.PollResult,
+          ctx: notificationCtx,
+        },
+      ];
+    },
+    parseMessage(message) {
```
AmarTrebinjac, line 58: Added this

src/workers/cdc/primary.ts
```diff
@@ -574,6 +591,26 @@ const onPostChange = async (
...
+    ) {
+      const poll = data as ChangeMessage<PollPost>;
+      const after = poll.payload.after!;
+
+      const createdDate = debeziumTimeToDate(after.createdAt).getTime();
+      const notificationTime =
+        after!.endsAt?.getTime() ||
+        new Date(createdDate + 14 * 24 * 60 * 60 * 1000).getTime();
+
+      await cancelEntityReminderWorkflow({
```
AmarTrebinjac, line 606: added missing await

capJavert:
Looks good, will do some manual checks.

PR #3140, opened 2025-09-19
feat: poll result topics

AmarTrebinjac:


omBratteng:
ah fml, test fails because it needs 1:1 map between workers and subscriptions

omBratteng (approved):

PR #3154, opened 2025-09-22
feat: add support for multiple squad posting and notification dedup

ilasw:
### Jira ticket
MI-1043


src/common/post.ts
```diff
@@ -1104,3 +1222,53 @@ export const ensurePostAnalyticsPermissions = async ({
     );
   }
 };
+
+export const createFreeformPost = async (
+  ctx: AuthContext,
+  args: CreatePostArgs,
+) => {
```
ilasw, line 1229: This is the old `createFreeformPost` function plus the checks that the endpoint was doing before calling it

src/schema/sources.ts
```diff
@@ -1113,7 +1111,6 @@ export const isPrivilegedMember = async (
 type PostPermissions = SourcePermissions.Post | SourcePermissions.PostRequest;
 
 export const canPostToSquad = (
-  ctx: Context,
```
ilasw, line 1116: This was not used by the function

src/schema/posts.ts
```diff
@@ -1437,6 +1458,44 @@ export const typeDefs = /* GraphQL */ `
       duration: Int
     ): SourcePostModeration! @auth @rateLimit(limit: 1, duration: 30)
 
+    """
+    Create multiple source posts
+    """
+    createMultipleSourcePosts(
```
AmarTrebinjac, line 1464: I would suggest renaming this to `createMultiSourcePost`. Current name feels a little ambiguous. To me it reads as "create multiple posts in a source" 😅

src/schema/posts.ts
```diff
@@ -3585,6 +3603,107 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+        );
+      }
+
+      const { sourceIds, sharedPostId, ...postArgs } = args;
+      const isSharingPost = !!sharedPostId;
+      const isMultiplePosting = sourceIds.length > 1;
+
+      await ensurePostRateLimit(ctx.con, ctx.userId);
+
+      return await Promise.all(
```
AmarTrebinjac, line 3632: Doesn't `Promise.all` fail all if one fails? I think we should process these synchronously and handle the errors if there are any

src/schema/posts.ts
```diff
@@ -3585,6 +3603,107 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+            sourceId,
+            sharedPostId,
+            type: sharedPostId ? PostType.Share : PostType.Freeform,
+          });
+          const { id } = await createSourcePostModeration({
+            ctx,
+            args: pendingPost,
+            options: { isMultiplePosting },
+          });
+          return {
```
AmarTrebinjac, line 3699: Can we not just change the type based on the condition, and just return once at the end?

AmarTrebinjac (commented):
Looking good, man! 😄 

Just a few comments from me! 🙂

src/common/post.ts
```diff
@@ -1104,3 +1222,53 @@ export const ensurePostAnalyticsPermissions = async ({
...
+  if (!title) {
+    throw new ValidationError('Title can not be an empty string!');
+  }
+
+  if (sourceId === userId) {
+    await ensureUserSourceExists(userId, con);
+  }
+
+  await Promise.all([
+    ensureSourcePermissions(ctx, sourceId, SourcePermissions.Post),
```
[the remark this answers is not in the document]
ilasw: Nice point, will check if is good removing it 👍

src/notifications/common.ts
```diff
@@ -481,18 +481,22 @@ export const generateUserNotificationUniqueKey = ({
...
   const uniqueKey = notificationTypeToUniqueKey[type];
 
   if (!uniqueKey) {
     return null;
   }
 
-  return [uniqueKey, referenceId, referenceType].filter(Boolean).join(':');
+  return [uniqueKey, referenceId, referenceType, dedupKey]
+    .filter(Boolean)
+    .join(':');
```
capJavert, line 499: since you keep referenceId, that would still be different post ids no and then it would generate multiple notifications? 

Best to add notification specific tests and verify it works

src/schema/posts.ts
```diff
@@ -361,8 +367,23 @@ const builderForCampaignV1 = (builder: QueryBuilder<CampaignPost>) =>
       'post',
     );
 
+enum MultiplePostItemType {
+  Post = 'post',
+  ModerationItem = 'moderationItem',
+}
```
capJavert, line 373: Let's move this to `types.ts` in root so its reused easily

src/schema/posts.ts
capJavert: it should all be in a transaction and I agree, process sync what we can and I would fail the whole mutation if any of the inserts fail

ilasw: Initially, I added this logic in a transaction, but many shared functions need refactoring since they accept `Context` as a parameter. We should replace ctx with a connection/entity manager so we can switch from using `ctx.con` to `queryRunner.manager` in order to support error rollbacks.

I can implement this way (it won’t take too long), but it will add a lot of changes to this PR. Is that okay? @capJavert

src/schema/posts.ts
```diff
@@ -3585,6 +3603,107 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      }>
+    > => {
+      const { error: inputErrors, data: args } =
+        multipleSourcePostArgsSchema.safeParse(input);
+
+      if (inputErrors) {
+        throw new ValidationError(
+          `Error '${inputErrors.issues[0].path}': ${inputErrors.issues[0].message}`,
+        );
+      }
```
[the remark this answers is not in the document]
ilasw: Nice, thanks for suggesting! Just updated ✔️

src/entity/posts/utils.ts
```diff
@@ -471,6 +471,7 @@ export const createSharePost = async ({
         private: privacy,
         visible,
       },
+      type: PostType.Share,
```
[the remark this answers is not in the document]
capJavert: Post type is part of dedup key?

src/schema/posts.ts
ilasw: Renamed to `createPostInMultipleSources`

src/common/post.ts
```diff
@@ -373,13 +378,90 @@ export interface CreateSourcePostModeration
...
+
+  return await queryReadReplica(con, async ({ queryRunner }) => {
+    const pendingQb = queryRunner.manager
+      .getRepository(SourcePostModeration)
+      .createQueryBuilder('p')
+      .where({
+        status: SourcePostModerationStatus.Pending,
+      })
+      .andWhere(`p.flags->> 'dedupKey' = :dedupKey`, { dedupKey })
+      .limit(1);
```
capJavert, line 401: can remove this and use exists() below, its more efficient

src/common/post.ts
```diff
@@ -373,13 +378,90 @@ export interface CreateSourcePostModeration
...
+    const postQb = queryRunner.manager
+      .getRepository(Post)
+      .createQueryBuilder('p')
+      .where(`p.flags->> 'dedupKey' = :dedupKey`, { dedupKey })
+      .limit(1);
```
capJavert, line 406: same comment as above

src/common/post.ts
```diff
@@ -446,9 +543,136 @@ export interface CreatePollPostProps
...
+              numVotes: 0,
+              order: option.order,
+              postId: id!,
+            }),
+          ),
+        },
+      });
+    }
+    default: {
+      return await createFreeformPost(
```
capJavert, line 664: should this default to freeform, I think safer if its case freeform, and default is throw because its something we don't expect?

src/common/post.ts
```diff
@@ -1104,3 +1328,46 @@ export const ensurePostAnalyticsPermissions = async ({
...
+  const { sourceId, image } = args;
+  const { con, userId } = ctx;
+  const id = await generateShortId();
+  const { title, content } = validatePost(args);
+
+  if (!title) {
+    throw new ValidationError('Title can not be an empty string!');
+  }
+
+  await (options?.entityManager || con).transaction(async (manager) => {
```
capJavert, line 1346: would it not mean that if entityManager is passed its already in transaction? Can we simplify it so it only accepts EntityManager and uses that whatever it is?

src/notifications/common.ts
```diff
@@ -481,18 +481,26 @@ export const generateUserNotificationUniqueKey = ({
...
 
   if (!uniqueKey) {
     return null;
   }
 
-  return [uniqueKey, referenceId, referenceType].filter(Boolean).join(':');
+  return [
+    uniqueKey,
+    dedupKey ? `dedup_${dedupKey}` : referenceId,
+    referenceType,
```
capJavert, line 500: yeah think this works now, basically (correct me if I am wrong):
- if dedup key is present we send only single notification for that post
- if no dedup key we dedup per postId (as before)

src/notifications/common.ts
capJavert: still missing post notification tests for dedup logic?

capJavert (commented):
Feels almost there, just please address final comments and I think we still miss tests for notification dedup part.

src/notifications/common.ts
ilasw: I've added this case https://github.com/dailydotdev/daily-api/pull/3154/files#diff-6e46397f1e2cc9f81f6d251217c5f625accc5596fbd5e11dcdb644922166ddb9R3436 , do you think we need to add more?

src/common/post.ts
ilasw: Updated ✔️

src/common/post.ts
ilasw: Updated ✔️

src/common/post.ts
```diff
@@ -373,13 +378,88 @@ export interface CreateSourcePostModeration
...
+
+  const isDuplicatedInSameSquad = await hasDuplicatedPostBy(con, {
+    dedupKey,
+    sourceId,
+  });
+  if (isDuplicatedInSameSquad) {
+    return WarningReason.DuplicatedInSameSquad;
+  }
+
+  const isDuplicatedAcrossSquads = await hasDuplicatedPostBy(con, {
```
capJavert, line 447: why do we call `hasDuplicatedPostBy` twice, once with sourceId once without? Would it not do the same thing if sourceId is not passed thus doing the query twice?

capJavert (approved):

rebelchris (approved):
Non blocking, but good to check that one by Ante.
Logic wise will approve from my side.

src/schema/posts.ts
```diff
@@ -3271,6 +3308,81 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      const hasExternalLink = !!postArgs.externalLink;
+      if (hasExternalLink) {
+        const { id } = await getPostIdFromUrlOrCreateOne(ctx, postArgs);
+        postArgs.sharedPostId = id;
+      }
+
+      return await ctx.con.transaction(async (entityManager) => {
+        const output: Array<MultipleSourcesPostResult> = [];
+        for (const sourceId of sourceIds) {
```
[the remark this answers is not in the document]
capJavert: works ok yes

src/common/post.ts
```diff
@@ -373,13 +383,93 @@ export interface CreateSourcePostModeration
...
+}): Promise<WarningReason | undefined> => {
+  if (isMultiplePosting) {
+    return WarningReason.MultipleSquadPost;
+  }
+
+  if (!dedupKey) {
+    return;
+  }
+
+  const isDuplicatedInSameSquad = await hasDuplicatedPostBy(con, {
```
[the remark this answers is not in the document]
capJavert: also weird on this https://github.com/dailydotdev/daily-api/pull/3154#discussion_r2401283518

src/common/post.ts
```diff
@@ -373,13 +378,88 @@ export interface CreateSourcePostModeration
...
+  }
+
+  const isDuplicatedAcrossSquads = await hasDuplicatedPostBy(con, {
+    dedupKey,
+  });
+  if (isDuplicatedAcrossSquads) {
+    return WarningReason.MultipleSquadPost;
+  }
+
+  return;
```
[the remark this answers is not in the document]
ilasw: Removed

PR #3158, opened 2025-09-24
feat: recruiter accepted event

rebelchris:
Recruiter accepted event.

Had to remove state tests as they now can trigger on both making them unreliable/wrong.

### DIFF
--- a/src/common/typedPubsub.ts
+++ b/src/common/typedPubsub.ts
@@ -33,6 +33,7 @@ import {
   ContentUpdatedMessage,
   MatchedCandidate,
   type OpportunityMessage,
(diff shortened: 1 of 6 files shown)


src/common/typedPubsub.ts
```diff
@@ -221,6 +222,7 @@ export type PubSubSchema = {
   'api.v1.member-joined-source': {
     sourceMember: ChangeObject<SourceMember>;
   };
+  'api.v1.recruiter-accepted-candidate-match': RecruiterAcceptedCandidateMatchMessage;
```
omBratteng, line 225: 💯

nensidosari (approved):

PR #3160, opened 2025-09-24
fix: add resolver for endsAt

AmarTrebinjac:
Missing resolver made the history page explode

Screenshot 2025-09-24 at 16 19 40

### Jira ticket
MI-1047

### DIFF
--- a/src/schema/posts.ts
+++ b/src/schema/posts.ts
@@ -3638,6 +3638,10 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       }
       return undefined;
     },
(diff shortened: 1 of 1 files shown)


src/schema/posts.ts
```diff
@@ -3638,6 +3638,10 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       }
       return undefined;
     },
+    endsAt: (post: GQLPost): Date | null => {
+      if (!post.endsAt) return null;
+      return new Date(post.endsAt);
+    },
```
capJavert, line 3644: we have `transformDate` helper for this

PR #3164, opened 2025-09-25
feat: edit job questions

capJavert:


nensidosari (approved):
Coming from polls, just a question to make sure we're doing it right, as I haven't been following recruiter that closely. 

When we save the answer for the user do we save the question as well? So when question is edited, it won't really be edited for users that have already answered?

PR looks good otherwise

PR #3169, opened 2025-09-26
feat: store application scoring

rebelchris:
Storing the application scoring

### DIFF
--- a/__tests__/workers/opportunity/storeCandidateApplicationScore.ts
+++ b/__tests__/workers/opportunity/storeCandidateApplicationScore.ts
@@ -26,15 +26,7 @@ describe('storeCandidateApplicationScore worker', () => {
     await saveFixtures(con, User, usersFixture);
     await saveFixtures(con, Opportunity, opportunitiesFixture);
   });
-
-  it('should be registered', () => {
-    const registeredWorker = typedWorkers.find(
-      (item) => item.subscription === worker.subscription,
-    );
-
-    expect(registeredWorker).toBeDefined();
-  });
-
+  
   it('should successfully store candidate application score', async () => {
     const applicationScoreData = new ApplicationScored({
       userId: '1',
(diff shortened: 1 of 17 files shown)


__tests__/workers/opportunity/storeCandidateApplicationScore.ts
```diff
@@ -0,0 +1,245 @@
...
+    await saveFixtures(con, Opportunity, opportunitiesFixture);
+  });
+
+  it('should be registered', () => {
+    const registeredWorker = typedWorkers.find(
+      (item) => item.subscription === worker.subscription,
+    );
+
+    expect(registeredWorker).toBeDefined();
+  });
```
omBratteng, line 36: We technically don't need these tests anymore, since we have a global test for workers.
https://github.com/dailydotdev/daily-api/blob/6ec435845c3a30017ca2f06fe09c390f84bb3ed5/__tests__/workers/workers.ts

src/workers/opportunity/storeCandidateApplicationScore.ts
```diff
@@ -0,0 +1,53 @@
...
+
+          return;
+        }
+
+        throw err;
+      }
+    },
+    parseMessage: (message) => {
+      return ApplicationScored.fromBinary(message.data);
+    },
```
omBratteng, line 52:
```suggestion
    parseMessage: (message) => ApplicationScored.fromBinary(message.data),
```

omBratteng: Also maybe we can start putting this further up? That way it's easier to see and not forget about it :lolsob:

omBratteng (approved):
Non-blocking comments

src/workers/opportunity/storeCandidateApplicationScore.ts
```diff
@@ -0,0 +1,51 @@
...
+    subscription: 'api.store-candidate-application-score',
+    handler: async ({ data }, con, logger): Promise<void> => {
+      try {
+        const { userId, opportunityId, score, description } = data;
+        if (!userId || !opportunityId) {
+          logger.warn(
+            { data },
+            'Missing userId or opportunityId in candidate application score',
+          );
+          return;
```
capJavert, line 18: if its expected think we just return, no need to log anything

src/workers/opportunity/storeCandidateApplicationScore.ts
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        const applicationRank = applicationScoreSchema.parse({
+          score,
+          description,
+        });
+
+        await con.transaction(async (manager) => {
+          await manager.getRepository(OpportunityMatch).upsert(
```
capJavert, line 27: since its a single call no need for transaction?

src/workers/opportunity/storeCandidateApplicationScore.ts
```diff
@@ -0,0 +1,51 @@
...
+            },
+          );
+        });
+      } catch (originalError) {
+        const err = originalError as TypeORMQueryFailedError;
+        if (err?.name === 'QueryFailedError') {
+          logger.error({ err, data }, 'could not update application score');
+
+          return;
+        }
```
capJavert, line 45: why would this not throw?

src/workers/opportunity/storeCandidateApplicationScore.ts
capJavert: image

src/workers/opportunity/storeCandidateApplicationScore.ts
rebelchris: It's just copy-paste of the other workers.
Better to log and catch no?

src/workers/opportunity/storeCandidateApplicationScore.ts
capJavert: if it should never happen then we should throw to know if it does happen

src/workers/opportunity/storeCandidateApplicationScore.ts
capJavert: which other worker? Usually if query fails we want to retry or inspect why it failed

capJavert: since sole purpose of worker is to update application score

capJavert: oh I see `storeCandidateOpportunityMatch`, honestly there I would also just throw

rebelchris: The other one I did haha,  removed it from both now.

PR #3181, opened 2025-10-02
fix: only publish one topic for now

rebelchris:
For now we should only use the new topic.
We'll keep the isUpdate call, eventually we might want to use it.


src/common/opportunity/pubsub.ts
```diff
@@ -264,18 +264,13 @@ export const notifyCandidateOpportunityMatchRejected = async ({
 export const notifyJobOpportunity = async ({
   con,
   logger,
-  isUpdate,
```
omBratteng, line 267: problably need to remove it from the caller too?

PR #3190, opened 2025-10-03
fix: using right imageUrl and commentary when sharing new external url

ilasw:
When sharing new/existing URLs use _commentary_ as shared post title


rebelchris (approved):

sshanzel (commented):
Can you write tests for this? That they are reflected on the right props accordingly?

Also, is it possible that you might have to update the way we save it through moderation? Or moderation have it right?

src/common/post.ts
```diff
@@ -588,7 +590,7 @@ export const getMultipleSourcesPostType = (
     return PostType.Poll;
   }
 
-  if (args.sharedPostId) {
+  if (args.sharedPostId || args.externalLink) {
     return PostType.Share;
   }
```
ilasw, line 595: Want to treat the same if sharing new/existing URLs

capJavert (approved):

PR #3191, opened 2025-10-03
fix: duped notification for streak restore

sshanzel:
# TLDR;
The reference for the notification's uniqueness was not a stable value.

# Duplicated notification
The reason why there are two notifications is not because the streak is getting cleared twice.

The notification worker listens to the streak changes update. Those updates are regardless of whether the streak went to 0, or read a new post and updated the `lastViewAt`. It is just that the worker validates if the cache was set to know that it is the right event. There is nothing wrong with that if we correctly set the right reference for the notification's uniqueness.

We used the `lastViewAt,` which was not a stable reference due to certain scenarios:

- The `lastViewAt` is 2 days ago.
- Streak gets cleared.
- CDC triggered.
  - Sets cache.
  - Sends pubsub event to the topic.
- Subscriber checks cache, restore is valid, we send notification.
- The notification was created and was referencing the **day** of the `lastViewAt` for uniqueness.
- However, the user read something. The `lastViewAt` gets updated.
- CDC is triggered.
  - No setting of cache to happen, but the cache is still fresh from the last setting.
  - Sends pubsub event to the topic.
- Subscriber says valid since cache is still there (which is true, user can read today then restore as long as it is valid).
- Here, it should not be able to generate since we reference the day for restoration through `lastViewAt` - **worker should just skip** due to uniqueness constraint.
- But the `lastViewAt` was just updated, and with a different day value now.
- New notification is created.

The idea to use a reference date was correct, but it was not realized that the CDC can be triggered again with a different value from the original reference.

## Solution
We will set the uniqueness of the generated notification on the actual expiry of the cache - there is only one restore expiry date whenever the streak is broken, regardless of how many times this gets run, the notification won't be repeated due to the uniqueness constraint. This can also help with debugging too.

Now that we can get this out of the way, knowing there was no issue with streak possibly being cleared twice, we can further check why the modal is not opening.

### DIFF
--- a/src/workers/notifications/userStreakResetNotification.ts
+++ b/src/workers/notifications/userStreakResetNotification.ts
@@ -45,7 +45,12 @@ const worker: TypedNotificationWorker<'api.v1.user-streak-updated'> = {
       return;
     }
 
-    if (settings?.optOutReadingStreak || !lastStreak || !isNumber(lastStreak)) {
+    if (
+      settings?.optOutReadingStreak ||
+      !lastStreak ||
+      !redisExpiryTime ||
+      !isNumber(lastStreak)
+    ) {
       return;
     }
 


src/workers/notifications/userStreakResetNotification.ts
```diff
@@ -48,12 +49,14 @@ const worker: TypedNotificationWorker<'api.v1.user-streak-updated'> = {
       return;
     }
 
-    const ctx: NotificationStreakContext = {
+    const restorationValidUntil = redisExpiryTime * 1000; // unix timestamp in ms
```
capJavert, line 52: `redisExpiryTime` can be null or empty I suppose?

capJavert: Can we handle it somehow and add test case on what happens when it is

capJavert: let's just add a check if its not in the future we early return as well

sshanzel: Cool. I'll add it in the conditions.

PR #3192, opened 2025-10-03
refactor: ignore campaigns that are not found

sshanzel:
I think we can just ignore these campaigns that are not found. We really have no reference as what kind they are anyway.

### DIFF
--- a/src/workers/notifications/campaignUpdatedAction.ts
+++ b/src/workers/notifications/campaignUpdatedAction.ts
@@ -14,7 +14,6 @@ import {
 } from '../../common/campaign/common';
 import { CampaignType } from '../../entity/campaign/Campaign';
 import { type DataSource } from 'typeorm';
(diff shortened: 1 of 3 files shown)


src/workers/notifications/campaignUpdatedAction.ts
```diff
@@ -44,9 +43,6 @@ const worker: TypedNotificationWorker<'skadi.v2.campaign-updated'> = {
       const err = originalError as TypeORMQueryFailedError;
 
       if (err?.name === 'EntityNotFoundError') {
-        // some campaigns do not exist on API so just warn in case we need to check later
-        logger.warn({ err, params }, 'could not find campaign');
-
```
capJavert, line 49: Why are we removing the logs?

The issue Chris reported was that the worker is retrying and logs have nothing to do with it.

PR #3195, opened 2025-10-06
feat: expose impressions analytics under post query

capJavert:
We will show them instead of views

image

### DIFF
--- a/src/graphorm/index.ts
+++ b/src/graphorm/index.ts
@@ -1427,7 +1427,6 @@ const obj = new GraphORM({
       },
     },
   },
-
   PostAnalytics: {
     requiredColumns: ['id', 'updatedAt'],
     fields: {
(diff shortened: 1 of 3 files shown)


src/graphorm/index.ts
```diff
@@ -1497,6 +1496,19 @@ const obj = new GraphORM({
...
     },
   },
+  PostAnalyticsPublic: {
+    from: 'PostAnalytics',
+    fields: {
+      impressions: {
+        rawSelect: true,
+        select: (_, alias) => {
+          return `
+            GREATEST(${alias}.impressions + ${alias}."impressionsAds", 0)
```
omBratteng, line 1506:
```suggestion
            GREATEST(${alias}."impressions" + ${alias}."impressionsAds", 0)
```
just to be consistent

capJavert: ah missed the comment, its ok since lowercase but yes generally ok to put `"` everywhere

PR #3197, opened 2025-10-06
feat: add temporary `cvParsedMarkdown` table to store extracted markdown

omBratteng:
A temporary column until CV parsing is fully in place


src/entity/user/UserCandidatePreference.ts
```diff
@@ -126,4 +126,7 @@ export class UserCandidatePreference {
     foreignKeyConstraintName: 'FK_user_candidate_preference_user_id',
   })
   user: Promise<User>;
+
+  @Column({ type: 'text', default: null })
```
rebelchris, line 130: Ah i thought we would just store this in the parsedJson as a string?
Eitherway not too worried about it, guess this is easier cleanup

rebelchris (approved):

src/entity/user/UserCandidatePreference.ts
omBratteng: cleaner to keep it separate :)

PR #3202, opened 2025-10-07
feat: private create opp

rebelchris:
Private endpoint to create opportunities.
It's needed to inject keywords without having to call manually many databased on Retool.

Also the markdown to html works nicely like this.

### DIFF
--- a/src/routes/private.ts
+++ b/src/routes/private.ts
@@ -29,6 +29,7 @@ import {
 } from '@dailydotdev/schema';
 import { OpportunityJob } from '../entity/opportunities/OpportunityJob';
 import { OpportunityKeyword } from '../entity/OpportunityKeyword';
+import { logger } from '../logger';
 
 interface SearchUsername {
   search: string;
@@ -89,7 +90,7 @@ export default async function (fastify: FastifyInstance): Promise<void> {
 
     const opportunity = opportunityCreateSchema.safeParse(req.body);
     if (opportunity.error) {
-      req.log.error(
+      logger.error(
         {
           opportunity,
         },


__tests__/private.ts
```diff
@@ -1083,3 +1088,191 @@ describe('GET /p/actions/:user_id/:action_name', () => {
...
+
+    // Verify keywords
+    const keywords = await con.getRepository(OpportunityKeyword).find({
+      where: { opportunityId: opportunity?.id },
+    });
+    expect(keywords).toHaveLength(2);
+    expect(keywords.map((k) => k.keyword).sort()).toEqual([
+      'golang',
+      'kubernetes',
+    ]);
```
capJavert, line 1276: there is also expect.arrayContaining, but not blocking

src/routes/private.ts
```diff
@@ -71,6 +82,72 @@ export default async function (fastify: FastifyInstance): Promise<void> {
...
     },
   );
+  fastify.post<{ Body: Opportunity }>('/newOpportunity', async (req, res) => {
+    if (!req.service) {
+      return res.status(404).send();
+    }
+
+    const opportunity = opportunityCreateSchema.safeParse(req.body);
+    if (opportunity.error) {
+      req.log.error(
```
capJavert, line 92:
```suggestion
      logger.error(
```

use our exported logger?

capJavert (approved):

src/routes/private.ts
rebelchris: It just generalizes the error right?
think the req has tracing?

CC @omBratteng probably knows best nuance.

capJavert: its the same config, but we use logger export all around to keep it consistent

omBratteng (approved):

PR #3204, opened 2025-10-07
feat: add script for backfilling user_candidate_preference

omBratteng:
A backfill script for fetching all blobs from the resume bucket, check if user exsits and wether UCP contains the blob.
If users exists and UCP does not contain blob, we upsert the row by adding the CV information. This should result in ~12k inserts.

Not really optimized the queries for speed, it'll only be ran once or twice to ensure all is upserted. And it's a known value if less than 18k blobs to iterate over.

### DIFF
--- a/bin/backfillUCP.ts
+++ b/bin/backfillUCP.ts
@@ -1,0 +1,141 @@
+import { createInterface } from 'node:readline/promises';
+import { stdin as input, stdout as output } from 'node:process';
+
+import { acceptedResumeFiles } from '../src/common/schema/files';
+import type { UserCandidateCV } from '../src/common/schema/userCandidate';
+import createOrGetConnection from '../src/db';
+import { Storage, type GetFilesOptions, File } from '@google-cloud/storage';
+import { logger } from '../src/logger';
+import { UserCandidatePreference } from '../src/entity/user/UserCandidatePreference';
+import { queryReadReplica } from '../src/common/queryReadReplica';
+import { JsonContains } from 'typeorm';
+import { User } from '../src/entity/user/User';
+
+const storage = new Storage();
+
+const getExtension = (contentType: string) => {
+  switch (contentType) {
+    case acceptedResumeFiles.pdf.mime[0]:
+      return 'pdf';
+    case acceptedResumeFiles.docx.mime[0]:
+      return 'docx';
+    default:
+      return null;
+  }
+};
+
+const ask = async (q: string) => {
+  const rl = createInterface({ input, output });
+  try {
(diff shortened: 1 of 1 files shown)


bin/backfillUCP.ts
```diff
@@ -0,0 +1,142 @@
...
+(async () => {
+  console.log('Starting script...');
+  const con = await createOrGetConnection();
+
+  const bucketName = 'daily-dev-resumes';
+  const bucket = storage.bucket(bucketName);
+
+  const maxResults = parseInt(
+    process.argv[2] || (await ask('Enter maxResults: ')),
+  );
```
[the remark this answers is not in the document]
capJavert: then you can easily have args like `--maxResults` and so on

bin/backfillUCP.ts
```diff
@@ -0,0 +1,142 @@
...
+        bucket: bucketName,
+        fileName: `${file.name}.${ext}`,
+        contentType: contentType,
+        lastModified: new Date(file.metadata.updated as string),
+      };
+      const [userExists, ucpExists] = await queryReadReplica(
+        con,
+        async ({ queryRunner }) => {
+          return await Promise.all([
+            queryRunner.manager.getRepository(User).exists({
```
capJavert, line 96: I know its not optimized for speed, which is fine but if it takes long you could query the whole batch in one go with `IN`

bin/backfillUCP.ts
```diff
@@ -0,0 +1,142 @@
...
+      await con.getRepository(UserCandidatePreference).upsert(
+        updates.map((cv) => ({
+          userId: cv.blob!,
+          cv: cv,
+        })),
+        ['userId'],
+      );
+    } catch (_err) {
+      const err = _err as Error;
+      logger.error({ err }, 'Error inserting user candidate preferences:');
```
capJavert, line 134: error would block the whole process again, is there offset functionality so you can restart from last page?

capJavert (approved):
just some optional comments

bin/backfillUCP.ts
omBratteng: it just logs it without throwing anything, so it passes. Ran this locally against prod bucket but adhoc db, so it doesn't have any of the users, so it just logs errors and continues. Which is what I want. Because it shouldn't fail (famous last words)

capJavert: but you have throw below?

omBratteng: oh yeah 😂 
it's fine, I want it to fail. Because it means I've probably messed something up and I need to adust the script

PR #3207, opened 2025-10-08
feat: autocomplete entity/query/schema

sshanzel:
Introduced the entity to handle generic autocompletes [as discussed](https://dailydotdev.atlassian.net/wiki/spaces/HAN/pages/1816657933/RFC-146+-+Profile+4.0+Backend#Autocomplete).


bin/import.ts
```diff
@@ -52,6 +52,14 @@ const viewsToRefresh = [
 
 const start = async (): Promise<void> => {
   const con = await createOrGetConnection();
+  const entityParam = process.argv[2];
+
+  if (entityParam) {
+    await importEntity(con, entityParam);
+    return;
+  }
```
sshanzel, line 60: Just to allow importing a specific entity only.

capJavert: nice idea, you can also use parseArgs from node:util + zod to avoid using raw process argv, check assignCores script, not blocking

src/schema/autocompletes.ts
```diff
@@ -0,0 +1,59 @@
...
+`;
+
+export const resolvers = traceResolvers<unknown, BaseContext>({
+  Query: {
+    autocomplete: async (
+      _,
+      payload: z.infer<typeof autocompleteSchema>,
+      ctx: AuthContext,
+    ): Promise<AutocompleteData> => {
+      const { data, error } = autocompleteSchema.safeParse(payload);
```
capJavert, line 40:
```suggestion
      const { data } = autocompleteSchema.parse(payload);
```

you can just use `parse`, API will now convert the error thrown by it automatically to zod error, no need to throw ValidationError below

src/common/schema/autocompletes.ts
```diff
@@ -0,0 +1,9 @@
+import z from 'zod';
+import { normalizeContentForDeduplication } from '../../entity/posts/hooks';
+import { AutocompleteType } from '../../entity';
+
+export const autocompleteSchema = z.object({
+  type: z.enum(AutocompleteType),
+  query: z.string().transform(normalizeContentForDeduplication),
```
capJavert, line 7: there is lowercase and trim in zod

bin/import.ts
sshanzel: Updated. Nicely done on type-strict script!

src/entity/Autocomplete.ts
```diff
@@ -0,0 +1,31 @@
...
+  FieldOfStudy = 'field_of_study',
+  Degree = 'degree',
+  Role = 'role',
+}
+
+const compositePrimaryKeyName = 'PK_autocomplete_value_type';
+
+@Entity()
+@Index('IDX_autocomplete_value_enabled_trgm', { synchronize: false })
+export class Autocomplete {
```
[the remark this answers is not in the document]
sshanzel: @rebelchris when it involves a separate entity, we won't store that here. If you only need a simple string from a single source (opposite example: location requires looking up three columns), we use this table.

> Autocomplete endpoint could still be uniform but take "Type" of dataset into account?

Autocompletions that involve a separate entity normally require a more complicated query; hence, I think it is better to be separated.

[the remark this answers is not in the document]
sshanzel: Even if the data grows very large for a specific type, as long as we only require a single column string for the lookup, we will store them here.

src/entity/index.ts
```diff
@@ -36,3 +36,4 @@ export * from './SquadPublicRequest';
 export * from './UserCompany';
 export * from './Organization';
 export * from './campaign';
+export * from './Autocomplete';
```
omBratteng, line 39: do we need to?

src/entity/Autocomplete.ts
```diff
@@ -0,0 +1,31 @@
...
+    type: 'text',
+    primaryKeyConstraintName: compositePrimaryKeyName,
+  })
+  value: string;
+
+  @PrimaryColumn({
+    type: 'text',
+    primaryKeyConstraintName: compositePrimaryKeyName,
+  })
+  type: string;
```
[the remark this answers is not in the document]
sshanzel: I think we should not need an index anymore since they are composite keys already?

src/entity/index.ts
sshanzel: For some reason I was getting some errors (can't exactly which script) when this was left out.

src/migration/1759767544399-Autocomplete.ts
```diff
@@ -0,0 +1,18 @@
...
+    );
+    await queryRunner.query(
+      `CREATE INDEX IF NOT EXISTS "IDX_autocomplete_value_enabled_trgm" ON "public"."autocomplete" USING gin (value gin_trgm_ops) WHERE enabled = true`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP TABLE "autocomplete"`);
+  }
+}
```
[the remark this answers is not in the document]
sshanzel: Thank you!

src/entity/Autocomplete.ts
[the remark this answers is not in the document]
omBratteng: It's possible to create a partitioned table if we want.

```sql
-- Parent partitioned table
CREATE TABLE autocomplete (
  value text NOT NULL,
  type text NOT NULL,
  "createdAt" TIMESTAMP NOT NULL DEFAULT now(),
  enabled boolean NOT NULL DEFAULT true,
  CONSTRAINT "PK_autocomplete_value_type" PRIMARY KEY (value, type)
) PARTITION BY LIST (type);

-- Partitions for known types
CREATE TABLE autocomplete_type_user PARTITION OF autocomplete
  FOR VALUES IN ('user');

CREATE TABLE autocomplete_type_repo PARTITION OF autocomplete
  FOR VALUES IN ('repo');

-- Optional catch-all so inserts for new types do not fail
CREATE TABLE autocomplete_type_default PARTITION OF autocomplete
  DEFAULT;

-- Indexes (per partition where needed)
CREATE INDEX autocomplete_user_value_idx ON autocomplete_type_user (value);
CREATE INDEX autocomplete_repo_value_idx ON autocomplete_type_repo (value);
-- For the default partition, consider a composite for common lookups
CREATE INDEX autocomplete_default_type_value_idx
  ON autocomplete_type_default (type, value);
```

src/entity/Autocomplete.ts
omBratteng, line 24:
```suggestion
  type: AutocompleteType;
```

src/entity/Autocomplete.ts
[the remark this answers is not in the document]
rebelchris: @sshanzel but what's the downside of doing the seperated tables now?
It feels like a safer/easy thing.

Im just not seeing why we wouldn't do it.

PR #3210, opened 2025-10-08
feat: import dataset location

sshanzel:
- Prepared the script to import locations once we buy the pro dataset.
- For local dev, I used the free dataset which is about **40k entries**. Should give us a close experience to what we'll have in prod.

### Jira ticket
MI-1061

### DIFF
--- a/src/schema/autocompletes.ts
+++ b/src/schema/autocompletes.ts
@@ -71,7 +71,7 @@ export const typeDefs = /* GraphQL */ `
 
 type FindLocation = FindOptionsWhere<DatasetLocation>;
 
-const getLocationCondition = (query: string) => {
+const getLocationCondition = (query: string): FindLocation[] => {
   const [country, subdivision, city] = query
     .split(',')
     .reverse()


bin/common.ts
```diff
@@ -24,3 +31,38 @@ export const runInQueueStream = async <T>(
   await insertQueue.drained();
   console.log('update finished with a total of: ', insertCount);
 };
+
+type AllowedZodTypes = z.ZodString | z.ZodBoolean;
+
+export const zodToParseArgs = <T extends z.ZodRawShape>(
```
sshanzel, line 37: Had fun with Zod. Not bulletproof but just enough to work with `parseArgs` specifically.

Defining node arguments can now be based solely on the Zod schema.

bin/export.ts
```diff
@@ -1,13 +1,31 @@
 import { writeFileSync } from 'fs';
 import createOrGetConnection from '../src/db';
+import z from 'zod';
```
sshanzel, line 3: Applied the same thing from import - to export specific entity only.

__tests__/schema/autocompletes.ts
```diff
@@ -358,3 +359,381 @@ describe('query autocompleteKeywords', () => {
     ]);
   });
 });
+
+describe('query autocompleteLocation', () => {
```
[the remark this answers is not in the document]
sshanzel: Good points! Let me summon my intern 😂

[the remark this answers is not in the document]
sshanzel: @omBratteng we can use iso2 and iso3 for when the query length is less than 4. Good call on keeping those 🚀

bin/common.ts
```diff
@@ -24,3 +31,38 @@ export const runInQueueStream = async <T>(
...
+      acc[key] = { type: 'string' };
+    }
+
+    const meta = field.meta() as { short?: string };
+
+    if (meta?.short) {
+      acc[key].short = meta.short;
+    }
+
+    const defaultValue = field.safeParse(undefined).data;
```
rebelchris, line 56: Why safeParse a undefined value? not getting why this is needed?

rebelchris (approved):

PR #3224, opened 2025-10-14
feat: verification workers for user work experience

sshanzel:
- user company table:
  - If user listed their work experience first before verification.
- user experience table:
  - When creating a new record with a pre-existing verification.
  - When updating experience to a pre-existing verified user company.
  - When changing the company from a verified one to a new unverified.

### Jira ticket
MI-1066


src/workers/cdc/primary.ts
```diff
@@ -1457,6 +1466,60 @@ export const onCampaignChange = async (
...
+    return;
+  }
+
+  const work = experience as ChangeObject<UserExperienceWork>;
+
+  if (!work.companyId) {
+    if (work.verified) {
+      await con
+        .getRepository(UserExperienceWork)
+        .update({ id: work.id }, { verified: false });
```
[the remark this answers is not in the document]
capJavert: sounds like it could be a trigger as well in that case, then its enforced on db level for sure and on all writes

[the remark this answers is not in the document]
sshanzel: Makes sense. Updated it to be a trigger.

sshanzel:
@capJavert would love to hear what you think now that it is a trigger.

src/migration/1760520784878-UserExperienceCompanyTrigger.ts
```diff
@@ -0,0 +1,95 @@
...
+      CREATE OR REPLACE FUNCTION validate_user_experience()
+      RETURNS TRIGGER AS $$
+      BEGIN
+          IF EXISTS (
+              SELECT 1
+              FROM user_company uc
+              WHERE uc."userId" = NEW."userId"
+              AND uc."companyId" = NEW."companyId"
+              AND uc.verified = true
+          ) THEN
```
capJavert: would FK not guard that company exists, we don't need to do subquery?

sshanzel: Part of the where condition is whether the company in question is verified. You are querying whether the user is verified in this company, not if the Company just exists.

capJavert: Or if we want to keep the trigger do it other way around, when user_company is set as verified we update all user_experience where companyId and userId match. That would only do additional write?

sshanzel: > By the way, I owe you a beer for this one.

All cool! I'll take that any day of the week in the Gathering 😆 

> Or if we want to keep the trigger do it other way around, when user_company is set as verified we update all user_experience where companyId and userId match. That would only do additional write?

We check boths sides, from user_company, and from user_experience table to ensure no sides are missed. The thing is that, the user_company might have existed first before the change, so when we trigger from user_company, the experience is not there yet. Then when we add the experience, user_company won't be triggered anymore.

This is the one for the user_company side:

https://github.com/dailydotdev/daily-api/blob/0e9fcc20a146d95ad438818c7cafd73f5cbb59bc/src/workers/cdc/primary.ts#L1201-L1206

capJavert (approved):

PR #3227, opened 2025-10-15
feat: remove user experience

sshanzel:
### Jira ticket
MI-1060

### DIFF
--- a/src/schema/profile.ts
+++ b/src/schema/profile.ts
@@ -332,11 +332,9 @@ export const resolvers = traceResolvers<unknown, AuthContext>({
       return getUserExperience(ctx, info, entity.id);
     },
     removeUserExperience: async (_, { id }: { id: string }, ctx) => {
-      const toRemove = await ctx.con
+      await ctx.con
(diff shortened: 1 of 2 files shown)


src/schema/profile.ts
```diff
@@ -331,5 +331,14 @@ export const resolvers = traceResolvers<unknown, AuthContext>({
 
       return getUserExperience(ctx, info, entity.id);
     },
+    removeUserExperience: async (_, { id }: { id: string }, ctx) => {
+      const toRemove = await ctx.con
+        .getRepository(UserExperience)
+        .findOneOrFail({ where: { id, userId: ctx.userId } });
+
+      await ctx.con.getRepository(UserExperience).remove(toRemove);
```
capJavert: you don't have to query, you can just do delete with where query above and that should do it

capJavert: with our other deletes we usually also silently succeed, its like removing upvote which was already removed, no need to fail since user wanted to delete it, and its gone already

PR #3229, opened 2025-10-17
fix: return streaks as null for user profile if they opted out of streaks

nensidosari:
Frontend [pr](https://github.com/dailydotdev/apps/pull/5006)

Instead of exposing the users settings in frontend for each user that we show the public profile, we are checking the users profile settings in the api, and returning streaks as null if user has opted out. 

Till today we were just hiding any reading streaks information about any user if the current logged in user had opted out. This now hides based on each user profile.

### DIFF
--- a/src/schema/users.ts
+++ b/src/schema/users.ts
@@ -1424,7 +1424,10 @@ const getUserStreakQuery = async (
       .addSelect('u.id', 'userId')
       .addSelect('u.timezone', 'timezone')
       .addSelect('u."weekStart"', 'weekStart')
-      .addSelect('COALESCE(s."optOutReadingStreak", false)', 'optOutReadingStreak')
+      .addSelect(
+        'COALESCE(s."optOutReadingStreak", false)',
+        'optOutReadingStreak',
+      )
       .innerJoin(User, 'u', `"${builder.alias}"."userId" = u.id`)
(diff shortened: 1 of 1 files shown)


src/schema/users.ts
```diff
@@ -1649,7 +1650,17 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+    ): Promise<GQLUserStreak | null> => {
+      // Check if user has opted out of reading streaks
+      const settings = await ctx.con.getRepository(Settings).findOne({
+        where: { userId: ctx.userId },
+        select: ['optOutReadingStreak'],
+      });
+
+      if (settings?.optOutReadingStreak) {
+        return null;
+      }
```
capJavert: Not the best solution since we are introducing another query that needs to run before querying streak

We need to do it in parallel, can you maybe add this as join in `getUserStreakQuery` and then select `settings.optOutReadingStreak` from there.

Then you can do `streak.optOutReadingStreak` to return null in places you want.

capJavert:
@nensidosari can we maybe make it somehow backward compatible then? So it does not break existing clients. Every breaking change on the API is clunky thing to handle down the line.

What happens exactly with current client, how does it break?

capJavert:
Nice, then this is not a breaking change after all 👌

PR #3233, opened 2025-10-22
feat: query cta and message

sshanzel:
- Introduced an `externalUserId` that must be required for better tracking.
- Query for getting the copy and the message.

### DIFF
--- a/src/schema/profile.ts
+++ b/src/schema/profile.ts
@@ -232,7 +232,7 @@ const getUserExperience = (
   );
 
 // These values are something that could come from growthbook for experimentations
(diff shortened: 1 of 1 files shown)


src/entity/user/referral/UserReferral.ts
```diff
@@ -39,6 +44,11 @@ export class UserReferral {
   })
   userId: string;
 
+  @Column({
+    type: 'text',
+  })
+  externalUserId: string;
+
```
sshanzel: @omBratteng introduced a property instead as it must be present at all times. Much easier to manage. Then the URL is kept in the flags.

PR #3260, opened 2025-11-06
fix: manually delete user post actions

omBratteng:
Should cover edge cases where user deletes their account, and when they have voted on their own polls on their own source.


omBratteng:
Problem was not db related this time, but client side because multiple clicks.
https://github.com/dailydotdev/apps/pull/5041

PR #3263, opened 2025-11-07
hotfix: update to 400 limit

rebelchris:
Changing limit to 400 from user feedback


omBratteng (approved):
Now imagine if we had a common zod schema package, then this change would just be needed one place.

PR #3264, opened 2025-11-09
docs: add comprehensive guides for agents, cron jobs, GraphORM, and background workers

idoshamun:


AGENTS.md
```diff
@@ -0,0 +1,85 @@
...
+- `pnpm run dev` - Start API server with hot reload on port 5000
+- `pnpm run dev:background` - Start background processor
+- `pnpm run dev:temporal-worker` - Start Temporal worker
+- `pnpm run dev:temporal-server` - Start Temporal server for local development
+
+**Database:**
+- `pnpm run db:migrate:latest` - Apply latest migrations
+- `pnpm run db:migrate:reset` - Drop schema and rerun migrations
+- `pnpm run db:seed:import` - Import seed data for local development
+- `pnpm run db:migrate:make` - Generate new migration based on entity changes
```
rebelchris: This comes with nuance that it imports everything retro, maybe just say prefer the create for agent mode? it's good at filling structured stuff anyway

AGENTS.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Core Services:**
+- `src/Context.ts` - Request context with user, permissions, and data loaders
+- `src/auth.ts` - Authentication middleware and user context resolution
+- `src/dataLoaderService.ts` - Efficient batch loading for related entities
+- `src/workers/` - Use workers for async, non-critical operations (notifications, reputation, external syncs). Prefer `TypedWorker` for type safety. Architecture uses Google Pub/Sub + CDC (Debezium) for reactive processing. See `src/workers/AGENTS.md` for more.
+- `src/integrations/` - External service integrations (Slack, SendGrid, etc.) 
+- `src/cron/` - Scheduled cron jobs for maintenance and periodic tasks. One file per cron, registered in `index.ts`, deployed via `.infra/crons.ts` Pulumi config. Each cron exports a `Cron` object with `name` and `handler(DataSource, logger, pubsub)`. Run locally with `pnpm run cli cron <name>`. See `src/cron/AGENTS.md` for more.
+
+**Business Domains:**
```
rebelchris: Squads I think as well, I had one agent ask me what it was

AGENTS.md
```diff
@@ -0,0 +1,85 @@
...
+- **Content**: Posts, comments, bookmarks, feeds, sources
+- **Users**: Authentication, preferences, profiles, user experience
+- **Organizations**: Squad management, member roles, campaigns
+- **Notifications**: Push notifications, email digests, alerts
+- **Monetization**: Paddle subscription management, premium features
+
+**Testing Strategy:**
```
rebelchris: There's also singular e2e test, but agent might be able to extend if it feels needed

src/workers/AGENTS.md
```diff
@@ -0,0 +1,380 @@
...
+- Updating reputation scores
+- Processing images and media
+- Syncing data to external services (CIO, Slack)
+- Analytics and metrics collection
+- Cleanup operations
+- Content processing (markdown, translations)
+
+## Worker Types
+
+### Typed Workers (`TypedWorker`)
```
rebelchris: Wondering if we should also mention itr on the main one, that we prefer strong typed (can't recall if i saw it)

src/workers/AGENTS.md
```diff
@@ -0,0 +1,380 @@
...
+
+1. **Publish**: Application code publishes a message to a Pub/Sub topic
+2. **Subscribe**: Worker subscribes to the subscription
+3. **Receive**: Pub/Sub delivers the message to the worker
+4. **Process**: Worker handler executes
+5. **Ack/Nack**: Worker acknowledges (ack) on success or negative-acknowledges (nack) on failure
+   - **Ack**: Message is removed from subscription
+   - **Nack**: Message is redelivered (with exponential backoff)
+
+### Error Handling
```
rebelchris: Something it's also good it validated params.
EG it's userId it could be user deleted in meantime so it should query internally to validate preferably?

rebelchris (approved):
Non blocking thoughts

AGENTS.md
idoshamun: I usually find it safer to use make, it's battle tested and deterministic. It's up to the developer to make sure their environment is up to date

rebelchris: Just afraid it will prompt ALL changes in a file *(pretty sure this is what happens now right?)

AGENTS.md
```diff
@@ -0,0 +1,85 @@
...
+- `NODE_ENV=test npx jest __tests__/specific-test.ts --testEnvironment=node --runInBand`
+- Use `--testEnvironment=node --runInBand` flags for database-dependent tests
+
+## High-Level Architecture
+
+**Core Framework Stack:**
+- **Fastify** - Web framework with plugins for CORS, helmet, cookies, rate limiting
+- **Mercurius** - GraphQL server with caching, upload support, and subscriptions
+- **TypeORM** - Database ORM with entity-based modeling and migrations
+- **PostgreSQL** - Primary database with master/slave replication setup
```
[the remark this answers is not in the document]
rebelchris: It depends what it's using, all queries have RR params, or you can wrap a whole block.
We always favor if it's stale,  if you just mutated we prefer master reads to ensure latest data.

src/workers/AGENTS.md
rebelchris: In some cases for instance now with external messages (gondul example where match can be made, but used deleted account in mean time)

The worker ideally should query user from database vs simply hoping it exists to avoid endless errors.

PR #3277, opened 2025-11-12
feat: endpoint for all opportunities

rebelchris:
New endpoint for fetch:

- Team members: all opportunities
- Recruiter: all LIVE opportunities for them
- Option to switch type (potential future thing)

### DIFF
--- a/src/schema/opportunity.ts
+++ b/src/schema/opportunity.ts
@@ -794,11 +794,22 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
                   OpportunityMatchStatus.RecruiterAccepted,
                   OpportunityMatchStatus.RecruiterRejected,
                 ],
-              });
+              })
+              // Order by candidate_accepted status first (priority 0), then others (priority 1)
+              // Then by createdAt ascending (oldest first) within each group
+              .addOrderBy(
+                `CASE WHEN ${builder.alias}.status = :candidateAcceptedStatus THEN 0 ELSE 1 END`,
+                'ASC',
+              )
+              .addOrderBy(`${builder.alias}.createdAt`, 'ASC')
+              .setParameter(
+                'candidateAcceptedStatus',
+                OpportunityMatchStatus.CandidateAccepted,
+              );
 
             return builder;
           },
-          orderByKey: 'DESC',
+          orderByKey: 'ASC',
           readReplica: true,
         },
       );


src/schema/opportunity.ts
```diff
@@ -217,6 +220,16 @@ export const typeDefs = /* GraphQL */ `
...
+
+    """
+    Get all opportunities filtered by state (defaults to LIVE)
+    """
+    getOpportunities(
+      """
+      State of opportunities to fetch (defaults to LIVE)
+      """
+      state: ProtoEnumValue
+    ): [Opportunity!]! @auth
```
idoshamun: Better to have pagination, just in case

src/schema/opportunity.ts
```diff
@@ -464,6 +477,47 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+
+      return await graphorm.query<GQLOpportunity>(
+        ctx,
+        info,
+        (builder) => {
+          builder.queryBuilder.where({ state: opportunityState });
+
+          // If fetching non-LIVE opportunities and user is not a team member,
+          // filter at database level to only show opportunities where user is a recruiter
+          if (opportunityState !== OpportunityState.LIVE && !ctx.isTeamMember) {
```
idoshamun: Why do we apply it only if the opp isn't live? It should apply on all opps regardless i think

src/schema/opportunity.ts
rebelchris: Ah yes good point :)

src/schema/opportunity.ts
```diff
@@ -464,6 +501,54 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+                  'ou',
+                  `ou.opportunityId = ${builder.alias}.id`,
+                )
+                .andWhere('ou.userId = :userId', { userId: ctx.userId })
+                .andWhere('ou.type = :type', {
+                  type: OpportunityUserType.Recruiter,
+                });
+            }
+
+            builder.queryBuilder.addSelect(`${builder.alias}.state`, 'state');
```
idoshamun: why do you need it? (non blocking)

src/schema/opportunity.ts
```diff
@@ -217,6 +276,42 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      Paginate first
+      """
+      first: Int
+    ): OpportunityConnection! @auth
+
+    """
+    Get all opportunity matches for a specific opportunity (includes only candidate_accepted, recruiter_accepted, and recruiter_rejected statuses)
+    """
+    getOpportunityMatches(
```
idoshamun: Same here

src/schema/opportunity.ts
```diff
@@ -330,6 +425,28 @@ export const typeDefs = /* GraphQL */ `
...
+      Id of the Opportunity
+      """
+      opportunityId: ID!
+      """
+      Id of the candidate user to accept
+      """
+      candidateUserId: ID!
+    ): EmptyResponse @auth
+
+    recruiterRejectOpportunityMatch(
```
idoshamun: Shouldn't we do one mutation that accepts a parameter?

src/schema/opportunity.ts
```diff
@@ -464,6 +711,109 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+                  OpportunityMatchStatus.RecruiterRejected,
+                ],
+              })
+              // Order by candidate_accepted status first (priority 0), then others (priority 1)
+              // Then by createdAt ascending (oldest first) within each group
+              .addOrderBy(
+                `CASE WHEN ${builder.alias}.status = :candidateAcceptedStatus THEN 0 ELSE 1 END`,
+                'ASC',
+              )
+              .addOrderBy(`${builder.alias}.createdAt`, 'ASC')
```
idoshamun: `updatedAt` is the right one. it will order by the matches according to the time it was shown to the recruiter, not when it was sent to the candidate

src/schema/opportunity.ts
rebelchris: I thought about it, but followed what we had for other mutations we seem to do duplication incase something needs to have exceptions. (under the hood they share the same internal function)

PR #3279, opened 2025-11-12
feat: snotra engagement provider

rebelchris:
Snotra client + engagement resolver.

(Kinda hacky on graphORM, dataLoader felt bit overkill and this also works)
Since it's "nice to have" I added timeout and early exit if snotra fails to return.

(Also some minor CIO fixes as it breaks local now)

### DIFF
--- a/src/schema/opportunity.ts
+++ b/src/schema/opportunity.ts
@@ -197,6 +197,11 @@ export const typeDefs = /* GraphQL */ `
     warmIntro: String
   }
 
+  type EngagementProfile {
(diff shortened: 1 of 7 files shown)


src/schema/opportunity.ts
```diff
@@ -197,6 +197,11 @@ export const typeDefs = /* GraphQL */ `
     warmIntro: String
   }
 
+  type EngagementProfile {
```
idoshamun: We should return just the text i think

PR #3289, opened 2025-11-18
refactor: location

AmarTrebinjac:
After a lot of back and forth we decided that we are going to use Mapbox API instead of seeded data, due to the data we got being very low quality.

I removed the timezone and ranking for two reasons:
1. RFC doesn't mention how or if we are using these at all
2. Mapbox API doesn't provide timezones, and does its own ranking when searching, so redundant for us.

### DIFF
--- a/src/integrations/mapbox/clients.ts
+++ b/src/integrations/mapbox/clients.ts
@@ -59,7 +59,7 @@ const garmrMapboxService = new GarmrService({
     minimumRps: 0,
   },
   retryOpts: {
-    maxAttempts: 3,
+    maxAttempts: 2,
     backoff: 100,
   },
 });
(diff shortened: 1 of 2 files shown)


src/entity/dataset/DatasetLocation.ts
```diff
@@ -32,12 +32,7 @@ export class DatasetLocation {
...
   iso3: string;
 
-  @Column()
-  timezone: string;
-
-  @Column({ default: 0 })
-  ranking: number;
-
-  @Column({ type: 'text', nullable: true })
+  @Index('IDX_dataset_location_externalId')
```
AmarTrebinjac: Added index since we are now querying it on profile save

src/schema/autocompletes.ts
```diff
@@ -148,22 +106,35 @@ export const resolvers = traceResolvers<unknown, BaseContext>({
...
-
-      return result;
+      const { query } = autocompleteBaseSchema.parse(payload);
+      const mapboxUrl = `${process.env.MAPBOX_GEOCODING_URL}?q=${encodeURIComponent(query)}&types=country,place&limit=5&access_token=${process.env.MAPBOX_ACCESS_TOKEN}`;
+
+      try {
+        const response = await fetch(mapboxUrl);
+
+        if (!response.ok) {
+          return [];
```
AmarTrebinjac: Not sure if we wanna throw an error here or just return empty? Cause if the user starts typing again it will trigger the endpoint again, and then we might have an OK response.

bin/import.ts
```diff
@@ -66,7 +66,6 @@ const start = async (): Promise<void> => {
     return await importEntity(con, params.entity);
   }
 
-  await importEntity(con, 'DatasetLocation');
```
capJavert: also remove seeds then?

src/entity/dataset/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  let mapboxGeocodingUrl = `${process.env.MAPBOX_GEOCODING_URL}?q=${encodeURIComponent(externalId)}&access_token=${process.env.MAPBOX_ACCESS_TOKEN}`;
+
+  if (process.env.NODE_ENV === 'production') {
+    mapboxGeocodingUrl += '&permanent=true';
+  }
+
+  const response = await fetch(mapboxGeocodingUrl);
+
+  if (!response.ok) {
+    throw new Error('Mapbox API error');
```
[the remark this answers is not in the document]
capJavert: Error is fine, it will be 5xx, but maybe you want to add alerts if you think its necessary we know when it fails. 

Also with 3party integrations we usually use Garmr for automatic retries, failover and so on so ideally you add that. Garmr will automatically notifiy on the channel.

src/integrations/mapbox/clients.ts
```diff
@@ -0,0 +1,73 @@
...
+  service: 'mapbox',
+  breakerOpts: {
+    halfOpenAfter: 5 * 1000,
+    threshold: 0.1,
+    duration: 10 * 1000,
+    minimumRps: 0,
+  },
+  retryOpts: {
+    maxAttempts: 3,
+    backoff: 1000,
```
capJavert: This will wait for 1000ms after each retry so if we hit rate limit and it keeps failing it will delay any request that calls this for 3+ seconds?

Just making sure its what you want?

PR #3295, opened 2025-11-21
feat: new endpoint to expose user opportuntity matches

rebelchris:
We want to add something like this:
Contemplated doing 2 queries, but since majority will be very few for now thought 1 would be better for now. (can always modify to accept types or something)

Screenshot 2025-11-21 at 13 44 45

### DIFF
--- a/src/schema/opportunity.ts
+++ b/src/schema/opportunity.ts
@@ -873,16 +873,8 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
       args: ConnectionArguments,
       ctx: AuthContext,
       info,
-    ) => {
-      if (!ctx.userId) {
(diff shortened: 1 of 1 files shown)


src/schema/opportunity.ts
```diff
@@ -854,6 +868,36 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
     },
+    userOpportunityMatches: async (
+      _,
+      args: ConnectionArguments,
+      ctx: AuthContext,
+      info,
+    ) => {
+      if (!ctx.userId) {
+        throw new NotFoundError('Not found!');
+      }
```
capJavert: should not be needed since in AuthContext its always there

capJavert (approved):

PR #3302, opened 2025-11-25
feat: customLocation for user experience

capJavert:
We currently don't have data in dataset_location, so saving unmatched locations to `customLocation` column so we can do enrichment later.

### DIFF
--- a/src/entity/user/experiences/UserExperience.ts
+++ b/src/entity/user/experiences/UserExperience.ts
@@ -69,6 +69,13 @@ export class UserExperience {
   @Column({ type: 'text', default: null })
   locationId: string | null;
 
+  @Column({ type: 'jsonb', default: {} })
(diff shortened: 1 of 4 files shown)


src/entity/user/experiences/UserExperience.ts
```diff
@@ -69,6 +69,13 @@ export class UserExperience {
   @Column({ type: 'text', default: null })
   locationId: string | null;
 
+  @Column({ type: 'jsonb', default: {} })
+  customLocation: Partial<{
```
capJavert: will expose this column and connect to frontend for readonly display on profile in different PR.

capJavert: Write should always go to `locationId`

AmarTrebinjac (approved):

PR #3304, opened 2025-11-25
feat: enable worker and notification for parse cv

capJavert:


rebelchris (approved):
The trigger was already merged i assume right

PR #3315, opened 2025-12-01
feat: create slack channel

rebelchris:
Option to create shared slack channel and invite the users. 
(option to add internal we need to look into)

### DIFF
--- a/.infra/index.ts
+++ b/.infra/index.ts
@@ -532,10 +532,12 @@ const [apps] = deployApplicationSuite(
           value: 'true',
         },
       ],
-      // TODO: split limit and request
+      requests: {
+        cpu: '50m',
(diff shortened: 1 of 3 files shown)


src/schema/opportunity.ts
```diff
@@ -576,6 +578,21 @@ export const typeDefs = /* GraphQL */ `
...
+      """
+      Email address of the user to invite
+      """
+      email: String!
+
+      """
+      Name of the channel to create (lowercase letters, numbers, hyphens, and underscores only)
+      """
+      channelName: String!
+    ): EmptyResponse @auth
```
capJavert: Should we check that user is recruiter/paid, like this any auth user can create slack channel?

PR #3316, opened 2025-12-01
feat: claim anon opportunities

capJavert:
When user registers while having anon opportunities waiting from self serve flow we register him as a recruiter for that opportunity.

Also added logic to create organization by name if not set on opportunity already.

### DIFF
--- a/__tests__/users.ts
+++ b/__tests__/users.ts
@@ -59,30 +59,32 @@ import {
 } from '../src/entity';
 import { sourcesFixture } from './fixture/source';
 import {
-  bskySocialUrlMatch,
   CioTransactionalMessageTemplateId,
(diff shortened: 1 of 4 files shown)


src/entity/Organization.ts
```diff
@@ -34,6 +34,7 @@ export class Organization {
   updatedAt: Date;
 
   @Column({ type: 'text' })
+  @Index('IDX_organization_name_unique', { unique: true })
```
capJavert: Added so that organization name is unique to avoid people adding  well known organizations that were already created.

Still anyone can add any organization so we need to be careful and monitor what gets added. Since we will be doing some manual approval before self serve goes live I am counting on that.

rebelchris (approved):
looks good to me

PR #3326, opened 2025-12-07
docs: update AGENTS.md

idoshamun:


src/graphorm/AGENTS.md
```diff
@@ -239,7 +241,7 @@ export const resolvers = {
             .limit(1);
           return builder;
         },
-        true // Use read replica
+        true // readReplica: Use read replica for better performance
```
capJavert:
```suggestion
        true // for better performance for read queries are required, take into account potential replication lag if doing reads right after writes
```

src/workers/AGENTS.md
```diff
@@ -69,15 +71,50 @@ import { MatchedCandidate } from '@dailydotdev/schema';
 
 const worker: TypedWorker<'gondul.v1.candidate-opportunity-match'> = {
   subscription: 'api.store-candidate-opportunity-match',
-  handler: async ({ data }, con): Promise<void> => {
-    // data is typed as MatchedCandidate
+  handler: async (message, con): Promise<void> => {
+    // message.data is typed as MatchedCandidate
+    const { data } = message;
   },
   parseMessage: (message) => {
```
capJavert: we need to say that adding `parseMessage` is required if `data` is protobuf or from `@dailydotdev/schema`, linter also warns

src/workers/AGENTS.md
```diff
@@ -69,15 +71,50 @@ import { MatchedCandidate } from '@dailydotdev/schema';
...
+### Experiment Workers (`ExperimentWorker`)
+
+For workers that need access to the GrowthBook experiment allocation client:
+
+```typescript
+import { ExperimentWorker, workerToExperimentWorker } from './worker';
+
+const worker: ExperimentWorker = {
+  subscription: 'api.experiment-allocated',
+  handler: async (message, con, logger, pubsub, experimentAllocationClient) => {
```
capJavert: make sure it calls `experimentAllocationClient.waitForSend` before exiting the worker logic so that allocation client sends the allocation to GB

src/workers/AGENTS.md
```diff
@@ -121,15 +159,25 @@ export default worker;
...
+// For TypedWorker instances (recommended)
 export const typedWorkers: BaseTypedWorker<any>[] = [
   // ... existing workers
   myNewWorker,
 ];
+
+// For legacy Worker instances
+export const workers: Worker[] = [
+  // ... existing workers
+];
```
capJavert: Basically never use this array, we only support typed workers moving forward

capJavert (approved):
Nice improvements, few comments

src/workers/AGENTS.md
idoshamun: The note above already mentions it. but I'll push a clarification in couple places

PR #3328, opened 2025-12-08
feat: enhance opportunityMatches query with status filtering and total count

idoshamun:
- Added optional status filter to the opportunityMatches query.
- Updated OpportunityMatchConnection to include totalCount for matches.

### DIFF
--- a/src/schema/opportunity.ts
+++ b/src/schema/opportunity.ts
@@ -65,6 +65,7 @@ import {
   opportunityUpdateStateSchema,
   createSharedSlackChannelSchema,
   parseOpportunitySchema,
+  opportunityMatchesQuerySchema,
 } from '../common/schema/opportunities';
 import { OpportunityKeyword } from '../entity/OpportunityKeyword';
 import {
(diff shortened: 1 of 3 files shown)


src/schema/opportunity.ts
```diff
@@ -1061,44 +1093,57 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
-          queryBuilder: (builder) => {
-            builder.queryBuilder
-              .where({ opportunityId: args.opportunityId })
-              .andWhere(`${builder.alias}.status IN (:...statuses)`, {
-                statuses: [
+      // If status is provided, filter by that status; otherwise use all allowed statuses
+      const statusesToFilter = args.status ? [args.status] : allowedStatuses;
+
+      const [connection, totalCount] = await Promise.all([
+        queryPaginatedByDate<GQLOpportunityMatch, 'updatedAt', typeof args>(
```
capJavert: Is this now used for tinder style matching? Do we want to make sure user can't browse multiple candidates if doing direct API requests and requesting more candidates, and only do one by one? (could be something we add later, just for shenanigans protection) 😆

capJavert (approved):

src/schema/opportunity.ts
```diff
@@ -1044,14 +1059,31 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
 
+      // Allowed statuses for this query
+      const allowedStatuses = [
+        OpportunityMatchStatus.CandidateAccepted,
+        OpportunityMatchStatus.RecruiterAccepted,
+        OpportunityMatchStatus.RecruiterRejected,
+      ];
+
+      // Validate status if provided
+      if (args.status && !allowedStatuses.includes(args.status)) {
```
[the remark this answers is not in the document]
capJavert: look at `editOpportunity` mutation in this same file

src/schema/opportunity.ts
rebelchris: Where do we show it as list? (just for my own reference)

PR #3342, opened 2025-12-11
fix: return all opportunities

rebelchris:
We should expose all opps not just live ones now in self serve


src/schema/opportunity.ts
```diff
@@ -1110,8 +1107,9 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
         { key: 'createdAt', maxSize: 50 },
         {
           queryBuilder: (builder) => {
-            builder.queryBuilder.where({ state: opportunityState });
-
+            if (args?.state) {
+              builder.queryBuilder.where({ state: args.state });
```
capJavert: we need to validate here with zod no, its protoenumvalue so people could send whatever?

PR #3345, opened 2025-12-11
feat: brief read links

capJavert:


src/schema/feeds.ts
```diff
@@ -1581,11 +1582,14 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
         builder.limit(limit).offset(offset),
       {
-        fetchQueryParams: async (ctx): Promise<void> => {
-          if (!ctx.isTeamMember) {
-            throw new ForbiddenError(
-              'Access denied! You need to be authorized to perform this action!',
-            );
+        fetchQueryParams: async (ctx, { postIds }): Promise<void> => {
+          // limit to 50 for non-team members, eg. for brief
+          if (postIds.length > briefingPostIdsMaxItems) {
```
capJavert: this feed was blocked for non team members, I now allow if only 10 posts are sent, which is max for brief for now

else we block to disallow abuse

PR #3350, opened 2025-12-15
feat: opportunity location links

rebelchris:
Added new location system, it basically uses dataset one so we can ensure the iso2 codes.

### DIFF
--- a/__tests__/helpers.ts
+++ b/__tests__/helpers.ts
@@ -500,6 +500,7 @@ export const createMockBrokkrTransport = () =>
                 country: 'USA',
                 city: 'San Francisco',
                 subdivision: 'CA',
+                iso2: 'US',
                 type: 1,
               }),
             ],
--- a/__tests__/schema/opportunity.ts
+++ b/__tests__/schema/opportunity.ts
@@ -5262,6 +5262,9 @@ describe('mutation parseOpportunity', () => {
 
     await deleteKeysByPattern(`${rateLimiterName}:*`);
 
+    // Ensure dataset locations are available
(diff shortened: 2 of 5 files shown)


src/common/opportunity/prompt.ts
```diff
@@ -6,19 +6,20 @@ import {
...
 import { textFromEnumValue } from '../protobuf';
 
-export const createOpportunityPrompt = ({
+export const createOpportunityPrompt = async ({
   opportunity,
 }: {
   opportunity: OpportunityJob;
 }) => {
+  const locations = await opportunity.locations;
+  const firstLocation = locations?.[0];
```
idoshamun: where are we planning to do with other locations? (i know that it used to be this way even before this PR)

src/graphorm/index.ts
```diff
@@ -1558,6 +1555,13 @@ const obj = new GraphORM({
           childColumn: 'opportunityId',
         },
       },
+      locations: {
```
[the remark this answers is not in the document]
rebelchris: Yep :)

capJavert (approved):

src/schema/autocompletes.ts
```diff
@@ -107,6 +107,15 @@ export const resolvers = traceResolvers<unknown, BaseContext>({
...
 
+      return [
+        {
+          id: 'custom-123',
+          country: 'The Netherlands',
+          city: 'Amsterdam',
+          subdivision: 'Noord-Holland',
+        },
+      ];
+
```
omBratteng: Just test data?

PR #3352, opened 2025-12-15
feat: recruiter organization seats

capJavert:
- move subscription info to organization tied to opportunity (still a bit clunky but allows us to move in a direction we want in the future)
- added seats check for moving opportunity to live
- added mutation to add more seats to opportunity organization (will be used for boost)
- currently not removing seat on cancellation, will finish frontend flow and then see how it looks

### DIFF
--- a/__tests__/fixture/opportunity.ts
+++ b/__tests__/fixture/opportunity.ts
@@ -19,6 +19,8 @@ import {
 import type { QuestionScreening } from '../../src/entity/questions/QuestionScreening';
 import type { QuestionFeedback } from '../../src/entity/questions/QuestionFeedback';
 import { demoCompany } from '../../src/common';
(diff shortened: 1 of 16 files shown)


src/common/paddle/recruiter/processing.ts
```diff
@@ -47,14 +49,33 @@ export const createOpportunitySubscription = async ({
...
+  const organization = await opportunity.organization;
+
+  if (!organization) {
+    throw new Error(
+      'Opportunity does not have organization during payment processing, can not assign subscription, manual fixup needed',
+    );
+  }
+
+  if (
+    organization.recruiterSubscriptionFlags?.status ===
```
rebelchris: We sure we have org at this point? as it's now a new flow where they can change lots of things basically?
I think so but might be some edge-cases?

PR #3356, opened 2025-12-17
fix: for analyze table location and open to work

rebelchris:
Analyze table:
- Open to work should be yes by default, only no if they explicitly opted out (status 1)
- Added location from geo as well with verified false option.

### DIFF
--- a/src/schema/opportunity.ts
+++ b/src/schema/opportunity.ts
@@ -5,7 +5,7 @@ import { GraphQLResolveInfo } from 'graphql';
 import { FileUpload } from 'graphql-upload/GraphQLUpload.js';
 import { traceResolvers } from './trace';
 import { AuthContext, BaseContext, type Context } from '../Context';
-import graphorm from '../graphorm';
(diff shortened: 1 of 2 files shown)


src/schema/opportunity.ts
```diff
@@ -471,6 +471,11 @@ export const typeDefs = /* GraphQL */ `
     """
     location: String
 
+    """
+    Whether the location is verified (from dataset_location or custom) or inferred from geo flags
+    """
+    locationVerified: Boolean
```
idoshamun: Another thought, should we consider an enum here for future facing? (non blocking)

idoshamun (approved):

src/schema/opportunity.ts
rebelchris: enum as in? I think it can only be true/false right (it's technically user-provided not user-verified)

What would you say the enum would hold?

rebelchris: Good idea, i'll do that quickly

PR #3357, opened 2025-12-17
feat: start using common orb

omBratteng:


.circleci/config.yml
```diff
@@ -71,96 +44,34 @@ jobs:
...
     jobs:
-      - build
+      - build-app
       - test:
           requires:
-            - build
+            - build-app
           filters:
             branches:
-              ignore: master
```
omBratteng: Noticed that at some point we might've not ran tests on main branch, but because it was named `master` before, this filter was never applied. Not sure if we want to always run it like we're doing now, or don't run on main branch?
@capJavert

omBratteng: 👍 I'll remove that filter then

PR #3380, opened 2025-12-29
feat: add mock functionality for opportunity-related services

idoshamun:
- Introduced mock implementations for opportunity parsing, preview, and screening questions to facilitate local development without external service dependencies.
- Updated SnotraClient to utilize mock engagement profiles when enabled.
- Added configuration options in Pulumi for enabling mock external services.

### DIFF
--- a/src/schema/opportunity.ts
+++ b/src/schema/opportunity.ts
@@ -120,7 +120,7 @@ import {
 } from '../types';
 import { garmScraperService } from '../common/scraper';
 import { Storage } from '@google-cloud/storage';
-import { randomUUID } from 'node:crypto';
+import { randomUUID, randomInt } from 'node:crypto';
 import { addOpportunityDefaultQuestionFeedback } from '../common/opportunity/question';
 import { cursorToOffset, offsetToCursor } from 'graphql-relay/index';
 import { getShowcaseCompanies } from '../common/opportunity/companies';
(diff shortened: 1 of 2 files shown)


src/schema/opportunity.ts
```diff
@@ -939,7 +945,6 @@ export const typeDefs = /* GraphQL */ `
     Parse an opportunity from a URL or file upload
     """
     parseOpportunity(payload: ParseOpportunityInput!): Opportunity!
-      @rateLimit(limit: 5, duration: 3600)
```
rebelchris: Test only

idoshamun: Good catch. Fixed

idoshamun:
@rebelchris i tried now to make it as close as possible, but the opportunity preview was more challenging due to its async nature. i had to hack it a bit

src/mocks/opportunity/services.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+export const isMockEnabled = (): boolean =>
+  process.env.MOCK_EXTERNAL_SERVICES === 'true';
+
+/**
+ * Mock response for Brokkr parseOpportunity service call
+ * Returns plain objects that match the expected protobuf message shape
+ */
+export const mockBrokkrParseOpportunityResponse = () => ({
+  opportunity: {
```
rebelchris: nit: doesn't this expect proto type response?

rebelchris (approved):
Nice! actually also made it a lot leaner this way 👀

PR #3387, opened 2026-01-01
refactor: sanitize Brokkr response by filtering invalid locations in parseOpportunity function

idoshamun:


src/common/opportunity/parse.ts
```diff
@@ -199,9 +200,19 @@ export async function parseOpportunityWithBrokkr(
...
-    const parsedOpportunity = await opportunityCreateParseSchema.parseAsync(
-      result.opportunity,
-    );
+    // Sanitize Brokkr response - filter out invalid locations
+    const sanitizedOpportunity = {
+      ...result.opportunity,
+      location: Array.isArray(result.opportunity?.location)
+        ? result.opportunity.location.filter((loc: Location) => {
+            // Only keep locations with valid country
+            return loc?.country && loc.country.trim().length > 0;
```
capJavert: think actually we should look at those with iso2, since that one is actually important for process down the line

capJavert: should be both, gondul looks only at ISO so it can't do anything with country name

PR #3397, opened 2026-01-05
feat: support HTML content from WYSIWYG editor for job content

idoshamun:
## Summary
- Add sanitizeHtml function to sanitize HTML from rich text editor
- Add isHtmlContent helper to detect HTML vs markdown content
- Update renderOpportunityContent to handle both HTML (new) and markdown (legacy)

## Test plan
- [ ] Verify new HTML content is properly sanitized
- [ ] Verify legacy markdown content still renders correctly
- [ ] Check that dangerous HTML tags/attributes are stripped

### DIFF
--- a/__mocks__/isomorphic-dompurify.ts
+++ b/__mocks__/isomorphic-dompurify.ts
@@ -4,6 +4,12 @@ const DOMPurify = {
     // Simple mock that returns the input - actual sanitization not needed in tests
     return html;
   },
+  addHook: (): void => {
+    // No-op in tests
+  },
+  removeHook: (): void => {
+    // No-op in tests
+  },
 };
 
 export default DOMPurify;


src/common/markdown.ts
```diff
@@ -8,6 +9,27 @@ import { EntityTarget } from 'typeorm/common/EntityTarget';
...
+    ADD_ATTR: ['target', 'rel'],
+  }).replace(/<a /g, '<a target="_blank" rel="noopener nofollow" ');
+};
+
+/**
+ * Detects if content appears to be HTML (contains HTML tags)
+ */
+export const isHtmlContent = (content: string): boolean => {
+  // Check for common HTML tags used in rich text
+  return /<\/?(?:p|strong|em|ul|ol|li|a|br)[^>]*>/i.test(content);
```
capJavert: We should match open and closing tag as well, for example `<strong>whatever else I want here!!! ### ....`. 

Think HTML content should be always wrapped in a section or parent div. Also markdown can have mix of html links and images, for example github allows `` tag. We don't have to support it but somebody pasting markdown from github to us could false detect.

src/common/markdown.ts
```diff
@@ -8,6 +9,27 @@ import { EntityTarget } from 'typeorm/common/EntityTarget';
...
+ * Sanitizes HTML content, allowing only safe tags for rich text content.
+ * Used for opportunity content sections (WYSIWYG editor output).
+ */
+export const sanitizeHtml = (html: string): string => {
+  return DOMPurify.sanitize(html, {
+    ALLOWED_TAGS: ['p', 'strong', 'em', 'ul', 'ol', 'li', 'a', 'br'],
+    ALLOWED_ATTR: ['href', 'target', 'rel'],
+    // Force all links to open in new tab with nofollow
+    ADD_ATTR: ['target', 'rel'],
+  }).replace(/<a /g, '<a target="_blank" rel="noopener nofollow" ');
```
rebelchris: Doesn't this add double target/rel if that already exists?

idoshamun:
@claude we don't need to the markdown support for this feature anymore, so you can remove it from here

PR #3406, opened 2026-01-07
fix: plus sub updated race condition

capJavert:


capJavert:
@rebelchris made some changes to throw out status boolean and deduce from subscription status

rebelchris (approved):
Oh even better!

PR #3415, opened 2026-01-08
feat: dataset location continent

capJavert:
- add continent to DatasetLocation
- this is now used to match continent based locations as Europe
- refactored autocomplete to support continent matching
- added same indexes and updated unique checks to include continent
- updated opportunity preview handling to send just continent to gondul (this will make proper match)
- updated parsing to match and save based on continent (if only returned without country)
- made iso2, iso3 and country nullable to allow only continent dataset


capJavert:
@claude review again, addressed critical feedback, z.literal in zod supports array of values, check https://zod.dev/api?id=literals

capJavert:
@claude please update relevant AGENT.md files to make sure that project is using zod 4.x to avoid further confusion.

PR #3423, opened 2026-01-09
perf: add PostgreSQL performance flags for CI tests

idoshamun:
Add performance flags to the PostgreSQL container in CircleCI:
- fsync=off - Don't wait for writes to hit disk
- synchronous_commit=off - Don't wait for WAL flush
- full_page_writes=off - Skip extra safety writes
- max_connections=200 - Allow more connections

These are safe for tests since we don't need durability. Expected impact: 5-10% faster database writes.

Closes ENG-281

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,12 @@ jobs:
         environment:
           POSTGRES_DB: api_test
           POSTGRES_PASSWORD: 12345
+        command: >
+          postgres
+          -c fsync=off
+          -c synchronous_commit=off
+          -c full_page_writes=off
+          -c max_connections=200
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -28,6 +28,12 @@ jobs:
         environment:
           POSTGRES_DB: api_test
           POSTGRES_PASSWORD: 12345
+        command: >
```
omBratteng: Also upgrade to psql 18!

omBratteng (approved):

.circleci/config.yml
idoshamun: good catch

PR #3427, opened 2026-01-11
feat: implement backwards compatibility for social links migration

idoshamun:
This commit implements full backwards compatibility for the flexible social links
migration, ensuring legacy social field columns continue to work while reading from
the new socialLinks JSONB column.

Changes:
- GraphQL (GraphORM): Added field resolvers for all 13 legacy social fields
  (twitter, github, linkedin, threads, roadmap, codepen, reddit, stackoverflow,
  youtube, bluesky, mastodon, hashnode, portfolio) that derive values from socialLinks
- Boot endpoint: Updated to read from socialLinks and populate legacy fields for
  backwards compatibility with older clients (especially browser extension)
- Common utilities: Moved extractHandleFromUrl to common/schema/socials.ts for
  code reuse across GraphQL and boot endpoint
- Tests: Added test to verify legacy fields are resolved from socialLinks, updated
  user fixtures to use socialLinks instead of legacy columns
- Dual-write: Maintained existing dual-write logic for mutations to ensure both
  socialLinks and legacy columns stay in sync when updates occur

This ensures the browser extension and other clients can continue using legacy fields
while the system internally reads from the new socialLinks column. Legacy columns can
be safely dropped in Phase 3 after all clients are updated.


rebelchris (approved):
seems valid enough

PR #3438, opened 2026-01-13
fix: anon user

rebelchris:


src/schema/opportunity.ts
```diff
@@ -1830,23 +1848,49 @@ export const resolvers: IResolvers<unknown, BaseContext> = traceResolvers<
...
+      const users = await ctx.con.getRepository(User).find({
+        where: { id: In(userIds) },
+        select: ['id', 'experienceLevel', 'flags'],
+      });
+
+      for (const user of users) {
+        const flags = (user.flags ?? {}) as Record<string, unknown>;
+        userContextMap.set(user.id, {
+          seniority: user.experienceLevel ?? null,
+          locationCountry: (flags.country as string) ?? null,
```
rebelchris: @claude I just realized country can be either this or from user_candidate_preference the latter should take priority, you can find some examples in the code already.

PR #3794, opened 2026-04-14
fix: ios logout issues

rebelchris:


capJavert (approved):

src/routes/boot.ts
```diff
@@ -925,12 +926,17 @@ export const getBootData = async (
...
-        ),
-      })) as BetterAuthSession | null;
+      const session = (await asyncRetry(
+        () =>
+          getBetterAuth().api.getSession({
+            headers: fromNodeHeaders(
+              req.headers as Record<string, string | string[] | undefined>,
+            ),
+          }),
+        { retries: 1, minTimeout: 50 },
```
idoshamun: why only one retry though?

rebelchris: We can do more but 2 tries in total seemed ok to catch edge-cases?

rebelchris:
```suggestion
        { retries: 3, minTimeout: 50 },
```

idoshamun: retries 1 means we will try only once and then stop, doesn't it?
it doesn't mean we will try once and then another time on failure. but maybe i'm wrong

idoshamun (approved):

src/routes/boot.ts
```diff
@@ -925,12 +926,17 @@ export const getBootData = async (
...
-        ),
-      })) as BetterAuthSession | null;
+      const session = (await asyncRetry(
+        () =>
+          getBetterAuth().api.getSession({
+            headers: fromNodeHeaders(
+              req.headers as Record<string, string | string[] | undefined>,
+            ),
+          }),
+        { retries: 3, minTimeout: 50 },
```
idoshamun:
```suggestion
        { retries: 3 },
```

PR #3798, opened 2026-04-15
feat: async user deletion via background worker

idoshamun:
## Summary
- Replace synchronous user deletion (CASCADE-locked 60 FK tables for 10-20min, blocking logins) with an async flow
- `deleteUser()` now sets `flags.inDeletion=true`, invalidates sessions, and returns immediately
- New `userDeletionCleanup` worker handles child table cleanup one table at a time
- Boot endpoint blocks login for users marked `inDeletion`
- Zombie cron marks users for deletion instead of deleting directly

## Test plan
- [x] Build passes
- [x] Lint passes (0 warnings)
- [x] `__tests__/users.ts` delete tests pass (19/19)
- [x] `__tests__/cron/cleanZombieUsers.ts` passes (3/3)
- [ ] Companion PR: dailydotdev/streams#93 (topic must be deployed first)
- [ ] Follow-up: migration to remove CASCADE from user FK constraints

### DIFF
--- a/src/common/user.ts
+++ b/src/common/user.ts
@@ -1,38 +1,20 @@
 import { DataSource } from 'typeorm';
-import {
-  Alerts,
-  ArticlePost,
-  Bookmark,
-  BookmarkList,
-  Comment,
-  DevCard,
-  Feed,
-  Organization,
-  Post,
-  PostReport,
-  Settings,
-  SourceDisplay,
-  SourceRequest,
-  User,
-  View,
-} from '../entity';
+import { Organization, User } from '../entity';
 import { OpportunityJob } from '../entity/opportunities/OpportunityJob';
 import { OpportunityUser } from '../entity/opportunities/user';
-import { ghostUser } from './index';
+import { updateFlagsStatement } from './utils';
 import { cancelSubscription } from './paddle';
 import type { AuthContext, Context } from '../Context';
 import { ForbiddenError } from 'apollo-server-errors';
 import { logger } from '../logger';
 import { CoresRole } from '../types';
 import { remoteConfig } from '../remoteConfig';
-import { UserTransaction } from '../entity/user/UserTransaction';
 import { SubscriptionProvider, SubscriptionStatus } from './plus';
 import {
   deleteEmploymentAgreementByUserId,
   deleteResumeByUserId,
 } from './googleCloud';
 import { ConflictError } from '../errors';
-import { DigestPost } from '../entity/posts/DigestPost';
 
 export const deleteUser = async (
   con: DataSource,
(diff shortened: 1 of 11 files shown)


src/common/user.ts
```diff
@@ -76,110 +58,60 @@ export const deleteUser = async (
...
+
+      if (orgsWithActiveSubscription > 0) {
+        throw new ConflictError(
+          'Cannot delete your account because one of your organizations has an active recruiter subscription. Please cancel the subscription first.',
+        );
+      }
+    }
+
     // Delete user's resume if exists
     await deleteResumeByUserId(userId);
```
rebelchris: Why not put these in the cleanup deletion as well?

rebelchris (approved):
Does it not still lock once it hits the cdc worker then?

src/workers/userDeletionCleanup.ts
```diff
@@ -0,0 +1,247 @@
...
+      [userId],
+    );
+    const affected = result?.[1] ?? 0;
+    totalDeleted += affected;
+    if (affected < BATCH_SIZE) {
+      break;
+    }
+  }
+  if (totalDeleted > 0) {
+    log.info({ table, totalDeleted, userId }, 'batch deleted rows');
```
idoshamun:
```suggestion
    log.debug({ table, totalDeleted, userId }, 'batch deleted rows');
```

src/workers/userDeletionCleanup.ts
```diff
@@ -0,0 +1,247 @@
...
+  }
+};
+
+const worker: TypedWorker<'api.v1.user-deletion-requested'> = {
+  subscription: 'api.user-deletion-cleanup',
```
capJavert, line 108: would there be any way where this fails and we have to enable user back?

src/workers/userDeletionCleanup.ts
```diff
@@ -0,0 +1,247 @@
...
+    await con
+      .getRepository(PostMention)
+      .update({ mentionedUserId: userId }, { mentionedUserId: ghostUser.id });
+    await con
+      .getRepository(CommentMention)
+      .update({ commentByUserId: userId }, { commentByUserId: ghostUser.id });
+    await con
+      .getRepository(CommentMention)
+      .update({ mentionedUserId: userId }, { mentionedUserId: ghostUser.id });
+    await con
```
idoshamun, line 149: We also need `grantFromId` here

src/workers/userDeletionCleanup.ts
```diff
@@ -0,0 +1,247 @@
...
+      .getRepository(ReputationEvent)
+      .update({ grantToId: userId }, { grantToId: ghostUser.id });
+    await con
+      .getRepository(UserTransaction)
+      .update({ senderId: userId }, { senderId: ghostUser.id });
+    await con
+      .getRepository(UserTransaction)
+      .update({ receiverId: userId }, { receiverId: ghostUser.id });
+    await con
```
idoshamun: Why do we have a statement for article post and post? also we should set the author as ghost user, not null

src/workers/userDeletionCleanup.ts
```diff
@@ -0,0 +1,247 @@
...
+      .getRepository(ArticlePost)
+      .update({ authorId: userId }, { authorId: null });
+    await con.getRepository(DigestPost).delete({ authorId: userId });
+    await con
```
idoshamun, line 162: why do we need this?

src/workers/userDeletionCleanup.ts
idoshamun: what do you mean?

src/common/user.ts
idoshamun: yap it's possible and better

idoshamun:
> Does it not still lock once it hits the cdc worker then?

@rebelchris no, because it's not a transaction and not one statement. so you can acquire lock per table, rather than ~60 locks

src/workers/userDeletionCleanup.ts
capJavert: For example we figure out that while this was processing we cant delete user after all due to some constraint?

idoshamun: I check the constraints in the endpoint already. I didn't touch it. If it fails there it's only for technical reasons we should fix without letting the user know

src/common/user.ts
idoshamun: Good point — moved resume and employment agreement deletion to the worker. They're not time-sensitive and don't need to block the user-facing response.

src/workers/userDeletionCleanup.ts
idoshamun: Fixed — changed to `log.debug`.

src/workers/userDeletionCleanup.ts
idoshamun: Added — now ghosting both `grantToId` and `grantById` on ReputationEvent.

src/workers/userDeletionCleanup.ts
idoshamun: Removed the separate ArticlePost line — the `Post.update({ authorId: userId }, { authorId: ghostUser.id })` already handles all post types including articles. No more null, everything goes to ghost.

src/workers/userDeletionCleanup.ts
idoshamun: Removed. The SourceUser is already ghosted (userId → ghost), and the general `Post.update({ authorId: userId }, { authorId: ghostUser.id })` catches all posts regardless of sourceId. The separate sourceId update was redundant.

PR #3804, opened 2026-04-16
chore: cleanup prompts

rebelchris:
Cleanup prompts


idoshamun (approved):
We just need to make sure old clients won't break all of a sudden (except when using smart prompts)

PR #3806, opened 2026-04-17
fix(auth): set up new user defaults in betterAuth hook

capJavert:
src/betterAuth.ts
  - resolveSignUpUserId + ensureNonDeletedUserId helpers — deleted-user collision handled once at the end of the before hook
  - user.create.after now sets: feed (id = user.id), notificationFlags (marketing muted when acceptedMarketing=false), UserAction CheckedCoresRole, DigestPost
  flags/referral/timezone/coresRole UPDATE
  - Inline calls to addClaimableItemsToUser + claimAnonOpportunities (each in its own try/catch)

  src/entity/user/utils.ts
  - addClaimableItemsToUser signature narrowed to Pick<AddUserData, 'id' | 'email'>

### DIFF
--- a/__tests__/routes/betterAuth.ts
+++ b/__tests__/routes/betterAuth.ts
@@ -175,13 +175,23 @@ describe('betterAuth routes', () => {
         overrides: {
           body?: Record<string, unknown>;
           cookie?: string;
+          ip?: string;
         } = {},
-      ) => ({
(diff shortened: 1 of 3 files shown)


rebelchris (approved):

capJavert:
@claude review

PR #3847, opened 2026-05-03
feat: add live room lobby api

idoshamun:
## Summary
- add scheduled live-room lobby fields, markdown descriptions, and post embed support
- add live-room subscriptions with hard-delete unsubscribe and cleanup after go-live
- add default-enabled live-room-started notifications that skip currently connected subscribers when Flyting can report them

## Validation
- `NODE_ENV=test pnpm run db:migrate:reset`
- `pnpm run lint`
- `pnpm run build`
- `pnpm exec jest __tests__/liveRooms.ts __tests__/workers/liveRoomLifecycle.ts --runInBand`

### DIFF
--- a/src/schema/liveRooms.ts
+++ b/src/schema/liveRooms.ts
@@ -3,6 +3,7 @@ import { ForbiddenError, ValidationError } from 'apollo-server-errors';
 import type { GraphQLResolveInfo } from 'graphql';
 import type { Context } from '../Context';
 import { toGQLEnum } from '../common';
+import { GQLEmptyResponse } from './common';
 import { createLiveRoomJoinToken } from '../common/liveRoom/token';
 import {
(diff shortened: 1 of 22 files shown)


src/schema/liveRooms.ts
```diff
@@ -220,6 +260,30 @@ export const resolvers: IResolvers = {
...
+      if (!ctx.userId) {
+        return false;
+      }
+
+      return ctx.con.getRepository(LiveRoomSubscription).existsBy({
+        roomId: room.id,
+        userId: ctx.userId,
+      });
+    },
+    contentEmbeds: async (room: GQLLiveRoom, _, ctx: Context) =>
```
idoshamun: We should use GraphORM for this. See post and comment

PR #3878, opened 2026-05-14
feat: pnpm 10 and deps hardening

capJavert:
Basically copy 🍝 from what @rebelchris did with apps. Have to test yet if all works.


rebelchris (approved):
looks ok on paper

PR #3974, opened 2026-07-02
feat(posts): support scheduled publishing

idoshamun:
## Summary

- Add scheduled freeform post creation using `flags.scheduledAt` and hidden visibility until publish time.
- Add scheduled post listing plus schedule editing through the existing post edit mutation.
- Wire CDC to Temporal so scheduled posts publish later, refresh `createdAt`/score on publish, and cancel stale workflows on edit/delete/publish.
- Add a partial index for scheduled post flag lookups and focused GraphQL/Temporal tests.

## Validation

- `pnpm run build`
- `pnpm run lint`
- `NODE_ENV=test npx jest __tests__/posts.ts --testEnvironment=node --runInBand`
- `NODE_ENV=test npx jest __tests__/temporal/notifications/activities.ts __tests__/temporal/notifications/utils.ts --testEnvironment=node --runInBand`
- `NODE_ENV=test npx jest __tests__/temporal/notifications/workflows.ts --testEnvironment=node --runInBand`


idoshamun (commented):
I see a lot `Object.prototype` tricks which I dont really like and we have no where else in the codebase.

Have we added the new workflow to our temporal service here so it won't be left out?

PR #4033, opened 2026-07-30
fix: resolve the trusted client IP behind the load balancer

rebelchris:
## Context

A security report claimed `daily.dev`'s password reset endpoint has no rate limiting, and that 100 reset emails could be sent to a user via Burp Intruder.

The literal claim is wrong — better-auth 1.5.0 applies a built-in rule of **3 requests / 60s** to `/request-password-reset`, it's enabled in prod (`nodeEnv: production`), and it's Redis-backed so it's shared across pods. But while verifying that, I found the limit doesn't actually bind, for a reason that turned out to be much broader than password reset.

## The bug

`src/index.ts` configured Fastify with `trustProxy: true`. That makes `request.ip` resolve to the **leftmost** `x-forwarded-for` entry — which is the one the client sends, not the one the load balancer vouches for.

The intuitive assumption is that a proxy strips a client-supplied `X-Forwarded-For`. GCP does **not** do this by default; it appends:

```
X-Forwarded-For: <client-supplied>,<client-ip>,<load-balancer-ip>
```

> The load balancer does not verify any IP addresses that precede `<client-ip>,<load-balancer-ip>` in this header.
> — [External Application Load Balancer overview](https://docs.cloud.google.com/load-balancing/docs/https)

Stripping is opt-in, via a backend-service custom request header (`--custom-request-header=x-forwarded-for:{client_ip_address},{server_ip_address}`). We have not configured that — there is no `custom-request-header`, `BackendConfig`, or Cloud Armor policy anywhere in `daily-infra`, and no Cloudflare in front of `api.daily.dev`.

### Confirmed against production

`/boot` derives `geo.region` from `request.ip`, so this is directly observable today:

| Request to `https://api.daily.dev/boot` | `geo.region` |
|---|---|
| no header | `ZA` (real egress IP) |
| `X-Forwarded-For: 133.11.0.1` | `JP` |
| `X-Forwarded-For: 200.160.2.3` | `BR` |

`request.ip` is whatever the caller claims.

**Password reset was the least of it.** Everything keyed on `request.ip` is affected:

| Consumer | Impact |
|---|---|
| `routes/webhooks/paddle` | IP allowlist bypassable (signature check still holds, so defense-in-depth only) |
| `routes/webhooks/apple` | Same — JWS verification still holds |
| `common/vordr.ts` | Anti-abuse IP checks bypassable |
| `routes/public`, `routes/outbound`, `directive/rateLimit.ts` | Rate limit keys rotatable at will |
| `Context.ts`, `entity/user/utils.ts`, `schema/users.ts`, `routes/boot.ts` | Geo → `coresRole` spoofable, i.e. self-selected pricing region at signup |

Sign-in throttling (3/10s) was bypassable the same way, which matters more than the reported email flood.

## The fix

`trustProxy: 2` — trust the two hops the load balancer guarantees and discard the client-supplied prefix. One setting, fixes every consumer above.

The hop count of 2 assumes nothing appends to `x-forwarded-for` after the load balancer. Verified: the ingress is a plain VPC-native GKE `Ingress` (container-native LB, GFE → pod directly), with no service mesh, no Envoy/nginx sidecar, and no init container in the api deployment. So the address list is exactly `[GFE, lb-vip, client, ...client-supplied]` and trusting 2 hops lands on the real client.

better-auth parses `x-forwarded-for` itself and takes the first entry, with no trusted-proxy support (`trustedProxies` exists only in a warning string in 1.5.0). Since `callBetterAuth` is the sole entry point into its handler, the header is collapsed there to the resolved IP.

## Verification

- New test asserts a client-supplied prefix is discarded, using the real LB header shape (`<spoofed>,<client>,<lb>`).
- Existing suites pass unchanged: `routes/betterAuth` (31), `webhooks/apple`, `webhooks/paddle`, `boot`, `outbound` (164 total), plus all 19 `routes/public` + `routes/private` suites. Tests that set a single-entry `x-forwarded-for` are unaffected by the hop count.
- `pnpm run build` and `pnpm run lint` clean.

**Post-deploy check** — this should stop tracking the spoofed header and report the real egress region instead of `JP`:

```
curl -s https://api.daily.dev/boot -H 'X-Forwarded-For: 133.11.0.1' | jq .geo
```

If it instead returns an empty/unexpected region for *every* caller, the hop count is off by one and `request.ip` is resolving to the load balancer VIP — revert and switch to trusting Google's LB ranges plus the ingress VIP by CIDR.

## Deliberately not included

- **Per-recipient limit on reset.** The key is still `(ip, path)`, so a distributed IP pool can still flood one inbox. `rateLimit.customRules` can't change the key, so this needs a small Redis counter in the existing `hooks.before` middleware.
- **Turnstile on reset.** Already integrated but scoped to `['/sign-up/email', '/sign-in/email']`. Adding reset requires a matching `apps` change to send the token, so it can't ship from this repo alone.

## Unrelated, worth a separate look

Kratos is still deployed in `heimdall` and still serves `/self-service/*` at `sso.daily.dev` via a catch-all ingress rule, with recovery enabled in every checked-in config — a second reset-email surface, rate-limited by Cloud Armor at 30/60s. `heimdall/services/ba_user_creator` also mirrors better-auth users into Kratos with their password hashes, so a Kratos-side recovery could set a credential better-auth doesn't honor. Confirming prod requires reading the live `kratos-config` secret, which is gitignored.

---

Replaces #4032, which was accidentally branched off `feat-squad-privileged-rate-limit` and so carried that PR's commits in its diff.


capJavert (approved):
Since we don't want added cost this is a solution for API 👍

PR #4055, opened 2026-08-06
feat: tool votes, discussion and sitemap

rebelchris:
Phase 2.5 of the tools surface + SEO plumbing (API side). Pairs with the apps PR (vote/discussion UI + JSON-LD).

### Votes
- `tool_vote` table (PK userId+toolId, smallint vote) with `voteTool(id, vote)` @auth — 1/-1 upserts, 0 deletes the row.
- `DatasetTool` gains `upvotes`/`downvotes` (GraphORM count subselects) and viewer-scoped `userVote` (nulled for anonymous).

### Discussion (reworked per review — no parallel comment table)
- Each tool lazily gets a **hidden discussion post** (`FreeformPost`, `showOnFeed: false`, new public `tools` machine source created in the migration like the `briefing` source) via `initToolDiscussion(id)` @auth — idempotent with a race-safe claim on `dataset_tool.discussionPostId`.
- Comments on tools are **ordinary post comments**: threading, comment votes, mentions, notifications, reporting and moderation all come from the existing machinery. No new comment table.

### Sitemap
- `/sitemaps/tools.xml`: the `/tools` hub + every tool with 3+ stacks (via `tool_stack_stats`), wired into the sitemap index next to tags.

## Testing
- 28 datasetTool integration tests (vote lifecycle incl. anonymous nulling, comment threading/permissions/markdown) + tools sitemap test (40 total in sitemaps suite).
- Comment notifications work out of the box (they are post comments).


src/schema/datasetTool.ts
```diff
@@ -398,4 +440,73 @@ export const resolvers: IResolvers<unknown, BaseContext> = {
...
+        .getRepository(DatasetTool)
+        .createQueryBuilder()
+        .update()
+        .set({ discussionPostId: post.id })
+        .where('id = :id AND "discussionPostId" IS NULL', { id: tool.id })
+        .execute();
+
+      if (!claimed.affected) {
+        // Lost a concurrent init; use the winner's post.
+        await ctx.con.getRepository(Post).delete({ id: post.id });
```
rebelchris: **Blocking:** the post insert and the claim on `dataset_tool.discussionPostId` are two separate transactions, so the write is not atomic in either direction.

- If the process dies (or the `UPDATE` fails) between the two statements, the freeform post is committed and never referenced by any tool. The next call creates a second one, so the orphan is permanent and invisible to any cleanup.
- The concurrency loser path compensates by hard-deleting a `post` row. That emits a CDC insert followed by a CDC delete for a post that was never real, and `notifyPostVisible` has already been published for it (`visible: true`), so `api.v1.post-visible` subscribers see a post that no longer exists by the time they fetch it.

Please do the whole thing in one `ctx.con.transaction(...)`: create the post and run the conditional `UPDATE ... WHERE "discussionPostId" IS NULL` inside it, and on `affected === 0` roll back and return the winner's id. That makes the compensating `Post` delete unnecessary and leaves no intermediate state. This also requires `createToolDiscussionPost` to accept `DataSource | EntityManager` (repo convention for reusable write helpers) instead of `DataSource` only.

Worth adding a test that runs two `initToolDiscussion` calls concurrently and asserts exactly one `post` row exists for the tool — the current test only covers sequential reuse, which passes even with the race.

Reviewed by AI.

src/common/datasetTool.ts
```diff
@@ -76,6 +80,34 @@ export const fetchAndUploadToolIcon = async (
...
+  const content = `Community discussion about ${tool.title} on daily.dev.`;
+
+  return con.getRepository(FreeformPost).save({
+    id,
+    shortId: id,
+    title: `${tool.title} discussion`,
+    sourceId: TOOLS_SOURCE,
+    content,
+    contentHtml: markdown.render(content),
+    visible: true,
```
rebelchris: **Blocking:** the post is saved with `visible: true` but no `visibleAt`, and no `origin`.

`visible && visibleAt IS NULL` is the state the codebase uses to mean "published flag set but never actually made visible" — see `src/workers/postUpdated/shared.ts:271` (`else if (data.visible && !databasePost.visibleAt) { data.visibleAt = ... }`) and every other creation path (`insertFreeformPost`, `createPollPost`, `createSharePost`, `channelDigest/generate.ts`, `liveRooms.ts`) which sets `visibleAt: new Date()` alongside `visible: true`. Persisting an inconsistent pair here means anything ordering or filtering on `visibleAt` treats these posts differently from every other post, and a later metadata update will silently backfill `visibleAt` with an unrelated timestamp.

Please set `visibleAt: new Date()` and an explicit `origin` (the other machine-generated paths set one rather than leaving the column null while the entity types it as non-null `PostOrigin`).

Reviewed by AI.

src/schema/datasetTool.ts
```diff
@@ -99,6 +128,19 @@ export const typeDefs = /* GraphQL */ `
...
+    """
+    Vote on a tool (1 up, -1 down, 0 to clear)
+    """
+    voteTool(id: ID!, vote: Int!): EmptyResponse! @auth
+
+    """
+    Get or create the tool's hidden discussion post; comments on it are
+    ordinary post comments
+    """
+    initToolDiscussion(id: ID!): ID! @auth
```
rebelchris: **Non-blocking:** `initToolDiscussion` is an authenticated post factory with no rate limit — any logged-in caller can walk every `dataset_tool` id and create one freeform post plus one `api.v1.post-visible` event per tool in a single loop. `createFreeformPost` and the other post-creating mutations carry `@rateLimit(limit: 1, duration: 30)`; please add the same directive here, or create the discussion lazily on the first `commentOnPost` instead of exposing a dedicated mutation.

Reviewed by AI.

rebelchris (commented):
Review of the tool votes / discussion / sitemap plumbing. The move to ordinary post comments on a hidden host post is the right call and the vote table + GraphORM subselects follow existing patterns. Two blocking items on the discussion-post write path, plus a few smaller notes inline.

**Question (cross-cutting):** the host post is `visible: true`, `private: false` in a public `tools` machine source, so it is a normal reachable post — `post(id)`/`/posts/:slug`, profile comment activity, related-post surfaces. Since this PR is the SEO half of the feature, can you confirm the apps side canonicals/noindexes that post page so `Next.js discussion` does not compete with `/tools/nextdotjs` in the index, and that these posts are excluded wherever posts are ranked or indexed downstream? `showOnFeed: false` covers the feed and `postVisibleInterestMatch`, but not the post page itself.

Checks: no tests or build executed here (review is static).

Reviewed by AI.

src/routes/sitemaps.ts
```diff
@@ -358,6 +363,23 @@ const buildTagsSitemapQuery = (
...
 
+const MIN_SITEMAP_TOOL_STACKS = 3;
+
+const buildToolsSitemapQuery = (
+  source: DataSource | EntityManager,
+): SelectQueryBuilder<DatasetTool> =>
+  source
+    .createQueryBuilder()
+    .select('dt."titleNormalized"', 'slug')
+    .addSelect('dt."updatedAt"', 'lastmod')
```
rebelchris: **Non-blocking:** `dataset_tool.updatedAt` is a plain column with `default now()` (`src/entity/dataset/DatasetTool.ts`) — it is not an `@UpdateDateColumn` and no trigger or write path touches it, so this `lastmod` is really `createdAt` and will never change when a tool is edited or gets new discussion activity. Either make the column actually update on write, or derive `lastmod` from something that moves (for example the discussion post's `metadataChangedAt` / stack stats refresh) so crawlers get a useful recrawl signal.

Reviewed by AI.

src/migration/1786100000000-ToolVoteAndComment.ts
```diff
@@ -0,0 +1,42 @@
...
+        CONSTRAINT "PK_tool_vote" PRIMARY KEY ("userId", "toolId"),
+        CONSTRAINT "FK_tool_vote_user_id" FOREIGN KEY ("userId") REFERENCES "user"("id") ON DELETE CASCADE ON UPDATE NO ACTION,
+        CONSTRAINT "FK_tool_vote_tool_id" FOREIGN KEY ("toolId") REFERENCES "dataset_tool"("id") ON DELETE CASCADE ON UPDATE NO ACTION
+      )`,
+    );
+    await queryRunner.query(
+      /* sql */ `CREATE INDEX IF NOT EXISTS "IDX_tool_vote_tool_id_vote" ON "tool_vote" ("toolId", "vote")`,
+    );
+    await queryRunner.query(
+      /* sql */ `INSERT INTO "public"."source" ("id", "name", "handle", "private", "type", "active", "image") VALUES ('tools', 'Tools', 'tools', false, 'machine', true, 'https://media.daily.dev/image/upload/s--LrHsyt2T--/f_auto/v1692632054/squad_placeholder_sfwkmj') ON CONFLICT DO NOTHING`,
```
rebelchris: **Non-blocking, three small things on this migration:**

1. Bare `ON CONFLICT DO NOTHING` swallows a conflict on any constraint, including `handle`. If a source already occupies handle `tools` under a different id, the insert silently no-ops and `initToolDiscussion` then fails at runtime with an FK violation on `sourceId = 'tools'`. Please confirm the handle is free, and consider targeting the conflict (`ON CONFLICT ("id") DO NOTHING`) so an unexpected handle collision fails loudly at deploy time instead of at first use.
2. Timestamp `1786100000000` is below `main`'s head migration (`1787085561243`), so this runs last in already-migrated environments but early on a fresh database — ordering differs between prod and CI/local. Please bump it to a current timestamp.
3. The migration name still says `ToolVoteAndComment`, but the review rework dropped the comment table. Rename to match what it does now.

Also: `tool_vote.vote` is an unconstrained `smallint` with `DEFAULT 0`. The mutation deletes instead of storing `0`, so a stored `0` row would count as neither upvote nor downvote and would surface as `userVote: 0` rather than `null`. A `CHECK ("vote" IN (-1, 1))` (with no default) would make the intended domain explicit.

Reviewed by AI.

src/schema/datasetTool.ts
rebelchris: Fixed in e582fa41: the post insert and the conditional `discussionPostId` UPDATE now run inside a single `ctx.con.transaction(...)`. On a lost race (`affected === 0`) we throw a local `ToolDiscussionRaceError` to roll back the transaction (no committed post, no compensating delete, no CDC insert+delete pair), then re-read the winner's `discussionPostId` outside the transaction and return that. `createToolDiscussionPost` now accepts `DataSource | EntityManager` so it can join the caller's transaction. `notifyPostVisible` is CDC-driven off the `post` table insert, so a rolled-back insert never reaches it. Added a concurrency test (two different users via `Promise.all`) asserting exactly one post row and the same returned id for both callers.

src/common/datasetTool.ts
rebelchris: Fixed in e582fa41: `createToolDiscussionPost` now sets `visibleAt: new Date()` alongside `visible: true`, and an explicit `origin: PostOrigin.UserGenerated` — matching the convention used by the other machine-generated host posts in this codebase (channelDigest/generate.ts, liveRooms.ts), which also use UserGenerated for system-created posts that host ordinary user interactions.

src/schema/datasetTool.ts
rebelchris: Added in e582fa41: `initToolDiscussion` now carries `@rateLimit(limit: 1, duration: 30)`, matching createFreeformPost and friends. Had to add a `deleteKeysByPattern(rateLimiterName:*)` reset in the test's beforeEach/between sequential calls to keep the existing reuse test working.

src/routes/sitemaps.ts
rebelchris: Fixed in e582fa41, without a migration: `dataset_tool.updatedAt` is now `@UpdateDateColumn` instead of a plain `default: now()` column. The underlying column type/default in Postgres is unchanged (TypeORMs @UpdateDateColumn is applied in application-level query building, not a DB-level ON UPDATE trigger, so there's no DDL drift). Verified every write path to dataset_tool goes through TypeORM (findOrCreateDatasetTool's repo.save, and initToolDiscussion's repo.createQueryBuilder().update()) rather than raw SQL, so @UpdateDateColumn actually fires on both creation and the discussionPostId claim. Confirmed via __tests__/sitemaps.ts (all 40 cases still pass) that this does not break existing sitemap behavior.

src/migration/1786100000000-ToolVoteAndComment.ts
rebelchris: All fixed in e582fa41. (1) Verified no other migration inserts a source with handle "tools" (the only other "tools" hits in the codebase are keyword/tag arrays, unrelated) — ON CONFLICT now targets ("id") DO NOTHING. (2) Renamed the migration to timestamp 1787130000000 (> mains current head 1787085561243) and (3) renamed the class/file to ToolVoteAndDiscussion1787130000000 since the comment table was dropped from this PR already. Confirmed via grep no other file in the branch references the old class name. (4) Added CHECK ("vote" IN (-1, 1)) and dropped DEFAULT 0 on tool_vote.vote in both the migration and the ToolVote entity, since voteTool deletes the row instead of storing 0. Re-ran db:migrate:reset locally against a fresh test db to confirm the renamed/edited migration applies cleanly, and the datasetTool/sitemaps test suites (76 tests) all pass against the new schema.

rebelchris:
Investigated both sides honestly — this is a real, currently-unaddressed gap on both api and apps.

**api side (src/routes/sitemaps.ts):** `buildPostsSitemapBaseQuery` (~line 161) filters by `POSTS_SITEMAP_EXCLUDED_TYPES` (Welcome, Collection, Share, Brief, SocialTwitter), `NOT private/banned/deleted`, `visible = true`, `createdAt > now() - 90 days`, and `(author is null OR reputation > 10)`. It does **not** filter on `sourceId` or `showOnFeed`. The tool discussion host post is a `PostType.Freeform` (not excluded), has no `authorId` (so `u.id is null` passes the reputation check), and is `visible: true`. So for up to 90 days after a tool page is first opened, its discussion post *will* be enumerated in `/api/sitemaps/posts-N.xml`, directly competing with `/tools/{slug}`. Concrete follow-up: add a `sourceId != 'tools'` exclusion to `buildPostsSitemapBaseQuery` (the `tools` source is dedicated solely to these host posts, so this is safe and precise).

**apps side (packages/webapp):** Checked `pages/posts/[id]/index.tsx` and `lib/seo.ts`. `shouldNoindexPost()` only noindexes on `post.private`, `post.source.public === false`, low-reputation author, or `post.type` in `THIN_NOINDEX_POST_TYPES` (Brief, SocialTwitter) — no check on `post.source.id`. Since the tools source is public and these posts are `private: false`, they pass through as indexable with `canonical` self-pointing to `/posts/{slug}`, same as any normal post. `/tools/[slug]` (built on this branch) does not reference the discussion post's canonical either way, so there's no cross-linking in either direction. No other machine/system post type (squad welcome, live room, digest) is excluded by `source.id` today either — the existing pattern is type-based, not source-based. Concrete follow-up: extend `shouldNoindexPost` in `packages/webapp/lib/seo.ts` with a check for `post.source?.id === 'tools'` (or a `THIN_NOINDEX_SOURCE_IDS` set), and optionally point `canonical` at `/tools/{slug}` for those posts once the tool slug is exposed on the post query.

Bottom line: nothing prevents indexing/competition today. I did not fix this in code here per the task scope (investigate + reply), but both fixes above are small and precise if you want them tracked as follow-ups.

rebelchris (commented):
Re-reviewed at `e582fa4`. All five previous findings verified fixed in code, not just claimed:

- Atomicity: post insert + conditional claim now run inside one `ctx.con.transaction`, the race loser rolls back via `ToolDiscussionRaceError` and re-reads the winner, and the compensating `Post` delete is gone. `createToolDiscussionPost` takes `DataSource | EntityManager`. The new concurrent test asserts a single `post` row.
- `visibleAt: new Date()` and `origin: PostOrigin.UserGenerated` now set, matching `channelDigest/generate.ts` and `liveRooms.ts`.
- `@rateLimit(limit: 1, duration: 30)` added.
- `dataset_tool.updatedAt` is now `@UpdateDateColumn`; I confirmed every write path to that table goes through TypeORM repositories/query builders (`src/common/datasetTool.ts`, `src/schema/datasetTool.ts`, `src/schema/autocompletes.ts`), so `lastmod` will actually move, and the DDL (`TIMESTAMP NOT NULL DEFAULT now()`) matches so there is no generated-migration drift.
- Migration renamed to `1787130000000-ToolVoteAndDiscussion` (above main's head), `ON CONFLICT ("id")`, and `CHECK ("vote" IN (-1, 1))` with the default dropped. DB-only check constraints without an `@Check` decorator match existing precedent (`dataset_location`, `keyword_niche`), so no change needed there.

Two items remain, one of them new.

**Blocking — the tool host posts land in the posts sitemap, in this repo.** Thanks for the investigation; I verified it independently. `buildPostsSitemapBaseQuery` (`src/routes/sitemaps.ts:161-173`) filters on type, `private`, `banned`, `deleted`, `visible`, a 90-day window and author reputation — no `sourceId` and no `showOnFeed` filter. The host post is `PostType.Freeform`, `visible: true`, `private: false`, `authorId` null (so it passes `u.id is null or u.reputation > 10`). So the moment a tool discussion is initialized, `/api/sitemaps/posts-N.xml` starts advertising `Next.js discussion` alongside `/tools/nextdotjs` from this same PR. Since the stated purpose of this PR is the SEO plumbing, shipping the sitemap that creates the duplicate-content pair without the exclusion is not something to defer: please add the `sourceId != 'tools'` (or `showOnFeed = true`) exclusion to `buildPostsSitemapBaseQuery` here, with a test asserting a tools-source post is absent from the posts sitemap. The apps-side `shouldNoindexPost` change is a fair follow-up for the paired PR, but the api-side exclusion belongs in this diff.

**Non-blocking — the rate limit now punishes the common read path.** The default key generator is per user per field, so `initToolDiscussion` is 1 call per 30s per user, and the limiter consumes before the resolver runs. That means the cheap idempotent branch (discussion already exists, early `return tool.discussionPostId`) is rate-limited too: a user opening two tool pages within 30 seconds gets a `RateLimitError` on the second one even though nothing would be created. Since `discussionPostId` is already exposed on `DatasetTool`, the clean fix is for apps to call the mutation only when `discussionPostId` is null and the user actually intends to comment. If it can be called on page load, either raise the limit or add a shared key in `src/directive/rateLimit.ts` scoped to creation intent. Worth confirming the apps call site before merge.

Checks: static review only, no tests or build executed. CI: `ci/circleci: test` still pending, `build-app`, `build-docker`, `pulumi-preview`, CodeQL green.

Reviewed by AI.

rebelchris:
Both handled.

**1. BLOCKING — sitemap exclusion.** Fixed in 6f81f4c6. Added `.andWhere('p."sourceId" != :toolsSourceId', { toolsSourceId: TOOLS_SOURCE })` to `buildPostsSitemapBaseQuery` (src/routes/sitemaps.ts:170) — this is the single query backing posts.txt, posts.xml and posts-N.xml, so the exclusion applies everywhere. Chose `sourceId != 'tools'` over `showOnFeed = true`: grepped for `showOnFeed: false` write paths and found src/common/interest/tools/writePost.ts sets it on legitimate, authored interest-digest summary posts that are still meant to be indexable — switching the filter to `showOnFeed = true` would have silently dropped those from the sitemap, which is out of scope here. `sourceId != 'tools'` is precise since that source exists solely to host these discussion posts. Added a test in __tests__/sitemaps.ts ("should exclude tools-source discussion posts while keeping other posts") asserting a tools-source post is absent from /sitemaps/posts.txt while a normal post (p4) is present — verified it fails without the fix and passes with it. Ran __tests__/sitemaps.ts (42 tests) and __tests__/datasetTool.ts + __tests__/common/datasetTool.ts (35 tests), all pass; lint and tsc clean on both changed files.

**2. NON-BLOCKING — rate limit / idempotent path.** Confirmed via packages/webapp/components/tools/ToolDiscussion.tsx (feat/tool-page branch): `initToolDiscussion` is called only inside the comment-submit flow, only when `postId` is null — `const targetPostId = postId ?? (await initToolDiscussion(toolId))` — with `postId` seeded from the SSG `discussionPostId` and cached after first init. It never runs on page load, and at most once per tool per user on their first comment. No code change made; this matches createFreeformPost's existing rate-limit semantics.

rebelchris (commented):
Re-reviewed at `6f81f4c6`. Both remaining items verified resolved.

**Blocking (sitemap exclusion) — fixed and verified.** `p."sourceId" != :toolsSourceId` is on `buildPostsSitemapBaseQuery`, which is the only source for the `.txt` list (line 278), the paginated `posts-N.xml` (line 197) *and* the sitemap-index page count (line 1068), so the exclusion and the advertised page counts stay consistent — no partially filtered page. Your `sourceId != 'tools'` over `showOnFeed = true` reasoning checks out: `src/common/interest/tools/writePost.ts` does set `showOnFeed: false` on authored interest-digest posts that should stay indexable, so the narrower filter is the right call. The new test asserts absence of the tools post while a normal post is still present, which would fail if the filter were dropped.

**Non-blocking (rate limit on the idempotent path) — verified on the apps side.** I read `packages/webapp/components/tools/ToolDiscussion.tsx` at `feat/tool-page`: `postId` is seeded from `discussionPostId`, and `initToolDiscussion` is only reached inside the comment-submit handler via `postId ?? (await initToolDiscussion(toolId))`, then cached in state. Nothing calls it on mount or in an effect, so the 1-per-30s bucket is only ever touched on a user's first comment for an uninitialized tool. No change needed; note that the paired apps PR must keep it out of page-load paths.

No blocking findings remain from my side. Outstanding non-api follow-up: `shouldNoindexPost` in `packages/webapp/lib/seo.ts` still treats the tools host posts as indexable — worth landing with the apps PR so the host post does not compete with `/tools/:slug` outside the sitemap.

Checks: static review only; I did not execute the suite. `ci/circleci: test` is still pending on this commit — that should be green before merge.

Reviewed by AI.

rebelchris (commented):
**Sign-off (posted as a comment — GitHub will not let this account formally approve, the review runtime authenticates as the PR author).**

Signing off at `6f81f4c6`. All findings from both review rounds are resolved and each fix was verified in code:

- `initToolDiscussion` creates the host post and claims `dataset_tool.discussionPostId` in one transaction, with a rollback on a lost race and a concurrency test asserting a single post row.
- Host post persists a consistent `visible`/`visibleAt` pair and an explicit `origin`.
- `@rateLimit(limit: 1, duration: 30)` in place, and the apps call site only hits it on a user's first comment for an uninitialized tool.
- `dataset_tool.updatedAt` is an `@UpdateDateColumn` with every write path going through TypeORM, so the tools sitemap `lastmod` actually moves; no DDL drift.
- Migration renamed/retimestamped above main's head, `ON CONFLICT ("id")`, and `CHECK ("vote" IN (-1, 1))` with the default dropped.
- Tools-source posts are excluded from `buildPostsSitemapBaseQuery`, covering `posts.txt`, `posts-N.xml` and the sitemap-index count, with a regression test.

Two things to keep on the radar rather than blockers: `ci/circleci: test` is still pending on this commit, so please only merge once it is green; and `shouldNoindexPost` in the apps repo should land with the paired PR so the host post page does not compete with `/tools/:slug` outside the sitemap.

Reviewed by AI.
Approved through AI review.

